PR #3, opened 2016-10-22
Add apache 2.0 license header to all source files

merope07:


merope07:
Unsure if the cuckoo cycle implementation should be MIT because it's a port of Tromp's MIT-licensed code

ignopeverell:
ASL 2.0 and MIT are compatible so that's not an issue. Actual copyright holder attribution can be handled in a separate NOTICE file. Can you also replace "The Developers" with "The Grin Developers" to make it a little more specific? That should be enough context.

merope07:
Changed "The Developers" to "The Grin Developers"

PR #7, opened 2016-10-29
fix merkle tree malleability

GarrickOllivander:
Duplicating last transaction hash for odd number of tree leafs allows for malleability, as
MerkleHash([a,b,c])=MerkleHash([a,b,c,c])
The receiver could not tell if the tree had odd number of leafs or the tree was corrupt containing transaction c twice. 

To fix, define
MerkleHash[a,b,c]=MerkleHash([a,b,c,0])

Assuming that it is impossible to create valid transaction with hash 0

This issue came up with Bitcoin, see https://bitcointalk.org/index.php?topic=102395.0

### DIFF
--- a/core/src/core/mod.rs
+++ b/core/src/core/mod.rs
@@ -163,13 +163,13 @@ impl Hashed for HPair {
 	}
 }
 /// An iterator over hashes in a vector that pairs them to build a row in a
-/// Merkle tree. If the vector has an odd number of hashes, duplicates the last.
+/// Merkle tree. If the vector has an odd number of hashes, it appends a zero hash
 struct HPairIter(Vec<Hash>);
 impl Iterator for HPairIter {
 	type Item = HPair;
 
 	fn next(&mut self) -> Option<HPair> {
-		self.0.pop().map(|first| HPair(first, self.0.pop().unwrap_or(first)))
+		self.0.pop().map(|first| HPair(first, self.0.pop().unwrap_or(ZERO_HASH)))
 	}
 }
 /// A row in a Merkle tree. Can be built from a vector of hashes. Calculates
--- a/core/src/core/mod.rs
+++ b/core/src/core/mod.rs
@@ -164,6 +164,8 @@ impl Hashed for HPair {
 }
 /// An iterator over hashes in a vector that pairs them to build a row in a
 /// Merkle tree. If the vector has an odd number of hashes, it appends a zero hash
+/// See https://bitcointalk.org/index.php?topic=102395.0 CVE-2012-2459 (block merkle calculation exploit)
+/// for the argument against duplication of last hash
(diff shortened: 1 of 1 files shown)


merope07:
Good catch Garrick -- another potential issue with the current Merkle tree code is that it may be possible to make an input or output that looks like a hash, which provides another sort of malleability. This is nearly a problem in Bitcoin (we're safe because it requires an impractical amount of grinding to create a transaction that looks like a hash) and not a problem in grin in practice (unlike Bitcoin we have no scripts or any other "freeform" fields so the grinding is even worse). But as a matter of defense in depth I'd like to kill this dead.

I've been working on a PR to replace the current Merkle tree code with [Peter Todd's MMR algorithm](https://github.com/opentimestamps/opentimestamps-server/blob/master/python-opentimestamps/opentimestamps/core/timestamp.py#L324) prepending every element with the current tree depth before hashing. This is described [https://github.com/opentimestamps/opentimestamps-server/blob/master/doc/merkle-mountain-range.md](here) and I have a working C implementation that shouldn't be too hard to port to Rust.

I also want to make the code a bit more efficient and generalize it a bit so that it will support sum-trees without needing additional algorithmic support.

If you're working on hashing/Merkle structures too we should probably coordinate. What are your thoughts on this?

~M

GarrickOllivander:
I am only poking around in the code for now and learning the language. Would not yet attack a bigger problem. As for MMR I like the idea. You are likely aware of Peter's more recent writeup on its use, but here for reference: https://petertodd.org/2016/delayed-txo-commitments

merope07:
Ok, cool, I'll write a general sum-MMR algorithm, then I propose the following data structures:

- transaction Inputs/Outputs in a tree where we sum all commitment values
- TxProof in a tree where all the "remainder" values are summed (then checking the sum means checking these two roots are the same). Signatures TBD, I understand Andrew is thinking about aggregate Schnorr signatures so maybe we will have one signature outside of the tree (?)
- UTXO commitment is a sum tree of "spent count" (and gets updated in place when txos are spent). Then given a hash at depth `n` a verifier knows she needs more data iff the spent-count is less than 2^n

I'll give more details in a PR once I've worked everything out.

ignopeverell:
Nicely spotted @GarrickOllivander, thanks for taking a look, I'll apply this pull request as it's a net improvement regardless of other additions.

The current Merkle tree code was mostly a placeholder, you'll see that I'm a firm believer in incremental implementation. And we're going to need soonish something to represent the UTXO Merkle tree. We'll need cheap lookups, additions and deletions as well as efficient store and update operations on storage (likely key/value). I'd also like a solution that's as simple as possible, Ethereum's Patricia tree makes me shudder. Believe @apoelstra had some ideas on the subject as well.

Any exploration and help here would be greatly appreciated.

PR #11, opened 2016-11-01
core: Clean up serialization and hashing

merope07:
Clean up the `Hashed` trait so that it does not require allocation anywhere; also make it reuse the `Writeable` trait where this is available. Add a `SerializationMode` property to `Writer` which signals whether or not witness data should be serialized. Replace `Option<Error>` with `Result<(), Error>` in core/ser.rs.

Sorry for these large PRs with no visible changes, but I'm working on cleaning up Merklization and I keep bumping into these problems.

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -60,7 +60,7 @@ impl Default for BlockHeader {
 // Only Writeable implementation is required for hashing, which is part of
 // core. Readable is in the ser package.
 impl Writeable for BlockHeader {
-	fn write(&self, writer: &mut Writer) -> Option<ser::Error> {
+	fn write(&self, writer: &mut Writer) -> Result<(), ser::Error> {
 		ser_multiwrite!(writer,
 		                [write_u64, self.height],
 		                [write_fixed_bytes, &self.previous],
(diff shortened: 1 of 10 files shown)


ignopeverell:
All the serialization changes look good, thanks for that.

Sadly I'm less happy with the Hashed and Writeable unification. In almost every case (i.e. anything but tests), transactions or blocks will come straight as byte arrays either from the network or the data store. Those can be hashed directly and memoized. This adds a fair amount of code that will basically be never used. Does my argument sound reasonable?

merope07:
By "a fair amount of code" do you mean the implementation of HashWriter? I can make every method of the trait Writer except write_bytes default now, since they can all be defined in terms of write_bytes. This will share code between HashWriter and BinWriter to reduce the amount of new code.

This code will be used when computing Merkle roots for transactions and for the UTXO set. The first step is always to hash the leaf data. Right now the Hashed implementation for Input and Output both allocate memory, which means that validating blocks will needs many allocations and deallocations. The simplest way to fix this is to use a Writer pattern, and the simplest way to do this is to reuse the Writer trait that already exists.

merope07:
I have added a commit that does this.

ignopeverell:
Yes , that works. Thanks for another great pull request!

PR #12, opened 2016-11-05
Add design document for Merkle tree structures

merope07:
I don't have an implementation yet but I'm posting this as a RFC of what I am going to do.

### DIFF
--- a/doc/merkle.md
+++ b/doc/merkle.md
@@ -1,0 +1,157 @@
+# Merkle Structures
+
+MimbleWimble is designed for users to verify the state of the system given
+only pruned data. To achieve this goal, all transaction data is committed
+to the blockchain by means of Merkle trees which should support efficient
+updates and serialization even when pruned.
+
+Also, almost all transaction data (inputs, outputs, excesses and excess
+proofs) have the ability to be summed in some way, so it makes sense to
+treat Merkle sum trees as the default option, and address the sums here.
+
+A design goal of Grin is that all structures be as easy to implement and
+as simple as possible. MimbleWimble introduces a lot of new cryptography
+so it should made as easy to understand as possible. Its validation rules
+are simple to specify (no scripts) and Grin is written in a language with
+very explicit semantics, so simplicity is also good to achieve well-understood
+consensus rules.
+
+## Merkle Trees
+
+There are four Merkle trees committed to by each block:
+
+### Total Output Set
+
+Each object is one of two things: a commitment indicating an unspent output
+or a NULL marker indicating a spent one. It is a sum-tree over all unspent
+outputs (spent ones contribute nothing to the sum). The output set should
+reflect the state of the chain *after* the current block has taken effect.
+
+The root sum should be equal to the sum of all excesses since the genesis.
+
+Design requirements:
+
+1. Efficient additions and updating from unspent to spent
+2. Efficient proofs that a specific output was spent
+3. Efficient storage of diffs between UTXO roots.
+4. Efficient tree storage even with missing data, even with millions of entries.
+5. If a node commits to NULL, it has no unspent children and its data should
+   eventually be able to be dropped forever.
+6. Support serializating and efficient merging of pruned trees from partial
+   archival nodes.
+
+### Output witnesses
+
+This tree mirrors the total output set but has rangeproofs in place of commitments.
+It is never updated, only appended to, and does not sum over anything. When an
+output is spent it is sufficient to prune its rangeproof from the tree rather
+than deleting it.
+
+Design requirements:
+
+1. Support serializating and efficient merging of pruned trees from partial
+   archival nodes.
+
+### Inputs and Outputs
+
+Each object is one of two things: an input (unambiguous reference to an old
+transaction output), or an output (a (commitment, rangeproof) pair). It is
+a sum-tree over the commitments of outputs, and the negatives of the commitments
+of inputs.
+
+Input references are hashes of old commitments. It is a consensus rule that
+there are never two identical unspent outputs.
+
+The root sum should be equal to the sum of excesses for this block. See the
+next section.
+
+In general, validators will see either 100% of this Merkle tree or 0% of it,
+so it is compatible with any design. Design requirements:
+
+1. Efficient inclusion proofs, for proof-of-publication.
+
+### Excesses
+
+Each object is of the form (excess, signature). It is a sum tree over the
+excesses.
+
+In general, validators will always see 100% of this tree, so it is not even
+necessary to have a Merkle structure at all. However, to support partial
+archival nodes in the future we want to support efficient pruning.
+
+Design requirements:
+
+1. Support serializating and efficient merging of pruned trees from partial
+   archival nodes.
+
+
+## Proposed Merkle Structure
+
+**The following design is proposed for all trees: a sum-MMR where every node
+sums a count of its children _as well as_ the data it is supposed to sum.
+The result is that every node commits to the count of all its children.**
+
+[MMRs, or Merkle Mountain Ranges](https://github.com/opentimestamps/opentimestamps-server/blob/master/doc/merkle-mountain-range.md)
+
+The six design criteria for the output set are:
+
+### Efficient insert/updates
+
+Immediate (as is proof-of-inclusion). This is true for any balanced Merkle
+tree design.
+
+### Efficient proof-of-spentness
+
+Grin itself does not need proof-of-spentness but it is a good thing to support
+in the future for SPV clients.
+
+The children-counts imply an index of each object in the tree, which does not
+change because insertions happen only at the far right of the tree.
+
+This allows permanent proof-of-spentness, even if an identical output is later
+added to the tree, and prevents false proofs even for identical outputs. These
+properties are hard to achieve for a non-insertion-ordered tree.
+
+### Efficient storage of diffs
+
+Storing complete blocks should be sufficient for this. Updates are obviously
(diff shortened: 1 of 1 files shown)


doc/merkle.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Merkle Trees
+
+There are three Merkle trees committed to by each block:
+
+### Total Output Set
+
+Each object is one of two things: a (commitment, rangeproof) pair indicating
+an unspent output or a NULL marker indicating a spent one. It is a sum-tree
+over all unspent outputs (spent ones contribute nothing to the sum).
```
ignopeverell: Do you mean the rangeproof would be in the tree? Why? It sounds like what you're describing here is closer to the fully cut-through block. Ideally the UTXO set would fit mostly in memory so recomputing the root for each block is cheap.

doc/merkle.md
```diff
@@ -0,0 +1,139 @@
...
+
+1. Support serializating and efficient merging of pruned trees from partial
+   archival nodes.
+
+
+## Proposed Merkle Structure
+
+**The following design is proposed for all trees: a sum-MMR where every node
+sums a count of its children _as well as_ the data it is supposed to sum.
+The result is that every node commits to the count of all its children.**
```
ignopeverell: Maybe include a link to Peter Todd's MRR description for other readers?

ignopeverell:
Forgot to mention: the design overall makes a lot of sense. Thanks for specifying it!

doc/merkle.md
GarrickOllivander: How is the identity of an output is derived and how does it link to its later spend?

merope07: By the hash of its commitment, as is done now. I'll update the doc to say this.

merope07: How can the UTXO set fit cheaply in memory? What if there are millions of outputs?

And the rangeproof has to be committed somewhere, putting it beside the commitment to the output seemed to make sense. We can have a separate Merkle tree for them if you like.

apoelstra: @ignopeverell can you clarify the distinction between "fully cut-through block" and the UTXO set?

I like this idea of maintaining a commitment to every output forever (even though nobody needs to keep it to validate the chain) and being able to produce spentness proofs (and also that every output has an efficiently verifiable index which will indirectly indicate its age).

apoelstra: I suppose you should also enforce that identical outputs don't exist at the same time, or introduce a tiebreaker rule ... I prefer saying they don't exist at once.

apoelstra: @merope07 I would prefer (though I'm not certain yet what Igno is thinking) that the rangeproofs go in a separate Merkle structure, which maybe mirrors the output structure, so that validators who don't want to check the rangeproofs (or don't want to check old ones, or whatever) have a cheap option.

merope07: I don't feel strongly about it, but yes, that makes sense @apoelstra

ignopeverell: Agree with @apoelstra, that'd be a good place to put that constraint.

apoelstra: I think that if we have unique indices for each output, these should be included in the references. This is simple and easy, prevents replay attacks, and might simplify lookups for some nodes.

ignopeverell: An output without a rangeproof is 32 bytes. Ten millions of them is only 320MB. Ten million outputs with the rangeproof (fully blinding) is over 50 GB.

I see the UTXO Merkle tree as a specialized index. Updating it and recomputing the root is going to be the most expensive validation operation. Someone will likely try to DDoS it at some point. Without the rangeproof it easily fits in memory, gives us fast block processing and fast reply to a good range of queries, and makes a node less susceptible to random reads types of attacks. Even with a billion output, because of the use of MMR, we can still fit the most recent parts in memory.

Is see the rangeproof as reference data. It will fit on disk with transactions and blocks, likely part of a nice full sequential write and rarely read, may be eventually pruned based on node policies, etc.

Makes sense?

apoelstra: Yes, I think we're in full agreeance: (a) the rangeproofs should be in an entirely separate structure, (b) otherwise this sum-MMR-where-spent-outputs-are-NULLed scheme seems good.

merope07: So I add a fourth tree with rangeproofs in it? And it doesn't sum over anything, unless @apoelstra knows a meaningful way to aggregate rangeproofs.

ignopeverell: Btw just a little gotcha to keep in mind: output unicity can only be enforced in a given fork. Two parallel forks may have the same output and that should be fine. Each fork should have its own tree which is mostly based off a common tree from a few blocks back

apoelstra: They need to be committed somehow to the blockchain, otherwise blocks can be presented to different peers with different validity and preventing this without DoS vectors is difficult. Committing them in a tree is the most efficient way to do this (it avoids rehashing too much old data) and also enables partial validation of the rangeproofs.

apoelstra: Oh, this is a good point, committing to the index means that output references are no longer reorg-safe. This might be a serious enough reason by itself to not do this.

GarrickOllivander: @apoelstra Why should references be reorg safe? Each fork had its own decisions of how it ordered transactions, cut-through or resolved conflicts.

GarrickOllivander: @apoelstra  I think transactions remain valid in their original form, but once cut-through or merged into the blocks they do not retain information for "relocation" into an other fork. The original transactions will have been merged differently on different forks and that is fine.

GarrickOllivander: @apoelstra Is it possible to use commitment derived identities of outputs on wire but index references in blocks?

apoelstra: Suppose I try to spend an output from block 1000 in block 1001. I make a transaction saying "spend output `abcd`, index 25000" and publish this, but then block 1000 gets replaced with one that makes output `abcd` have index 24999. Now my transaction is invalid, even though I didn't double-spend and nobody was acting maliciously. That's what I'm concerned about.

While its true that complete blocks don't have enough information for this sort of reorg to happen, in a model where miners do aggregation "at the last minute" and have some cache of unmerged transactions, this can certainly happen.

apoelstra: @GarrickOllivander yes, you can do that, but it means that the indices don't get signed, so you lose the replay protection. (I'm not super concerned about replay protection to be honest, any software that's reusing commitments is broken, but if we can get it we should.)

Oh -- to be clear, by "replay protection" I'm referring to a situation where somebody makes an output, spends it, then later makes the same output, which gets spent by a replay of the first transaction. I'm not talking about forks at all.

GarrickOllivander: @apoelstra you'd sign "spend output abcd" the index would be worked out by the miner.

apoelstra: Right, but this doesn't gain you much, it's just a lookup hint, and (I'd argue) shouldn't be taking up blockspace. It doesn't protect against replays.

apoelstra: I don't see any data that could be added to an input that is (a) reorg safe, and (b) protects against replays.

GarrickOllivander: If transaction is of form "spend output abcd" then index is not signed, but it does not need to be since index can be used to retrieve abc and check the signature. Reorg can assign other index as long as it checks out with the sig.

apoelstra: Right, if another index can be assigned then this is no replay protection at all.

ignopeverell: I haven't been able so far to find a working replay protection that doesn't add too many limitations (like making the spending of unconfirmed transactions impossible). But like @apoelstra I'm not super worried about it for now.

GarrickOllivander: For the replay you describe it must be possible to create the same commitment again, right? I suggested to avoid that by requiring that every commitment has some digest of its input(s) worked in so it is not feasible to create same commitment that is not also a double spend. This would be similar to Bitcoin's replay protection. There reference to output is by transaction id + index where transaction id hashes the input references, therefore output could only be recreated if it were also a double spend.

ignopeverell: Can you elaborate on the DoS vector? Right now the range proofs aren't part of the output, transaction or even block hashes. A block with altered range proofs would be the same block.

ignopeverell: The "worked in" part is the issue. You'd either need to add signatures so the spending reference can be authenticated (using more blockchain space) or break cut-through if you added it somehow to the commitment.

GarrickOllivander: @ignopeverell you are right. My previous idea is wrong as it would leak association with inputs. I assume you used block height in your replay protection idea that therefore made spend of unconfirmed impossible. @apoelstra is right that replay protection needs to be addressed. I'd prefer that commitments would be unique by construction. Can we assume that in the construction of the tree?

doc/merkle.md
merope07: Added.

merope07:
Update:

1. Add a link to MMR.
2. Remove rangeproofs from output tree.
3. Add note about uniqueness of outputs, and explain that inputs refer to old outputs by hash.
4. Add fourth tree, to be debated, with rangeproofs.

ignopeverell:
Sounds good to me. Would you prefer I merge this now and create new pull requests for additions or continue on this pull request? I have a slight preference toward merging now and more pull requests but either way work for me.

merope07:
I would prefer to merge now. This comment thread is noisy enough already. :)

ignopeverell:
Done. Thanks for a great start on this work with the writeup!

PR #14, opened 2016-11-09
Slight changes to transaction signature.

ignopeverell:
The transaction fee gets signed (instead of the empty string) and the fee amount is attached to the transaction proof in blocks. This allows fee accounting and the tracking of fees within a block.

The background for this is to either enforce a certain maturity for the block reward as well as the fees reward, or not allow any fee reward at all. Without the signing of transaction fees, a miner would be able to introduce any output to claim the fees without any possibility for the system to know about, or check its maturity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,11 @@ Find an area you can help with and do it. Open source is about collaboration and
 
 Additional tests are rewarded with an immense amount of positive karma. So is documentation.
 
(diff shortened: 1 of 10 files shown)


ignopeverell:
Pinging @merope07 and @apoelstra to make sure they're aware of this.

PR #15, opened 2016-11-13
Safer serializations

GarrickOllivander:
Removed Hash::from_vec as it was working with any length of vec, creating ambiguity. Instead introduced read_hash to avoid the need for repeated conversion of vec to Hash. Used Hashed trait instead of repeating hash algorithm. 

Renamed Proof to POW. Removed POW::from_vec for similar reasons. introduced read_pow and write_pow.

I am not sure [0; x] is interchangeable with [0u8; x]. made explicit to be on the safe side.

what is read_33_bytes for?

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -29,7 +29,7 @@ pub const REWARD: u64 = 1_000_000_000;
 pub const BLOCK_TIME_SEC: u8 = 30;
 
 /// Cuckoo-cycle proof size (cycle length)
-pub const PROOFSIZE: usize = 42;
+pub const POWSIZE: usize = 42;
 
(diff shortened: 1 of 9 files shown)


ignopeverell:
Both read_33_bytes and write_33_bytes are used to serialize commitments.

Rust will infer the u8 part so if you use [0; x] in a context that requires byte arrays, the u8 is unnecessary.

I like the move of read and write to POW structure, but I'm a little conflicted with the naming. It was originally Proof because it's the Cuckoo proof, which represents a cycle in the graph. Strictly speaking the proof of work is the hash of that. We just happen to have a lot of different types of proofs :) It might be better for simplicity to conflate the two or that might make it more confusing, I'm not sure. What do you think?

PR #18, opened 2016-12-03
Safer serializations

GarrickOllivander:
Safer serializations. 
Removed xx:from_vec constructors as they did not check length or paniced.


ignopeverell:
Going in the right direction, you're right about removing the from_vec constructors. I just don't like adding all those specialized readers and dependencies in ser.rs and the related base traits. How about doing Proof::read or Hash::read instead of read_proof and read_hash?

GarrickOllivander:
review please. Can't move Commitment and RangeProof out of ser.rs as doing so would introduce cyclical dependency between core and secp.

ignopeverell:
As there were a lot of small additive changes I ended up squashing the merge to keep the history cleaner. Hope you don't mind.

GarrickOllivander:
That was the right thing to do, thank you.

ignopeverell:
Thanks for another great pull request!

chrismayo23:
hello

Issue #34, opened 2017-03-28
Replace AsFixedBytes<T> with AsRef<T>


Latrasis:
Since [`AsRef`](https://doc.rust-lang.org/std/convert/trait.AsRef.html) already exists within the standard library, it would make sense to use it instead of adding another trait.

ignopeverell:
The trait AsFixedBytes adds the requirement that structs implementing it must be fixed size. Implementing it for a struct with a &[u8] representation of variable size would violate the trait. Maybe it would be better expressed in terms of AsRef forcing Sized though.

Latrasis:
I assume if the `Sized` constraint is added to the `Writer::write_fixed_bytes` method, that may work?

PR #35, opened 2017-03-29
Replace AsFixedBytes with Sized + AsRef<[u8]>

Latrasis:
Implementation for #34

### DIFF
--- a/core/src/core/hash.rs
+++ b/core/src/core/hash.rs
@@ -20,8 +20,9 @@
 use byteorder::{ByteOrder, BigEndian};
 use std::fmt;
 use tiny_keccak::Keccak;
+use std::convert::AsRef;
 
-use ser::{self, AsFixedBytes, Reader, Readable, Writer, Writeable, Error};
+use ser::{self, Reader, Readable, Writer, Writeable, Error};
 
 pub const ZERO_HASH: Hash = Hash([0; 32]);
 
@@ -50,6 +51,12 @@ impl Hash {
 	}
 }
 
+impl AsRef<[u8]> for Hash {
+    fn as_ref(&self) -> &[u8] {
+        &self.0
+    }
+}
+
 impl Readable<Hash> for Hash {
 	fn read(reader: &mut Reader) -> Result<Hash, ser::Error> {
 		let v = try!(reader.read_fixed_bytes(32));
@@ -62,7 +69,7 @@ impl Readable<Hash> for Hash {
 }
 
 impl Writeable for Hash {
-	fn write(&self, writer: &mut Writer) -> Result<(), Error> {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), Error> {
 		writer.write_fixed_bytes(&self.0)
 	}
 }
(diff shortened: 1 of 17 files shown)


core/src/ser.rs
```diff
@@ -140,14 +132,14 @@ pub trait Writer {
 
 	/// Writes a variable number of bytes. The length is encoded as a 64-bit
 	/// prefix.
-	fn write_bytes(&mut self, bytes: &AsFixedBytes) -> Result<(), Error> {
-		try!(self.write_u64(bytes.as_fixed_bytes().len() as u64));
+	fn write_bytes<T: Sized + AsRef<[u8]>>(&mut self, bytes: &T) -> Result<(), Error> {
```
Latrasis: Note that this makes the `Writer` trait generic.

ignopeverell:
Instead of defining <T: Sized + AsRef<[u8]>> on everything (especially outside of the module, like in ser.rs), I'd still introduce a separate type, maybe something like (not set on the name, if you can find something better):

    trait SizedBytes: Sized + AsRef<[u8]>

That should help encapsulating it a bit.

Also in secp256k1zkp I think it makes sense to implement AsRef as part of the impl_array_newtype macro.

Latrasis:
Considering that `[u8]` is unsized, why is it being included as part of `AsFixedBytes`? I'm having problems currently with using implementing it as a marker trait, because `[u8]` is unsized.

core/src/core/hash.rs
```diff
@@ -121,7 +121,7 @@ impl Hashed for [u8] {
 	fn hash(&self) -> Hash {
 		let mut hasher = HashWriter::default();
 		let mut ret = [0; 32];
-		ser::Writer::write_fixed_bytes(&mut hasher, &self).unwrap();
+		ser::Writer::write_fixed_bytes(&mut hasher, &self.to_owned()).unwrap();
```
Latrasis: Had to box the slice to have it `Sized`. This doesn't seem like a solution though...

ignopeverell:
Can you describe to problem maybe, see if I can help (or if my review comments were just wrong)? :)

Latrasis:
`impl Hashed for [u8]` relies on `ser::Writer::write_fixed_bytes` however `self` (which is `[u8]` in this case) can't be`AsFixedBytes`, because any `[T]` is not `Sized`... So i'm currently slightly stuck on this because i'm not sure what solution would be a better fit. The previous `FixedBytes` seems to have been contradictory on `impl FixedBytes for [u8]` because slices can't be guaranteed to have fixed size. Perhaps you mean something else than `Sized`?

ignopeverell:
I pushed a change that removes `impl Hashed for [u8]`, it was really only there to be able to hash an empty array. In your branch you should be able to remove the `AsFixedBytes` implementations for `[u8]` and `String` as they're only required for `Hash`, which in your branch should implement `AsRef`. I believe that should help.

Latrasis:
Thank you!

ignopeverell:
Are you ok with where the pull request is at now? If so I'll go ahead and merge it, it looks fine to me.

Latrasis:
Yup

PR #36, opened 2017-04-04
Create the beginning of the wallet library.

melaurent:
The wallet library will be used to manage the keys of the user. Currenty it only contains a minimal implementation of the extended keys. The format of the extended keys is subject to change

### DIFF
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "wallet"
+version = "0.1.0"
+authors = ["Laurent Meunier <laurent.meunier95@gmail.com>"]
+
+[dependencies]
+
+secp256k1zkp = { path = "../secp256k1zkp" }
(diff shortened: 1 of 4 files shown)


wallet/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "wallet"
```
ignopeverell: Should be grin_wallet to be consistent with the other modules.

ignopeverell:
Should I let this pull request go for a bit or you'd rather merge early? If so you'll have to merge master so it compiles and I have to review it for a bit.

ignopeverell:
Should I let this pull request go for a bit or you'd rather merge early? If so you'll have to merge master so it compiles and I have to review it for a bit.

melaurent:
You can leave it open, it was so that others could see which direction it was going

ignopeverell:
Merged this as I'm going to need it.

PR #40, opened 2017-04-14
[WIP] Implement Merkle-sum-tree

merope07:


core/src/core/sumtree.rs
```diff
@@ -0,0 +1,922 @@
...
+
+/// Enum indicating whether or not the node has the maximum 2^n leaves
+/// under it. Leaves count as being full, so partial nodes are always
+/// internal nodes.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+enum SubtreeFull {
+    Yes,
+    No
+}
+
```
ignopeverell: I'm all for strong typing but not sure this type does a lot for you over a boolean :)

core/src/core/sumtree.rs
```diff
@@ -0,0 +1,922 @@
...
+    type Sum = NullSum;
+    fn sum(&self) -> NullSum { NullSum }
+}
+
+#[derive(Clone)]
+enum NodeData<T: Summable> {
+    /// Node with 2^n children which are not stored with the tree
+    Pruned(T::Sum),
+    /// Actual data
+    Leaf(T),
```
ignopeverell: As discussed on Gitter, I think we want to tree to only deal with hashes to keep it as succinct as possible. A slight improvement would be to be able to take a commitment directly (as it's only 1 byte over a hash) but it may be a premature optimization at this point.

core/src/core/sumtree.rs
```diff
@@ -0,0 +1,922 @@
...
+    pub fn unpruned_len(&self) -> usize {
+        match self.root {
+            None => 0,
+            Some(ref node) => node.n_children()
+        }
+    }
+
+    /// Add an element to the tree. Returns true if the element was added,
+    /// false if it already existed in the tree.
+    pub fn push(&mut self, elem: T) -> bool {
```
ignopeverell: This method signature could stay the same but elem wouldn't be stored, just hashed and summed.

core/src/core/sumtree.rs
```diff
@@ -0,0 +1,922 @@
...
+        // Can add
+        assert!(tree.push(elem));
+        // Cannot double-add
+        assert!(!tree.push(elem));
+        // Can prune but not double-prune
+        assert!(tree.prune(&elem));
+        assert!(!tree.prune(&elem));
+        // Can re-add
+        assert!(tree.push(elem));
+    }
```
ignopeverell: Maybe some basic serialization/deserialization tests?

merope07:
Rebased but made no other changes. Checked that every commit compiles and passes unit tests. Maybe I should be doing something different with serialization since the `Codec` stuff, but I have not followed those changes, sorry.

merope07:
Remove `Yes`/`No` enum

Latrasis:
We decided to undo the codec pr, please ignore it! Thank you!

ignopeverell:
Given that it has very little impact on the rest of the codebase I'm going to merge this PR as a first pass at the problem and then work from there. I'll implement the changes I requested and start integrating with the rest of the core package.

PR #43, opened 2017-04-18
Correct minor typo - "circumstances"

Rob-pw:


ignopeverell:
Ended up removing the file altogether as it was outdated and the ML is a better resource. Thanks still!

PR #44, opened 2017-04-19
Delegate Block Encoding to Store

Latrasis:
Move Block and Transaction encoding to separate codecs.

Blocks via `BlockCodec`:
* [x] BlockHeader
* [x] Input
* [x] Output
* [x] TxKernel
* [x] Hash
* [x] Proof

Transactions via `TxCodec`
* [x] Transaction


Latrasis:
@ignopeverell: Just to make sure, the `BlockCodec` only needs to encode a `Block` and nothing else, correct? Not including (`Tip`, `PeerData`, or `Transaction`).

ignopeverell:
@Latrasis correct, BlockCodec only needs to encode a Block. And Tip and PeerData encoding should be separate, one is used by the chain implementation while the other is only relevant to the p2p module.

Latrasis:
@ignopeverell: Just a heads up, the rough implementations of `TxCodec` and `BlockCodec` are done. Could probably use some refactoring, but otherwise I think ok. Let me know how you would like to move forward.

ignopeverell:
The code looks good. A little cleanup as you mention would be nice (license header, spurious carriage return diffs, etc) nothing major at all.

My main question would be regarding the serialization mode, which the current serializer supports. The block serializer for example has something like this:

https://github.com/ignopeverell/grin/blob/master/core/src/core/block.rs#L159

Depending on the consumer (hashing, network, storage), serialization can be slightly different. I didn't see anything like this in the patch but I may have missed it. Otherwise, how can we do the same thing with the codec? Looks like an simple member in the struct could do the job.

Once that's sorted out, I'm happy to merge this PR, and maybe do the migration to codecs and removal of the old serialization in another PR? Does that work for you?

And thanks for the great work, it's much nicer separated from core!

Latrasis:
Sounds good! In terms of two different encoding formats though, how do you know which is which during decoding? Otherwise it looks like that should be in a separate codec then.

ignopeverell:
It's context specific. The store knows which mode was used to encore and so is the p2p protocol.

Latrasis:
Since you are currently only encoding the `BlockHeader`, it looks like it should be separate codec for the `BlockHeader` then.

ignopeverell:
The link I provided was just one example but there are multiple uses of this mechanism. Serializing for hashing uses it and so do transactions. Grep the code for SerializationMode. Not sure we'll want a separate codec for every one of these...

Latrasis:
Wouldn't we need just one additional codec though? Doesn't the store only encode/decode blocks?

ignopeverell:
Oh sorry, misunderstood what you meant. Yes, different codec may be fine.

Latrasis:
I'm still confused however, how is a hashed block being decoded though? I don't see a Hashed Block being decoded.

Latrasis:
Looking at how a `BlockHeader` handles `Writeable`, it's not a really hashing, but simply omitting parts. Is this really hashing or should it be called something else?

ignopeverell:
The hashing is only one way, it's basically which parts of serialization need to be fed to the digest, which are almost entirely the same as standard serialization. And it is omitting parts because a block (and transactions) have a lifecycle. Some of the data (like the nonce) is only added after and so some things can't relay on it. Does that help?

Latrasis:
I see, thanks that helps!

Latrasis:
Implemented `BlockHasher` at 15d581c. Let me know if it's good to go. I'm not sure how to unit test it however, so if you have any suggestions let me know.

Latrasis:
Can you try running checks again? Some port issue it looks like.

ignopeverell:
One of them is a little too long and sometimes abort. Restarted it and looks ok now.

ignopeverell:
So there's no way to reconcile a bit the BlockHasher and BlockHeader codecs? Maybe just a utility function? The generics may be a little tricky to figure out but I can already see people modifying one and forgetting to modify the other. And thanks for the great continuing work!

PR #50, opened 2017-05-03
adding some basic build instructions to help people get started

dan-da:
I've never used rust before so it wasn't immediately obvious to me how to build and get started.

Fortunately it is not too complicated but anyway I added these instructions as a starting point for others.


ignopeverell:
That's great, thank you!

PR #51, opened 2017-05-07
[WIP] Partial Transition from Writeable/Readable to Codecs

Latrasis:
Continuing from #44, we need to transition from `grin_core::ser` to tokio codecs. This involves:
1. Re-implementing left-over `Writeable`/`Readable` types to codecs, including: 
* [x] `grin-chain`
* [x] `grin-p2p`
  * [x] Setup `MsgCodec` + Tests
  * [x] Setup`PeerDataCodec` + Tests
2. Redefining how `grin_store::Store` is used.


Latrasis:
@ignopeverell : Due to how `ChainKVStore` has methods requiring more granular types such as `Output` and `Hash`. I decided to make `BlockCodec` generic at 683eccb. Let me know if you have any concerns.

There's a lot of churn, so I'm open for suggestions on how to make this PR better for auditing.

ignopeverell:
No concern. I'm not too worried about churn, this is still an early project, sometimes things take some time to figure out. I'll just clean up the squash comments on merge.

p2p/src/msg_codec.rs
```diff
@@ -230,6 +241,81 @@ impl MsgDecode for Hand {
 	}
 }
 
+// #[derive(Clone, Debug, PartialEq)]
```
Latrasis: Seems like you forgot to remove some comments

p2p/src/msg_codec.rs
```diff
@@ -304,18 +302,46 @@ impl MsgDecode for Shake {
...
+			return Ok(None);
+		}
+		let mut buf = src.split_to(4).into_buf();
+		let capabilities = Capabilities::from_bits(buf.get_u32::<BigEndian>()).unwrap_or(UNKNOWN);
+		Ok(Some(GetPeerAddrs { capabilities: capabilities }))
+	}
+}
+
 const SOCKET_ADDR_MARKER_V4: u8 = 0;
 const SOCKET_ADDR_MARKER_V6: u8 = 1;
```
Latrasis: Should probably move to top

Latrasis:
Hi @ignopeverell! I think we should merge this as a partial transition.

Due to how extensive `Writeable/Readable` are used, I have realized making a full transition within this PR isn't feasable. Instead of a full transition, we have kept the scope strictly within implementing the encoding format and pruning as much as we can. 

Here is the rundown:

There are 4 codecs implemented:
 * `BlockCodec` : for encoding/decoding `Block` _and_ sub-types.
 * `TxCodec` : for encoding/decoding `Transaction`
 * `PeerCodec`: for encoding/decoding `PeerData`
 * `MsgCodec`* : for encoding/decoding a `Message`

* `BlockHashCodec` : for encoding `Block` _and_ sub-types

Three stores have been changed to use above codecs:
 * `store::Store`
 * `p2p::PeerStore`
 * `chain::ChainStore`

What has been pruned, (not much).
 * Most of `chain` types
 * Some of `p2p` types

What is left and should be split into separate PR's:

1.  The main blocker for pruning `core` types is the `Hash` type. The most straightforward direction is to use the `BlockHashCodec`. However due to how extensive the `Hash` type is, I believe this should be in a separate PR.

2.  The `p2p` crate. With the benefits of using codecs, it would make sense to use them in the context of the networking which they are designed for. However, currently the`p2p` only marginally uses `tokio`, and transitioning to a tokio-based system will require extensive changes in itself ( if we decide to do so).

Once these two blockers are resolved, we should be able to have a full transition and be able to remove the `core::ser` module completely.

Latrasis:
On the other hand, I realize that while a full-transition PR is concernedly large, having a partial-transition may induce unnecessary churn for the main branch. It's a trade-off, so I'm open either way.

ignopeverell:
I'm fine with splitting in multiple PR and you separation in 2 additional ones makes sense. The full sync tests in the grin module are failing on this branch, is that due to the hash type?

Latrasis:
Most likely a regression from 82b824c. If it is, we should probably then revert it since there's insufficient test granularity anyways.

Latrasis:
Looks like that's a different error, so that regression is removed. I think it's good to go.

Issue #55, opened 2017-06-07
Implement Rate Limiting


Latrasis:
From #54: 
> For DoS protection the first rule should be rate limiting based. No peer should send us more than X MB per second and we shouldn't send more (assuming it was all requested) than Y MB per second to any peer either. We can start with X=200 and Y=100. That should be measurable at the protocol or connection level easily.

* [x] Implement memory throughput counter for each peer connection
* [x] Setup tests for rate limits

Latrasis:
@ignopeverell: For testing, any pointers for mocking nodes that spam too much but valid data?

ignopeverell:
* The easiest is probably using the PeerAddrs message (when a peer sends the list of peers it knows to another peer). See: https://github.com/ignopeverell/grin/blob/master/p2p/src/protocol.rs#L200
* What do you mean by "handled"? How to implement them? Maybe a per-second counter?

PR #57, opened 2017-06-08
[WIP] Rate Limiting

Latrasis:
For #55


Latrasis:
Hi @ignopeverell, sorry for the delay, it took me a while to figure out the best solution, here's a rundown:
* Implemented `ThrottledRead` and `ThrottledWrite` [here](https://github.com/ignopeverell/grin/pull/57/files#diff-01225f277758cfd0455433dc85a0c8d2)
* Current binding at [`Connector`](https://github.com/ignopeverell/grin/pull/57/files#diff-8d32dcd9ab02f899032561413230fc3fR97)

I scrapped the previous tests, so i'll write inline ones. I think that's all for now, let me know what you think.

ignopeverell:
After a first read, this looks pretty good. It looks like some imports could be cleaned up (seems that rate_limit.rs doesn't need to depend on any other internal module) but that's a detail. The implementation looks nice and reasonably tight, I like it!

Latrasis:
Thank you! :smile:

I haven't tested them yet, but it should be reasonably easy. Oh and I added the pub imports simply to see the docs - I can fix that.

PR #65, opened 2017-06-16
Testing Framework [WIP] and new command line options to configure wallet ports and directories

yeastplume:
While working on the testing framework in grin/tests/simulnet.rs (which is still a work-in-progress, below,) I've made a few changes in the rest of the code to better support tests running multiple servers on the same host. They reach across a few modules, so I thought I'd put them in a PR before going too much further.

Changes are:

1) Added a stop method to the rest API server, as any test running a receiving a wallet will hang indefinitely without stopping the server.
2) Modified the wallet.dat and wallet.lock files, (and the with_wallet function) to also take a directory path, and added the directory path to the WalletConfig struct. Also now require a WalletConfig to start a wallet server, and added it into a few more places so the data directory is available when needed.
3) Modified the wallet module to make the listener port configurable from the command line (via an addition to the MinerConfig structure)
4) Modified the server module to make the listening wallet for mining rewards configurable on the command line.
5) Added switches to the command line and associated help documentation.
6) Changes to the simulnet.rs tests, still a WIP

RUST_LOG=debug cargo test simulate_parallel_miners

Although a WIP, these changes can be exercised by running the simulate_parallel_miners test, which doesn't prove much at the moment, but it starts up 5 separate mining servers in 5 separate directories (under test_servers), running a separate wallet receiver for each and placing their respective wallet.dat files in the right directories.

Also considering refactoring the test framework structs (LocalServerContainer and LocalServerContainerPool) out of the grin module and somewhere accessible to all modules, perhaps in a separate module set up for running tests. Happy to discuss this further.

### DIFF
--- a/api/src/rest.rs
+++ b/api/src/rest.rs
@@ -24,11 +24,13 @@ use std::io::Read;
 use std::net::ToSocketAddrs;
 use std::string::ToString;
 use std::str::FromStr;
+use std::mem;
 
-use iron::{Iron, Request, Response, IronResult, IronError, status, headers};
+use iron::{Iron, Request, Response, IronResult, IronError, status, headers, Listening};
 use iron::method::Method;
 use iron::modifiers::Header;
 use iron::middleware::Handler;
+use iron::error::HttpResult;
 use router::Router;
 use serde::{Serialize, Deserialize};
 use serde::de::DeserializeOwned;
@@ -227,6 +229,8 @@ fn extract_param<ID>(req: &mut Request, param: &'static str) -> IronResult<ID>
 pub struct ApiServer {
 	root: String,
 	router: Router,
+	server_listener: Option<Listening>,
+
 }
 
 impl ApiServer {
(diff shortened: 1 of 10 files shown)


ignopeverell:
Just merged, thanks a lot, very useful. I was already planning on pushing up the wallet config and now you'd done it, it's great! Being able to stop the server was also necessary.

I may just add a couple commits to tweak your code here and there, for example Cargo has a special section for test dependencies instead of adding them for everything. But nothing big, hence the direct merge.

yeastplume:
Thanks, and yes I thought I was pulling too much unnecessary nonsense into the grin module, so I'll keep an eye out for the tweaks for future reference.

I'm going to keep plodding away at the test framework I've started for more learning exercise fun, hopefully getting to a point where the POW difficulty can be easily tested. Still working my way around the codebase and language, but if you have any similar peripheral/quality of life improvements you could do with, let me know and I'll be happy to take them on

PR #69, opened 2017-06-28
Mining test debug output, fixes to diff. adjustment and start of POW documentation

yeastplume:
-Modified the POW difficulty adjustment to cover the case where the difficulty has been 1 for the last *window* blocks.
-In the case of a fresh blockchain without enough of a window to make a difficulty calculation, use the block height as the difficulty.
-Added a check if the difficulty adjustment somehow comes up with a zero for the difficulty, which is unlikely in the wild but quickly destroyed any blockchains I had going, so thought it best to test for it and cause a panic, just in case.
-Modification made to miner debug output, to better track which server is doing what when running multiple servers locally on different threads.
-A fix to a mining 'gimping' adjustment which is supposed to artificially delay mining threads each iteration.
-There's a start here on documenting Grin's POW, still needs quite a bit of review and editing, but it's clearly marked as a work in progress and should be no harm to include at the moment.


JollyMort:
I read the doc, looking forward to first CC with Cuckoo Cycle. I have one question - what's the rationale behind using SHA256? Did you consider alternatives like Scrypt, Keccak or maybe CryptoNight to keep it ASIC-resistsnt. CN needs 2MB of high-speed memory per thread to be competitive, so it's either CPU or GPU for now. Thanks!

ignopeverell:
First thing is we're actually using SHA3 (Keccak) thoughout, not SHA256. Thanks for catching that.

The second thing is that SHA3 is not the proof of work, the actual proof of work is Cuckoo Cycle. We're likely to set the Cuckoo size to 30, so memory requirements should be in the 2GB to 4GB ballpark.

But because Cuckoo generates 42xu32 sized proofs (168 bytes), it's easier to hash the proof so we can maintain a 32 bytes target and use that for comparison. It's just a final trick so our target/difficulty is shorter and looks more like what people are used to.

JollyMort:
Cool, thanks! Do I understand correctly that the hashing part is what will actually enable pool mining?

ignopeverell:
Indirectly yes. What enables pool mining is something desirable in any proof of work called progress-freeness. The idea is that if you have a 1 min block time (for example) and have to spend 1 min to find a solution on average, this still satisfies the PoW requirement but gives a strong advantage to big miners. In that setup, small miners will generally lose at least 1 min every time while larger miners can move on as soon as they found a solution. So you want a PoW that requires multiple solution attempts with each attempt being a relatively smaller amount of time.

What makes us progress-free is the fact that a solution to Cuckoo with our default parameters can be found typically in under a second on most GPUs and that we add an additional requirement on top of that. Our requirement is that the SHA3 of the solution should be lower than some value, just because it's easy and common (and smaller). But we could directly require the Cuckoo proof to be lower than some value. Or the first n bits to be a prime number.

Once you have progress-freeness, then pooled mining comes somewhat easily on top of that. In our case, pools could not even require the additional SHA3 step and make any valid Cuckoo Cycle solution a valid share. Or a small bundle of those.

Makes sense?

yeastplume:
I'll update the doc with the SHA 3 correction and fill out the section on pooled mining with something based on your progress freeness description.... wanted to put more in that section anyhow, just wasn't sure how to put it

JollyMort:
@ignopeverell makes perfect sense, thanks!

PR #79, opened 2017-07-12
Addition of configuration file

yeastplume:
Added a configuration file (just placed in root directory in grin.toml for the time being,) and a separate crate (config) to manage it. The rules for where/how the config file is read is described within grin.toml, but it basically searches the current directory, executable directory, and then home directory for a configuration file.

The command line options are still there, and the config file is only picked up if no command line switches are given. If switches are there, the config file is ignored and everything works as before.

I originally wanted to include the wallet as part of the config, but I don't think it makes sense at the moment because there are sensitive values there (keys) and I think (correct me if I'm wrong,) that the wallet should be split out at some point, or at least treated separately. I thought it was better to keep the config file relevant to the server for the time being.

Also some internal changes to allow for better serialisation and for some values to be optional in the configuration file.

Just a first pass and will grow and change, obviously, but I think this will work to make grin easier to play with and less painful to start adding configuration values to.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,11 +4,12 @@ version = "0.1.0"
 authors = ["Ignotus Peverell <igno.peverell@protonmail.com>"]
 
 [workspace]
-members = ["api", "chain", "core", "grin", "p2p", "store", "util", "pool", "wallet"]
+members = ["api", "chain", "config", "core", "grin", "p2p", "store", "util", "pool", "wallet"]
 
 [dependencies]
 grin_api = { path = "./api" }
 grin_wallet = { path = "./wallet" }
+grin_config = { path = "./config" }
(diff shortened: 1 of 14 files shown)


ignopeverell:
This seems to break compilation of tests in the grin crate, can you check?

Ultimately I think we'll want to merge the config file (if it exists) with the command line arguments (those taking precedence over the file). But that's a good first pass at it.

yeastplume:
Apologies, I submit the PR then ran off without checking the CI status, I need to remember to run all the tests locally in future. Fixed now.

It is absolutely just a first pass, no doubt it will evolve as needs become more evident. But for the moment I mostly wanted a way to start adding parameters to start exercising cuckoo-miner without having to type a dozen command line switches each time.

PR #80, opened 2017-07-14
Cuckoo-miner better merged into Grin

yeastplume:
With these changes, cuckoo-miner (a particular tag made just for testing) is fully integrated into grin, and all of the variants and options can be controlled via the grin.toml config file. Cuckoo-miner is turned off by default, but if you change the use_cuckoo_miner to true in the config file, it'll pick up and use the simple memory-hog miner. The edge trimming variant miner and its options are also included. Cuda miner should work if you build it from the cuckoo-miner repository and configure properly, but it's not included here as it would break many environments.

Couple of points:
-Removed conditional dependencies for including cuckoo miner. It's now being pulled from a specific tag in the cuckoo_miner repository.
-Removed the cuckoo_size field from the mining config, because it was confusing and conflicting with the consensus values. The cuckoo size is always driven by either TEST_SIZESHIFT or DEFAULT_SIZESHIFT now, with test mode turned on and off via the config file
-Changed the TEST_SIZESHIFT default value to 16, as that's John Tromp's recommended minimum for the edge trimming miner.

All miner parameters are documented in the config file, and if anything's misconfigured there should be a helpful log::error! message indicating what the problem is. Also tested cuckoo-miner integration on both linux and mac, all seems working as intended for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ members = ["api", "chain", "config", "core", "grin", "p2p", "store", "util", "po
 [dependencies]
 grin_api = { path = "./api" }
 grin_wallet = { path = "./wallet" }
+grin_grin = { path = "./grin" }
 grin_config = { path = "./config" }
 secp256k1zkp = { path = "./secp256k1zkp" }
 
@@ -20,14 +21,5 @@ serde = "~1.0.8"
 serde_derive = "~1.0.8"
 serde_json = "~1.0.2"
 tiny-keccak = "1.1"
-
-[dependencies.grin_grin]
-path = "./grin"
-version = "*"
-default-features = false
-#Comment this in to use the cuckoo-miner package
-#ensure cuckoo-miner is cloned next to the 
-#grin directory
-#features = ["cuckoo_miner", "use-cuckoo-miner"]
 
 
--- a/chain/Cargo.toml
+++ b/chain/Cargo.toml
@@ -19,3 +19,8 @@ secp256k1zkp = { path = "../secp256k1zkp" }
 [dev-dependencies]
 env_logger="^0.3.5"
 rand = "^0.3"
+
+#just to instantiate a mining worker during a test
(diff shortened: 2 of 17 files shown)


yeastplume:
Just a note, CI tests are failing because I've gone and changed the TEST_SIZESHIFT value to 16 and they're taking too long to mine a chain using the default miner. Can either change this back to 12 or enable cuckoo_miner by default, but I thought it better to use a minimum cuckoo_size that all implementations can handle

gazhayes:
I haven't looked at the test so this might be a dumb idea but could you just test block verification using the same library instead of trying to mine?

ignopeverell:
@yeastplume I'd lean toward enabling the optimized cuckoo miner by default, the more mileage this gets the better

yeastplume:
@ignopeverell that's absolutely fine by me, I've changed the config file to use cuckoo-miner by default and updated the tests to use cuckoo-miner as well.

One issue remaining with CI though, by default cargo runs all tests in parallel, which is fine when everything is static, but in this case many threads are trying to load and unload the same DLL and symbols all at random times causing the segfault error message at the end of the tests.

I'd considered wrapping the dll and symbol loading and function access within cuckoo-miner in a mutex locked structure, but I don't think this desirable, because a) the mining plugins themselves should keep control of how many threads they want to open or close, b) having loads of threads in the client (in this case grin) competing for the 'mine' function of a plugin and waiting for mutexes to unlock would just slow all mining to a crawl, and c) normal grin operation shouldn't involve many threads doing the same thing in the way that all of the multithreaded tests do.

It's also a bit awkward to write tests in there because of conflicting ports, etc. It may eventually be more desirable to have a larger test harness that spawns different processes rather than trying to run many instances of grin in the same process, but for now could we consider adding RUST_TEST_TASKS=1 for the grin/grin CI tests to ensure they're run sequentially rather than all in parallel?

PR #86, opened 2017-08-05
Ci mode fixes

yeastplume:
For... https://github.com/ignopeverell/grin/issues/85 

Just testing speed of build, for now. Not sure I like the Global RWLock approach, but not sure there's a better way without awkwardly passing parameters all over the place.

Also, since proof size is now dynamic, need to change the p2p server to reject blocks with proofs that don't match the current size being run.


yeastplume:
Okay, this seems to work fine, the internal tests work very quickly and accurately when everything is put into CI mode but there are a few things I'd like to discuss a bit before I advocate merging it:

-It's using a 'singleton-esque' global static to set the cuckoo parameters (proofsize and cuckoo size).. the reason I've done it this way is because there were many places in the code where the proofsize was considered static. I've changed it to be dynamic, but there are a lot of places where there is no real configuration object, or functions where calls aren't instanced alongside any configuration object (like pipe.rs), and getting that value into there would have meant a lot of hacking quasi-unrelated arguments along chains of function calls and/or creating configuration objects. The global.rs more or less acts as a global configuration object that all modules can access (and contains non-consensus values, to keep consensus.rs uncluttered). These parameters will be share across all threads spawned by the process, which might make testing some cases difficult, but in practice I don't think it will be a problem. Interested in hearing opinions here.

-I had to change the POW serialisation in the header to include the POW length, which has the side effect that the server will consider blocks with any POW size valid. To counter this, I've just added a check to the block validation to reject blocks with POW sizes that don't match what's currently running. Are there any other implications/issues with doing this?

ignopeverell:
Thanks a lot for the patch, very helpful in the short term but will also save us a lot of time in the long run. Addressing your 2 points:

* The static globals are a little ugly but as you said, I don't think there's a much better way. I have a minor nit with naming, where `get_global_proofsize` in a `global` module is a little redundant. I think `global::proofsize()` and `global::sizeshift()`, using the module name, is a little more Rustish.
* I'm not sure I would do this. I don't see the advantage, as we'll never need to read/write headers with different proof sizes than what a server is running with. Why not use the globals again there?

yeastplume:
Point 1, yes agree, had the same thoughts myself but was more focused on getting it working, will change.

On the second, since I'd changed the internal representation of a proof to a variable length vector instead of a known-size byte array, it seemed more consistent/flexible to store the length when de/serialising. But you're right I can't think of a particular case either where you'd want to support variable length PoWs in a running blockchain, except perhaps in testing cases where you might want to blast a blockchain into existence quickly with automated test parameters then perform validation/pruning later using a size closer to production values? No issues changing it to use the global values instead, just thought it might give some flexibility later.

ignopeverell:
I just don't see us changing the solution length without a hard fork. And I'm worried about possible weaknesses that can open. For example I can pass MAX_U32 as proof size in the current code.  This would have to be caught by the serialization. The other thing is that a header has the convenient property of being fixed size right now. I don't think this is worth making it variable sized.

yeastplume:
Yep, understood, will change accordingly.

ignopeverell:
Thank you!

yeastplume:
Should be good to go now

PR #87, opened 2017-08-07
cleanup build warnings

antiochp:
Took some time to clean up some `cargo build` warnings - let me know if you think is worth spending a bit of time on. 

I'm taking some time to get familiar with the code and would like to contribute if I can.

Ideally these will all run cleanly without warnings - 

- [x] `cargo build -p grin_api`
- [x] `cargo test -p grin_api`
- [x] `cargo build -p grin_chain`
- [x] `cargo test -p grin_chain`
- [x] `cargo build -p grin_core`
- [x] `cargo test -p grin_core`
- [x] `cargo build -p grin_p2p`
- [x] `cargo test -p grin_p2p`
- [x] `cargo build -p grin_pool`
- [x] `cargo test -p grin_pool`
- [x] `cargo build -p grin_config`
- [x] `cargo test -p grin_config` <- not actively in use?
- [x] `cargo build -p grin_grin`
- [x] `cargo test -p grin_grin`
- [x] `src`
- [x] `util`
- [x] `wallet`

PR #91, opened 2017-08-09
POW Blake 2 changeover

yeastplume:
Just a quick one to change all POW calculations/verifications to blake2, to keep in sync with John Tromp's repository. Updated to use a newer version of cuckoo miner, which contains updated plugins using blake 2 instead of SHA256 for key generation.

Also a tweak to genesis block creation, starting it at the magic nonce depending on the mining parameters, to prevent having to wait around for the genesis block POW to be discovered.


ignopeverell:
Sorry, I applied #87 first, a general build warnings cleanup PR, and naturally it created multiple conflicts with other PRs. Can I get a rebase?

yeastplume:
No worries, rebased.

Issue #27, opened 2017-01-09
Ensure new coins can't be spent for X blocks


ignopeverell:
We need to prevent miners from being able to spend their reward for enough blocks so that they can't benefit from forking the chain. A sensible default may be 1000 blocks, which would be 10 times Bitcoin's (given that our block time is 1/10th).

JohnnSebastian:
i have been staring at the code for a while now ...
   do you follow a grand design ?

ignopeverell:
First part of this implemented in 38d5d67e, by explicitly flagging coinbase outputs.

antiochp:
Question about this -
I see we have `OutputFeatures` (for txn outputs) and `KernelFeatures` for (txn kernels), identifying an output and a transacton that spends a coinbase.
What stops these bits being modified or altered during use? If a miner simply flipped the "this is a coinbase" bit then would this not defeat any attempt at preventing it from being spent?

Would signing the `features` alongside the `fee` on a transaction eliminate this?

ignopeverell:
A miner can't mess with these bits without invalidating the coinbase sums of outputs and kernels:

https://github.com/ignopeverell/grin/blob/master/core/src/core/block.rs#L418

If you want to convince yourself of it, adding a unit test would be great :-)

antiochp:
I see some discussion around a possibly related idea in #25? 

Feels like these two requirements "prevent spending of coinbase for n blocks" and "time locked transactions" may be examples of a more generalized solution somehow - like tracking a "minimum block height before spendable" on outputs of a transaction. Maybe every transaction has this (and normally this is the current block height)?

antiochp:
Ah got it - on the block containing the coinbase itself we check that the coinbase output (presumably a single output normally?) and the coinbase transaction kernels are all consistent.

I think my question was poorly written - I'm thinking more in terms of what happens when that coinbase output becomes an input into a subsequent transaction, say in the next block vs. 1001 blocks later.
My initial rough idea was to track the features on the input itself, based on the features from the underlying output. There would be nothing to prevent these bits from being modified though or am I missing something?

(And yes a unit test here is something I'd definitely like to put together to convince myself of some of this...)

ignopeverell:
No, you have to track the features on the output the input is referencing. If the output is a coinbase, then you check the block distance.

antiochp:
Throwaway spike/experiment PR up here - https://github.com/antiochp/grin/pull/1
[...]

PR #99, opened 2017-08-21
unit tests for coinbase flags (output and tx kernel)

antiochp:
* add unit test for what an empty block looks like (coinbase only)
* add unit test for invalid flipping of `COINBASE_OUTPUT` flag on txn output
* add unit test for invalid flipping of `COINBASE_KERNEL` flag on txn kernel
* add unit test for serialize_deserialize of a minimal Block

TODO - investigate why the block header does not serialize/deserialize consistently (UTC related?)


antiochp:
Related: #27

PR #101, opened 2017-08-22
Tests for input output commitments

antiochp:
This PR includes the following - 

* fixup `grin_wallet` tests 
  * add `rustc-serialize` to Cargo.toml in wallet dir
  * zero'd out fingerprint was causing test to fail
* add new test to verify equality of input and output commitments when spending a coin

That last point was not immediately obvious to me. With hindsight it makes more sense but it is still far from intuitive that the commitments are actually _the same_ without writing it down here as a test.


ignopeverell:
Couple issues:

* I would prefer not to check in Cargo.lock yet. I understand that it's better to do that eventually, but right now we don't have any "release" to speak of and dependencies are still evolving quite regularly. So until we reach a more stable stage, it's much more convenient without a committed Cargo.lock.
* I'm happy to take multiple test fixes in a single PR, but would rather no have mixed test fixes with build environment changes.

antiochp:
@ignopeverell makes sense to me - will revert the `Cargo.lock` changes.
In terms of build env changes - you want the `travis.yml` changes in a separate PR?

ignopeverell:
That's still test-related to that's fine.

ignopeverell:
Just saw your other PR. That works too :-)

PR #104, opened 2017-08-25
Store block header hash by input and output commitment

antiochp:
This PR adds a new index to the chain store - storing block_headers (actually the block header hash) by output commitment.
This allows lookup of the block header by output commit for the purpose of verifying the block height is valid when spending a coinbase output (see #27).

* Index in store to allow block_header lookup by output commitment
* Check current chain state via the `get_header_by_height()` to double check the block_header is what we think it is
* added some test coverage around `get_block_header_by_output_commitment`

NOTE: this index is not exhaustive - we cannot guarantee every block header is in the index. We can (I believe) guarantee this for a recent period of time (2 months?).
This time period *is* long enough to verify a coinbase block height.

Q) Does checking the hash of the block_header stored via `get_header_by_height()` give us enough of a guarantee to be sure the current chain state is correct in `get_block_header_by_input_commit()` from `get_unspent()`? Is this even necessary?

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -211,26 +211,35 @@ impl Chain {
 		}
 	}
 
-	/// Gets an unspent output from its commitment. With return None if the
-	/// output doesn't exist or has been spent. This querying is done in a way that is
-	/// consistent with the current chain state and more specifically the
-	/// current branch it is on in case of forks.
+    /// Gets an unspent output from its commitment. With return None if the
+	/// output
+	/// doesn't exist or has been spent. This querying is done in a way that's
+	/// constistent with the current chain state and more specifically the
+	/// current
+	/// branch it is on in case of forks.
 	pub fn get_unspent(&self, output_ref: &Commitment) -> Option<Output> {
-        // this will only return a block header based on the current chain state
-        // as get_block_header_by_input_commit looks at the block heights
-        let header = self.store.get_block_header_by_input_commit(output_ref);
-
-        match header {
-            Ok(_) => None,
-            Err(NotFoundErr) => {
-                let output = self.store.get_output_by_commit(output_ref);
-                match output {
-                    Ok(out) => Some(out),
-                    Err(_) => None
-                }
-            },
-            Err(_) => None
-        }
+		// TODO use an actual UTXO tree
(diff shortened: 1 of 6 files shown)


antiochp:
@ignopeverell thoughts/feedback on this? Is this a reasonable approach to help with #27?

ignopeverell:
I'm sorry to catch this only now, but for I'm not sure what get_block_header_by_input_commit is for.
The output version I'm guessing you'll reuse to check the age of an output referenced by an input
but it looks like you're reusing the input version to check if something is spent, and the UTXO tree will be a much better fit for that (there was actually a comment to that effect in the code).

antiochp:
The more I think about this the more I feel like something is missing here.
With this approach the "block header by output commit" index is *only* populated when a node stores a block, this need to be a full block with associated inputs and outputs - so any node that does not process the block will not have this index populated for this specific coinbase.
Maybe I'm missing something here but I don't see a way for all nodes to receive enough sync'd data, in particular nodes that receive the "compact blockchain" as described in the whitepaper.
The coinbase output will be in the UTXO set but I don't see any way of knowing which block it originated in without processing that full block on that node?

So I took another pass through the whitepaper (it is slowly making a bit more sense now) and noticed this - 

```
Definition 7. A Mimblewimble transaction is the following data:
• A list of homomorphic commitments termed the inputs, ...
• A list of homomorphic commitments termed the outputs, ...
• A blockheight x.
• An excess commitment to zero, ...
```

That third point explicitly calls out the blockheight.
Is there any way of including the blockheight on a coinbase output directly (in addition to the features etc.)?

Or am I thinking through this wrong - and there *is* a way of indexing a block_header by a given output commitment without a node processing a full block?

antiochp:
Not an issue. Even with a compact chain sync we would sync "recent two months" of full blocks. And this would cover the period that coinbase outputs are locked up for.
So we can index block_headers by output commit for all outputs that we have seen.
If we don't find a block_header for a given output header then its either - 
* not actually an output, or
* older than 2 months (or whatever the cutoff is)

ignopeverell:
Practically, on most nodes we'll even remove the output related entries when they are spent as they're not needed by the node anymore. So yeah, not an issue.

PR #105, opened 2017-08-25
Use UTC in timestamps and fix broken header test

seamusfinnigan:
babb6110c4a7014a631239420b44eba4f2c3b848 by @antiochp introduced a new TODO to fix a failing test of block headers. This PR implements the proper fix and also addresses the large cause.

The fix is two part:
1. The standard seems to be that timestamps are assumed to be at 0 nsec (which seems reasonable) but was being stored with non-zero values for `tm_nsec` in `time::Tm` objects.
2. `time::at()` was being called instead of `time::at_utc()`, and as such the time was not in UTC (as suggested by the TODO).

The second issue appears elsewhere in the code base as well, and so this PR addresses those instances also. In some cases, this fix will not be important in practice, but should be fixed nonetheless for uniformity.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -130,7 +130,7 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		return Err(Error::InvalidBlockTime);
 	}
 	if header.timestamp >
-		time::now() + time::Duration::seconds(12 * (consensus::BLOCK_TIME_SEC as i64))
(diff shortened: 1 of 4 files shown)


ignopeverell:
Thank you for those fixes. All timestamps should normalize on UTC.

antiochp:
:+1:

antiochp:
@ignopeverell did you mean to merge this? I think the PR just got closed?

ignopeverell:
My bad, I must have been too tired when doing this, I closed instead of merging. Should be good now.

seamusfinnigan:
All good. Thank you for speaking up @antiochp. Will see how and where else I can help.

PR #106, opened 2017-08-27
run rustfmt on chain

antiochp:
I see we have `rustfmt.toml` in various dirs.

This is the result of running `cargo +nightly fmt` on the `chain` dir.

Q) Is this something we want to be doing? Regularly? And is it something we may want to consider automating somehow?

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -28,7 +28,7 @@ use pipe;
 use store;
 use types::*;
 
-use core::global::{MiningParameterMode,MINING_PARAMETER_MODE};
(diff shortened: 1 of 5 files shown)


ignopeverell:
It'd be great to automate. I try to run it regularly but it tends to drift here and there with more patches and PRs. Also a new version of rustfmt was released recently, with some changes in styles, so practically now it should be run on all modules. Would you mind checking the best way to automate this?

antiochp:
See #107 for one possible solution (not mergeable as there are a lot of rustfmt issues).

antiochp:
Closing this PR - about to reopen a replacement with `rustfmt` run across all modules.

PR #107, opened 2017-08-27
run rustfmt from travis

antiochp:
One possible approach to "automating" `rustfmt` - run it via travis and fail the build if anything is not correctly formatted.
i.e. to keep things clean you would need to run rustmft locally before pushing to github.
This *may* be preferable to attempting to fully automate the running of `rustfmt`?

See - https://github.com/rust-lang-nursery/rustfmt#checking-style-on-a-ci-server

* `rustup install nightly` (needed for `rustfmt`)
* install `rustfmt-nightly`
* run `rustfmt` as part of the travis script

This causes the build to fail if `rustfmt` returns with anything other than 0.
But the rest of the build script will continue to execute (so we will still see test failures etc.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,4 +26,12 @@ env:
     - TEST_DIR=wallet
     - RUST_TEST_THREADS=1 TEST_DIR=grin
 
-script: cd $TEST_DIR && cargo test --verbose
+install:
+  - rustup install nightly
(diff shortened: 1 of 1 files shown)


antiochp:
See #108 for (significantly larger) PR that also cleans up all modules via a full run of `rustfmt`.

antiochp:
Closing this PR - investigating possibility of a local git hook to run rustfmt pre-commit.

PR #108, opened 2017-08-28
Rustmt cleanup all dirs

antiochp:
* run `rustfmt` automatically from travis
* full `rustfmt` pass across all modules so the build is clean
* disable `error_on_line_overflow_comments` in `core` to handle big url in comment
* manually fixup some lines that exceed 100 chars

screen shot 2017-08-28 at 12 05 55 pm

Note: This is against the current (as of right now) `rustfmt` version - 
```
cargo +nightly fmt -- --version
0.2.2-nightly ( )
```

One complication here might be the nightly status of rustfmt - it does change regularly, and from what I can tell the formatting rules do appear to change regularly as rustfmt gets tightened up - so we would need to be ok with this failing the build at random times...


antiochp:
Closing for now - separate PR up for the git pre-commit hook
and if we go with the pre-commit hook this can be an incremental cleanup (modified files).

PR #109, opened 2017-08-28
Switch commitments in libsecp fork

yeastplume:
This adds switch commitment support to our fork of libsecp as outlined in mailing list discussions. It should be okay to include this functionality regardless of whether we choose to use it.

The operation itself is a simple operation, just multiplying a key using generator point 3, which I've defined for the time being as sha256(sha256(g)), where g is the uncompressed DER representation of generator G. I've added a SageMath script into secp256k1zkp/depend/secp256k1-zkp/sage/generators.sage that shows the process for deriving both of our 'extra' generator points, H and J, which should be reviewed carefully. I think it's comfortably NUMS, but this needs independent verification. I've just refactored code from the pedersen commit slightly to allow the multiplication against another generator (as the internal context structure is exactly the same), so hopefully all should be well, but review is advised.

There's a (commented out) example of how it would be called in core/core/block.rs around line 460, but it's not being called from grin code as of yet. The further hashing of the switch commit value with blake2 would be handled on the grin side and that value added to the output struct.


core/src/core/block.rs
```diff
@@ -457,8 +457,11 @@ impl Block {
 		let msg = try!(secp::Message::from_slice(&[0; secp::constants::MESSAGE_SIZE]));
 		let sig = try!(secp.sign(&msg, &skey));
 		let commit = secp.commit(REWARD, skey).unwrap();
+		//let switch_commit = secp.switch_commit(skey).unwrap();
 		let rproof = secp.range_proof(0, REWARD, skey, commit);
 
+		//println!("{:?}", switch_commit);
+
```
ignopeverell: Maybe a little more cleanup required? :)

secp256k1zkp/depend/secp256k1-zkp/sage/generators.sage
```diff
@@ -0,0 +1,34 @@
...
+b = 7
+F = FiniteField(p)
+
+o = '%x' % (EllipticCurve ([F (a), F(b)]).order())
+print o 
+
+gen2 =  hashlib.sha256('0479be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8'.decode('hex'))
+gH = EllipticCurve ([F (0), F(7)]).lift_x(int(gen2.hexdigest(),16))
+print 'gH: %s' % gH
+gen_h = '%x %x'%gH.xy()
```
ignopeverell: Doesn't look like this is used anywhere, are you keeping it just for documentation/context?

secp256k1zkp/src/pedersen.rs
```diff
@@ -199,6 +199,20 @@ impl ::std::fmt::Debug for RangeProof {
 }
 
 impl Secp256k1 {
+
+/// Creates a switch commitment from a blinding factor
```
ignopeverell: Minor indentation issue.

ignopeverell:
I have a few very minor comments but beside that the patch looks good to me. I'm wondering whether @apoelstra would also want it for his secp256k1-mw fork.

We can discuss the integration of this into our outputs separately after.

core/src/core/block.rs
yeastplume: :0

secp256k1zkp/depend/secp256k1-zkp/sage/generators.sage
yeastplume: This is just for documentation/reference. The point is supposed to be provably NUMS so I thought it best to have the code that generates it there for reference/review by any interested parties.

yeastplume:
Okay, thanks, should be good to go now. I won't actually barge ahead with putting them into our outputs until we've discussed and decided whether we actually want to.

PR #110, opened 2017-08-29
git pre-commit hook to run rustfmt on modified files

antiochp:
This PR adds a git pre-commit hook file in `hooks/pre-commit`.

To enable this you need to copy this file across to the `./.git/hooks` dir
and check it is executable (it should be).

```
cd .git/hooks
ln -s -f ../../.hooks/pre-commit ./pre-commit
```

**Note**: this does not fail or halt `git commit`. 
It displays out a list of files that should be formatted to clean them up as a suggestion.
It only looks at modified files in the commit so cleanup will be incremental over time.

Assumes rustfmt-nightly installed via rustup -
https://github.com/rust-lang-nursery/rustfmt#installation
```
rustup update
rustup install nightly
rustup run nightly cargo install rustfmt-nightly
```

Example output if rustfmt is not installed as expected - 

screen shot 2017-08-29 at 2 49 53 pm

Example output if rustfmt exits with non-zero - 

screen shot 2017-08-29 at 2 51 25 pm

example output if everything looks good - 

screen shot 2017-08-29 at 2 54 23 pm

TODO - 
- [x] check this works for other devs
- [x] add some docs around initial setup above


antiochp:
Hey @ignopeverell @yeastplume - mind checking this out locally to see if it works for you?

yeastplume:
I've just tested it locally, it seems to work and I like this approach, it's not too intrusive and serves as a gentle reminder to run rustfmt instead of making it a hard rule. I'll definitely take note and run rustfmt when checking in using this approach.

Just a couple of comments:
-I'm not sure I like having a 'hooks' directory in the root that's not a rust crate, because it's incongruous with everything else, but also acknowledge there's no immediately more obvious place to put them. Perhaps in a hidden directory prefaced with a '.', which will make it more evident that it's configuration/build related.

-There appears to be a core.hooksPath git setting, https://git-scm.com/docs/git-config#git-config-corehooksPath.. perhaps this might be a better way of configuring these rather than a manual copy, which is liable to fall out of sync if any of the hooks change? Perhaps just a one line shell script that sets this variable once can be created, so the setup becomes running a command a single time then forgetting about it.

antiochp:
Thanks for the feedback - good comments on both points.
1) Agreed - the `hooks` dir feels wrong - I'll look around and see what other repos are doing but `.hooks` makes a lot of sense (particularly if we can leverage `core.hooksPath`).
2) This is great - I'll work on getting this to work.

antiochp:
Path is now `.hooks` in the repo to keep the root dir clean.

I'm actually leaning against automating the installation of the git pre-commit hook too much - I think there's too much trust put on the script by the dev (I know would be wary of running an "install" script that messes with my git config).

I think we (at least initially) should just aim to document this clearly and have the dev create the symlink themselves so they are comfortable with what is going on.

ignopeverell:
I've been meaning to add a CONTRIBUTING file at the root with general instructions, I think that would be a good place to strongly nudge people toward using the hook.

antiochp:
Sounds good - I can move the existing "contributing" section out of README and link to it.

PR #111, opened 2017-08-30
Cannot spend coinbase for N blocks

antiochp:
* introduce core errors in core/types.rs - this is WIP (not sure where to put these)
* map errors in adapters (chain error -> core error)
* use `get_block_header_by_output_commit` and `head_header` in `add_to_memory_pool`
* introduce `COINBASE_MATURITY` const in consensus (currently `10` for testing/dev)
* introduce `PoolError::ImmatureCoinbase`
* return `output` in `Parent::BlockTransaction`
* add same coinbase maturity check to `validate_block`
* test `process_block` with unique `reward_key` per block (not enforced but by convention for it to work correctly)

TODO - 
- [x] add similar consensus check when adding a new block - `validate_block()` I think?
- [x] confirm we are not off-by-1 on block validation vs. txn validation rules
- [x] test coverage around `add_to_memory_pool` to cover `ImmatureCoinbase` path
- [x] test coverage around `validate_block` to cover `ImmatureCoinbase` path
- [x] we do not currently prevent "duplicate" coinbase outputs (to be revisited)
- [x] reduced the COINBASE_MATURITY down from `1_000` to `10` for testing (revisit later)


chain/src/chain.rs
```diff
@@ -275,10 +279,12 @@ impl Chain {
 	}
 
     /// Gets the block header by the provided output commitment
-    pub fn get_block_header_by_output_commit(&self, commit: &Commitment) -> Result<BlockHeader, Error> {
-        self.store.get_block_header_by_output_commit(commit).map_err(
-            &Error::StoreErr,
-        )
+    pub fn get_block_header_by_output_commit(&self, commit: &Commitment) -> Option<BlockHeader> {
```
ignopeverell: I'm not a huge fan of the error suppression here. Consumers should know when an output commitment is absent (i.e. spent) vs a data store read error. Errors can be converted by the adapter for the pool. I understand you probably try to follow the signature of get_unspent but that 's not something I had thought through for too long (meant it as temporary :-))

chain/src/pipe.rs
```diff
@@ -167,15 +168,29 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
...
-	try!(b.validate(&curve).map_err(&Error::InvalidBlockProof));
-
-	// TODO check every input exists as a UTXO using the UTXO index
+	try!(block.validate(&curve).map_err(&Error::InvalidBlockProof));
+
+	for input in &block.inputs {
+		// TODO check every input exists as a UTXO using the UTXO index
+
+		// now check that any coinbase inputs have matured sufficiently
+		if let Ok(output) = ctx.store.get_output_by_commit(&input.commitment()) {
```
ignopeverell: I'd reverse this condition and the next. Calling get_output_by_commit first incurs a db cost that's useless for non-coinbase.

chain/src/types.rs
```diff
@@ -57,6 +57,8 @@ pub enum Error {
 	InvalidBlockTime,
 	/// Block height is invalid (not previous + 1)
 	InvalidBlockHeight,
+    /// coinbase can only be spent after it has matured (n blocks)
+    ImmatureCoinbase,
```
ignopeverell: rustfmt? :-)

grin/src/adapters.rs
```diff
@@ -295,4 +296,12 @@ impl pool::BlockChain for PoolToChainAdapter {
 	fn get_unspent(&self, output_ref: &Commitment) -> Option<Output> {
 		self.chain.borrow().get_unspent(output_ref)
 	}
+
+	fn get_block_header_by_output_commit(&self, commit: &Commitment) -> Option<BlockHeader> {
```
ignopeverell: Same observation regarding the use of Option here. The pool::BlockChain trait should be modified to include some error type.

chain/src/types.rs
antiochp: Hmm - soft tabs or hard tabs?
Got my editor to show invisibles and it's hard tabs right?
(except the bitflags up top had spaces in them, which confused my editor...)

core/src/types.rs
```diff
@@ -0,0 +1,21 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// TODO - what should this be called
+/// TODO - where should this live?
+#[derive(Debug)]
+pub enum CoreError {
+    /// TODO - rename this
+    GenericCoreError,
```
antiochp: I added this as a placeholder. We need a set of errors that can be shared between `pool` and `chain` and I wanted to avoid requiring one from the other.
1) This needs renaming and/or making more specific
2) Not sure where to put this (so put it in `types.rs` for now)

chain/src/pipe.rs
antiochp: We need to call get_output_by_commit first, as we need the output to know if it is a coinbase output or not (unless I'm misunderstanding what you are suggesting).

ignopeverell: Yes, ignore me, mind fart :P

core/src/types.rs
ignopeverell: I think we can have the adapter do the translation. So the chain throws its error. The pool trait requires its error as well. And the adapter implements the trait and translate the chain error into the pool one.

pool/src/types.rs
```diff
@@ -88,17 +89,38 @@ pub enum PoolError {
...
+        header: block::BlockHeader,
+        /// The unspent coinbase output
+        output: Commitment,
+    },
     /// An orphan successfully added to the orphans set
     OrphanTransaction,
+    /// TODO - wip, just getting imports working, remove this and use more specific errors
+    GenericPoolError,
+    /// TODO - is this the right level of abstraction for pool errors?
+    OutputNotFound,
```
antiochp: Alternatively - do we just want a more generic `ChainError` here to represent something going wrong with accessing the chain?

ignopeverell: I'd stick with the more specific errors as long as there are only a few of them.

core/src/core/block.rs
```diff
@@ -307,28 +307,37 @@ impl Block {
...
-			.map(|&out| out)
-			.collect::<Vec<Output>>();
+        let in_set = self.inputs
+            .iter()
+            .map(|inp| inp.commitment())
+            .collect::<HashSet<_>>();
+
+        let out_set = self.outputs
+            .iter()
+            .filter(|out| !out.features.contains(COINBASE_OUTPUT))
```
antiochp: I believe we always want to exclude coinbase outputs (and corresponding inputs) from being compacted away.
If we ever compact away a coinbase output the block will no longer validate successfully as the outputs and the kernels will no longer be consistent.
So added this additional filter to ensure we always leave coinbase inputs and outputs untouched and uncompacted.

core/src/core/block.rs
```diff
@@ -429,12 +438,12 @@ impl Block {
 	fn verify_coinbase(&self, secp: &Secp256k1) -> Result<(), secp::Error> {
 		let cb_outs = self.outputs
 			.iter()
-			.filter(|out| out.features.intersects(COINBASE_OUTPUT))
+			.filter(|out| out.features.contains(COINBASE_OUTPUT))
```
antiochp: `intersect` here will not work as expected if we ever compare against a bitflag that consists of more than a single bit - we need it to be a full `contains` to match correctly.

antiochp:
@ignopeverell thoughts/feedback on how this is progressing?
What's outstanding from your perspective to get this finished?

Went into this thinking this would be a relatively minor change... definitely given me a good deep-dive into the existing txn and block processing code.

ignopeverell:
So what's the status on this? Are we giving it a little time to see if a solution that's also applicable to time locks can be found?

antiochp:
@ignopeverell I *think* this is good to go. As you mentioned in another thread we should be able to migrate this to a more generalized solution for time locks. But for coinbase specific locks I think this works pretty well.
Thoughts/feedback?

PR #112, opened 2017-08-31
WIP: Prunable MMR storage

ignopeverell:
Full storage of the prunable MMR sum-tree, including the main file store containing sums and hashes, the remove log and a prune list structure to keep track of what has been pruned already.


ignopeverell:
Pinging @merope07 for a review if you have some time. It's not done yet and needs to be integrated with the rest of the system but it's getting there.

PR #114, opened 2017-09-04
Range Proof doc and intro update

yeastplume:
Don't merge this as of yet, putting this here for review and discussion first.

I've created a document intending to serve as a primer on range proofs, which outlines what they are as well as provide the curious with an illustration why all private keys must be known to create a valid range proof for an output. I've also referenced this doc from the intro doc.

Please review and comment and point out anything I may have got wrong. There's also a single-question FAQ at the end, that I genuinely don't have an answer for at the moment :D

### DIFF
--- a/doc/rangeproofs.md
+++ b/doc/rangeproofs.md
@@ -1,8 +1,8 @@
 # Range Proofs - A Primer
 
-This document is intended to give the reader a high level conceptual overview of what range proofs are and how they're contructed. While understanding range proofs is not 100% necessary in order to understand Mimblewimble, a few points about them should be noted, even if just from a 'black box' perspective:
+This document is intended to give the reader a high level conceptual overview of what range proofs are and how they're constructed. While understanding range proofs is not 100% necessary in order to understand Mimblewimble, a few points about them should be noted, even if just from a 'black box' perspective:
 
-* Range Proofs are used to prove, with zero-knowledge (i.e. without revealing the amount or the blinding factor), that the value committed to in a given commitment falls within a certain range. The inclusion of a range proof in a Mimblewimble commitment demonstrates that the amount commmited to is positive, and therefore doesn't create any new money.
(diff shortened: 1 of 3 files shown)


ignopeverell:
The answer to your question comes from the hiding property of Pedersen commitments. Specifically in your example, for every `v` there is a `b` that will satisfy `C`.

doc/rangeproofs.md
```diff
@@ -0,0 +1,132 @@
+# Range Proofs - A Primer
+
+This document is intended to give the reader a high level conceptual overview of what range proofs are and how they're contructed. Understanding range proofs is not 100% necessary in order to understand Mimblewimble, however it is necessary to understand that the blinding factor used to create an output on the block chain must be known in order to create a valid range proof. An explantion of why this is the case follows through an elementary example of the creation of a range proof.
```
ignopeverell: Typo: explantion.

doc/rangeproofs.md
```diff
@@ -0,0 +1,132 @@
...
+C' = C - 1H
+```
+
+Following from the example:
+
+```
+C' = 113G + 1H - 1H
+```
+
+If a ring signature is provided over {C, C'}, it proves that the signer knows the private key for either C or C', but not both. If C was a commitment to zero, the signer was able to use C to create the ring signature. If C was a commitment to one, the signer was able to use C' to create the ring signature. If C is neither a commitment to one or zero, the signer wouldn not have been able to create the ring signature at all. However, no matter how the ring signature was created, the private key that was used to create it was not revealed. If a ring signature exists at all, C must have been either a commitment to zero, OR a commitment to one.
```
ignopeverell: Typo: wouldn

doc/rangeproofs.md
ignopeverell: Maybe also reiterate the main reason for a range proof to exist at all: proving with zero-knowledge the range of the value (otherwise there are simpler ways to prove ownership of the private key).

yeastplume:
I've included these changes and suggestions, and done another pass myself and made some further clarifications, refinements. Have another read at your leisure (or if anyone else cares to comment, please do).

PR #115, opened 2017-09-04
Adding kcov code coverage to chain and core modules

percytheprefect:
This pull request implements code coverage for the core and chain modules of grin. It follows [this guide](http://sunjay.ca/2016/07/25/rust-code-coverage) to implement the test coverage. 

If anyone knows of a better way to make code coverage easily accessible for grin let me know.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,12 @@ addons:
     packages:
       - g++-5
       - cmake
-
+      - libcurl4-openssl-dev
+      - libelf-dev
+      - libdw-dev
+      - cmake
+      - gcc
+      - binutils-dev 
 env:
   global:
     - RUST_BACKTRACE="1"
@@ -28,5 +33,41 @@ env:
     - TEST_DIR=keychain
     - TEST_DIR=core
     - TEST_DIR=util
+
+after_success: |
+  pwd &&
+  wget https://github.com/percytheprefect/kcov/archive/master.tar.gz &&
+  tar xzf master.tar.gz &&
+  cd kcov-master &&
+  mkdir build &&
+  cd build &&
+  cmake .. &&
+  make &&
+  sudo make install &&
+  cd ../.. &&
+  rm -rf kcov-master &&
+  cargo test --no-run &&
+  cd ../
+  pwd &&
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_core* && 
+  echo "Finished coverage for grin_core"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_chain* && 
+  echo "Finished coverage for grin_chain"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_keychain* &&
+  echo "Finished coverage for grin_keychain"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_p2p* &&
+  echo "Finished coverage for grin_p2p"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_pow* &&
+  echo "Finished coverage for grin_pow"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_store* &&
+  echo "Finished coverage for grin_store"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_wallet* &&
+  echo "Finished coverage for grin_wallet"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_util* &&
+  echo "Finished coverage for grin_util"
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/mine_simple* &&
+  echo "Finished coverage for mine_simple"
+  bash <(curl -s https://codecov.io/bash) &&
+  echo "Uploaded code coverage"
 
 script: cd $TEST_DIR && cargo test --verbose


.travis.yml
```diff
@@ -26,4 +31,29 @@ env:
     - TEST_DIR=wallet
     - RUST_TEST_THREADS=1 TEST_DIR=grin
 
+after_success: |
+  pwd &&
+  echo "HERE" &&
+  wget https://github.com/percytheprefect/kcov/archive/master.tar.gz &&
```
ignopeverell: Why did you need your own fork? Any other solution (like depending on some official release)?

percytheprefect: The guide has this line: 

> wget https://github.com/SimonKagstrom/kcov/archive/master.tar.gz

I think we want to be able to update kcov at will instead of waiting on `SimonKagstrom` to do it. If you want to fork the repo so that we can use your fork as our reference repository that is fine too.

sesam:
@percytheprefect Nice initiative.

- Guide is dated July 25th, 2016.
  - Could you try it out with the official repo instead of the SimonKagstrom fork?
  - Maybe you (or me) should ask the author for ideas what's since changed?
- Did I understand right, that this is something I'd use locally only? An alternative, with less overhead on each grin developer, could be to use some outside service. If we add a badge or "shield" in README.md (recent article https://medium.com/@Razican/continuous-integration-and-code-coverage-report-for-a-rust-project-5dfd4d68fbe5) that might induce more PRs that adds tests.
I've been looking at the article above. Now I also found what might be the most official commentary on how to track code coverage https://github.com/rust-lang/crates.io/issues/706 (dated May 2017).

I'm interested to work more on this. I'm sesam on grin's gitter chat.

sesam:
I found https://github.com/huonw/travis-cargo which probably does more or less the same, but may be even more outdated. The opensource way to solve this would be upstreaming fixes to @huonw or someone who can take over that burden.

I also found https://crates.io/crates/cargo-tarpaulin (linux x86-only):
````

apt-get update && apt-get install libssl-dev
cargo install cargo-tarpaulin
cargo tarpaulin -v # no success (tested on mostly vanilla Debian Jessie)
````

**Comment by sesam** [comment]
@codecov-io aboev has outdated links. The link to the "full report" is https://codecov.io/gh/mimblewimble/grin/pull/115 and that is still reporting "missing base report".
Maybe loggin in as the profile https://codecov.io/gh/mimblewimble and doing (something) can fix this?

**Comment by sdtsui** [comment]
@sesam :
Thank you for the links and research -- saved me some time getting oriented. :)

I think tarpaulin is a good option: actively maintained and aiming for functionality that `kcov` won't support. Unfortunately, it's not complete.  An `0.5.6` [release](https://github.com/xd009642/tarpaulin/releases) was pushed last week and the build is still failing. Here's the [original Reddit thread](https://www.reddit.com/r/rust/comments/6ohd9y/introducing_tarpaulin_a_code_coverage_tool_for/). What do you think of looking at it after a `1.0`, perhaps only if we need more coverage options (like condition/branch coverage)?

---

Regarding [`huonw/travis-cargo`](https://github.com/huonw/travis-cargo), the main negative I see is it's not actively maintained. Here's one of its most recent PRs, from a few months ago:

Most recent commits are >2yrs old.

**Comment by ignopeverell** [comment]
Hey @sdtsui, thanks for picking this up with @sesam. I would have no problem having our own fork of `kcov` but it sounds like `cargo-cov` is a better option now. Am I reading this right? The report you pasted looks pretty good to me and seems about right. Is there any publicly hosted option for `cargo-cov`? Could we somehow integrate it with `travis`?

**Comment by sesam** [comment]
I might try to convince @hendi once back from newyears busyness to set up a
cargo-cov on his machines

**Comment by sdtsui** [comment]
@ignopeverell  I did some more reading today, looking deeper into the different projects we could use.
- `cargo-cov` maintainer's progress on coveralls support:
https://github.com/kennytm/cov/issues/3
> Anyway this project (kennytm/cov) is currently very inactive because the coverage is useless when unused generic functions are simply ignored...
- also parity's solidity testing framework, `sol-rs` on this topic:
https://github.com/paritytech/sol-rs/pull/20
>Apparently there is an issue with Rust profiling picking up on generic functions (see this issue in cargo-cov). So this approach may work for some basic coverage statistics, but not a long-term / ideal solution by any stretch.

Hosted publically on crates.io: 
https://crates.io/crates/cargo-cov

I'm not sure if we should use this after all, especially given `kennytm`'s comment.  Let me know how you feel about losing coverage for unused generic funcs: could be a deal breaker. 

Also note that `kennytm` also has another project: [cargo-kcov](https://github.com/kennytm/cargo-kcov) that has <20k downloads (`kcov`, according to its github badge has 900k+ now), and is Linux only (no future macOS or other support).

@ignopeverell 
So perhaps what we have here (+forking `kcov`) is best? Individual devs can still use `cargo-cov` as-is. If so, I can take this commit and add changes (including the badge) in anticipation of a `mimblewimble/kcov` repo, wrap this up tomorrow.

Sound ok?
/cc @sesam

**Comment by percytheprefect** [comment]
I think we should choose the simplest solution. We do not want to allocate many resources to fiddle with code coverage. @sdtsui you will need to rebase your #545 onto this new commit. I have added coverage to other modules in the grin codebase. 

>mentioned adding a fuzzer using quickcheck -- I'd be interested in exploring this as well. Do we have the same concerns with ownership of forks? It feels like a separate project to me -- might it make sense as a separate issue?

This is a good idea. Grin's testing procedures have been lacking. Property based testing, in conjunction with a few hard coded test vectors, is a proven way to build high quality software. I'm guessing quickcheck will unearth some tricky bugs in our implementation. We should start with writing properties for our cryptographic primitives and serializers and work our way out from there. 

https://github.com/BurntSushi/quickcheck

**Comment by ignopeverell** [comment]
Ended up merging #545, please PR on top of master to add new modules, improvements, etc.

Definitely agree we should ramp up testing as much as possible and quickcheck has its place there.

File: .travis.yml [hunk 2]
```diff
@@ -29,4 +34,40 @@ env:
...
+  cd build &&
+  cmake .. &&
+  make &&
+  sudo make install &&
+  cd ../.. &&
+  rm -rf kcov-master &&
+  cargo test --no-run &&
+  cd ../
+  pwd &&
+  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_core* &&
```
sdtsui: @percytheprefect thanks for adding the other directories.

Are you sure these exist?  I have been trying to follow [the tutorial you linked](http://sunjay.ca/2016/07/25/rust-code-coverage) locally: after running `cargo test --no-run`, my `target/debug` dir looks like this:


I think this is why none of our builds are uploading coverage. When we run
`kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin-chain*`, there is no `grin-chain` executable.  Travis should behave the same.  I also tried a [few times](https://github.com/sdtsui/grin/pulls) with different configurations on my fork.

Issue #28, opened 2017-01-10
Determine wallet import/export format


ignopeverell:
One difficulty with MimbleWimble is that to find one's UTXOs, both the private key and the amounts must be known. There's plenty of solution for private key backup and exports, including HD, but amounts is a MW specific issue. Exporting a list of amounts is possible but ergonomically complicated, especially if we still want to support paper wallets.

Brute-forcing the value space is a possibility, but with 8 or 10 significant numbers could take a prohibitively long time. One improvement could be to have wallets restrict the ranges of amounts it handles and maybe enforcing it through the range proof. A better idea may be to combine a HD type root key with a small bloom filter (32 bytes for example) for values. Brute forcing the Bloom filter would be very fast, reducing drastically the space of values that have to be explored with each HD derivation.

JohnnSebastian:
Can UTXO get pushed back on to the visible chain ? So the set you spend is actually the ECC of (STXO+ key) and a Previous (ECCkey+key), again one input of a (ECCkey+key) etc where the above ECCkey is a (Private key +  (ECCkey+key)), public UTXO could be used to cryptographically lock previous ECCkeys , where one 'key' is derived from the wallet's virtual keyboard and UTXO are pushed back on the visible chain to motivate the spend of older UTXO.  Coinbase transaction can create a data set of IN and OUT points, there uniqueness protected with block chain block nr and the mercle tree. Miners verify the singularity of a coin questioning validity of the public UTXO hash (Embedded is the ECC key) against - true number of IN en OUT created on time stamp / block chain nr / amount of ECCkeys in block + mercle root. So a UTXO pushed publicly is 1unit of account and hash signature with additional hash value ...  ? Coins would become 1 unit of account - again with the possibility to submerse some math ?   Does this solve the problem ?

Niondir:
To get this clear: Let's assume I have my wallet A and a backup wallet B. Containing an address X (private key) with 2 coins.

Now I spend 1 coin using wallet A. I can no longer spend any coin from wallet B since it does not know the updated amount of address X?

ignopeverell:
Correct. And forget the bloom filter idea from above, it doesn't make anything more compact.

However, using denominations, I believe it's possible from the information in wallet B to find wallet A again. Assuming you're using HD derivation, A would have X1 and you know B would be using X2. Denominations reduce the search space enough that brute forcing all possible amounts to find the output from X2 should be doable.

apoelstra:
[...]

PR #123, opened 2017-09-12
WIP - support message in range proof (and recover value via rewind_range_proof)

antiochp:
The PR includes two things (might make sense to split it up) -
  * adds `message` support to `range_proof` (was previously just using zero'd out message)
  * uses `rewind_range_proof` to allow for recovery of the original unblinded output `value`

We use the secret key (blinding factor) originally used to generate the commitment as the `nonce` for the range proof - this "known key" allows the owner and only the owner to rewind the range proof and recover the value (and the message).

Related: #28 (wallet import/export format, specifically needing the privkey & value to recover wallet UTXOs)

Takes inspiration from Elements -
https://github.com/ElementsProject/elements/blob/8fed03f63ca2f46ea97e2f286a3bc46ce3087d6c/src/blind.cpp#L145
https://github.com/ElementsProject/elements/blob/8fed03f63ca2f46ea97e2f286a3bc46ce3087d6c/src/blind.cpp#L421

Some takeaways - 

The benchmarks indicate a single `rewind_range_proof` takes approx 14ms.
Without any way to identify UTXOs we'd need to try every UTXO with every possible secret key in the wallet - this is going to add up really quickly.

### DIFF
--- a/core/benches/output_value_recovery.rs
+++ b/core/benches/output_value_recovery.rs
@@ -1,0 +1,70 @@
+// Copyright 2017 The Grin Developers
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


antiochp:
Will revisit this once the new `rust-secp256k1-zkp` repo is integrated with grin - looks like the c changes in here are not necessary when we start using the updated version of `secp256k1-zkp`.

antiochp:
Closing this - will replace with PRs on `mimblewimble/secp256k1-zkp` and `mimblewimble/rust-secp256k1-zkp` once they are ready to go.

PR #132, opened 2017-09-23
expose "wallet info" on CLI

antiochp:
This PR adds a new `wallet info` subcommand to the CLI and fixes up various things 

* `grin wallet -p "password" info`
* add sleep and retry logic when obtaining wallet.lock
* reduce `COINBASE_MATURITY` down to 3 for testing
* cleanup maturity/lock_height handling in api
* rework logic around `Locked` status on output data (we were setting Locked on a cloned copy)
* fix url for `get_tip`
* map error correctly in `receive_json_tx`

```
grin wallet -p "password" info

Outputs - 
fc385d28, 1, 1, 4, Spent, 1000000000
d17777d9, 2, 2, 5, Unspent, 1000000000
b03b50c1, 3, 3, 6, Unspent, 1000000000
3d225151, 4, 4, 7, Unspent, 1000000000
eb68319f, 5, 5, 8, Immature, 1000000000
17669c25, 6, 6, 9, Immature, 1000000000
a937af5a, 7, 7, 10, Immature, 1000000000
49dc5efb, 8, 8, 0, Unspent, 200000000
44e24a8b, 9, 8, 0, Unspent, 800000000
e7980d0c, 10, 8, 11, Immature, 1000000000
5901ac54, 11, 0, 0, Unconfirmed, 1000000000
```

The `info` output is far from perfect but definitely useful when testing out sending txns.

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -29,7 +29,7 @@ pub const REWARD: u64 = 1_000_000_000;
 /// Number of blocks before a coinbase matures and can be spent
 /// TODO - reduced this for testing - need to investigate if we can lower this in test env
 // pub const COINBASE_MATURITY: u64 = 1_000;
-pub const COINBASE_MATURITY: u64 = 10;
+pub const COINBASE_MATURITY: u64 = 3;
 
 /// Block interval, in seconds, the network will tune its next_target for. Note
 /// that we may reduce this value in the future as we get more data on mining
--- a/api/src/endpoints.rs
+++ b/api/src/endpoints.rs
@@ -132,7 +132,10 @@ impl<T> ApiEndpoint for PoolApi<T>
 			.write()
 			.unwrap()
 			.add_to_memory_pool(source, tx)
-			.map_err(|e| Error::Internal(format!("Addition to transaction pool failed: {:?}", e)))?;
+			.map_err(|e| {
+				Error::Internal(format!("Addition to transaction pool failed: {:?}", e))
+			})?;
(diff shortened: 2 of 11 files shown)


antiochp:
Txn sending and receiving tested against a mining enabled node checking outputs.
Wallet info appears to be reliably doing the right thing for both coinbase and regular transaction outputs.
Wallet is not attempting to spend Immature coinbase outputs.
Wallet is not attempting to spend Locked outputs before they are confirmed as Spent.

core/src/consensus.rs
```diff
@@ -29,7 +29,7 @@ pub const REWARD: u64 = 1_000_000_000;
 /// Number of blocks before a coinbase matures and can be spent
 /// TODO - reduced this for testing - need to investigate if we can lower this in test env
 // pub const COINBASE_MATURITY: u64 = 1_000;
-pub const COINBASE_MATURITY: u64 = 10;
+pub const COINBASE_MATURITY: u64 = 3;
```
ignopeverell: We should track this somewhere so we don't forget it.

ignopeverell:
Looks like reducing the coinbase maturity broke the test you wrote on the tx pool :-)

core/src/consensus.rs
antiochp: Opened an issue to think about making this configurable based on mining mode - #133 
and to document this somewhere.

antiochp:
Fixed that test (at least they are good for something).

PR #137, opened 2017-09-25
add new CoinbaseReserved output status in wallet

antiochp:
Every time the miner process is started up we create a new coinbase output in the wallet.
This was really confusing as these were labeled "Unconfirmed" yet may never be confirmed.

This introduces a new `MinerReserved` output status to help differentiate between these miner reserved coinbase outputs and real unconfirmed outputs.

Also added a (hardcoded) `amount=1` miner fee to partial transactions created by the wallet to exercise the transaction fee handling code and to make outputs + change outputs more realistic during testing.

A basic txn generates - 
* output to receiving party
* change output to sender
* a fee (no output, burnt by the network)
* a coinbase reward output in the block containing the txn

Coinbase outputs now look like this in `wallet info` prior to the miner actually mining a block and successfully claiming the reward - 

```
grin wallet info
Outputs - 
fc385d28, 1, 0, 0, CoinbaseReserved, 1000000000
d17777d9, 2, 0, 0, CoinbaseReserved, 1000000000
b03b50c1, 3, 0, 0, CoinbaseReserved, 1000000000
3d225151, 4, 1, 4, Immature, 1000000000
eb68319f, 5, 0, 0, CoinbaseReserved, 1000000000
```

I can split this into 2 separate PRs if we want to keep the fee out of this for now.

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -460,9 +460,10 @@ impl Block {
 
 	/// Builds the blinded output and related signature proof for the block
 	/// reward.
-	pub fn reward_output(skey: secp::key::SecretKey,
-	                     secp: &Secp256k1)
-	                     -> Result<(Output, TxKernel), secp::Error> {
+	pub fn reward_output(
+		skey: secp::key::SecretKey,
+		secp: &Secp256k1
+	) -> Result<(Output, TxKernel), secp::Error> {
 		let msg = try!(secp::Message::from_slice(&[0; secp::constants::MESSAGE_SIZE]));
 		let sig = try!(secp.sign(&msg, &skey));
 		let commit = secp.commit(REWARD, skey).unwrap();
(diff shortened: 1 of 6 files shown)


ignopeverell:
I'm not sure this warrants a special status. All sorts of things can happen to an unconfirmed transaction that will lead to it never confirmed as well. It could even not make it to a transaction, the local computer can crash, in a multi-step exchange the other party may not finish the transaction, etc. So I see the coinbase output in a very similar way and it's handled in a very similar way. Does that make sense? Maybe we can solve this with better code doc?

The fee improvement should still be merged.

antiochp:
Ahh ok - that makes sense.

I'll clean this up to just include the fee improvements.

antiochp:
Closing - will reopen a PR to cover the `fee amount=1` change in wallet.

PR #138, opened 2017-09-25
[WIP] Replace rust-crypto with sodiumoxide/libsodium in wallet

antiochp:
This PR proposes a possible replacement of rust-crypto with sodiumoxide (rust bindings for libsodium)

https://github.com/dnaq/sodiumoxide
https://www.gitbook.com/book/jedisct1/libsodium/details

libsodium does not include ripemd160 so this PR introduces a dependency on https://crates.io/crates/ripemd160
to remove the remaining dependency on ripemd160.

Note: the published crate for sodiumoxide assumes libsodium is installed on the system.
Travis uses Ubunutu Trusty and there is no official package for libsodium on Trusty.
There is a PR open on sodiumoxide to build libsodium (similar to how we build secp256k1-zkp) via git submodule.
See - https://github.com/dnaq/sodiumoxide/pull/167

This PR specifies the following in Cargo.toml - 
```
sodiumoxide = { git = "https://github.com/oherrala/sodiumoxide", branch = "compile-from-git" }
```

If/when this is merged into sodiumoxide we would be able to use the published crate.
Alternatively we could fork sodiumoxide ourselves and manage the git submodule configuration on our own terms.

The tests in wallet have been intentionally left unchanged to verify these changes maintain current behavior.

We may want to consider moving away from sha256+ripemd160 for key identifiers? But this is far outside the scope of this PR.


wallet/Cargo.toml
```diff
@@ -8,7 +8,10 @@ authors = ["Ignotus Peverell <igno_peverell@protonmail.com>", "Laurent Meunier <
 byteorder = "1"
 log = "^0.3"
 rand = "^0.3"
-rust-crypto = "^0.2"
+ripemd160 = "~0.6.0"
+# TODO - use the crate once this get merged in (assuming it does)
+# sodiumoxide = "~0.0.15"
+sodiumoxide = { git = "https://github.com/oherrala/sodiumoxide", branch = "compile-from-git" }
```
antiochp: We probably want to either fork the official sodiumoxide repo into mimblewimble ourselves and manage it there or wait for the related PR to be merged in.
https://github.com/dnaq/sodiumoxide/pull/167

wallet/src/extkey.rs
```diff
@@ -205,31 +212,21 @@ impl ExtendedKey {
...
 		BigEndian::write_u32(&mut n_bytes, n);
+		let mut seed = self.key[..].to_vec();
+		seed.extend_from_slice(&n_bytes);
 
-		hmac.input(&self.key[..]);
-		hmac.input(&n_bytes[..]);
-
-		let mut derived = [0; 64];
-		hmac.raw_result(&mut derived);
+		let derived = hmacsha512::authenticate(&seed, &hmac);
```
antiochp: This is a good example of how sodiumoxide/libsodium api is more user friendly and less prone to error.

ignopeverell:
Before addressing this specific PR I just wanted to explain how I'd like to handle crypto library dependencies. At a high level I'd like them as narrow as possible. I much prefer having a single tractable dependency against one SHA256-only library that's optimized for that than those everything-crypto libraries. Also, I'd rather have as few crypto primitives we depend on.

Here's what we can't avoid:

* libsecp256k1-zkp - we're pretty much married to it and have our own extensions, we all know what it's for.
* blake2-rfc: that's the hash function we standardized on for core.

So then my first question would be, can we get rid of SHA256 and both rust-crypto and sodiumoxyde so we still have have the 2 above?

yeastplume:
I can get behind that logic. However, I don't think this can really be answered without more detail on what the final wallet is likely to look like. BIP38esque key derivation functions can obviously be handled with libsecp, but some sort of symmetric encryption will be needed as well, in the very minimal case to store the seed. That's also the part most vulnerable to side-channel attacks, so needs to be selected very carefully.

Pure rust libs are out for many reasons, so are there any standalone/rust friendly symmetric key encryption libs (AES or anything else suitable) with a solid, well vetted, C core?

JollyMort:
Check this out if you need ideas: https://github.com/monero-project/monero/pull/2257

The seed is a scalar encoded into base1626 number where each digit is a letter of the dictionary. Last word is a checksum. The passhphrase is hashed, EC added to the seed and encoded into base1626 and checksum word calculated. Both mnemonics can result in a wallet, and there's no way to tell whether a given seed is encrypted or not.

Maybe not everyone will agree, but I find it rather clever - especially since you can put some little cash on the wallet resulting from the encrypted seed to make the thief think that he found the actual stash.

antiochp:
Wallet code today uses - 
* `HMACSHA512` for key derivation
* `SHA256` for key identifiers
* `RIPEMD160` for key identifiers

Disclaimer: I am *far* from an expert on cryptographic primitives.
My initial thought was to suggest BLAKE2 as a potential replacement for our usage of SHA256 in wallet. But I decided against that because we needed an HMAC for key derivation.

But just been reading more about BLAKE2 and came across this - 
> BLAKE2 does not require
   a special "HMAC" (Hashed Message Authentication Code) construction
   for keyed message authentication as it has a built-in keying
   mechanism.

https://tools.ietf.org/html/rfc7693

I don't yet fully understand the implications of this - but it _sounds_ like we can use BLAKE2 as-is as a drop in replacement for both SHA256 and HMAC-SHA512 in wallet.

So to answer your question - I _think_ we can get rid of the dependency on rust-crypto and simply use `blake2-rfc` to achieve something very similar to what wallet does today for key derivation and key identifiers.
We would still need something for `ripemd160` - but as @apoelstra mentioned in 
> but also, it really doesn't matter who implements a hash function. you can't do it wrong

https://gitter.im/grin_community/Lobby?at=59c8fa88b59d55b8233583ee

So we could potentially use - 

* `BLAKE2b` for key derivation
* `BLAKE2s` + `RIPEMD160` for key identifiers (roughly analogous to BTC pubkeys/addresses but less useful)

This would definitely be beneficial if we could standardize on BLAKE2 for the one hash function across all of Grin.

antiochp:
@yeastplume - yeah symmetric encryption of the wallet seed was another reason to suggest libsodium as a possible lib.
It is a _very_ opinionated library in terms of the primitives it exposes - and provides for example `crypto_secretbox` (XSalsa20 + Poly1305) to handle authenticated symmetric encryption/decryption.
https://download.libsodium.org/doc/secret-key_cryptography/authenticated_encryption.html

So even if we decide to go down the path of just using BLAKE2 for hashing - at what point do we need to start thinking about symmetric encryption?
And asymmetric encryption primitives when we start thinking more about interactively building transactions across multiple wallets?
So is there going to be a need to re-introduce a crypto library then?

antiochp:
I do like the idea of having BIP-39 style mnemonic seeds at some point in the wallet impl for import/export assuming recovery of UTXOs can be solved.
But this can be handled separately to the key derivation and hashing right?

ignopeverell:
Yes about mnemonics being handled separately.

So yes, before going into picking a library, let's see what we need for the wallet. I don't think we need ripemd160, it's generally only to have shorter hashes and blake2 can output hashes of any size.

antiochp:
Sounds good - let me take a look at a possible approach of simplifying this and using blake2 exclusively in the wallet.
I can get a PR up for discussion shortly.

antiochp:
Opened #141 for discussion around a pure `blake2b` approach.

antiochp:
Closing - replaced with simpler usage of blake2 in #141.

PR #141, opened 2017-09-26
use blake2b for wallet key derivation

antiochp:
* use blake2b in wallet for key derivation and key identifiers
* remove dependency on rust-crypto

Original impl used ripemd160 and generated an identifier with digest length 20.
Do we want to stick with this or does another length make more sense (32 for example)?

This PR replaces #138 as an alternative solution.


wallet/src/extkey.rs
```diff
@@ -202,43 +205,30 @@ impl ExtendedKey {
...
-		let mut shres = [0; 32];
-		sha.result(&mut shres);
-
-		let mut ripe = Ripemd160::new();
-		ripe.input(&shres[..]);
-
-		let mut identifier = [0; 20];
-		ripe.result(&mut identifier);
-		Identifier::from_bytes(&identifier)
+		let identifier = blake2b(20, &[], &self.key[..]);
```
antiochp: Do we want to stick with `20` here to match original ripemd160 impl?

wallet/src/extkey.rs
```diff
@@ -202,43 +205,30 @@ impl ExtendedKey {
...
 		BigEndian::write_u32(&mut n_bytes, n);
+		let mut seed = self.key[..].to_vec();
+		seed.extend_from_slice(&n_bytes);
 
-		hmac.input(&self.key[..]);
-		hmac.input(&n_bytes[..]);
-
-		let mut derived = [0; 64];
-		hmac.raw_result(&mut derived);
+		let derived = blake2b(64, &self.chaincode[..], &seed[..]);
```
antiochp: Can someone double check the key and the msg are in the right places here in this call?

wallet/src/extkey.rs
ignopeverell: I think so, seems like a sane default.

wallet/src/extkey.rs
ignopeverell: Looks good to me

Issue #140, opened 2017-09-26
Proposal - decouple public wallet receiver from wallet private keys


antiochp:
Recent topics have got me thinking a bit about how the current wallet implementation could be improved.
Today we have a single "wallet receiver" that sits listening for http requests on a given port.
This process needs to be publicly accessible _and_ requires access to the wallet private key.

A sending wallet posts a partial txn to the receiver wallet (so receiver needs to be publicly accessible).
Schnorr signatures will change the flow but the wallets will still need to interact.

Ideally the process that has access to the wallet private keys is not publicly accessible and this suggests a slightly different topology - 

```
(wallet sender) --> [internet] --> (wallet receiver) --> (private wallet process)
```

* The `wallet receiver` listens on `0.0.0.0:13417`
* The `private wallet process` listens on `127.0.0.1:13416`
* A mining node talks to the `private wallet process` locally to build coinbase outputs.
* A wallet sender could talk to both the `private wallet process` (to build the partial txn) and the wallet receiver on the other side of the txn.

At the cost of additional complexity this would allow the process with access to the private key/seed to be locked down to only accept local connections, while allowing a more open wallet receiver to be run as a standalone process (effectively proxying requests to the private wallet).

Thoughts?

ignopeverell:
I agree with your separation, it's something that needs to be done.

Separately (and just because this seems like a good time to mention it), @apoelstra mentioned a trick with the range proofs where we can introduce both a spending key and a view key, both having different characteristics. It has impacts on the range proof size though.

Regardless, the change you propose is a good first step.

antiochp:
Yeah I remember seeing that mentioned somewhere (probably the mailing list).
I'm assuming this would be done down in the secp256k1-zkp lib?

yeastplume:
I agree with the need for this, but really don't like the idea of having yet another port to configure and manage.. we're now talking about having to manage two processes to run a simple server, and three to run a mining server, and having to lock them all down appropriately. Given MW's already a bit more cumbersome than usual to run due to the need for a wallet listener, I think we should actively aim to keep the setup complexity to a minimum.

Are there other ways to do ensure that the key is kept of the public-facing process without having to manage another port? Forking, execing, spawning a separate daemon process... I know we're working with rust as opposed to posix directly, and also have to consider the fact we'd potentially want this to run on windows one day, so another port might just be the only solution (unless we come up with something clever), but I'd still like to have a think about other ways of addressing this.

antiochp:
[...]

PR #146, opened 2017-09-29
keychain crate (no more secretkeys in core)

antiochp:
Edit: should be good to go - assuming the tests on travis pass this time.

TODO - 
- [x] get grin building
- [x] get all tests passing on travis
- [x] another cleanup pass (unused imports)
- [x] naming - `keychain`? (works for now)
- [x] naming - `Identifier` -> `PublicKey`? (works for now)

This PR introduces a new `grin_keychain` crate to explore the following - 
* Can we encapsulate all private key activity in a small component with limited surface area?
* Could this keychain component be subsequently moved out into a separate process in the future?

Specifically - 
* Can we separate `core` from any knowledge of `secp::key::SecretKey`?
* Can we separate `wallet` (i.e. a public facing wallet receiver) from `ExtendedKey`?

Idea: an `extkey::Identifier` can be thought of as roughly analogous to a public key in bitcoin.
Given an Identifier (pubkey) and a way to derive the corresponding `ExtendedKey` (and `secp SecretKey`) - can we instead pass around a reference to a `keychain` and the `pubkey`?

The keychain api is still taking shape in this PR but it looks approximately like the following (effectively a wrapper around secp + management of the SecretKeys ) - 

```
pub fn derive_pubkey(&self, derivation: u32) -> Result<Identifier, Error>

pub fn commit(&self, amount: u64, pubkey: &Identifier) -> Result<Commitment, Error> 

pub fn range_proof(
    &self,
    amount: u64,
    pubkey: &Identifier,
    commit: Commitment,
) -> Result<RangeProof, Error> 

pub fn blind_sum(&self, blind_sum: &BlindSum) -> Result<BlindingFactor, Error>

pub fn sign(&self, msg: &Message, pubkey: &Identifier) -> Result<Signature, Error>

pub fn sign_with_blinding(
    &self,
    msg: &Message,
    blinding: &BlindingFactor,
) -> Result<Signature, Error> 
```

There is still one place we need to pass a `secp::key::SecretKey` around and that is during txn construction (`sign_with_blinding` in the api above) - we send a partial txn from one wallet to another and this includes the "blinding factor" which is in practice a `secp::key::SecretKey`.

This PR introduces a `BlindingFactor` struct that encapsulates a SecretKey to make it explicit that this is what is being passed around.
This SecretKey does not exist in either wallet (so cannot be derived from a pubkey) as it is a sum of blinding factors involved in the construction of the transaction.

See #140 for some rationale on why I started down this rabbit hole...

I think if we can get to a point where both `core` and `wallet` interact with `keychain` via pubkeys and no longer pass SecretKeys around it will be significantly easier to extract `keychain` out into a separate secure process that the rest of Grin interacts with only when necessary.

Additional benefit - we have a lot less low-level interaction with the secp lib spread throughout the Grin codebase. The interaction points are mainly via `keychain` for txn and block creation. The majority of secp usage now is when verifying txn and blocks are valid.


antiochp:
Rebased against master and got a bunch of compile errors - working through these.

antiochp:
Hey @ignopeverell - thoughts/comments on this?
This gets us to a point where we can take `keychain` and put it in a separate process.
Neither `wallet` nor `core` need to access extended_keys or secret_keys directly and that secret_keys can remain exclusively in the keychain process.

antiochp:
cc @yeastplume and @apoelstra

ignopeverell:
Just had a look at the code changes, I like it. Cleans up a fair amount of stuff from core and makes our various key manipulations more elegant.

yeastplume:
Yes, I think this is a good first step to isolate the private keys. 

Should probably merge sooner rather than later due to the number of files changed.

antiochp:
@ignopeverell - if we're good with the naming (and everything else in here) then this is good to go.
All tests in travis are green.
I've done reasonably extensive testing locally - mining, sending coins to wallet etc.

PR #147, opened 2017-09-29
removed the $ symbol in the "export $PATH" example

marzig76:
running that as "export $PATH" will result in a "not a valid identifier" bash error.  I just removed the $ from the example.

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -56,7 +56,7 @@ from the build directory will run grin using the defaults in the grin.toml file,
 For the time being, it's recommended just to put the built version of grin on your path, e.g. via:
 
 ```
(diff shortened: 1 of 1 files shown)


BloddyBaron:
The export command is the wrong path. It should be be /path/to/grin/dir/target/debug as it needs to include the debug folder and not the grin executable or main grin folder either as it implies in the description. 

The correct command to add to PATH is therefore: export PATH=/home/username/directory/grin/target/debug/:$PATH or type `pwd` to see absolute path to the debug folder. 

Alternatively you can run: `cargo run help` instead of grin help after you add grin to the path.

ignopeverell:
I believe this has been fixed independently, but thanks for the report!

PR #148, opened 2017-09-30
Fix deprecated futures items

seamusfinnigan:
`futures` v0.1.15 deprecated two items used in Grin, replacing them with more verbose names to avoid confusion and simplify functionality. As a result, new installs of Grin grabbing the latest version of `futures` gave numerous warnings like the following:

    warning: use of deprecated item: renamed to `unbound_send`

and

    warning: use of deprecated item: implementation moved to `iter_ok`

This PR updates the minimum requirement for `futures` to the most recent release, v0.1.16, and takes care of these warnings without breaking functionality. I have also bumped the minimum version for `futures-cpupool`, primarily so both projects are synchronized.

More info on the relevant changes to `futures` here: https://github.com/alexcrichton/futures-rs/blob/master/CHANGELOG.md#0115---2017-08-24

### DIFF
--- a/grin/Cargo.toml
+++ b/grin/Cargo.toml
@@ -17,8 +17,8 @@ grin_pow = { path = "../pow" }
 secp256k1zkp = { git = "https://github.com/mimblewimble/rust-secp256k1-zkp" }
 
 env_logger="^0.3.5"
-futures = "^0.1.9"
-futures-cpupool = "^0.1.3"
+futures = "^0.1.16"
+futures-cpupool = "^0.1.16"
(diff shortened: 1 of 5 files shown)


seamusfinnigan:
It would appear Travis CL had issues using `futures` and `futures-cpupool` v0.1.16. I reverted `futures-cpupool` to what it originally set at, and bumped `futures` down to v0.1.15, where the deprecation issue was first introduced. This lets it pass all Travis CL checks but maintains the fixes introduced in the PR.

PR #156, opened 2017-10-05
wee warning cleanup + build doc updates

yeastplume:
Because loads of warnings imply the developers don't care, which is not the case in this instance. 

Edit: fixed, no warning now, except for an annoying 'redundant linker flag' thing, which I think is coming from rocksdb, so might be stuck there. Also slipping in these fixes to the build doc:

* information on what to do if cuckoo miner build fails
* state that target arch is currently x86
* path statement corrected from conversation in https://github.com/ignopeverell/grin/pull/147 (which can be closed as well)

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use core::{core, consensus, global};
+use core::{core, global};
 use chain;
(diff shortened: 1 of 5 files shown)


yeastplume:
whoops.. have some version confict here.. hold off a bit

yeastplume:
okay to go now

PR #158, opened 2017-10-06
Fix duplicate wallet coinbase

### LINKED ISSUE #157: wallet appending duplicate coinbase outputs from miner
The "ask for a new coinbase if the block fees change" appears to have introduced a bug in the way the wallet tracks outputs.
The wallet is always appending the output to its internal list, regardless of if the coinbase is actually new (or just rebuilt because the fees may have changed).

```
Outputs - 
fd997546, 1, 1, 4, Immature, 1000000000
15d2f09b, 2, 2, 5, Immature, 1000000000
6040547d, 3, 0, 0, Unconfirmed, 1000000000
6040547d, 3, 0, 0, Unconfirmed, 1000000000
```

Fix coming shortly - I think the wallet just need to check to existing outputs before appending a new one.
I'm also playing around with passing the `pubkey` to the miner and not the `derivation` when we build the coinbase - I think we can keep the `derivation` internal to the wallet/keychain and not need to leak it outside.

antiochp:
This PR -
* reworks `wallet_data` to store output_data in hashmap (keyed by identifier)
* reworks `Identifier` internals to store them as a hex string (this needs revisiting at some point)
* store identifier in `output_data`
* make the placeholder fee `value=10` in wallet so we can split the fee safely

This resolves #157 (the hashmap ensures we can update the existing output data easily).

Also cleaned up `wallet info` output - n_child is less important now, order the outputs by n_child but don't display it -

```
Outputs - 
identifier, height, lock_height, status, value
fd997546..., 1, 4, Unspent, 1000000000
15d2f09b..., 2, 5, Spent, 1000000000
6040547d..., 3, 6, Unspent, 1000000000
8d3f2247..., 4, 7, Unspent, 1000000000
abfb3af7..., 5, 8, Unspent, 1000000000
ac6a91ce..., 6, 9, Immature, 1000000000
3177752d..., 7, 10, Immature, 1000000000
712e0ddc..., 8, 0, Unspent, 995000000
4cc0d030..., 8, 0, Unspent, 4999990
948ca690..., 8, 11, Immature, 1000000005
732ce7e4..., 9, 12, Immature, 1000000000
d306989e..., 0, 0, Unconfirmed, 1000000000
```

Note the `1000000005` coinbase in the above output.
1) spent `1000000000`, sending `5000000` back to same wallet
1) received `4999990` (`10` in fees)
1) coinbase for block containing this fee was `1000000005` (`5` burnt, `5` reward)

### DIFF
--- a/keychain/src/extkey.rs
+++ b/keychain/src/extkey.rs
@@ -188,8 +188,9 @@ impl ExtendedKey {
 
 		let mut secret_key = SecretKey::from_slice(&secp, &derived.as_bytes()[0..32])
 			.expect("Error deriving key");
-		secret_key.add_assign(secp, &self.key)
-			.expect("Error deriving key");
+		secret_key.add_assign(secp, &self.key).expect(
+			"Error deriving key",
+		);
 		// TODO check if key != 0 ?
 
 		let mut chain_code: [u8; 32] = [0; 32];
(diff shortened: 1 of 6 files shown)


keychain/src/extkey.rs
```diff
@@ -75,36 +75,25 @@ impl fmt::Display for Fingerprint {
 	}
 }
 
-#[derive(Serialize, Deserialize, Clone)]
-pub struct Identifier([u8; 20]);
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub struct Identifier(String);
```
antiochp: I did this so we can easily serialize/deserialize identifiers in and out of the wallet.dat file.
I'm not totally happy with this but think it is ok for now.

antiochp:
Some pool and chain tests are failing - looking into these.

antiochp:
@ignopeverell this is ready for review (assuming the tests pass on travis).

wallet/src/types.rs
```diff
@@ -248,27 +263,29 @@ impl WalletData {
 		let mut data_file = File::create(data_file_path).map_err(|e| {
 			Error::WalletData(format!("Could not create {}: {}", data_file_path, e))
 		})?;
-		let res_json = serde_json::to_vec_pretty(self).map_err(|_| {
+		let res_json = serde_json::to_vec_pretty(self).map_err(|e| {
+			println!("***** serialization error - {:?}", e);
```
ignopeverell: Want to clean that println? :-)

antiochp: Done - thanks!

PR #159, opened 2017-10-06
Pass pubkey for coinbase

antiochp:
Building on the wallet changes from #158 we can now pass an `Identifier` to and from the miner, instead of passing the `derivation`.
This allows us to keep `derivation` internal to wallet/keychain.

* Use `[u8; 20]` internally in Identifier so we can be sure we are serializing/deserializing in `ser` correctly via `Readable/Writeable`.
* Add serde_json `Serialize/Deserialize/Visitor` for `Identifier` so wallet.dat is clean and the api calls between miner and wallet serialize identifiers in a sane way (as hex encoded strings).
* Add some tests for json serialization/deserialization.

Example `block_fees` in the miner -

```
 BlockFees { fees: 0, pubkey: Some(Identifier(4cc0d030afeaa6413e7862dcf0b924fbf802fd0f)) }
```

Example wallet file -

```
cat wallet.dat
{
  "outputs": {
    "ac6a91ce7564bd4be32a485c4737519c2f5d2bf4": {
      "fingerprint": "bee24629",
      "identifier": "ac6a91ce7564bd4be32a485c4737519c2f5d2bf4",
      "n_child": 6,
      "value": 1000000005,
      "status": "Immature",
      "height": 6,
      "lock_height": 9
    },
    "6040547de5ffa4e01b29d47b336a8c31050e3dc8": {
      "fingerprint": "bee24629",
      "identifier": "6040547de5ffa4e01b29d47b336a8c31050e3dc8",
      "n_child": 3,
      "value": 1000000000,
      "status": "Immature",
      "height": 3,
      "lock_height": 6
    },
    ...
}
```

`wallet info` after spending an output - 

```
Outputs - 
identifier, height, lock_height, status, value
fd997546..., 1, 4, Spent, 1000000000
15d2f09b..., 2, 5, Unspent, 1000000000
6040547d..., 3, 6, Immature, 1000000000
8d3f2247..., 4, 7, Immature, 1000000000
abfb3af7..., 5, 8, Immature, 1000000000
ac6a91ce..., 6, 9, Immature, 1000000005
3177752d..., 6, 0, Unspent, 995000000
712e0ddc..., 6, 0, Unspent, 4999990
4cc0d030..., 0, 0, Unconfirmed, 1000000000
```

### DIFF
--- a/chain/tests/mine_simple_chain.rs
+++ b/chain/tests/mine_simple_chain.rs
@@ -80,7 +80,7 @@ fn mine_empty_chain() {
 	for n in 1..4 {
 		let prev = chain.head_header().unwrap();
 		let pk = keychain.derive_pubkey(n as u32).unwrap();
-		let mut b = core::core::Block::new(&prev, vec![], &keychain, pk).unwrap();
+		let mut b = core::core::Block::new(&prev, vec![], &keychain, &pk).unwrap();
 		b.header.timestamp = prev.timestamp + time::Duration::seconds(60);
 
 		let difficulty = consensus::next_difficulty(chain.difficulty_iter()).unwrap();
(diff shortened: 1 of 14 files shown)


antiochp:
@ignopeverell got those tests to pass.
Feels like a lot of code to serialize/deserialize that hex string into json but I guess that's just how serde_json does it.

PR #161, opened 2017-10-08
Add missing values to config file for pool config

casey:
It looks like the current config is missing a few values. I picked random values, since I didn't know what was reasonable.

### DIFF
--- a/grin.toml
+++ b/grin.toml
@@ -48,6 +48,17 @@ capabilities = [7]
 host = "127.0.0.1"
 port = 13414
 
+#Transaction pool configuration
+[server.pool_config]
+
+#Base fee for a transaction to be accepted by the pool. The transaction
+#weight is computed from its number of inputs, outputs and kernels and
+#multipled by the base fee to compare to the actual transaction fee.
+accept_fee_base = 100
+
+#Maximum capacity of the pool in number of transactions
+max_pool_size = 100
+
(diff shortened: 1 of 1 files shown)


ignopeverell:
Weren't the defaults here properly picked up?

https://github.com/ignopeverell/grin/blob/master/pool/src/types.rs#L43

casey:
I think not, I got this error when building and running a freshly cloned repo:

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
         Running `target/debug/grin`
         thread 'main' panicked at 'Error parsing config file: Error parsing configuration file at /Users/rodarmor/src/grin/grin.toml - missing field `pool_config` for key `server`', src/bin/grin.rs:83:2
```

ignopeverell:
My bad, I mistakenly assumed that the config package would use defaults when a section isn't present. Looks like it can't as serde can't deserialize into a default struct, which is vexing. We don't want to have to specify every single configuration property in the config file.

Ended up fixing this in 0d26daa with default serde tags. A little more boilerplate but does the job.

ignopeverell:
Oh, and thanks for the nice error report!

yeastplume:
I haven't at this particular case in detail, but if you make something in a config struct in the code an Option<T>, then it doesn't need to be present in the config file, and can use the default values.

casey:
Yup, you bet!

PR #162, opened 2017-10-08
Identifier from public key

antiochp:
Identifiers of keys in the wallet were originally - 

```
ripemd160(sha512(secret_key))
```

When we standardized on blake2 they became the following - 

```
blake2b(20, secret_key) #20 byte digest to arbitrarily keep identifiers the same length as before
```

This PR proposes using the _public_key_ of the underlying _secret_key_ as the input to the blake2 hash function - 

```
blake2b(20, public_key(secp, secret_key))
```

The "feels" better to me (like we shouldn't ever be operating on the bytes in the secret_key directly) - but I'd like to open this PR up for discussion. But I have nothing to point to indicating this is any more (or less) secure.

Is there any reason why this would be undesirable?

### DIFF
--- a/keychain/src/extkey.rs
+++ b/keychain/src/extkey.rs
@@ -58,7 +58,7 @@ impl error::Error for Error {
 			// TODO change when ser. ext. size is fixed
 			Error::InvalidSliceSize => "keychain: serialized extended key must be of size 73",
 			Error::InvalidExtendedKey => "keychain: the given serialized extended key is invalid",
-			Error::Secp(_) => "keychain: secp error"
+			Error::Secp(_) => "keychain: secp error",
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


ignopeverell:
Can't think of any. The security (more/less) argument would be dubious at best I think. One could argue that ripemd+sha256 is two rounds of hashing while blake2b(20) is a single round. But all these functions already include multiple internal rounds anyway and blake2b is generally thought to be more secure than sha256. And besides we're only worried about collision resistance here, not really pre-image attacks.

So I'm good with that change.

ignopeverell:
Revisiting this a little. I have a couple more remarks:

* I don't think the Identifier needs to be 20 bytes. The previous value was inherited from bitcoin and its use if ripemd160, but in our case this is only an internal id used by the wallet and we have full control about the length. So seems 10 bytes would be plenty enough to avoid internal collisions. What do you think?
* The code still has variables named `pubkey` everywhere for identifiers, which is confusing.

antiochp:
pubkey renamed to key_id in #174

antiochp:
Agreed on the identifier length.
If we use 10 bytes for identifiers then I think its probably worth cleaning this up further and getting rid of the concept of "fingerprints" (the shortened 4 byte representation of an identifier).
We may as well just use the identifier everywhere consistently.

Issue #25, opened 2016-12-23
Time locked transaction outputs


ignopeverell:
How do we support them? Would likely require another output field and a signature. Maybe add to the transaction proof signature.

This would be great to have for higher level contract supports like payment channels, lightning, etc.

apoelstra:
This belongs in the transaction signature. Yeah, it'd have to be encoded in another field.

ignopeverell:
Given that we already sign the fee, it'd either have to get composed or added as another signature.

antiochp:
Been thinking a little about this.

We currently sign the `fee` on the transaction itself - so presumably we could include a height or time in the message being signed.

We do the following to create the message (with just the fee currently) -
```
fn u64_to_32bytes(n: u64) -> [u8; 32] {
	let mut bytes = [0; 32];
	BigEndian::write_u64(&mut bytes[24..32], n);
	bytes
```

So there is space in there to include the timelock info.

But - this limits the timelock to the entire transaction, so all outputs would be affected by this, change outputs as well.

Is this an acceptable approach?
Ideally the timelock would be per-output but we have no way of doing this in MW - we don't sign anything limited to just a single output.

I guess the general approach would be to make the necessary transactions within the wallet to build a single output of the right amount and _then_ send the timelocked transaction creating a single timelocked output (and avoid ending up with a potentially large change output also encumbered with a timelock).

antiochp:
I also saw some mention on the ML of keeping the `TxKernel` as small as possible and aiming to avoid needing to persist a timelock on every `TxKernel`.

We have the `features` bitflags in there - so I'm thinking we could use one of the bits in `features` to flag timelocked transactions - only serialize/deserialize the timelock if present.

This would mean the majority of TxKernels would remain the same size (assuming most are not encumbered by timelocks).

apoelstra:
Yes, the only thing MW can do that I'm aware of is put a timelock on a kernel, which effectively timelocks the whole transaction.

I'd prefer that every kernel have a timelock to make "real" ones inconspicuous, for privacy's sake. But it's not a strong preference.

antiochp:
@apoelstra I thought I saw you recommend on the ML somewhere to avoid putting a timelock on every kernel to minimize the storage cost?
But yeah I agree it makes sense to have it on all of them to keep them all similar.

Except - is it not going to be trivially easy to identify no-op timelocks or very short lock periods by just comparing the kernels to the height or timestamp of the blocks they originate from?
[...]

PR #167, opened 2017-10-10
add lock_height to Transaction and TxKernel

antiochp:
Initial attempt at introducing "time locked" transactions.

------------------

TODO -
- [x] rename `transaction_fee_and_lock_height_to_bytes` -> `kernel_sig_msg`
- [x] add tests to cover adding invalid tx to pool
- [x] add tests to cover building invalid block

-----------------

* add `lock_height` to `Transaction` and `TxKernel`
* msg being signed in a kernel now includes both `fee` and `lock_height`
* verify on kernel checks both `fee` and `lock_height` against the signature
* wallet sender sets current chain height as the lock_height (as proof of concept)
* pool checks the tx is valid (in terms of lock_height) before adding to pool
* block is invalid if built with a tx with lock_height in future is included

Note: this is _similar_ to the `lock_height` on a coinbase output but not the same.
This is a "timelock" on the tx/kernel itself, not an output.

See #25 for more info/context/discussion.


antiochp:
working on fixing up those `pool` test failures

core/src/core/transaction.rs
```diff
@@ -453,19 +489,93 @@ impl ops::Add for SumCommit {
 	}
 }
 
-fn u64_to_32bytes(n: u64) -> [u8; 32] {
+/// TODO - consider putting this in consensus.rs so we can reuse it
+/// consistently across Transaction and TxKernel
+/// TODO - add some test coverage for this to make sure the bytes are in there
+/// correctly
+fn transaction_fee_and_lock_height_to_bytes(fee: u64, lock_height: u64) -> [u8; 32] {
```
ignopeverell: I'd use a name more descriptive of what's produced rather than consumed. Maybe `kernel_sig_msg`?

ignopeverell:
Good in principle. I'm a little bummed we have to have another clear value in addition to fees but I don't have a better option for now.

core/src/core/transaction.rs
antiochp: yeah good catch - that's on my todo list actually (named it that so I knew what was going on...)

antiochp:
@ignopeverell any further feedback/thoughts on this?
I added some test coverage for both adding timelocked txs to the pool and building a block from timelocked txs.

ignopeverell:
Looks good to me!

PR #171, opened 2017-10-11
slog-rs logging

yeastplume:
This replaces all instances of env_logger with a slog logger implementation.

This adds a crate called log, which contains a global static reference to a logger and has no dependencies, so it can be included and should be callable from anywhere else in the code. Using slog drains, it can write to the console, or a log file, or both, with different logging levels for both. 

Logging can be controlled via the new configuration options in grin.toml (which are all documented there). I've tried to set the logging configuration struct to sensible defaults if no grin.toml file is being used.

This is a start anyhow, and we'll see how we get on with this. Thought it best to merge earlier rather than later, as all log! macros throughout have been updated with a reference to the logger instance.

Also doc updates to remove the RUST_LOG=grin=debug from everywhere, as that's no longer necessary.

Some rustfmt changes there as well, as I tended to run rustfmt on files after sedding all of the log statements (I guess the rules in rustfmt.toml have changed slightly)


ignopeverell:
Looks good, thanks for taking that on. How about putting the log stuff under the util crate instead of having its own? I'm trying to prevent too many micro-crates at the root of the repository, I think it gets quickly confusing to newcomers.

yeastplume:
Makes sense, done. We should be able to log from anywhere in the code now with impunity.

Issue #168, opened 2017-10-11
sort inputs/outputs lexicographically during serialization


antiochp:
I know we're doing it in some places (we sort in a couple of places for inputs/outputs in blocks).

But we should make sure we sort before we serialize (specifically for transactions, but presumably for blocks also).
This is relatively low-hanging fruit.

From @apoelstra in gitter - 
> it occurs to me that even on the wire, transaction inputs and outputs should be lexicographically ordered.. one big privacy leak in bitcoin is wallet fingerprinting, we should try to make transactions as unique as possible

https://gitter.im/grin_community/Lobby?at=59dcc0d4210ac26920cb02f7

apoelstra:
To clarify, I think it should be a consensus rule that everything on the wire is sorted (and everything in blocks should be). This is super cheap to verify and will prevent broken software.

antiochp:
Ah right - of course, we can actually verify this via a consensus rule...

antiochp:
Resolved in #173.

PR #173, opened 2017-10-12
fail deserialization if inputs or outputs are not sorted for txs

antiochp:
Added `VerifySortOrder` trait and a `read_and_verify_sorted` function so we can call the following during deserialization - 

```
let inputs = read_and_verify_sorted(reader, input_len)?;
```

`VerifySortOrder` iterates over `window(2)` over the collection so we can efficiently compare each value with the previous. If previous is ever larger then we are not sorted and we fail.

Added `WriteableSorted` trait so we can call the following during serialization -

```
try!(inputs.write_sorted(writer));
```

-----------

This PR addresses #168 and adds a consensus rule to ensure input and outputs are sorted for txs. Similarly inputs, outputs and kernels are sorted for blocks.

- fail in deserialization if inputs or outputs are not sorted for txs
- fail in deserialization if inputs, outputs or kernels are not sorted for blocks
- make sure we sort them before serialization

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -20,10 +20,10 @@ use std::collections::HashSet;
 
 use core::Committed;
 use core::{Input, Output, Proof, TxKernel, Transaction, COINBASE_KERNEL, COINBASE_OUTPUT};
-use consensus::{MINIMUM_DIFFICULTY, REWARD, reward, exceeds_weight, VerifySortOrder};
+use consensus::{MINIMUM_DIFFICULTY, REWARD, reward, exceeds_weight};
 use core::hash::{Hash, Hashed, ZERO_HASH};
 use core::target::Difficulty;
-use ser::{self, Readable, Reader, Writeable, Writer, WriteableSorted};
+use ser::{self, Readable, Reader, Writeable, Writer, WriteableSorted, read_and_verify_sorted};
 use global;
 use keychain;
 
@@ -189,11 +189,11 @@ impl Writeable for Block {
 				[write_u64, self.kernels.len() as u64]
 			);
 
-			// Consensus rule that everything is sorted in lexicographical order on the wire.
 			let mut inputs = self.inputs.clone();
 			let mut outputs = self.outputs.clone();
 			let mut kernels = self.kernels.clone();
 
+			// Consensus rule that everything is sorted in lexicographical order on the wire.
 			try!(inputs.write_sorted(writer));
 			try!(outputs.write_sorted(writer));
 			try!(kernels.write_sorted(writer));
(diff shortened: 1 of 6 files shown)


antiochp:
Not sure why the tests failed on TravisCI - not sure it was related to this PR.

core/src/core/block.rs
```diff
@@ -212,9 +218,27 @@ impl Readable for Block {
...
+		//
+		let mut sorted_inputs = inputs.clone();
+		sorted_inputs.sort_by_key(|input| input.hash());
+		if sorted_inputs != inputs { return Err(ser::Error::BadlySorted); }
+		let mut sorted_outputs = outputs.clone();
+		sorted_outputs.sort_by_key(|output| output.hash());
+		if sorted_outputs != outputs { return Err(ser::Error::BadlySorted); }
+		let mut sorted_kernels = kernels.clone();
+		sorted_kernels.sort_by_key(|kernel| kernel.hash());
+		if sorted_kernels != kernels { return Err(ser::Error::BadlySorted); }
```
ignopeverell: I'm worried about the expense of the vector comparison here. It likely checks every value within each struct.

core/src/core/block.rs
```diff
@@ -189,14 +189,20 @@ impl Writeable for Block {
...
+			sorted_outputs.sort_by_key(|output| output.hash());
+			for out in sorted_outputs {
 				try!(out.write(writer));
 			}
-			for proof in &self.kernels {
-				try!(proof.write(writer));
+			let mut sorted_kernels = self.kernels.clone();
+			sorted_kernels.sort_by_key(|kernel| kernel.hash());
+			for kernel in sorted_kernels {
+				try!(kernel.write(writer));
```
ignopeverell: I'd factor this code out a bit.

core/src/core/block.rs
antiochp: Yeah that makes sense - let me take another look and see if we can check the order of the serialized values directly.

ignopeverell: Instead of the `(0..len).map` it'll likely be better to have a full blown loop that compares to the previous iteration. That can be its own little utility function.

core/src/core/block.rs
```diff
@@ -212,9 +211,15 @@ impl Readable for Block {
 		let (input_len, output_len, proof_len) =
 			ser_multiread!(reader, read_u64, read_u64, read_u64);
 
-		let inputs = try!((0..input_len).map(|_| Input::read(reader)).collect());
-		let outputs = try!((0..output_len).map(|_| Output::read(reader)).collect());
-		let kernels = try!((0..proof_len).map(|_| TxKernel::read(reader)).collect());
+		let inputs: Vec<_> = try!((0..input_len).map(|_| Input::read(reader)).collect());
```
antiochp: We could probably take this a step further and wrap this up in something like `read_sorted` to match `write_sorted` above.

antiochp:
@ignopeverell - assuming the tests pass on TravisCI this is good to go.
Thoughts on this approach?

I keep looking at this line and rust is blowing my mind...  

```
let inputs = read_and_verify_sorted(reader, input_len)?;
```

It only knows how many bytes to read for each input based on the type of the return value.
And it only knows the type of `inputs` because we eventually pass it into the Block constructor below.
It just doesn't feel like rust has enough information here to be able to do this.

ignopeverell:
Type systems can be *really* sophisticated :-)

PR #174, opened 2017-10-12
rename pubkey -> key_id

antiochp:
these are key identifiers, not public keys

PR #175, opened 2017-10-12
Shorter identifiers and call them key_id (not pubkey)

antiochp:
- Identifiers are now 10 bytes long.
- Remove `Fingerprint` and just the shorter `Identifier`.
- Wallets now store `root_key_id` and `key_id` on output data (was `fingerprint` and `identifier` which was confusing).
- renamed `pubkey` -> `key_id` (these are identifiers, not public keys)

Also fixes a bug in the fee code when sending a partial tx from a wallet.
* Fee needs to come out of the amount being _received_ (not the total being _spent_).
* Sender calculates the fee
* Recipient verifies the fee is correct and rejects the partial tx if fee is disputed

### DIFF
--- a/wallet/src/sender.rs
+++ b/wallet/src/sender.rs
@@ -100,7 +100,7 @@ pub fn issue_burn_tx(
 
 	// operate within a lock on wallet data
 	WalletData::with_wallet(&config.data_file_dir, |mut wallet_data| {
-		
+
 		// select all suitable outputs by passing largest amount
 		let (coins, _) = wallet_data.select(fingerprint.clone(), u64::max_value());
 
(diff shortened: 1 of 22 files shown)


antiochp:
This is going to need rebasing once #174 has been merged in (renaming pubkey to key_id).

wallet/src/sender.rs
```diff
@@ -118,19 +120,34 @@ pub fn issue_burn_tx(
...
+	// sender
+	let fee = tx_fee(coins.len(), 2, None);
 	parts.push(build::with_fee(fee));
-	let change = total - amount - fee;
+
+	// if we are spending 10,000 coins to send 1,000 then our change will be 9,000
+	// the fee will come out of the amount itself
+	// if the fee is 80 then the recipient will only receive 920
+	// but our change will still be 9,000
+	let change = total - amount;
```
antiochp: I think a bug was introduced here when the fee calc was put in place.
The fee comes out of the amount, not in addition to it.
This was producing an invalid tx as the amounts were no longer summing to 0 correctly.

ignopeverell:
Can you merge master into your branch? I merged @yeastplume PR first and it touched quite a fiew files :-)

antiochp:
@ignopeverell This should be good to go now.
Also fixes the "should not be mut" build error that we are seeing on master.

wallet/src/sender.rs
ignopeverell: Ah yes, good catch, sorry for the oversight. Maybe some more wallet tests would be good? I can create a github issue for it.

PR #178, opened 2017-10-13
refactor burn key into key_overrides on keychain

antiochp:
Possible approach to make the burn key implementation more flexible - 

* introduce `burn_enabled` on Keychain to take a keychain and add a key override
* introduce a map of identifer->secret_key overrides to keychain
* key overrides on a keychain take precedence (if present)
* fix fee calc in `issue_burn_tx` (which I broke when fixing fee calc in `inputs_and_change` previously...)

Right now we only use key_overrides within `issue_burn_tx` and we throw the "burner" keychain away when we are finished. We don't touch the original keychain so it is safe to subsequent use.

### DIFF
--- a/keychain/src/keychain.rs
+++ b/keychain/src/keychain.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use rand::{thread_rng, Rng};
+use std::collections::HashMap;
 
 use secp;
 use secp::{Message, Secp256k1, Signature};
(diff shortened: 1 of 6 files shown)


keychain/src/keychain.rs
```diff
@@ -42,29 +43,36 @@ impl From<extkey::Error> for Error {
...
-	pub enable_burn_key: bool,
+	key_overrides: HashMap<Identifier, SecretKey>,
 }
 
 impl Keychain {
 	pub fn root_key_id(&self) -> Identifier {
 		self.extkey.root_key_id.clone()
 	}
 
+	// for tests and burn only, associate the zero fingerprint to a known
```
ignopeverell: How about making it more flexible by allowing to provide the Identifier that will go in the map as well? May be useful in unit tests, to have multiple outputs to different parties for example.

wallet/src/sender.rs
```diff
@@ -92,8 +94,11 @@ fn build_send_tx(
 }
 
 pub fn issue_burn_tx(config: &WalletConfig, keychain: &Keychain, amount: u64) -> Result<(), Error> {
+	let keychain = &Keychain::burn_enabled(keychain, &Identifier::zero());
```
antiochp: Pass in the Identifier here for flexibility.

antiochp: We could potentially generalize this further and pass in multiple Identifiers.
I think we'd need to randomly generate secret keys at that point (and not rely on a single known value).
Maybe this is a subsequent PR.

antiochp:
@ignopeverell This is ready for review.

I added a `zero_ok` flag on OutputData so we can identify outputs with zero confirmations that we are comfortable spending from a wallet. But this is not is use right now (see #180 for issue relating to this).

Let me know if you want me to split this out into a couple of PRs as there are a couple of different things going on here - 
* refactoring the burn key logic in wallet
* identifying zero confirmation change outputs in the wallet

keychain/src/keychain.rs
```diff
@@ -42,29 +43,42 @@ impl From<extkey::Error> for Error {
...
 impl Keychain {
 	pub fn root_key_id(&self) -> Identifier {
 		self.extkey.root_key_id.clone()
 	}
 
+	//
+	// For tests and burn only, associate a key identifier with a known secret key.
+	//
+	// TODO - does this need to be a "known" key? Or can we just generate one randomly
+	// and *really* burn the coins?
```
ignopeverell: I think we could have a use for both. The advantage of a known key in testnet is we can generate a bunch of these outputs with different amounts and easily claim them all at once to test large input set transactions and pruning.

ignopeverell:
Looks good. Want to just cleanup the TODO?

antiochp:
Should be good to go now.

PR #179, opened 2017-10-14
Switch commitments

yeastplume:
This defines a new 'SwitchCommitHash' structure, which uses the recently added switch commit functionality in secp256k, and adds it to the Output structure. I've called it SwitchCommitHash as it's more accurate; it's the hash of a switch commit.

Length of the blake2 hash is 20 bytes, but the constant can be changed if necessary.

We'll need to add a 'get_output_by_switch_commit' function (which would seem the logical place to put tests) into chain/src/store.rs as some stage, but I though I'd leave that for another task.

Also a minor doc fix, and minor fix for logging output when no grin.toml file is present.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -618,7 +618,7 @@ mod test {
 		let commit = keychain.commit(5, &key_id).unwrap();
 		let switch_commit = keychain.switch_commit(&key_id).unwrap();
 		let switch_commit_hash = SwitchCommitHash::from_switch_commit(switch_commit);
-			let msg = secp::pedersen::ProofMessage::empty();
+		let msg = secp::pedersen::ProofMessage::empty();
 		let proof = keychain.range_proof(5, &key_id, commit, msg).unwrap();
 
 		let out = Output {
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -511,9 +511,6 @@ impl Output {
 pub struct SumCommit {
 	/// Output commitment
 	pub commit: Commitment,
-	/// Switch commitment hash, which should be included in the hash
-	/// but not the sum
-	pub switch_commit_hash: SwitchCommitHash,
 	/// Secp256k1 used to sum
 	pub secp: Secp256k1,
 }
@@ -525,7 +522,6 @@ impl Summable for SumCommit {
 	fn sum(&self) -> SumCommit {
 		SumCommit {
 			commit: self.commit.clone(),
-			switch_commit_hash: self.switch_commit_hash.clone(),
 			secp: self.secp.clone(),
 		}
 	}
@@ -538,7 +534,6 @@ impl Summable for SumCommit {
 impl Writeable for SumCommit {
 	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
 		self.commit.write(writer)?;
-		self.switch_commit_hash.write(writer)?;
 		Ok(())
 	}
 }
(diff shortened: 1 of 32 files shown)


ignopeverell:
One complication I hadn't thought about: right now there's no root hash that covers all switch commitment hashes and so they could get mutated. We should probably add them to the commitment sum tree, only on the hash computation side (not the actual sum).

yeastplume:
Yep, of course. 

This (git) commit will actually fail tests at the moment, but I just wanted to check on the right path as I haven't dug into that part of the code much (but glad to have an excuse to do so).

So is this just a matter of adding the switch_commit_hash to the SumCommit field, then making sure it's included it in the Writeable trait implementation (for which the Hashed trait is defined) as I've done here?

If so, for some reason, I'm getting 
```
ERRO Corrupted storage, could not read an entry from sum tree store: IOErr(Error { repr: Custom(Custom { kind: UnexpectedEof, error: StringError("failed to fill whole buffer") }) })
```
When trying to add a new block after adding the switch commit to the writeable/readable implementations for SumCommit.. just wondering if you can guess why offhand, just don't have the chance to dig into it right this moment as I need to go off and do other things.

yeastplume:
Actually, thinking about this away from the machine for a bit, rather than storing the whole extra 20 bytes in the sum tree, is it not better somehow just manually ensure the hash field in the summable element is actually the hash of the commitment+switch_commitment_hash? Or does that potentially confuse the design?

core/src/core/transaction.rs
```diff
@@ -469,15 +537,21 @@ impl Summable for SumCommit {
...
 	fn read(reader: &mut Reader) -> Result<SumCommit, ser::Error> {
 		let secp = secp::Secp256k1::with_caps(secp::ContextFlag::Commit);
+		let commit = Commitment::read(reader)?;
+		let switch_commit_hash = SwitchCommitHash::read(reader)?;
+
 		Ok(SumCommit {
-			commit: Commitment::read(reader)?,
+			commit: commit,
+			switch_commit_hash: switch_commit_hash,
 			secp: secp,
```
antiochp: Just reading through this - not directly related to your changes, but do we need to maintain an instance of `secp` in every `SumCommit`?
Seems like this the only place we actually use this is in `add` in `impl ops::Add for SumCommit` (and we could just init `secp` there on demand?)

ignopeverell:
@yeastplume correct, we should only incorporate the switch hash to the commitment hash
@antiochp I think I had intended to make this a reference instead of a full copy...

yeastplume:
Okay, this is a better approach, I think, and requires no additions to the sum node. I've added a 'hash_with' argument to the 'hash' function of the 'Hashed' trait, that optionally takes a type implementing 'Writeable'. Then if this option contains a value, it gets appended to the hash.

I've had to extend the pmmr push method to take that Option to pass into the created node, but I don't think that adds too much cruft. More annoyingly, all of the calls to .hash() throughout the code have to be updated with None::<Type> along with a type annotation. However, this approach is reusable in future if ever we feel we need to hash anything with anything else.

yeastplume:
... looking at this again, I could have added a hash_with method to the Hashed trait instead of changing the 'hash' method signature, then check for that within the sumtree 'push' and 'from_summable' methods.. actually hold off until I do that instead.

yeastplume:
Yes, that feels much better. Ready for any further review/comments.

ignopeverell:
The downside of rustfmt is it has a way for making PRs extremely difficult to review but hopefully that's only until everything is properly formatted... Changes look good though, thanks!

core/src/ser.rs
```diff
@@ -527,7 +528,11 @@ impl AsFixedBytes for [u8; 8] {
 		return 8;
 	}
 }
-impl AsFixedBytes for [u8; 32] {
+impl AsFixedBytes for [u8; 20] {
+	fn len(&self) -> usize {
+		return 20;
+	}
+}impl AsFixedBytes for [u8; 32] {
```
kargakis: Does this come from rustfmt?

Issue #181, opened 2017-10-16
zero confirmation spend - fees incorrect? build_block/process_block mismatch?


antiochp:
During some local testing of zero confirmation spends from a wallet, I am seeing outputs generated that never appear to get confirmed (so suspect the values are incorrect in the wallet).

For example - 

```
Spent 1000000000 
To send 5000000 
Received 4999920 (fees 80)
Change output 995000000
```

```
Spent 995000000 (zero confirmation from above)
To send 5000000 
Received 4999920 (fees 80) <-- never gets confirmed?
Change output 990000000 <-- never gets confirmed?
```

We see this in node logs -

```
Oct 15 20:18:27.867 DEBG block_fees here: BlockFees { fees: 80, height: 6, key_id: Some(Identifier(f26dbc24171432f88888)) }
Oct 15 20:18:27.994 DEBG (Server ID: Port 13414) Built new block with 1 inputs and 3 outputs, difficulty: 10
```
This is the _only_ block in the logs with inputs, so both txs are presumably in there (one spending the change output of the other).

I think the fees on this block are being calculated incorrectly.
There should be 80 in fees _for each_ tx here, 160 in total?

`wallet info` looks like - 

```
Outputs - 
key_id, height, lock_height, status, zero_ok, value
----------------------------------
17ebdeb665384ae5db17, 1, 4, Spent, false, 1000000000
a9d252771e29beb5cce8, 2, 5, Unspent, false, 1000000000
dd930ae6719a382182b2, 3, 6, Unspent, false, 1000000000
4636dd289d8f23d93a0c, 4, 7, Unspent, false, 1000000000
3f9002bfecc722ade9bc, 5, 8, Unspent, false, 1000000000
9b06dbf15475d98a65bf, 0, 0, Spent, true, 995000000
06c43cafdb59ae893b4e, 6, 0, Unspent, false, 4999920
86767a2b6d44795db540, 0, 0, Unconfirmed, true, 990000000
44a325a40d2c726bf074, 0, 0, Unconfirmed, false, 4999920
f26dbc24171432f88888, 6, 9, Unspent, false, 1000000040
4b443bf73a62d814d4b2, 7, 10, Immature, false, 1000000000
7f7d3f5ea9dcf0e1b0ee, 8, 11, Immature, false, 1000000000
b41e22cd4204c0454381, 9, 12, Immature, false, 1000000000
268081b71c66134a8c91, 0, 0, Unconfirmed, false, 1000000000
```

[...]

PR #182, opened 2017-10-16
update_roots on pool to keep it consistent after reconciliation

antiochp:
This fixes the issue I was seeing in #181 
We can now successfully spend a zero confirmation change output and it will get mined after a couple of blocks.
We currently only mine `roots` from the pool - this PR ensures `roots` is consistent after we call `reconcile_block`.

* introduced `update_roots` to ensure the pool is consistent after reconciliation
* remove duplicates before marking txs (leverage sets here)
* use HashSet instead of HashMap when marking txs

### DIFF
--- a/pool/src/graph.rs
+++ b/pool/src/graph.rs
@@ -28,6 +28,7 @@ use core::core::hash::Hashed;
 
 /// An entry in the transaction pool.
 /// These are the vertices of both of the graph structures
+#[derive(Debug, PartialEq, Clone)]
 pub struct PoolEntry {
 	// Core data
 	/// Unique identifier of this pool entry and the corresponding transaction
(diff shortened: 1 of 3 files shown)


pool/src/graph.rs
```diff
@@ -179,6 +180,31 @@ impl DirectedGraph {
 		}
 	}
 
+	/// Promote any non-root vertices to roots based on current edges.
+	/// For a given tx, if there are no edges with that tx as destination then it is a root.
+	pub fn update_roots(&mut self) {
+		let mut new_roots = vec![];
+		for x in &self.vertices {
+			let edges = &self.edges.values().collect::<Vec<_>>();
```
MoaningMyrtle: I think you can move this outside the for loop to avoid allocating for every vertex.

pool/src/graph.rs
```diff
@@ -179,6 +180,31 @@ impl DirectedGraph {
...
+				Some(_) => (),
+				None => new_roots.push(x.clone()),
+			}
+		}
+
+		for x in new_roots {
+			self.roots.push(x.clone());
+		}
+
+		let mut new_vertices = vec![];
```
MoaningMyrtle: I think you can avoid allocating a new vertices vector here and also avoid comparing against all of `roots` by reusing `new_roots`. Something like (not tested):

```
		for x in new_roots {
			self.roots.push(x.clone());
                        match self.vertices.iter().position(|v| v.transaction_hash == x.transaction_hash) {
                                   Some(i) => self.vertices.swap_remove(i),
                                    None=> (),
                         }
                  }
```

Even better, since you know where vertices are matching from the above loop (L187) you could do without this second position search altogether- although the ways I can think of to do that now are messy (would be really nice with Rust nightly's `drain_filter`, which isn't stable yet). That feels like an optimization that can wait, though.

antiochp:
Thanks for the feedback - good suggestions.
This was definitely more "make it work" and less "make it fast".
But let me take another pass tomorrow and I'll clean some of this up. I think adding the HashMaps so we can index the sources and destinations makes a lot of sense.

antiochp:
@MoaningMyrtle @ignopeverell - this is ready for review.
A *lot* less iterating and allocating going on in `update_roots` now.
I also noticed during testing that `conflicting_txs` and `conflicting_outputs` contained a lot of duplicate hashes (multiple outputs of a given tx etc.) So made some changes to de-dupe these during the "mark" process.

antiochp:
Not sure why TravisCI failed here - it failed on "address in use" during the `grin_grin` tests.
@ignopeverell can you restart the tests in TravisCI? (I do not have permission).

PR #197, opened 2017-10-23
Bind to services to all interfaces rather than loopback

yeastplume:
The p2p and endpoints were binding to 127.0.0.1, i.e. the loopback adapter, and not accessible from outside machines. I think we're going to need a way to explicitly specify the adapter to bind to, but this just solves the issue for now to allow to test multi-machine setups via docker.

### DIFF
--- a/grin/src/miner.rs
+++ b/grin/src/miner.rs
@@ -603,7 +603,6 @@ impl Miner {
 				"{}/v1/receive/coinbase",
 				self.config.wallet_receiver_url.as_str()
 			);
-			trace!(LOGGER, "Coinbase receiver URL: {}", url);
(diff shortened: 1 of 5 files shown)


dandcg:
Hi, can this change applied to the wallet receiver as well?

yeastplume:
I've applied it to all receivers/listeners throughout, including the wallet.

dandcg:
Ok thanks that's great news.

yeastplume:
Just clarifying, this will make services accessible on all interfaces, not just the default, so if this is merged we should ensure another issue is put up immediately to keep track of the need to configure what IP address or interface the services will bind to.

ignopeverell:
@yeastplume thanks and yes, it should be configurable and I believe the default should be loopback (especially for the wallet stuff).

PR #198, opened 2017-10-23
"grin wallet init" to create the initial wallet.seed file

antiochp:
We have been using the wallet "password" as the actual wallet seed for deriving keys.
As we get closer to `devnet/testnet` we're going to want wallets to be a bit less deterministic...

* introduce new `wallet.seed` file (hex representation of the seed)
* introduce new `grin wallet init` to initialize a new seed file
* keychain is derived from combination of seed file + password
  * for a given seed file, different passwords will generate different keychains/extended keys

TODO -
- [x] better error handling around seed file
- [x] better CLI error messages around seed file operations
- [x] `wallet init` does not actually use the password, so do not require it on CLI?
- [x] update docs

```
# must run wallet init first
grin wallet init

# then specify password to use in combination with the seed file for subsequent wallet operations
grin wallet -p "password" burn 1000
```

### DIFF
--- a/src/bin/grin.rs
+++ b/src/bin/grin.rs
@@ -163,7 +163,8 @@ fn main() {
 			.short("p")
 			.long("pass")
 			.help("Wallet passphrase used to generate the private key seed")
-			.takes_value(true))
+			.takes_value(true)
+			.default_value("mimblewimble"))
 		.arg(Arg::with_name("data_dir")
 			.short("dd")
 			.long("data_dir")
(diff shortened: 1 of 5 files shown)


src/bin/grin.rs
```diff
@@ -357,8 +350,26 @@ fn wallet_command(wallet_args: &ArgMatches) {
...
 
-	match wallet_args.subcommand() {
+	// Derive the keychain based on seed from seed file and specified passphrase.
+	// Generate the initial wallet seed if we are running "wallet init".
+	let keychain = match wallet_args.subcommand() {
+		("init", Some(_)) => {
+			let wallet_seed = wallet::WalletSeed::init_file(&wallet_config)
+				.expect("Failed to init wallet seed file.");
+			wallet_seed.derive_keychain(&"")
+				.expect("Failed to derive keychain from seed file and passphrase.")
```
ignopeverell: Maybe just return here? Someone may be tempted to add code underneath without thinking twice about it that wouldn't be applicable to init.

antiochp: Oh right... I was wondering how to stop here, never thought to just `return`...

antiochp:
@ignopeverell should be good to go now.

I'm proposing making the passphrase optional now with this.
There's a use case for a seed file with no passphrase - easy backup/restore of a wallet via the seed file. The use of a passphrase makes backup more complex but more secure (so a tradeoff there).

Particularly if we move to a mnemonic scheme at some point - it would be nice to default to generating a wallet that can be restored easily without the additional passphrase (assuming we can recover the utxos via switch commitments and rewound rangeproofs etc.)

At which point I'm imagining some kind of `grin wallet recover` command would be a nice proof of concept.

ignopeverell:
Very nice! And yes, we will need some recovery through the switch commitments for mainnet no matter what. Want to create an issue to track it?

Kixunil:
Do I understand correctly that this doesn't involve any kind of checksum - so that if you type incorrect password, the only way you find out about it is by seeing zero balance?

antiochp:
@Kixunil yeah that's correct. This is roughly analogous to the "25th seed word" in a mnemonic - any password will generate a valid wallet, and its just that they will all be empty except for the one with the correct passphrase.
https://blog.trezor.io/hide-your-trezor-wallets-with-multiple-passphrases-f2e0834026eb

This was mainly to move away from a fully deterministic wallet - it is likely the wallet passphrase behavior will go through several more iterations before this is final.

Kixunil:
That's nice, thanks for explanation!

PR #205, opened 2017-10-24
refresh multiple wallet outputs in single api call

antiochp:
Refreshing ~50 unspent outputs in the wallet takes a _significant_ amount of time - each output required an api call to the node to get the latest state (`/v1/chain/utxo/:id`).

This PR reworks this so we make a single api call, asking for multiple outputs from the node in one go.
Also experimented with skipping the `ApiEndpoint` and building this directly as an Iron handler on the Iron router - primarily for flexibility (no id param in the url but need to parse query params).

* reworked /chain/utxos endpoint to support retrieving multiple utxos in one api call
  * `GET /v2/chain/utxos?id=xxx&id=yyy&id=zzz`
  * "v2" endpoints are based on iron router and handlers (removing the need for an `ApiEndpoint`)
* moved wallet api out of `bin/grin.rs` and into `wallet` crate
  * call `wallet::server::start_rest_apis` from `bin.grin.rs`

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -15,6 +15,8 @@ hyper = "~0.10.6"
 slog = { version = "^2.0.12", features = ["max_level_trace", "release_max_level_trace"] }
 iron = "~0.5.1"
 router = "~0.5.1"
+mount = "~0.3.0"
+urlencoded = "~0.5.0"
 serde = "~1.0.8"
 serde_derive = "~1.0.8"
 serde_json = "~1.0.2"
--- a/api/src/endpoints.rs
+++ b/api/src/endpoints.rs
@@ -20,9 +20,9 @@ use chain;
 use core::core::Transaction;
 use core::ser;
 use pool;
+use handlers::UtxoHandler;
 use rest::*;
 use types::*;
-use secp::pedersen::Commitment;
 use util;
 use util::LOGGER;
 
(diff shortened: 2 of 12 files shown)


yeastplume:
I was just going though this part of the code to add an endpoint to return sumtree roots and some more information about the head, and was then looking at how inflexible the ApiEndpoint was (needing to provide an ID even though you want to query something without an ID). So if this provides more flexibility there I'm happy to see it. I'll hold off on my changes until this is merged.

antiochp:
I remember discussing this with @ignopeverell a few weeks ago - I get why we introduced the `ApiEndpoint` abstraction but I suspect it may be premature right now to commit to building this intermediate layer of api abstraction on top of iron/hyper.
I suspect it would be more flexible to use Iron handlers directly and potentially introduce an app level api abstraction at a later time once things are more stable.

antiochp:
Looking into those test failures.

ignopeverell:
Regarding the `ApiEndpoint`: you may be right, that may have been a premature optimization given that very few endpoints will be CRUD and most of them will be plain GETs.

Going to comment on #206 regarding wallet lock.

antiochp:
OK - hopefully those tests should pass now.
Got into some trouble there with the `mount` and the `router` and the borrow checker was not happy.

antiochp:
Backed out some of the changes around ownership of the ApiServer - borrow checker was not happy in the simulnet tests.

antiochp:
@ignopeverell should be good to go now. Tests are (finally) passing on TravisCI...

antiochp:
@ignopeverell Hold off on merging this - I'm seeing a problem locally with a mining node failing to connect to a wallet receiver.

antiochp:
@ignopeverell - I tested pretty throughly locally and this is now working. 
All tests are passing on TravisCI.

ignopeverell:
Looking good. Two peripheral remarks:

* We may want to profile this a little bit. There's no reason a local HTTP request and a local db lookup should take more than a few milliseconds. Maybe it's the debug compilation (as opposed to --release)?
* We may want to automate the tests you're doing manually around the wallet now.

PR #210, opened 2017-10-25
only hold wallet write lock for write operations on the wallet

antiochp:
* Add new `WalletData::read_wallet` to go with existing `WalletData::with_wallet`
* Rework the various wallet data operations so we hold the write lock via `with_wallet` as little as possible

Sample output now that we limit the use of the write lock (15ms here for a wallet refresh) - 

```
Oct 25 11:52:12.400 DEBG Refreshing wallet outputs
Oct 25 11:52:12.678 INFO acquired wallet lock ...
Oct 25 11:52:12.693 INFO ... released wallet lock
```

### DIFF
--- a/api/src/rest.rs
+++ b/api/src/rest.rs
@@ -295,7 +295,7 @@ impl ApiServer {
 
 		// declare a route for each method actually implemented by the endpoint
 		let route_postfix = &subpath[1..];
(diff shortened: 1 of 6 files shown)


antiochp:
@ignopeverell this should be good to go. Big(ish) PR but the changes themselves are actually pretty reasonable.

PR #213, opened 2017-10-26
add retry logic to miner when hitting wallet coinbase API

antiochp:
The miner needs to get a new coinbase output __reliably__ from the wallet API.
Failing to obtain a coinbase output effectively kills the miner - it cannot mine a block without obtaining one.

This PR adds retry logic on the miner client side when hitting the wallet API.
It also reworks the `receive/coinbase` endpoint to use an iron handler (for flexibility in terms of proposed timeout logic and handling of response codes for errors etc.).

* uses `tokio_retry` to implement the retry logic
* uses new `wallet_client` with latest hyper crate (0.11, api uses 0.10.6)
* new `CoinbaseHandler` in wallet API to replace wallet `ApiEndpoint` (for flexibility around returning different response codes, wallet unavailable etc.)
* wallet lock logic now uses `tokio_retry` internally


antiochp:
Hey @ignopeverell - early draft of a proposed impl for retry logic (and more robust) communication between the miner and the wallet API for obtaining coinbase outputs.

Feedback/thoughts greatly appreciated.

grin/src/wallet_client.rs
```diff
@@ -0,0 +1,66 @@
...
+	let retry_strategy = FibonacciBackoff::from_millis(250).take(3);
+	let retry_future = Retry::spawn(core.handle(), retry_strategy, f);
+	let res = core.run(retry_future).unwrap();
+	Ok(res)
+}
+
+///
+/// Call the wallet API to create a coinbase output for the given block_fees.
+///
+pub fn create_coinbase(url: &str, block_fees: &BlockFees) -> Result<CbData, Error> {
```
ignopeverell: Wouldn't the wallet client make more sense in the wallet crate?

api/src/handlers.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 The Grin Developers
+// Copyright 2017 The Grin Developers
```
ignopeverell: Thanks for remembering to update the copyright. By the time I remember it will be 2018...

wallet/src/handlers.rs
```diff
@@ -0,0 +1,88 @@
...
+use bodyparser;
+
+use receiver::receive_coinbase;
+use core::ser;
+use api;
+use keychain::Keychain;
+use types::*;
+use util;
+
+
```
ignopeverell: Looks like you were right, using a hyper handler looks cleaner.

ignopeverell:
I'm not sure there's a point in ever stopping polling for a wallet receiver. After backoff I think it should retry every 5-10 sec or so.

api/src/handlers.rs
antiochp: I'm never 100% sure if these _should_ be updated?

antiochp:
Updated - we now retry forever. The miner retries the API call forever, backing off to every 10s.
Internally the retry logic for acquiring the lock file uses a similar tokio_retry loop.

So we can now start the mining node before the wallet receiver. We'll see the following in the logs until we can make a successful API call - 
```
Oct 27 12:32:23.269 ERRO Failed to get coinbase via wallet API (will retry)...
Oct 27 12:32:23.371 ERRO Failed to get coinbase via wallet API (will retry)...
Oct 27 12:32:23.477 ERRO Failed to get coinbase via wallet API (will retry)...
```

Internally to the wallet we can jam it up with a `touch wallet.lock` and we will see the same thing in the mining node logs.
We will also see the following in the wallet receiver logs - 
```
Oct 27 12:28:54.794 INFO Acquiring wallet lock ...
Oct 27 12:28:54.794 DEBG Attempting to acquire wallet lock
Oct 27 12:28:54.807 DEBG Attempting to acquire wallet lock
Oct 27 12:28:54.820 DEBG Attempting to acquire wallet lock
...
Oct 27 12:28:56.266 ERRO Failed to acquire wallet lock file (multiple retries)
```

antiochp:
@ignopeverell I think this is good to go.
We can get more sophisticated with the retry logic and error codes back from the API but I think this works good enough as-is.

PR #214, opened 2017-10-27
API Endpoints

yeastplume:
Bit of WIP, but enough there now to be useful.

This adds the following endpoints (naming is up for debate):

* /v2/sumtrees/roots (gets all sumtree roots)
* /v2/sumtrees/lastutxos?n=10 (gets last x insertions into utxo sum tree, retrieves the associated outputs and formats for display friendliness) param is optional, returns 10 by default
* /v2/sumtrees/lastrangeproofs?n=10 (as above, but just rangeproof hashes)
* /v2/sumtrees/lastkernels?n=10 (as above, just kernel hashes)
* /v2/chain (adds more useful output, no longer need to supply param)

A lot of helper functions throughout to support all of this, most notably another method added directly into the PMMR struct to get the last n inserted nodes. There's definitely more elegant ways of doing this, but what's there is still O(n) and works.

### DIFF
--- a/api/src/endpoints.rs
+++ b/api/src/endpoints.rs
@@ -20,37 +20,11 @@ use chain;
 use core::core::Transaction;
 use core::ser;
 use pool;
-use handlers::UtxoHandler;
+use handlers::{UtxoHandler, ChainHandler, SumTreeHandler};
 use rest::*;
 use types::*;
 use util;
 use util::LOGGER;
-
-/// ApiEndpoint implementation for the blockchain. Exposes the current chain
-/// state as a simple JSON object.
-#[derive(Clone)]
-pub struct ChainApi {
-	/// data store access
-	chain: Arc<chain::Chain>,
-}
-
-impl ApiEndpoint for ChainApi {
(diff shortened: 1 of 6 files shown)


ignopeverell:
Looks good! But not sure it deserves a new version as no one really relies on our APIs yet?

yeastplume:
It's @antiochp 's fault.. he started it!

One of his previous PRs used 'v2' to distinguish between the previous ApiEndpoint (v1) and just setting up the router manually. There's still one endpoint that needs to be converted, but once that's done, I agree.. they should all just be v1. I also didn't want to rename v2 back to v1 since his wallet code is now expecting v2.

Suggest merging this for now and then we can take an action to change them all back to v1 by consensus?

antiochp:
Ha! - yeah they should all just be v1 at some point.
I just wanted a quick and easy way of keeping the handler based endpoints separate.

ignopeverell:
Ah sorry for missing that. Okay I'll merge and we can change back to v1 afterward.

PR #222, opened 2017-11-01
fix the wallet receiver api

antiochp:
This fixes the issue we were seeing with the wallet receiver api.

* POST to /receive/transaction (was GET)
* introduce `wallet::client::send_partial_tx`
* simplify json handling in wallet receiver

### DIFF
--- a/api/src/client.rs
+++ b/api/src/client.rs
@@ -40,19 +40,16 @@ where
 /// object as body on a given URL that returns a JSON object. Handles request
 /// building, JSON serialization and deserialization, and response code
 /// checking.
-pub fn post<'a, IN, OUT>(url: &'a str, input: &IN) -> Result<OUT, Error>
+pub fn post<'a, IN>(url: &'a str, input: &IN) -> Result<(), Error>
 where
 	IN: Serialize,
-	for<'de> OUT: Deserialize<'de>,
 {
 	let in_json = serde_json::to_string(input).map_err(|e| {
 		Error::Internal(format!("Could not serialize data to JSON: {}", e))
 	})?;
 	let client = hyper::Client::new();
-	let res = check_error(client.post(url).body(&mut in_json.as_bytes()).send())?;
-	serde_json::from_reader(res).map_err(|e| {
-		Error::Internal(format!("Server returned invalid JSON: {}", e))
-	})
+	let _res = check_error(client.post(url).body(&mut in_json.as_bytes()).send())?;
+	Ok(())
 }
 
 // convert hyper error and check for non success response codes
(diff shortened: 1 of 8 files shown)


api/src/client.rs
```diff
@@ -40,19 +40,16 @@ where
 /// object as body on a given URL that returns a JSON object. Handles request
 /// building, JSON serialization and deserialization, and response code
 /// checking.
-pub fn post<'a, IN, OUT>(url: &'a str, input: &IN) -> Result<OUT, Error>
+pub fn post<'a, IN>(url: &'a str, input: &IN) -> Result<(), Error>
```
antiochp: There's a mismatch right now between the old style Api stuff and the newer Iron handlers.
We don't have any api clients that expect anything but `()` as a response so I have simplified this for now.
We were seeing issues parsing empty responses from Iron handlers here.

src/bin/grin.rs
```diff
@@ -372,7 +372,7 @@ fn wallet_command(wallet_args: &ArgMatches) {
 			let mut contents = String::new();
 			file.read_to_string(&mut contents)
 				.expect("Unable to read transaction file.");
-			wallet::receive_json_tx(&wallet_config, &keychain, contents.as_str()).unwrap();
+			wallet::receive_json_tx_str(&wallet_config, &keychain, contents.as_str()).unwrap();
```
antiochp: Split `receive_json_tx` out into `receive_json_tx` and `receive_json_tx_str`.
One takes a `JSONPartialTx` and one take json string and first converts it to `JSONPartialTx`.
We need the string version for wallet send/receive via a file.
The other works cleaner (and simpler) via the api (wth parsing handled via bodyparser).

wallet/src/receiver.rs
```diff
@@ -65,25 +74,18 @@ pub struct WalletReceiver {
...
-						)
-					})
-					.unwrap();
-
-				Ok(Response::with(status::Ok))
-			}
-			_ => Ok(Response::with(
-				(status::BadRequest, format!("Incorrect request data.")),
-			)),
+		let struct_body = req.get::<bodyparser::Struct<JSONPartialTx>>();
```
antiochp: Use bodyparser to turn the request body directly into a JSONPartialTx (and avoid passing the json string representation around).

wallet/src/server.rs
```diff
@@ -37,9 +37,9 @@ pub fn start_rest_apis(wallet_config: WalletConfig, keychain: Keychain) {
 	};
 
 	let router = router!(
-    receive_tx: get "/receive/transaction" => receive_tx_handler,
+		receive_tx: post "/receive/transaction" => receive_tx_handler,
```
antiochp: This was the first problem - it needed to be registered as a `POST` and not `GET`...

antiochp:
Making an executive decision and going to merge this in (wallet api is effectively broken for everyone without this).

PR #233, opened 2017-11-03
iterate once over txs in a block

antiochp:
Spent some time going over (and getting clear in my head) how a `TxKernel` relates to a `Transaction` - particularly in the context of recent discussions around multi-kernel txs.

I took the liberty to reworking some of the block building code to make kernel building a bit more explicit and cleaned up some of the error handling around blocks and txs.

* only iterate once over the txs when building a block
  * collect all inputs, outputs and kernels in a single pass
* split out kernel building from `verify_sig`
  * `verify_sig` now returns a commitment (the excess)
  * `build_kernel` is now more explicit in what it does (and what it needs)
* call `validate` and not just `verify_sig` when building all the kernels for a block


ignopeverell:
Yes, looks cleaner, thanks!

Issue #237, opened 2017-11-03
Profile retrieving UTXOs from the data store


ignopeverell:
Running `grin wallet info` can take a very long time when there are hundreds or thousands of UTXOs. Normally reading those from rocksdb (even including related headers) should be extremely fast. Profile it and check what's wrong.

yeastplume:
Well that's interesting, from an initial perf run it looks like the bulk of time is spent parsing a hex string

-    1.99%     0.00%  grin     grin           [.] grin_api::handlers::UtxoHandler::get_utxo                                                                                                                                                                                
   - grin_api::handlers::UtxoHandler::get_utxo                                                                                                                                                                                                                             
      - 1.31% grin_util::hex::from_hex                                                                                                                                                                                                                                     
           grin_util::hex::split_n                                                                                                                                                                                                                                         
           core::iter::iterator::Iterator::collect                                                                                                                                                                                                                         
           <collections::vec::Vec<T> as core::iter::traits::FromIterator<T>>::from_iter                                                                                                                                                                                    
           <collections::vec::Vec<T> as collections::vec::SpecExtend<T, I>>::from_iter                                                                                                                                                                                     
         + <collections::vec::Vec<T> as collections::vec::SpecExtend<T, I>>::spec_extend                                                                                                                                                                                   
      - 0.67% grin_chain::chain::Chain::get_unspent                                                                                                                                                                                                                        
           grin_chain::sumtree::SumTrees::is_unspent                                                                                                                                                                                                                       
           <grin_chain::store::ChainKVStore as grin_chain::types::ChainStore>::get_output_pos                                                                                                                                                                              
           grin_store::Store::get_ser                                                                                                                                                                                                                                      
           grin_store::Store::get_ser_limited                                                                                                                                                                                                                              
           grin_store::Store::get                                                                                                                                                                                                                                          
           rocksdb::db::<impl rocksdb::DB>::get                                                                                                                                                                                                                            
           rocksdb::db::<impl rocksdb::DB>::get_opt                                                                                                                                                                                                                        
           rocksdb_get                                                                                                                                                                                                                                                     
           rocksdb::DB::Get                                                                                                                                                                                                                                                
           rocksdb::DB::Get                                                                                                                                                                                                                                                
           rocksdb::DBImpl::Get                                                                                                                                                                                                                                            
           rocksdb::DBImpl::GetImpl                                                                                                                                                                                                                                        
           rocksdb::Version::Get                                                                                                                                                                                                                                           
[...]

PR #238, opened 2017-11-06
Wallet output selection performance

yeastplume:
Half the solution to: https://github.com/mimblewimble/grin/issues/237

This should solve the wallet-side performance issue during a wallet_info command, it just allows for providing the key index stored in the wallet when retrieving a commit, as opposed to iterating to find it each time. 

Would appreciate a quick review to ensure I haven't misunderstood anything.

Server side still seems sluggish when actually selecting the UTXOs via the API, (as in there's a scroll delay as it outputs each of them in the log) so that's the next bit to address.

### DIFF
--- a/keychain/Cargo.toml
+++ b/keychain/Cargo.toml
@@ -7,6 +7,7 @@ authors = ["Antioch Peverell"]
 byteorder = "~1"
 blake2-rfc = "~0.2.17"
 rand = "~0.3"
+slog = { version = "^2.0.12", features = ["max_level_trace", "release_max_level_trace"] }
 serde = "~1.0.8"
 serde_derive = "~1.0.8"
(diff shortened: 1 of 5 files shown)


yeastplume:
More changes to speed up the server side somewhat, mostly the inclusion of a lazy static reference to a libsecp instance, and removed the instance from SumCommit. This static is only exposed via a function that explicitly re-randomises the context on every call (for side-channel attack resistance), so there hopefully shouldn't be any disadvantages to calling the lib in this manner.

yeastplume:
Yes, all good, was just waiting for a bit of a review.

ignopeverell:
Ok cool. I usually dislike statics but for this it makes sense, especially given that the context can be randomized. The small API tweak and the index on the keychain also make sense.

PR #240, opened 2017-11-07
a single character typo fix.

JFrankfurt:
"up on" => "upon" because the first sounds awkward. ¯\\__(ツ)_/¯


yeastplume:
Thanks for pointing this out, I've made the change while doing other things and updated the doc.

Issue #232, opened 2017-11-03
Wallet amount formatting (user-facing input and output), wallet Quality-of-life issues


yeastplume:
A couple of simple things on the wallet to be done, might be good for anyone looking to start getting familiar with the code:

-In advance of testnet, I think wallet amounts as well as user input and output should be converted to use decimal places. Currently 50 grins must be input as 5000000000, whereas the user should be able to input and see 50 or 50.00 or similar.

-Change the (NotEnoughFunds) error to be a recoverable message nicely printed to the client, instead of a panic dump. 

-Provide a response to the sender when the receiving a transaction via http post. Right now nothing is retuned to the sender, leaving sender to wonder whether the transaction was sent successfully.

yeastplume:
addressed in https://github.com/mimblewimble/grin/pull/241

PR #241, opened 2017-11-07
Wallet amounts

yeastplume:
Basically https://github.com/mimblewimble/grin/issues/232 

-You now enter coins as 59.0032 etc and they're displayed as such in wallet info
-A couple of functions to do conversions, and tests in place.. should be okay for the time being, though more future-proofing might be needed at some stage
-Better message shown client-side when sending wallet transaction
-NotEnoughFunds displays an info message with current total funds displayed, all other errors still panic
-and some rustfmt spam!!!


yeastplume:
Just going to merge, nothing game-breaking here

ignopeverell:
Just had a look, seems all good to me.

PR #244, opened 2017-11-08
more robust peer handling

antiochp:
This PR resolves both duplicate peers and "self peers" in the `v1/peers/connected` list.
Also makes sure a seedless node monitors any peers that start talking to us.

* track connected peers in a map (avoid duplicates)
* reuse the handshake in the server (to persist nonces for self connection avoidance)
* monitor peers even in seedless mode (peers may still talk to us as a seed)
* fixup simulnet tests (seeds/peers now more robust in the tests)
* start the syncer if we know any seeds *or* peers
* exit the sync cleanly even if we cannot connect to any peers to sync with

### DIFF
--- a/grin/src/server.rs
+++ b/grin/src/server.rs
@@ -144,8 +144,7 @@ impl Server {
 			_ => {}
 		}
 
-
-		// now attempt to sync unless we have no known seeds and no known peers
+		// If we have any known seeds or peers then attempt to sync.
 		if config.seeding_type != Seeding::None || peer_store.all_peers().len() > 0 {
 			let sync = sync::Syncer::new(shared_chain.clone(), p2p_server.clone());
 			net_adapter.start_sync(sync);
--- a/grin/src/server.rs
+++ b/grin/src/server.rs
@@ -144,10 +144,8 @@ impl Server {
 			_ => {}
 		}
 
-		if config.seeding_type != Seeding::None {
-			let sync = sync::Syncer::new(shared_chain.clone(), p2p_server.clone());
-			net_adapter.start_sync(sync);
-		}
+		let sync = sync::Syncer::new(shared_chain.clone(), p2p_server.clone());
+		net_adapter.start_sync(sync);
 
 		evt_handle.spawn(p2p_server.start(evt_handle.clone()).map_err(|_| ()));
 
--- a/grin/src/server.rs
+++ b/grin/src/server.rs
@@ -124,6 +124,7 @@ impl Server {
 					LOGGER,
 					"No seed configured, will stay solo until connected to"
 				);
+				seed.monitor_only(evt_handle.clone());
 			}
 			Seeding::List => {
 				seed.connect_and_monitor(
--- a/grin/src/server.rs
+++ b/grin/src/server.rs
@@ -122,9 +122,12 @@ impl Server {
 			Seeding::None => {
 				warn!(
 					LOGGER,
-					"No seed configured, will stay solo until connected to"
+					"No seed(s) configured, will stay solo until connected to"
 				);
-				seed.monitor_only(evt_handle.clone());
+				seed.connect_and_monitor(
+					evt_handle.clone(),
+					seed::predefined_seeds(vec![]),
+				);
 			}
 			Seeding::List => {
 				seed.connect_and_monitor(
@@ -133,7 +136,10 @@ impl Server {
 				);
 			}
 			Seeding::WebStatic => {
-				seed.connect_and_monitor(evt_handle.clone(), seed::web_seeds(evt_handle.clone()));
+				seed.connect_and_monitor(
+					evt_handle.clone(),
+					seed::web_seeds(evt_handle.clone()),
+				);
 			}
 			_ => {}
 		}
--- a/grin/src/server.rs
+++ b/grin/src/server.rs
@@ -144,8 +144,12 @@ impl Server {
 			_ => {}
 		}
 
-		let sync = sync::Syncer::new(shared_chain.clone(), p2p_server.clone());
-		net_adapter.start_sync(sync);
+
+		// now attempt to sync unless we have no known seeds and no known peers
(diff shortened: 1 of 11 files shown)


p2p/src/store.rs
```diff
@@ -94,11 +95,8 @@ impl PeerStore {
 	}
 
 	pub fn save_peer(&self, p: &PeerData) -> Result<(), Error> {
-		// we want to ignore any peer without a well-defined ip
```
antiochp: I *think* this is safe to remove now that we are correctly extracting ip addresses from the connection.
This was making testing locally really difficult as we were not saving anything with 127.0.0.1 ip address (even if we were testing locally where this was actually desired).

p2p/src/server.rs
```diff
@@ -63,7 +64,8 @@ impl NetAdapter for DummyAdapter {
 pub struct Server {
 	config: P2PConfig,
 	capabilities: Capabilities,
-	peers: Arc<RwLock<Vec<Arc<Peer>>>>,
+	peers: Arc<RwLock<HashMap<SocketAddr, Arc<Peer>>>>,
+	handshake: Arc<Handshake>,
```
antiochp: We want to maintain a single handshake for the server so we can track all generated nonces.

p2p/src/server.rs
```diff
@@ -91,7 +94,7 @@ impl Server {
 		let socket = TcpListener::bind(&addr, &h.clone()).unwrap();
 		warn!(LOGGER, "P2P server started on {}", addr);
 
-		let hs = Arc::new(Handshake::new());
+		let handshake = self.handshake.clone();
```
antiochp: Reuse the existing handshake here.

p2p/src/server.rs
```diff
@@ -174,13 +178,13 @@ impl Server {
...
 				// connect to the peer and add it to the server map, wiring it a timeout for
-	// the handhake
+				// the handshake
 				let connect = Peer::connect(
 					socket,
 					capab,
 					total_diff,
 					self_addr,
-					&Handshake::new(),
+					handshake.clone(),
```
antiochp: And reuse the handshake here as well so the nonces are preserved.

grin/src/server.rs
```diff
@@ -124,6 +124,7 @@ impl Server {
 					LOGGER,
 					"No seed configured, will stay solo until connected to"
 				);
+				seed.monitor_only(evt_handle.clone());
```
antiochp: We want the server to monitor peers even if we start without any seeds (somebody else may communicate with us).

antiochp:
Looks like `grin_grin` tests are failing for some reason.
Investigating.

antiochp:
Tested this pretty extensively on a "private devnet".
Running overnight we have multiple nodes all still in sync.
`v1/peers/connected` and `v1/peers/all` appear to reliably show peering working as expected across multiple scenarios of node stop|start|restart behavior.

* Stopping a node and immediately restarting it.
* Stopping a node and waiting for peers to see it drop (and remove it from their connected list).
* Stopping all nodes but one (both single seed and single non-seed).
* Stopping and restarting the single seed node (non-seed nodes still peering correctly).
* Stopping a node and seeing the seed node update its peer list correctly.
* Stopping and clearing `.grin` forcing a full re-sync.
* Stopping and restarting in middle of re-sync.

@ignopeverell ready for review.

grin/src/server.rs
```diff
@@ -122,7 +122,11 @@ impl Server {
 			Seeding::None => {
 				warn!(
 					LOGGER,
-					"No seed configured, will stay solo until connected to"
+					"No seed(s) configured, will stay solo until connected to"
+				);
+				seed.connect_and_monitor(
```
antiochp: I have been testing with nodes that do not necessarily reach out to peer with others - but are themselves treated as a seed (or a known peer).
So - want to make sure this node monitors its own peer list in case somebody else peers with it (and subsequently disappears).

grin/src/server.rs
```diff
@@ -132,15 +136,16 @@ impl Server {
...
 			}
 			_ => {}
 		}
 
-		if config.seeding_type != Seeding::None {
-			let sync = sync::Syncer::new(shared_chain.clone(), p2p_server.clone());
-			net_adapter.start_sync(sync);
-		}
+		let sync = sync::Syncer::new(shared_chain.clone(), p2p_server.clone());
+		net_adapter.start_sync(sync);
```
ignopeverell: I understand you might still want to monitor connection but why seeding in this case?

antiochp: I'm not totally sure I understand the question - but are you asking about why I'm proposing to run the sync _from_ a node with no seeds?

This is what i was seeing in testing - 

3 nodes -
* S (seed)
* M (miner)
* N (regular node)

Let them all run for an hour -
* S: height 60
* M: height 60
* N: height 60

Now stop the seed and let the network run for a few minutes -
* S: height 60
* M: height 65
* N: height 65

Restarting the seed (current behavior, no sync running on S) -
* S: height 60
* M: height 65
* N: height 65

Then stop the regular node N, let the network run for a few mins -
* S: height 60
* M: height 70
* N: height 65

When N restarts it successfully peers against S and M and will request blocks from both as it attempts to resync.

S is out of sync and will never resync.
N re-syncs slowly as it is is unaware that S is not a useful peer to sync from.

ignopeverell: Makes sense, thanks for clarifying. However this will still be a problem for a single mining node, as sync will try to get peers forever and block the start of the miner loop.

antiochp: Ahh ok - I see your concern. Let me take another look at this.

ignopeverell: One suggestion: instead of removing that condition, just add a check on the number of peers in db and only skip sync if we have no seeds and no known peers.

antiochp:
@ignopeverell added in the check like you suggested.
Also added a check near the start of the sync run to exit out cleanly if we have no connected peers to sync from.
Testing various combinations of starting/stopping/restarting nodes - sync appears to run as expected now.

Issue #245, opened 2017-11-09
Replace Secp256k1::with_caps calls with secp_static


antiochp:
We still have a few places where we call `Secp256k1::with_caps(secp::ContextFlag::Commit)`.
It probably makes sense to consolidate all these into the new `secp_static`.

antiochp:
Resolved.

PR #250, opened 2017-11-09
use the static secp instance everywhere (except the wallet)

antiochp:
We were creating new instances of `secp` in various places via 
```
Secp256k1::with_caps(secp::ContextFlag::Commit);
```

This PR replaces these with calls to -
```
let secp = static_secp_instance();
let secp = secp.lock().unwrap();
```

Also wraps some usage up in code blocks to limit how long we keep the lock for secp access.
And we can simplify a lot of the code by not passing around raw instances of `secp` into `verify` etc.

Note: we are trading the overhead of re-initializing secp against some added complexity around needing to manage locks and contention for access to the static secp instance.
We may want to consider making these threadlocal (or possibly a pool of instances) at some point if we do see contention.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -16,7 +16,6 @@
 
 use std::sync::{Arc, RwLock};
 
-use util::secp;
 use time;
 
 use core::consensus;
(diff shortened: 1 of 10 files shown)


antiochp:
Resolves #245

yeastplume:
I'm actually very okay with statics in instances like this, especially as we've found that libsecp initialisation is an expensive process. So long as it's ensured the randomisation is called whenever retrieved (which I think it the main benefit of initialising it every time it's used,) I don't think this would cause further problems, and will possibly give a performance benefit in places. The only thing I can think of is that if it's loaded into a static place on the heap, it makes it slightly easier for a malicious process to scanning it, but I don't think that's a blocker so long as the randomisation is used every time.

I think a pool is a better bet if there turns out to be contention.

antiochp:
Ok - I'm going to merge this.
@ignopeverell we can definitely revert this if you have any concerns about reusing the static instance everywhere.

ignopeverell:
I'm good with this.

Issue #243, opened 2017-11-08
miner mining blocks but wallet not seeing anything later than height n


antiochp:
Ran a very simple "private testnet" locally.
* 1 seed node
* 1 mining node (single seed node in seed list)
* 1 non-mining node (single seed in seed list)

This ran overnight.
In the morning two nodes were at height 269 and one node was stuck at approx 85 (forgot to write it down). Stopped and restarted various nodes to see if I could get them all to sync consistently.

All 3 nodes are now at height 269 - 
```
curl 18.220.204.168:13413/v1/chain
{"height":269,"last_block_pushed":"496d074e6abf14ea206a4698e484041fc1ca98b1349847f257cd3740dca44f08",
"prev_block_to_last":"a3529afb0b0c71e6572634124b086f92153150aed9a9e141d9028aa7126b7e8f","total_difficulty":29552}                                       

curl 52.14.128.150:13413/v1/chain 
{"height":269,"last_block_pushed":"496d074e6abf14ea206a4698e484041fc1ca98b1349847f257cd3740dca44f08",
"prev_block_to_last":"a3529afb0b0c71e6572634124b086f92153150aed9a9e141d9028aa7126b7e8f","total_difficulty":29552}%                                        

curl 18.216.195.126:13413/v1/chain
{"height":269,"last_block_pushed":"496d074e6abf14ea206a4698e484041fc1ca98b1349847f257cd3740dca44f08",
"prev_block_to_last":"a3529afb0b0c71e6572634124b086f92153150aed9a9e141d9028aa7126b7e8f","total_difficulty":29552}%
```

Peer lists are not consistent (some self duplicates, some nodes not speaking to each other) - 
```
curl 18.220.204.168:13413/v1/peers/connected
[{"capabilities":{"bits":7},"user_agent":"MW/Grin 0.1","version":1,"addr":"18.220.204.168:13414","total_difficulty":29552},
{"capabilities":{"bits":7},"user_agent":"MW/Grin 0.1","version":1,"addr":"18.220.204.168:13414","total_difficulty":29552},
{"capabilities":{"bits":7},"user_agent":"MW/Grin 0.1","version":1,"addr":"18.216.195.126:13414","total_difficulty":22396}]% 

# seed node
curl 52.14.128.150:13413/v1/peers/connected
[{"capabilities":{"bits":7},"user_agent":"MW/Grin 0.1","version":1,"addr":"18.216.195.126:13414","total_difficulty":22396}]% 

[...]

PR #251, opened 2017-11-09
use consensus rule that minimum difficulty is 10

antiochp:
See #243 for a mining node apparently dropping to `difficulty==0` and then getting jammed on sumtree rollbacking.

This PR ensure `next_difficulty` never returns anything lower than `MINIMUM_DIFFICULTY`.


antiochp:
@ignopeverell - thoughts/feedback on this?
Is `10` the lowest valid difficulty?

ignopeverell:
That looks good, thanks, and sorry for the late merge.

PR #252, opened 2017-11-09
rename JSONPartialTx -> PartialTx

antiochp:
`JSONPartialTx` is a hold-over from when we passed json representations around in the ApiEndpoint. We're now consistently using serde for json serialization/deserialization.
Rename to `PartialTx` to keep things clean.


antiochp:
This is minor - self merging.

PR #253, opened 2017-11-10
cache key_id->derivation in the keychain

antiochp:
Introduce a new `key_derivation_cache` in the keychain so we can maintain a cache of `derivation` by `Identifier`.
This allows us to efficiently lookup the `derivation` and then quickly derive the key based on this.

This seemed better than caching the keys themselves as the `derivation` itself is significantly smaller than the full extkey itself.

From visual inspection of the logs with `Trace` enabled - we see a lot of "cache hits" as keys are used multiple times.

### DIFF
--- a/keychain/src/keychain.rs
+++ b/keychain/src/keychain.rs
@@ -14,6 +14,7 @@
 
 use rand::{thread_rng, Rng};
 use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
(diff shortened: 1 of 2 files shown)


antiochp:
Tested this locally pretty extensively - wallet working as expected for both coinbase rewards and sending/receiving regular txs.
Going to merge.

PR #260, opened 2017-11-12
add premined pow solution for cuckoo16/UserTesting

antiochp:
faster startup time for `UserTesting`

Related - We probably need to update the genesis nonce and pow solution for `Production` also.


antiochp:
Going to merge this. 
Changes are pretty straightforward.
Tested locally to ensure we can safely start both mining and non-mining nodes up with this.

PR #269, opened 2017-11-15
Genesis block for known chains, mine dev chains

ignopeverell:
Renamed mining parameter mode to chain type, with existing types
of CI testing, user testing, testnet1 or mainnet. The public
chains (testnet1 and mainnet) come with their fully pre-defined
genesis block.

Still need to set the nonce and cycle for testnet1 genesis.

### DIFF
--- a/core/src/genesis.rs
+++ b/core/src/genesis.rs
@@ -17,18 +17,16 @@
 use time;
 
 use core;
-use consensus::MINIMUM_DIFFICULTY;
+use consensus;
 use core::target::Difficulty;
 use global;
 
-/// Genesis block definition. It has no rewards, no inputs, no outputs, no
-/// fees and a height of zero.
-pub fn genesis() -> core::Block {
-	let proof_size = global::proofsize();
-	let empty_hash = core::hash::Hash([0x0; 32]);
+/// Genesis block definition for development networks. The proof of work size
+/// is small enough to mine it on the fly, so it does not contain its own
+/// proof of work solution. Can also be easily mutated for different tests.
+pub fn genesis_dev() -> core::Block {
 	core::Block {
 		header: core::BlockHeader {
-			version: 1,
 			height: 0,
 			previous: core::hash::Hash([0xff; 32]),
 			timestamp: time::Tm {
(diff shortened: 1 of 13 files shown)


ignopeverell:
@antiochp @yeastplume if you guys can have a look, just want to make sure you're okay with the approach.

core/src/genesis.rs
```diff
@@ -17,18 +17,16 @@
...
 
-/// Genesis block definition. It has no rewards, no inputs, no outputs, no
-/// fees and a height of zero.
-pub fn genesis() -> core::Block {
-	let proof_size = global::proofsize();
-	let empty_hash = core::hash::Hash([0x0; 32]);
+/// Genesis block definition for development networks. The proof of work size
+/// is small enough to mine it on the fly, so it does not contain its own
+/// proof of work solution. Can also be easily mutated for different tests.
+pub fn genesis_dev() -> core::Block {
```
antiochp: Is this `UserTesting`? I think we should define the nonce and pow for this one (otherwise it can take a couple of minutes to startup a UserTesting node (mining or non-mining).

core/src/genesis.rs
```diff
@@ -37,13 +35,56 @@ pub fn genesis() -> core::Block {
...
+		},
+		inputs: vec![],
+		outputs: vec![],
+		kernels: vec![],
+	}
+}
+
+/// Placeholder for mainnet genesis block, will definitely change before
+/// release so no use trying to pre-mine it.
+pub fn genesis_main() -> core::Block {
```
antiochp: Maybe just be explicit about each genesis block? `main|testnet1|user_testing|automated_testing`?

yeastplume:
Yep, all looks fine and makes sense. I can throw the yeastmonster+PC box at the testnet1 nonce and cycle to fill that out shortly.

ignopeverell:
@yeastplume that would be great. I'm starting to worry a bit at having testnet1 at cuckoo30 now, given that we don't have a GPU mean miner yet.

core/src/genesis.rs
ignopeverell: There is a nonce for this in the globals. With a proper nonce, it's just a single try so it doesn't seem we would need the solution?

core/src/genesis.rs
ignopeverell: Seemed that everything non-public could share the same, as they'd have the exact same definition.

core/src/genesis.rs
antiochp: Oh I see - thought we still needed the pow solution in addition to the nonce in there. :+1:

Issue #264, opened 2017-11-14
network splits and 2 miners are unaware of each other - what happens?


antiochp:
@ignopeverell from testing #244 it is possible to get a couple of miners into a state where neither know about each other.
Artificial example - they are both seeded off each other but we run them alternately with the other stopped.
The end result is each mining node has a chainstate that is the consistent up to and including `height n` but they both (independently) mine `block n+1` onward.

What is in place to resolve this when the network fixes itself and the nodes can communicate again?

The sync process results in orphan blocks (node with lower difficulty syncing headers from node with higher difficulty).

I'm following through the `locate_headers` code path.
I see the node requesting what appears to be the correct headers based on the locators.
I see it receiving the headers as requested.
But it does not appear to do the right thing with these to "un-fork" at the appropriate height.

antiochp:
Some more context. We see this in the logs of the node that is attempting to sync (peer difficulty is > self difficulty) on startup - 

```
Nov 14 12:53:34.725 DEBG Sync: peer 4974 vs us 4309
Nov 14 12:53:34.725 DEBG Sync: blocks to download 0, block downloading 0
Nov 14 12:53:34.725 DEBG Loc heights: [81, 79, 77, 73, 65, 49, 17]
Nov 14 12:53:34.729 DEBG Asking peer 127.0.0.1:13414 for more block headers starting from 985ce32d at 81.
Nov 14 12:53:34.732 INFO Received 17 block headers
Nov 14 12:53:34.732 INFO Starting validation pipeline for block header 196c174c at 74.
Nov 14 12:53:34.733 INFO Received unfit block header 196c174c at 74: already in store.
Nov 14 12:53:34.733 INFO Starting validation pipeline for block header 3cb4dd99 at 75.
Nov 14 12:53:34.734 INFO Received unfit block header 3cb4dd99 at 75: already in store.
Nov 14 12:53:34.734 INFO Starting validation pipeline for block header a46ea7fb at 76.
Nov 14 12:53:34.735 INFO Received unfit block header a46ea7fb at 76: already in store.
Nov 14 12:53:34.735 INFO Starting validation pipeline for block header b955b4dc at 77.
Nov 14 12:53:34.736 INFO Received unfit block header b955b4dc at 77: already in store.
Nov 14 12:53:34.736 INFO Starting validation pipeline for block header ad096e49 at 78.
Nov 14 12:53:34.737 INFO Received unfit block header ad096e49 at 78: already in store.
[...]

PR #271, opened 2017-11-15
bad case of trying to rewind to block header height 0

antiochp:
* Always include height 0 (genesis) in the locator (to guarantee at least one header hash in common between the nodes)
* `ext.rewind` now takes a block (so we can handle it being empty if necessary)

To reproduce the issue - 
1) mine two completely independent chains 
    * clear out .grin dirs for miner1 and miner2
    * run miner1 for a couple of blocks
    * stop miner1
    * run miner2 for a couple of blocks
2) then try and sync them up (start them both up, may need to restart one with less total work)

See related - #264

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -237,13 +237,25 @@ fn validate_block(
 			}
 		}
 
-		// rewind the sum trees up the forking block, providing the height of the
-  // forked block and the last commitment we want to rewind to
+		// rewind the sum trees up to the forking block, providing the height of the
+		// forked block and the last commitment we want to rewind to
 		let forked_block = ctx.store.get_block(&current)?;
+
+		debug!(
+			LOGGER,
+			"validate_block: forked_block: {} at {}",
+			forked_block.header.hash(),
+			forked_block.header.height,
+		);
+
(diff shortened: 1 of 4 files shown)


antiochp:
@ignopeverell I think this is ready for review and I _think_ this fixes the issue I was seeing with mining nodes diverging.
It definitely appears to fix the pathological case where the chains diverge from the genesis (two miners mining in isolation, then later attempting to sync).

ignopeverell:
I'm wary of `rewind_to_genesis` because it assumes the genesis block is empty, which is true now but is unlikely to be true for testnetX with X>1 and mainnet. It may be better to set the rewind positions to 0 if none have been found and relax the `height > 0` condition.

antiochp:
Not sure I 100% follow but are you saying there may be a valid situation where we have a genesis block that contains an output?
Interesting - I'd never considered this.

ignopeverell:
I'm not 100% sure why but my guts tell me that would be better. I don't really like the idea of genesis not setting up the sum tree roots.

antiochp:
@ignopeverell take another look? I think this reads better now as well.

Issue #201, opened 2017-10-23
Cut-through in transaction pools


ignopeverell:
Allow for cut-through to append before transactions are included in a block. The difficulty is to resolve (or avoid) conflicts when the same transaction gets cut-through in 2 different ways by 2 different peers.

antiochp:
What's the benefit of doing this in the transaction pool rather than letting this happen during block compaction?
Is it that we have the opportunity to do this across a potentially larger set of txns (miner may not select txs that can cut-through)?
Or to have the opportunity to reduce the pool size?

ignopeverell:
There is no scalability gain, but there's a significant privacy gain as we end up with a non-interactive coinjoin with cut-through.

quentinlesceller:
Fixed by https://github.com/mimblewimble/grin/pull/1067 👍

PR #277, opened 2017-11-16
[WIP] tx now contains a kernel, step closer to supporting multiple kernels

antiochp:
It feels like a lot of interesting things can start to happen when we support txs with multiple kernels:
One example is coin-join + cut-through in tx pool (see #201).

The PR is an attempt to move us closer to supporting multiple kernels in a tx.
While still retaining the restriction of 1 tx <-> 1 kernel for now. 

* tx now contains a Vec of kernels
  * currently fails validation if kernels.len() != 1
  * ser/deser in place
* build the tx kernels in `build::transaction` (not when we mine the block)
* validating the tx now verifies the kernel(s) and requires the kernel to have been built previously
* pull out `build::partial_transaction` from `build::transaction`
  * only build the kernel when we build the full transaction
  * `build::transaction` returns the tx (with a built kernel)
  * `build::partial_transaction` returns the tx and the blinding key

@ignopeverell interested in getting your thoughts/feedback on this - it is throwaway and may be down the wrong path entirely.

### DIFF
--- a/chain/tests/test_coinbase_maturity.rs
+++ b/chain/tests/test_coinbase_maturity.rs
@@ -102,7 +102,7 @@ fn test_coinbase_maturity() {
 	let prev = chain.head_header().unwrap();
 
 	let amount = consensus::REWARD;
-	let (coinbase_txn, _) = build::transaction(
+	let coinbase_txn = build::transaction(
 		vec![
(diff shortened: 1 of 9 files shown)


antiochp:
Closing - this was experimental. Superseded by #681.

PR #290, opened 2017-11-17
Minor reformatting.

dongcarl:
Reformatted some files I looked at according to `rustfmt` recommendations.

### DIFF
--- a/grin/src/server.rs
+++ b/grin/src/server.rs
@@ -91,12 +91,12 @@ impl Server {
 
 		let mut genesis_block = None;
 		if !chain::Chain::chain_exists(config.db_root.clone()) {
-      let chain_type = config.chain_type.clone();
-      if chain_type == global::ChainTypes::Testnet1 {
-        genesis_block = Some(genesis::genesis_testnet1());
-      } else {
-			  genesis_block = pow::mine_genesis_block(config.mining_config.clone());
-      }
+			let chain_type = config.chain_type.clone();
+			if chain_type == global::ChainTypes::Testnet1 {
+				genesis_block = Some(genesis::genesis_testnet1());
+			} else {
+				genesis_block = pow::mine_genesis_block(config.mining_config.clone());
+			}
 		}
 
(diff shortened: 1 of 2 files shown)


yeastplume:
Thanks for looking at these, but we generally wouldn't create a separate PR just for rust formatting issues. We have a pre-commit hook here that can be used to warn you if a file you're committing as part of some other work fails rustfmt. In that case you can just run rustfmt against the file (or fix manually if you like) and re-commit:

https://github.com/mimblewimble/grin/blob/master/doc/style.md

dongcarl:
@yeastplume I've actually been seeing a lot of formatting issues in the codebase. If I'm understanding you correctly, is the current process to wait for code changes and fix the formatting along with the code?

yeastplume:
Correct, preferably via invoking rustfmt instead of manually. However.. rustfmt default rules have changed from time to time, and I think we need to take another look at how well this is working (though it’s probably a bit lower priority right this moment). We can run rustfmt on the entire codebase at some point if necessary, but for now, feel free to tidy up as you go along. Be warned though that sometimes running rustfmt makes is difficult to see what’s actually been changed in github diff output (another reason we need to revisit this).

sesam:
I suggest a scheduled formatting timepoint. Maybe right after each release
tag. And Contrib guidelines that any new code should be rustfmt:ed, while
updated code if rustfmt:ed should be split to separate reformatting commits
from code change.

mvh
/Simon

ignopeverell:
All good suggestions. But as @yeastplume mentioned, rustfmt is a little unstable and formatting rules have changed multiple times, so we will need to address this as well. Everytime their formatting rules change, it makes a bit of a mess. Not to mention that it doesn't seem to be working on a subset of files anymore.

dongcarl:
@ignopeverell Thanks for clarifying 😄

sesam:
rustfmt +nightly does massacre on // code comments and spaces/tabs.

If someone knows about some parameters or settings to make rustfmt not touch comments and spaces/tabs

that would just be great

PR #291, opened 2017-11-17
Added --release flag to build documentation.

Kixunil:
The `--release` build creates significanltly faster binaries, which is
what probably most people want.


yeastplume:
Thanks for submitting this, but we definitely don't want people creating release binaries by default right now. It strips them of debug info, which means any stack backtraces will be stripped of line number making tracking down issues more difficult. In later releases we might advocate release builds, but for now it's much more important for us to be able to see debug info if necessary.

Kixunil:
Good point. In that case, shouldn't use of `export RUST_BACKTRACE=1` be advocated?

PR #300, opened 2017-11-17
use a short timeout when opening the connection to possible peers

antiochp:
lots of peers in the db are unavailable and we do not want to wait 60s for these
when trying to find our first 8 peers

also shuffle the peers from the db so we do not always try and connect to the poor node at `100.38.91.52`...

### DIFF
--- a/grin/src/seed.rs
+++ b/grin/src/seed.rs
@@ -118,7 +118,7 @@ impl Seeder {
 					if peers.len() > 0 {
 						debug!(
 							LOGGER,
-							"Got {} more peers from db, trying to connect.",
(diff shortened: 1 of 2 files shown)


antiochp:
@ignopeverell we get peers faster with the spawn changes but we still wait 60s for some of these connection attempts - this drops the timeout down to 500ms so we iterate over the peers a lot faster.

ignopeverell:
It's a good improvement but I think 500ms is unrealistic. That will fail all connections for someone with a slow network. I would opt for 5 sec as a middle ground.

antiochp:
Sounds good - will update, test, then merge.

PR #316, opened 2017-11-18
less chatty Testnet1 outputs

### LINKED ISSUE #281: Cleanup testnet log
Default should be to log to console at WARN. Also the "Starting validation pipeline..."message is noisy.

sesam:
Aimed at #281

- "Starting validation pipeline for " -> "Processing "
- stdout logging default = Warning
- ERR -> INFO for "Transaction rejected: Already in pool" -- only for netadapter transaction_received, which (hopefully!) only gets tx from peers, and lots of those will be duplicates, that's good and not an error.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -53,7 +53,7 @@ pub fn process_block(b: &Block, mut ctx: BlockContext) -> Result<Option<Tip>, Er
 
 	info!(
 		LOGGER,
-		"Starting validation pipeline for block {} at {} with {} inputs and {} outputs.",
+		"Processing block {} at {} with {} inputs and {} outputs.",
 		b.hash(),
 		b.header.height,
 		b.inputs.len(),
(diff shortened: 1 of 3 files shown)


grin/src/adapters.rs
```diff
@@ -58,7 +58,11 @@ impl NetAdapter for NetToChainAdapter {
 		);
 
 		if let Err(e) = self.tx_pool.write().unwrap().add_to_memory_pool(source, tx) {
-			error!(LOGGER, "Transaction rejected: {:?}", e);
+			if e.kind() == PoolError::AlreadyInPool {
+				info!(LOGGER, "Transaction rejected: {:?}", e);
```
ignopeverell: I think a debug here would be just fine. Thanks!

chain/src/pipe.rs
```diff
@@ -53,7 +53,7 @@ pub fn process_block(b: &Block, mut ctx: BlockContext) -> Result<Option<Tip>, Er
 
 	info!(
```
ignopeverell: These can be changed to debug as well. There's a log for when the block is added (or errors when it fails) which should be enough.

grin.toml
```diff
@@ -63,7 +63,7 @@ port = 13414
 log_to_stdout = true
 
 # Log level for stdout: Critical, Error, Warning, Info, Debug, Trace
-stdout_log_level = "Debug"
+stdout_log_level = "Warning"
```
cryptagoras: I would argue that `Info` should be the default at this stage, but minor thing.

kargakis: Agreed about "Info"

Issue #318, opened 2017-11-18
Suggestion: change wallet 'receive' to wallet 'listen'


andybellenie:
'Receive' is an action whereas 'listen' is passive so more accurately describes the wallet state. This also makes 'receive' available for reverse transactions that are initiated by the receiving wallet (if such a thing is possible)

dougyoung:
Hey all, I submitted a PR for this: https://github.com/mimblewimble/grin/pull/322

andybellenie:
LOL #323

dougyoung:
Ha, I see that too! Sorry if I'm stepping on your toes! Reviewing that one now.

sesam:
Looks like a good idea. "Listener" is the traditional name for a TCP listening service.

This change looks like it's a deprecation. There are some standards around how deprecations are usually done, to minimise confusion and "customer support" issues.

Nice-to-haves:
- add a warning! whenever the old key is found, or the new key is missing. To inform and help users to fix this. See config.rs, maybe right after `let decoded: ...` (around line 130).

- make it work also with the old key, for a while (typical deprecation behaviour, that old keys/api:s are available for the next ~2 minor version upgrades, so likely only inside Testnet1 milestone. So the old config would be accepted if the new is missing. But maybe not the command line thing, to keep it simple. Save some time for users, and some time for you @dougyoung and/or @andybellenie

andybellenie:
@sesam added to PR

antonio-fr:
Better to put `Arg::with_name("input")` back in receive, so handling a received file would be done in 'received', as before, not as 'listen'. 'wallet listen -i FILE' might be very confusing.
Also this requires change in the relevant documentation part.

andybellenie:
@antonio-fr see https://github.com/mimblewimble/grin/pull/337

antonio-fr:
Great, so it will be 'wallet file -f FILE' ?

andybellenie:
yes

PR #322, opened 2017-11-18
ISSUE-318: Change wallet receive to wallet listen where used passively

dougyoung:
This is meant as a response to the feedback given in issue #318.

### DIFF
--- a/grin/tests/framework/mod.rs
+++ b/grin/tests/framework/mod.rs
@@ -131,7 +131,7 @@ impl Default for LocalServerContainerConfig {
 
 
 /// A top-level container to hold everything that might be running
-/// on a server, i.e. server, wallet in send or receive mode
+/// on a server, i.e. server, wallet in send or listen mode
 
 pub struct LocalServerContainer {
 	// Configuration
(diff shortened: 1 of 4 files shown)


dougyoung:
Note that this requires updating any copies of `grin.toml` outside of the one contained in the root of the project from 

```
wallet_receiver_url = "http://..."
```

to:

```
wallet_listener_url = "http://..."
```

E.g. if a user followed the `cp /path/to/project/root/grin.toml .` instruction at [0] then they would need to update their `server/grin.toml` or see the following error:

```
thread 'main' panicked at 'Error parsing config file: Error parsing configuration file at /Users/thelostspore/projects/oss/grin-node/server/grin.toml - missing field `wallet_listener_url` for key `mining`', src/bin/grin.rs:82:2
```

[0] https://github.com/mimblewimble/grin/blob/master/doc/build.md#running-a-node

thelostspore:
`Tested ACK`

dougyoung:
Some of the documentation and comments are off after the merge of #323. This updated PR aims to semantically correct these instances.

Rebased off [milestone/testnet1](https://github.com/mimblewimble/grin/tree/milestone/testnet1).

Issue #286, opened 2017-11-17
endpoint to link all the other endpoints


yeastplume:
I have trouble remembering the exact format for all of the available endpoints and keep having to dig into the code to get them. Provide a single 'index' endpoint that displays all of the available endpoints. Can probably just serve up a simple html page from the disk, or if we want to get really fancy, probe all of the available endpoints from the router and build it up dynamically

ignopeverell:
The REST purists always point out that all endpoints should be discoverable through links. So maybe automatically have `/v1` list all endpoints that have been added to hyper?

antiochp:
Would be nice if the iron router allowed some kind of introspection to list the registered endpoints but it looks like all the routing internals are kept private.
Shouldn't be too much work to maintain our own state as we go through the routing initialization.

PR #324, opened 2017-11-18
Quick placeholder api route index (#286)

sesam:
This is a quick n dirty solution to have at least something to remind of existing routes.

Which routes have I missed?
Is this the only grin component with HTTP / REST api endpoints?

Missing nice to haves:
- Make it a HTML response with links and descriptions
- DRY up the code, possibly by replacing the router! macro with Router new(hashmap) and use same hashmap to populate the IndexHandler route list too.


api/src/handlers.rs
```diff
@@ -317,7 +332,19 @@ pub fn start_rest_apis<T>(
 			p2p_server: p2p_server.clone(),
 		};
 
+		let route_list = vec!(
+			"get /".to_string(),
+			"get /chain".to_string(),
+			"get /chain/utxos".to_string(),
+			"get /sumtrees/*".to_string(),
+			"get /pool".to_string(),
+			"get /pool/push".to_string(),
```
antiochp: This one is `post /pool/push`

api/src/handlers.rs
```diff
@@ -317,7 +332,19 @@ pub fn start_rest_apis<T>(
 			p2p_server: p2p_server.clone(),
 		};
 
+		let route_list = vec!(
+			"get /".to_string(),
+			"get /chain".to_string(),
+			"get /chain/utxos".to_string(),
+			"get /sumtrees/*".to_string(),
```
antiochp: Might be worth being explicit about these as well.

Issue #288, opened 2017-11-17
Add 'sent' to wallet info output


yeastplume:
It's going to become a bit difficult to track wallet send/receive issues, especially seeing as how it's difficult to see what's been sent out. Perhaps just total amount spent for now, but I think it would be desirable to have the entire send history available, which breaks down exactly how much  was sent (or the user attempted to send) and where the outputs went.

antiochp:
Maybe we can have an additional log file that the wallet can log stuff to?
That way we have an audit trail of activity from a users wallet.

antiochp:
Agreed on the difficulty of debugging issues - the whole "confidential transactions" thing makes it kind of impossible.

heunglee:
I am not sure if @sesam or someone else is working on this issue. 
If so, please disregard my comment.

Otherwise, I propose a initial draft of requirements and some questions to clarify.
it is based off from quoting @ignopeverell's statement in #326:
1) _I think the idea in #288 was to make this accessible to wallet info, which means a separate data file, similar to wallet.dat that captures transaction-level information. It could say something like "sent X coins on Y date to Z addres, outputs A, B, C where consumed and D change was created"._
2) _To be easily usable by the wallet it should be some JSON data. With that the wallet could display some info but also rollback some failed transactions for example._
3) _Maybe extending the format of wallet.dat would make the most sense._

Possible drawback of adding "sent" and "received" data into wallet.data
1. As time passes, the file size increases quickly
2. Increased rate of write locks
3. The current format of wallet.data can be impacted to incorporate JSON format (Root node of "outputs" can be changed?)

So having separate file for sent and received, such as wallet_sent.data and wallet_received.data?
Or extending wallet.data?

Formats:
Need to support JSON.

SENT Transaction
key_id, amount, date, destination, outputs, change, status (pending, success, failed)

RECEIVED Transaction
key_id, amount, date, sent_by(?) and ???

Impacts:
TBD

Target PR date:
TBD

sesam:
Good. I'm not working on this.

There's a code line
>				Ok(_) => { debug!(LOGGER, "{} coins sent using strategy {} to {}", amount.to_string(), selection_strategy, dest) },

[...]

PR #326, opened 2017-11-19
Log "nn.nnn coins spent using strategy ..." on log level Debug

sesam:
Issue #288 says that, for now, we'd want to log total amount sent.

Should I change any of
- log level = Debug
- include destination? Currently not listed for log file brevity and privacy

### DIFF
--- a/src/bin/grin.rs
+++ b/src/bin/grin.rs
@@ -435,7 +435,7 @@ fn wallet_command(wallet_args: &ArgMatches) {
 				(selection_strategy == "all"),
 			);
 			match result {
-				Ok(_) => {}, //success messaged logged internally
+				Ok(_) => { debug!(LOGGER, "{} coins sent using strategy {} to {}", amount.to_string(), selection_strategy, dest) }, //success messaged logged internally
 				Err(wallet::Error::NotEnoughFunds(_)) => {},
 				Err(e) => panic!(e),
 			};
--- a/util/src/logger.rs
+++ b/util/src/logger.rs
@@ -33,10 +33,11 @@ fn convert_log_level(in_level: &LogLevel) -> Level {
 }
 
 lazy_static! {
-	/// Flag to observe whether logging was explcitly initialised (don't output otherwise)
+	/// Flag to observe whether logging was explicitly initialised (don't output otherwise)
 	static ref WAS_INIT: Mutex<bool> = Mutex::new(false);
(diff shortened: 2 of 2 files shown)


ignopeverell:
I think the idea in #288 was to make this accessible to `wallet info`, which means a separate data file, similar to `wallet.dat` that captures transaction-level information. It could say something like "sent X coins on Y date to Z addres, outputs A, B, C where consumed and D change was created".

sesam:
tl;dr roadworks. Will try adding a new and separate outputs_spent_logger or maybe better transactions_logger to get going.

Oh I see. To log (at minimum) total amount spent via util::logger, using slog, is probably cleanest done by instantiating an additional logger, completely separate from the util::logger helper.

There less clean option would be to bolt in one extra drain that is craftily filtered. But that means also terminal and file output filters must be filtered, so the filtering latency will be applied to _all_ logging. Filtering by referencing the wanted logger instance from the beginning is of course more efficient, and a lot less messy.

# [slog](https://crates.io/crates/slog) quick review
* + nice
* + nifty drain abstraction (but "cost free", really?)
* + supports filtering by module and line number (via environment variables, so that feature is compiled in by default) as well as runtime log filtering (google: Drain for RuntimeLevelFilter)
* + key-values can be attached to each log line, and be saved machine-readable, which could be a nice way to get good data to use for batch- and CI verification of executed transactions vs final wallet.dat contents
* - likely overkill
* - any filtering latency/penalties are paid by all log! / debug! / ... calls
* - attack surface
* - release builds contain all code, whether statically filtered out (release_max_level_...) or not
* + (?) release builds contain all code, just like the core bitcoind builds also prefer to do
* - no log rotation support, nor compression. slog docs point to this being a suitable tag:easy tag:helpwanted future task)

I'm imagining we can find a simpler logging solution with smaller potential attack surface later on, or roll our own, though I'm not suggesting to refactor and disturb everyone's stashes and branches right now.

ignopeverell:
Okay sorry, I should have been more explicit. To be easily usable by the wallet it should be some JSON data. With that the wallet could display some info but also rollback some failed transactions for example. So I didn't mean "it could say" literally, I was just trying to describe the type of information that would fit in there.

ignopeverell:
Still going to merge this as having some minimal logging is better than nothing until we can have a more formal solution. Maybe extending the format of wallet.dat would make the most sense.

Issue #261, opened 2017-11-14
Limit number of inputs in a block


ignopeverell:
A miner could potentially build a block with *many* inputs and just one output, which would likely be very expensive to validate (each spent output has to be checked for existence). I haven't tried yet to calculate what would make sense based on block size and number of outputs but that should be straightforward. This is a "self-defense" measure.

PR #328, opened 2017-11-19
[CONSENSUS] Limit number of inputs in a block #261

sesam:
Add a consensus rule to limit about of inputs in a block (and also for each individual transaction, so we don't create a transaction that is invalid to ever be in a block!)

consensus::MAX_BLOCK_INPUTS.

Actual number should start out too high, since it can be softforked down but not up. Highest possible would be MAX_BLOCK_WEIGHT / BLOCK_INPUT_WEIGHT I guess? Ignoring the weight of a single output.

- Q: Can a "crafty" miner make a 40000 inputs transaction with a single kernel and a single output, or are there other consensus rules that could block that?

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -22,6 +22,7 @@ use util::secp::pedersen::{Commitment, RangeProof};
 use std::cmp::Ordering;
 use std::ops;
 
+use consensus;
 use core::Committed;
 use core::hash::Hashed;
 use core::pmmr::Summable;
@@ -72,6 +73,8 @@ pub enum Error {
 	/// Underlying Secp256k1 error (signature validation or invalid public
 	/// key typically)
 	Secp(secp::Error),
+	/// Restrict number of incoming inputs
+	TooManyInputs,
 }
 
 impl From<secp::Error> for Error {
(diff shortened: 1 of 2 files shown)


ignopeverell:
As you mentioned this should start high. The reason for this limit is to not have block verification take too long due to checking whether inputs match existing UTXOs. There is some impact on block commitment summing as well but that's CPU bound and not I/O bound, so we can likely safely assume that to be negligible.

So how fast can we retrieve UTXOs? Let's assume a 5MB/s disk random reads, which isn't that great but isn't catastrophic either. Being optimistic about output lookup, that's about 150,000 of them. So I'd go with the double of that.

core/src/core/transaction.rs
```diff
@@ -343,10 +346,14 @@ impl Transaction {
 	/// Validates all relevant parts of a fully built transaction. Checks the
 	/// excess value against the signature as well as range proofs for each
 	/// output.
+	// TODO: this is Consensus critical code. Move to consensus.rs ?
```
ignopeverell: A lot of code is consensus critical in `core` and pretty much all `pipe.rs` is as well. The idea with `consensus.rs` is more to hold constants and limits, as well as simple logic. But we can't fit 1/4 of the codebase in there :-)

sesam: Ok, removed for now. Breathing space until testnet2 might be good, alas habits will form. Soon enough it might become necessary to extract separate .rs files containing the consensus rules and unit tests that ensure that consensus isn't broken by some code refactoring. I realize that forks can be caused by almost anything. ”Le battement d'ailes d'un papillon au Brésil peut-il provoquer ...” - a bit of fun for the French among us

PR #337, opened 2017-11-19
Breakout listen -i to separate method

andybellenie:
This PR creates a new command for the wallet: file - which takes in the -f arg to process a json transaction. This keeps it separate from 'listen'. I've also moved the port argument to the listen command as it is only applicable there

See issue #318 

TODO: Add documentation for file method

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -35,6 +35,21 @@ use util;
 use util::LOGGER;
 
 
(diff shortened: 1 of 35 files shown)


ignopeverell:
Maybe better name for the PR :-)

ignopeverell:
Sorry for the lateness, got a little busy!

PR #338, opened 2017-11-19
Wallet Restore feature

yeastplume:
This adds a wallet restore command, which works as follows:

grin wallet -p password -k [max_key_derivations] restore

The max key derivations should be a number a bit larger than the number of outputs you think you have in your wallet, or set larger if you no longer have the older wallet.bak file. The restore feature works a bit faster than I'd thought it would at this stage, so setting this to a relatively high number like 10000 is okay.. it'll just take a moment to build a potential switch commitment cache at the start, but then go through outputs from the entire chain in linear time.

Note the wallet utxo endpoints have changed, split into chain/utxos/byids and chain/utxos/byheight

First pass for now, so will contain issues, however it won't overwrite a wallet.dat file (it insists it be empty before starting), so should be non-descructive for now. It's restored all of my testnet wallets that had been corrupted through failed transactions earlier.

Will also submit similar PR for testnet1 branch.

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -25,6 +25,7 @@ use serde_json;
 
 use chain;
 use core::core::Transaction;
+use core::core::hash::Hashed;
 use core::ser;
 use pool;
 use p2p;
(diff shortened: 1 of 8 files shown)


antiochp:
Yay! This looks awesome!  🚀

PR #339, opened 2017-11-19
Wallet Restore feature testnet1

yeastplume:
As per #338, for testnet1 branch


yeastplume:
Doesn't impact any critical code, so going to merge.

PR #345, opened 2017-11-20
rollback the change output in the wallet on tx failure

### LINKED ISSUE #287: Funds apparently locked/gone forever on transaction failure
Tried to send the first few transactions just now, all failed due to port issues on the receiver side. However, after a failure all of the swept outputs are permanently locked in the wallet (locked by previous transactions), and any funds that were supposedly sent but failed basically appear gone.

I don't think these cases are explicitly handled and we'll probably some design decisions to make on when to reclaim unclaimed partial transactions, but opening issue here before the floodgates open from testnet participants.

antiochp:
If we attempt to send a tx but it fails on the receiving side - we want to remove our `Unconfirmed` change output.
Also took the opportunity to clean up the user output when sending tx - pulled this out into `grin.rs` as this is part of the CLI - either on success on when it fails for various reasons.

Tested locally under various conditions - 

```
# insufficient funds to build tx
Nov 20 12:54:42.932 ERRO Tx not sent: insufficient funds (max: 449.976000000)
```

```
# error during api call to receiver
# change output is removed correctly from wallet here
Nov 20 12:55:11.530 ERRO Tx not sent: Hyper(Io(Error { repr: Os { code: 61, message: "Connection refused" } }))
```

```
# success
Nov 20 12:55:58.039 INFO Tx sent: 10.000000000 grin to http://127.0.0.1:13415 (strategy 'all')
```

Resolves #287.

### DIFF
--- a/src/bin/grin.rs
+++ b/src/bin/grin.rs
@@ -42,6 +42,7 @@ use daemonize::Daemonize;
 use config::GlobalConfig;
 use wallet::WalletConfig;
 use core::global;
+use core::core::amount_to_hr_string;
 use util::{init_logger, LoggingConfig, LOGGER};
 
 fn start_from_config_file(mut global_config: GlobalConfig) {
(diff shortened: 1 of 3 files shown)


antiochp:
Resolves #287.

PR #359, opened 2017-11-21
[WIP] Syncer keeps running in background

antiochp:
First attempt at getting `syncer` to run in the background and stay running ever after successful first sync.

Some observations - 
* the syncer in the net_adapter is lazily instantiated (but the OneTime never returns now)
* the syncer is not futures aware (and feels like it doesn't fit cleanly with rest of the server)
* there's a big circular dependency between syncer <-> net_adapter <-> grin server (hence the lazy instantiation of syncer?)
* we run `init_download` within the sync loop and this causes `blocks_to_download` to grow as we are not checking if we are asking for duplicate blocks
* chain_opts is not handled correctly - SYNC vs. NONE (just realized this)

That last commit that I reverted was an attempt to to "simplify" and replace the OneTime wrapping the syncer with an Arc<RwLock<>> but I think it went off the rails...

### DIFF
--- a/grin/src/adapters.rs
+++ b/grin/src/adapters.rs
@@ -81,13 +81,13 @@ impl NetAdapter for NetToChainAdapter {
 		}
 
 		// TODO - revisit the logic here around initial sync
-		if self.initial_syncing() {
+		// if self.initial_syncing() {
 			match res {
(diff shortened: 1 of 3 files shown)


antiochp:
Closing, replaced with #389 .

PR #364, opened 2017-11-22
add some randomness to most_work_peer

antiochp:
If we have several peers with most work then choose a random one.
So shuffle before sorting so the sort order is less deterministic.


ignopeverell:
Can you cherry pick to `tesnet1` when you get a chance? Thanks!

antiochp:
Done - cherry-picked.

Issue #361, opened 2017-11-21
Accidental Fork Details


yeastplume:
Spent the evening collecting details of a fork that appears to have thrown the network into chaos, causing syncs to hang. Collecting details here instead of gitter:

Yeastplume @yeastplume 21:46
Right, trying to restart node stuck trying to sync, last block it accepts is 7356 -8a7eab34, everything after is an orphan according to the node.
First we heard of that block was:
```
Nov 21 19:05:21.578 DEBG Received block 8a7eab34 at 7356 from network, going to process.
Nov 21 19:05:21.579 DEBG Processing block 8a7eab34 at 7356 with 6 inputs and 12 outputs.
Nov 21 19:05:21.580 DEBG Block 8a7eab34 refused by chain: Orphan
.
.
.
Nov 21 19:05:21.688 DEBG locate_headers: common header: 1258fdb7
Nov 21 19:05:21.695 DEBG locate_headers: returning headers: 74
Nov 21 19:05:22.255 DEBG Received block 8a7eab34 at 7356 from network, going to process.
Nov 21 19:05:22.256 DEBG Processing block 8a7eab34 at 7356 with 6 inputs and 12 outputs.
Nov 21 19:05:22.257 DEBG Block 8a7eab34 refused by chain: Orphan
Nov 21 19:05:22.630 DEBG peer total_diff (ping/pong): 73.81.122.26:13414, 26860673 vs us 28873113
Nov 21 19:05:24.017 DEBG Received block 8a7eab34 at 7356 from network, going to process.
```
This one appears to be the missing link between me and the rest of the chain... when trying to sync it ends up in a loop here:
```
Nov 21 21:27:28.318 INFO Received unfit block header 6c022946 at 7354: already in store.
Nov 21 21:27:28.318 DEBG Processing header e4b4cd15 at 7355.
Nov 21 21:27:28.321 INFO Received unfit block header e4b4cd15 at 7355: already in store.
Nov 21 21:27:28.321 DEBG Processing header 8a7eab34 at 7356.
Nov 21 21:27:28.322 DEBG Validating block with cuckoo size 16
Nov 21 21:27:28.327 DEBG Processing header 822896e5 at 7357.
Nov 21 21:27:28.327 INFO Invalid block header 822896e5: Orphan.
```
It looks like even though we rejected that earlier as an Orphan, when trying to rebuild the chain we're trying to include it... is there an off-by-one error somewhere?

Yeastplume @yeastplume 21:54
[...]

PR #368, opened 2017-11-22
[DNM] investigtion into sumtree rewind issues (duplicate commits)

antiochp:
See #361 for context - syncing was failing if we ever encounter a reorg and need to switch to a chain with greater difficulty.

* Make sure we go back to the pmmr to actually check for duplicate outputs and kernels (not just the index).
* Make sure the pmmr backend safely handles "out of bound" gets (cannot return results later than the rewound position).

TODO - 
- [ ] cleanup logging/debugging
- [ ] check we are doing the right thing for kernels in the pmmr
- [ ] can we compare commitments directly without hashing them?
- [ ] are there any other places we need to go look in the pmmr?

### DIFF
--- a/chain/src/sumtree.rs
+++ b/chain/src/sumtree.rs
@@ -89,13 +89,29 @@ impl SumTrees {
 		})
 	}
 
-	// TODO - is it enough to just get the hashsum from the pmmt backend?
-	// TODO - or do we actually need to check the hash in the hashsum?
 	/// Whether a given commitment exists in the Output MMR and it's unspent
 	pub fn is_unspent(&self, commit: &Commitment) -> Result<bool, Error> {
 		let rpos = self.commit_index.get_output_pos(commit);
 		match rpos {
-			Ok(pos) => Ok(self.output_pmmr_h.backend.get(pos).is_some()),
+			Ok(pos) => {
+				debug!(
+					LOGGER,
+					"found it in the index, checking the pmmr, {}, {:?}, {:?}",
+					pos,
+					commit.hash(),
+					commit,
+				);
+
+				if let Some(hashsum) = self.output_pmmr_h.backend.get(pos) {
+					debug!(LOGGER, "hashsum from pmmr {:?}", hashsum);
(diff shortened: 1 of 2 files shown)


chain/src/sumtree.rs
```diff
@@ -252,7 +287,8 @@ impl<'a> Extension<'a> {
 		}
 
 		for out in &b.outputs {
-			if let Ok(_) = self.commit_index.get_output_pos(&out.commitment()) {
+			// if let Ok(_) = self.commit_index.get_output_pos(&out.commitment()) {
+			if let Ok(true) = self.is_duplicate_output(&out.commitment()) {
```
antiochp: is_duplicate_commit checks both the index and the rewound pmmr - checking to see if the commit at that pos is what we expect.

store/src/sumtree.rs
```diff
@@ -306,6 +306,13 @@ where
 			return None;
 		}
 
+		if let Some(rewind) = self.rewind {
+			println!("sumtree backed get: rewind exists {:?}, position {}", rewind, position);
+			if rewind.0 < position {
```
antiochp: I don't know if this is the right thing to do but here we short-circuit and return None if we are are trying to get beyond the rewound position.

antiochp:
@ignopeverell @yeastplume - I think this fixes the issues we are seeing.
I can get two nodes to sync locally even if on forked chains with duplicate coinbase commitments on the 2 forked chains.

JollyMort:
Is this related? My node crashed, I built from latest testnet1 branch and attemted to start again, and now the below log message keeps repeating:

```
...
Nov 22 20:10:10.433 INFO Invalid block header fe88463f: Orphan.
Nov 22 20:10:10.433 INFO Invalid block header 629e06e7: Orphan.
Nov 22 20:10:10.434 INFO Added 0 headers to the header chain.
```

yeastplume:
Fixed in https://github.com/mimblewimble/grin/commit/31f8f2ad3e5da5ce91204bb226d1c285bfd1b969

PR #380, opened 2017-11-23
fix chain_header_by_height error

yeastplume:
fixes `StoreErr(NotFoundErr, "chain get header by height")',` errors when trying to sync.. I think this was caused by trying to access headers in the chain that the server hasn't actually received yet.

### DIFF
--- a/grin/src/sync.rs
+++ b/grin/src/sync.rs
@@ -281,17 +281,29 @@ impl Syncer {
 
 		debug!(LOGGER, "Sync: locator heights: {:?}", heights);
 
-		let locator = heights
+		let locator:Vec<Option<Hash>> = heights
 			.into_iter()
 			.map(|h| {
-				let header = self.chain.get_header_by_height(h).unwrap();
-				header.hash()
+				trace!(LOGGER, "sync getting header: {}", h);
+				let mut header = self.chain.get_header_by_height(h);
+				if let Err(e) = header.as_mut() {
+					trace!(LOGGER, "Sync: no header in local chain \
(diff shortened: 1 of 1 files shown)


antiochp:
This is weird - if we are height `h` then we should _definitely_ have all headers with lower height than this.

i.e if we have `height=3` then by definition we have `height=2` and `height=1`.

Is this a timing/race condition on the latest header? We've updated the height but we haven't stored the header in the store yet?

I'm not convinced we need to check for existence earlier than this?

yeastplume:
Yes, quite possible... that change was enough to allow sync to continue, but there’s obviously a
deeper underlying issue here. was looking into it more, will share findings later

antiochp:
:+1:
If you have chance please look over my "fixes" from earlier as well - tested locally and on testnet1 and things look better to me, but I'd like some reassurance I didn't mess anything serious up.

ignopeverell:
The longer locator and checking on misses both look good to me. And as @yeastplume mentions, there's still the issue of why we can't get the headers we supposedly just saved.

ignopeverell:
[Okay, I figured that out. 6718d9 intro](url)

ignopeverell:
Okay, figured that out. The implementation of the locator extension introduced the use of height lookups for headers. Height-based keys are only saved when the full block is saved. Ended up reverting all that and just extending the locator length.

ignopeverell:
Okay, figured that out. The implementation of the locator extension introduced the use of height lookups for headers. Height-based keys are only saved when the full block is saved. Ended up reverting all that and just extending the locator length.

PR #382, opened 2017-11-23
more readable get_locator

bit:
move debug into get_header_by_height, follow up to cf909f28a7


yeastplume:
Thanks, but I hadn't quite finished yet. It evolved that way and was just testing it worked with a few people before cleaning it up.

PR #388, opened 2017-11-24
first pass on better peer affinity

yeastplume:
Updated with changes discussed yesterday:

-Select a peer with the longest work and stick to it, changing only when a longer peer becomes available.
-Use the last valid block height a particular peer gave you to derive the locator, as opposed to your own locator
-Keep track of the last height given to you by each peer separately, to enable flopping back and forth if needed.

MOSTLY there, it would seem, only there remains an issue whereby when you get a new set of headers from your peer because you validated far enough up the chain, you seem to keep repeatedly requesting blocks you already have. (Restart fixes, but I'm sick of restarting).

If you want to merge and cherrypick as is @ignopeverell , then run with the block issue, (it's probably something minor, I'm just running out of time now) go ahead. It's feeling very near to being able to properly sync.

### DIFF
--- a/grin/src/sync.rs
+++ b/grin/src/sync.rs
@@ -21,11 +21,14 @@
 const MAX_BODY_DOWNLOADS: usize = 8;
 
 use std::ops::{Deref, DerefMut};
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, RwLock};
 use std::thread;
 use std::time::{Duration, Instant};
+use std::collections::HashMap;
+use std::net::SocketAddr;
 
 use core::core::hash::{Hash, Hashed};
+use core::core::BlockHeader;
 use chain;
 use p2p;
 use types::Error;
@@ -49,6 +52,10 @@ pub struct Syncer {
 	last_header_req: Mutex<Instant>,
 	blocks_to_download: Mutex<Vec<Hash>>,
 	blocks_downloading: Mutex<Vec<BlockDownload>>,
+
+	//keep a map of the last header successfully received
+	//from each peerj
+	last_headers_from_peers: Mutex<HashMap<SocketAddr, u64>>,
 }
 
 impl Syncer {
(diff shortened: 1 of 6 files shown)


grin/src/sync.rs
```diff
@@ -99,18 +141,33 @@ impl Syncer {
 		// check if we have missing full blocks for which we already have a header
 		self.init_download()?;
 
-		// main syncing loop, requests more headers and bodies periodically as long
+	//Keep track of which peer we're currently syncing with
```
ignopeverell: Tabbing seems a bit off here.

grin/src/sync.rs
```diff
@@ -99,18 +141,33 @@ impl Syncer {
...
 		// as a peer with higher difficulty exists and we're not fully caught up
 		info!(LOGGER, "Sync: Starting loop.");
 		loop {
 			let tip = self.chain.get_header_head()?;
 
-			// TODO do something better (like trying to get more) if we lose peers
-			let peer = self.p2p.most_work_peer().expect("No peers available for sync.");
-			let peer = peer.read().unwrap();
+			// select the best peer to be peering with
+			let current_sync_buddy=self.select_buddy(current_sync_buddy.clone());
```
ignopeverell: Sorry can't help it, but I think it's `<space>=<space>`. It's silly but it's been bothering me for too long :-S

grin/src/sync.rs
```diff
@@ -236,38 +294,55 @@ impl Syncer {
...
 				LOGGER,
 				"Sync: Asking peer {} for more block headers, locator: {:?}",
-				p.info.addr,
+				peer.info.addr,
 				locator,
 			);
-			if let Err(e) = p.send_header_request(locator) {
+			if let Err(_) = peer.send_header_request(locator) {
 				debug!(LOGGER, "Sync: peer error, will retry");
+			} else {
```
ignopeverell: Remove if useless?

grin/src/sync.rs
```diff
@@ -294,21 +369,17 @@ impl Syncer {
...
-				heights = heights[1..].to_vec();
-				locator.push(header.hash());
-			}
-			if header.height > 0 {
-				header = self.chain.get_block_header(&header.previous)?;
-			}
-		}
+	let locator = heights
+		.into_iter()
+		.map(|h| self.chain.get_header_by_height(h))
```
ignopeverell: This still won't work for 2 reasons:

* The header isn't stored under a height under the full block is received is validated. Could be changed but would introduce some more complexity and still not solve the next point.
* If you're syncing fork A, and your buddy flips (or you have 2 alternating buddies), you'll get the wrong headers and wrong hashes.

Iterating back from the tip hash is your only safe bet here.

grin/src/sync.rs
```diff
@@ -49,6 +52,10 @@ pub struct Syncer {
 	last_header_req: Mutex<Instant>,
 	blocks_to_download: Mutex<Vec<Hash>>,
 	blocks_downloading: Mutex<Vec<BlockDownload>>,
+
+	//keep a map of the last header successfully received
+	//from each peerj
+	last_headers_from_peers: Mutex<HashMap<SocketAddr, u64>>,
```
ignopeverell: I don't think this is enough. Check the comment about the `get_locator` impl below. You could either have a full header clone here or a subset of (height, hash). And while you're at it, you should probably throw in the TD at that block as well.

grin/src/sync.rs
yeastplume: :0

grin/src/sync.rs
yeastplume: Yes, was got confused again between the header chain and the 'validated' chain, makes perfect sense.

grin/src/sync.rs
yeastplume: I must be .... corrrrrrrected, sir.

grin/src/sync.rs
yeastplume: Yep, understood, will change that u64 to a 'BuddyInfo' struct or the like and put all that info in there.

PR #389, opened 2017-11-25
[DNM] minimal sync (2 threads, headers and blocks)

antiochp:
<b>
This "fixes" sync but introduces subtle but interesting problems with the handling of forks.
I am not convinced the tradeoff is worth it here.

There is a lot of complexity around the possible states we can be in between the "head of the block chain" and the "head of the header chain". 

I solved sync by breaking the link between these two heads during the sync process to allow headers to be retrieved "in bulk" and letting the blocks catch up. But this gets really messy once we encounter a fork (heights are off, relative total_difficulty between the two heads and the current block being processed is off etc.

The sync stuff is pretty straightforward in isolation - some of this may make sense to move forward with.
The `update_head` changes I'm less convinced about. This affects sync, and forks, and consensus in general and I don't have a clear picture of how this should work. Some of this is necessary for the sync to work but this definitely needs revisiting.

tl;dr - I don't think we should merge this to either master or testnet1. This does not handle forks well.
</b>

&nbsp;

TODO - 
- [x] fix `update_head` logic (see below) (fixed)
- [x] `broadcast_block` is being run throughout the sync...
- [x] investigate suppress mining during sync?
    * from observing a sync - it actually works (wallet ends up with some unconfirmed outputs) and several small forks are created and rewound successfully
- [x] bigger issue - if we experience a fork, we pull max 512 headers from the most_work_peer,
      do we then stay on this fork as we pull blocks? Do we flipflop between them?
     And if 512 headers is not enough to become the longest fork, what happens?
    `update_header_head` vs `update_head` (header chain diverges from block chain)

---------
* syncs headers by requesting headers based on locator every 10s
* syncs missing blocks based on difference between header head and blockchain head
    * requests 10 blocks every 1s until we have caught up with headers
* syncs in the background if we ever get out of sync or behind `most_work_peer`

Full sync (at least according to my peers) from height 0 to height 8938 - 
Approx 20mins from my laptop.

---------

Why do I think this is worth further investigation?
1) fast (20mins full sync to height ~9000)
1) handles background sync (to handle the case of a peer with more work appearing)
1) this is the __first__  time I have seen a full sync from 0 to ~9000 without error or restart
1) a lot simpler - a lot less chance of errors or panics given the simplicity of thread interaction
1) a lot less code and it still does what we expect it to do

------------

The following logs are useful for monitoring progress (grep for `update_head`).
First number and hash is max header we know about (based on headers).
Second number and hash is where we are so far with downloading blocks to the chain.

```
Nov 25 03:31:55.346 DEBG pipe: update_head: 8984, 4404212d, 4787, 9d3b4735
Nov 25 03:31:55.991 DEBG pipe: update_head: 8984, 4404212d, 4788, 5b5cf78a
Nov 25 03:31:56.182 DEBG pipe: update_head: 8984, 4404212d, 4789, 91b257aa
Nov 25 03:31:56.207 DEBG pipe: update_head: 8984, 4404212d, 4790, 1dca414e
Nov 25 03:31:56.275 DEBG pipe: update_head: 8984, 4404212d, 4791, e050e6c1
```

Edit -

full sync appears to reliably get jammed at block 4687, but a restart frees it up again.


grin/src/server.rs
```diff
@@ -209,15 +215,11 @@ impl Server {
...
-		let net_adapter = self.net_adapter.clone();
 
 		let mut miner = miner::Miner::new(config.clone(), self.chain.clone(), self.tx_pool.clone());
 		miner.set_debug_output_id(format!("Port {}", self.config.p2p_config.unwrap().port));
 		thread::spawn(move || {
-			let secs_5 = time::Duration::from_secs(5);
-			while net_adapter.syncing() {
-				thread::sleep(secs_5);
-			}
+			// TODO - sync always runs, so need to consider how the miner interacts with the sync process
```
ignopeverell: Miner may build blocks on the current chain view, which should be immediately orphaned and can cause confusion. So better not to run the miner when we're not close to the latest.

antiochp: So I'm pretty sure we can do the same thing what we're doing with `update_head` where we compare the head of the header chain to the head of the block chain.
And if there's too much difference then we're in "sync/catchup" mode and we should pause the miner loop (may still sit mining the current block until the 90s timeout).

grin/src/adapters.rs
```diff
@@ -279,37 +266,23 @@ impl NetToChainAdapter {
...
-			chain::SYNC
-		} else {
-			chain::NONE
-		};
-		opts
+		let header_head = self.chain.get_header_head();
+		let block_header = self.chain.head_header();
+		if let Ok(header) = header_head {
+			if let Ok(block_header) = block_header {
+				if header.height < block_header.height + 100 {
```
ignopeverell: Looks like this is the wrong comparison direction.

grin/src/simple_sync.rs
```diff
@@ -0,0 +1,173 @@
...
+
+	let mut current = chain.get_block_header(&tip.last_block_h);
+	loop {
+		if let Ok(header) = current {
+			if heights.contains(&header.height) {
+				locator.push(header.hash());
+			}
+			current = chain.get_block_header(&header.previous);
+		}
+		break;
```
ignopeverell: Huh, surely you don't want to break after just one loop iteration...?

grin/src/adapters.rs
antiochp: Thanks! Good catch.

antiochp:
@ignopeverell @yeastplume I pushed a few more changes to this branch.
Thoughts on this?

antiochp:
I've been pushing changes to the simple_sync_testnet1 branch.
I'm pretty happy with how it appears to be working right now.
Will work on porting the changes across to this branch/PR (against master).

ignopeverell:
I'd like to hear @yeastplume thoughts but I tend to think we should merge this. It works well and is simpler than the existing code, with almost half the number of lines. We still need to tackle the issue of multiple forks, some of them invalid, but this seems to be a simpler base to start with.

antiochp:
Ok @ignopeverell @yeastplume - this is good to go now.
I can cherry-pick this final set of changes across to testnet1 if we merge this to master (and get rid of my branched version).

yeastplume:
Absolutely fine with this approach as well. The peer-affinity version does sync to the latest as well, but still needs some work. Happy to replace it with this and just keep the buddy version it in the back pocket if it's needed.

antiochp:
<b>Commented at top of PR - I'm not convinced we should merge this until we we have a better understanding of the interaction between sync and fork handling.</b>

I'm concerned this will actually make things worse (particularly on testnet).

ignopeverell:
Partially applied, closing now.

antiochp:
@ignopeverell thanks for running with this and getting the useful parts merged in.

ignopeverell:
Thanks to you @antiochp for doing all the work :-)

Issue #403, opened 2017-11-29
MW scaling: Abridged sync


ignopeverell:
Current node model is one of a full archival node. To realize the scaling benefits of a MimbleWimble type chain we need 2 additional features:

* Pruning of chain data when it's not relevant to the UTXO set or the recent chain state (tracked in #404).
* Sync only based on the header chain, the MMR sumtrees.

This issue is for the latter.

The abridged sync should only need to download:

* The header chain of the most work chain.
* The MMR sum trees as of enough blocks back from the head (10,000 for example, to avoid confusion with a recent fork).
* Only recent block data.

Using this, a full chain validation can take place.

PR #405, opened 2017-11-29
initial docs and images to describe receiving a block and simple sync.

antiochp:


antiochp:
Oof. Might need to re-export these images as pngs... (was hoping this would work as svgs).

ignopeverell:
This is great. We can also add #403 once we start on it.

antiochp:
I want to keep referring back to this as we cover more scenarios - so none of us goes down a path of assuming a head points at one place when it actually does something else (and maybe helps us track down weird behavior when its not doing what we expect).

kargakis:
Should this doc be collapsed into https://github.com/mimblewimble/grin/blob/master/doc/chainsync.md or at least grouped under the same dir?

antiochp:
Yeah quite likely. Still iterating on this.

PR #407, opened 2017-11-30
Ping heights

yeastplume:
Adds the height to the ping/pong, to make it easier to interpret log output.


yeastplume:
Just in test branch for now, not going to pull this across to master as it's extra network load that probably won't be needed in the end.

PR #410, opened 2017-11-30
[DNM] introduce new sync_head for tracking header chain during sync

### LINKED ISSUE #402: Fix block locator with long forks
We generate our local block locator based on our current chain. A peer that tries to share its longer fork will only send 500 block headers at a time. If a fork is longer, once we receive these 500 blocks, the locator we generate will still be the same, leading to an endless loop of useless locator requests.

We need to bridge that gap. The most straightforward way is likely to keep track of the headers we got from a given peer during sync and generate our locator based the latest.

antiochp:
#413 was just merged onto testnet1 branch and contains some additional changes.
I'll port these across, either here or on another PR to replace this one.

* init sync_head during chain init
* rename `process_block_header` to `sync_block_header` (more explicit that this is sync only)
* in `sync` use `sync_header` to track the header chain that we are syncing to
* update `sync_head` as we process headers during the sync
* update `header_head` during sync when processing headers if new header is now the one with most work
* pass in 2x ctxs (sync and header) to `sync_block_header` to maintain both head states

Resolves #402.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -99,6 +99,17 @@ impl Chain {
 			Err(e) => return Err(Error::StoreErr(e, "chain init load head".to_owned())),
 		};
 
+		// make sure sync_head is available for later use
+		let _ = match chain_store.get_sync_head() {
+			Ok(tip) => tip,
+			Err(NotFoundErr) => {
+				let tip = Tip::new(genesis.hash());
+				chain_store.save_sync_head(&tip)?;
+				tip
+			},
+			Err(e) => return Err(Error::StoreErr(e, "chain init sync head".to_owned())),
+		};
+
 		info!(
 			LOGGER,
 			"Chain init: {:?}",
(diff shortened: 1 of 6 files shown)


ignopeverell:
In don't think you should completely to get rid of the header head here. I think you still want to track the currently most worked head somewhere.

Here is a scenario:

* You sync all good.
* A peer either lies about its total work or has an invalid block. You'll follow it a bit before realizing the issue. Your sync head is at a weird position.
* Someone else comes in with more work than you but now your sync head is  in a weird state, so locator generation and work comparisons will get weird.

antiochp:
We still maintain `header_head` when we process blocks though.
In `process_block` we call `update_head` and this updates both heads (header and body) in non-SYNC mode (just body head in SYNC mode).

Oh - are you saying this is problematic because `header_head` gets "left behind" in SYNC mode?

antiochp:
@ignopeverell pushed some more changes to keep `header_head` updated as we sync new headers (if more work on new header).

ignopeverell:
I think you need to reset the sync head to the header head each time we restart sync (in the adapter). Or am I missing something?

antiochp:
Yeah I think you're right - I don't think its absolutely necessary but there's a lot of wasted effort if we don't do this.

antiochp:
Replaced by #425 (porting the changes back from testnet1).

PR #414, opened 2017-12-01
body -> header -> sync

antiochp:
Use `sync_head` to guide the retrieval of `headers`.
But use `header_head` to inform which `blocks` we need to ask for.

This way we can continue to catch up to our last know `header_head`, even as the `sync_head` is off syncing with with a fork that advertises more work (but we are only pulling headers in batches of 512 at a time).

Also - now we have sync mode enabled/disabled I think it is safe to sleep for less than 30s on every loop of the header sync.

### DIFF
--- a/grin/src/sync.rs
+++ b/grin/src/sync.rs
@@ -77,8 +77,8 @@ fn body_sync(
 
 	let mut hashes = vec![];
 
-	if sync_head.total_difficulty > body_head.total_difficulty {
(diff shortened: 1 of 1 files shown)


antiochp:
@ignopeverell @yeastplume 
We used to sync blocks based on catching up with `header_head` so this is closer to how it used to work anyway.

antiochp:
Merging this - shout if any concerns.

PR #416, opened 2017-12-03
API endpoints to browse blocks

### LINKED ISSUE #397: API endpoints to browse blocks
Allow retrieving block data by full hash or by height. Something like:
```
// block at height 10
/v1/blocks/10

// block with hash d3c2bc...
/v1/blocks/d3c2bc65eb58c94c8f5c048283e907afbe37d53959a6add37f8a7757542ab75e
```

Data returned should at least include everything that's in the block header and idea the list of input and ouput commitments (no range proof though, that's too big for an API).

jgannonjr:
Added a /block endpoint to the API.  This can take either a height number or block address and return the block info.

Fixes #397

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -291,8 +291,9 @@ impl BlockHandler {
 	// Try to decode the string as a height or a hash address.
 	fn parse_input(&self, input: String) -> Result<Hash, Error> {
 		if let Ok(height) = input.parse() {
-			if let Ok(header) = self.chain.clone().get_header_by_height(height) {
-				return Ok(header.hash())
+			match self.chain.clone().get_header_by_height(height) {
+				Ok(header) => return Ok(header.hash()),
+				Err(_) => return Err(Error::NotFound),
 			}
 		}
     lazy_static! {
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -447,7 +447,7 @@ pub fn start_rest_apis<T>(
 
 		let route_list = vec!(
 			"get /".to_string(),
-			"get /block".to_string(),
+			"get /blocks".to_string(),
 			"get /chain".to_string(),
 			"get /chain/utxos".to_string(),
 			"get /sumtrees/roots".to_string(),
@@ -462,7 +462,7 @@ pub fn start_rest_apis<T>(
 		let index_handler = IndexHandler { list: route_list };
 		let router = router!(
 			index: get "/" => index_handler,
-			block: get "/block/*" => block_handler,
+			blocks: get "/blocks/*" => block_handler,
 			chain_tip: get "/chain" => chain_tip_handler,
 			chain_utxos: get "/chain/utxos/*" => utxo_handler,
 			sumtree_roots: get "/sumtrees/*" => sumtree_handler,
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -296,14 +296,13 @@ impl BlockHandler {
 				Err(_) => return Err(Error::NotFound),
 			}
 		}
-    lazy_static! {
-        static ref RE: Regex = Regex::new(r"[0-9a-fA-F]{64}").unwrap();
-    }
-    if !RE.is_match(&input) {
-    	return Err(Error::Argument(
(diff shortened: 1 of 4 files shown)


jgannonjr:
I wasn't too sure extra fields to add in the output, so if there are any please let me know!

Also, it seemed from the bug that this was intended to be human readable, so I used "printable" versions of the outputs.

api/src/handlers.rs
```diff
@@ -273,6 +274,46 @@ impl Handler for ChainHandler {
...
+		let block = self.chain.clone().get_block(h).unwrap();
+		BlockPrintable::from_block(&block)
+	}
+
+	// Try to decode the string as a height or a hash address.
+	fn parse_input(&self, input: String) -> Hash {
+		if let Ok(height) = input.parse() {
+			return self.chain
+				.clone()
+				.get_header_by_height(height)
```
ignopeverell: This should address the `NotFoundErr` case so we return a nice HTTP error code `404`.

api/src/handlers.rs
```diff
@@ -273,6 +274,46 @@ impl Handler for ChainHandler {
...
+				.unwrap()
+				.hash();
+		}
+		let vec = util::from_hex(input).unwrap();
```
ignopeverell: Similar here, we should handle the error instead of unwrapping and return a `400` with a nicer message (like "invalid hex"). Maybe also check the length (32 bytes so 64 chars).

api/src/handlers.rs
```diff
@@ -403,6 +447,7 @@ pub fn start_rest_apis<T>(
 
 		let route_list = vec!(
 			"get /".to_string(),
+			"get /block".to_string(),
```
ignopeverell: I would use plural. `chain` and `pool` are singular because there's ever only one, but note `utxos` and `sumtrees`.

api/src/types.rs
```diff
@@ -239,6 +239,21 @@ impl OutputSwitch {
 		}
 	}
 }
+
+// Printable representation of a block
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct TxKernelPrintable {
+	pub excess: String,
```
ignopeverell: I'd add the fee, the sig, `lock_height` (if present) and features (at this time, only feature is whether it's a coinbase kernel or not).

api/src/types.rs
```diff
@@ -251,14 +266,60 @@ pub struct BlockHeaderInfo {
 }
 
 impl BlockHeaderInfo {
-	pub fn from_header(block_header: &core::BlockHeader) -> BlockHeaderInfo{
+	pub fn from_header(block_header: &core::BlockHeader) -> BlockHeaderInfo {
 		BlockHeaderInfo {
 			hash: util::to_hex(block_header.hash().to_vec()),
 			previous: util::to_hex(block_header.previous.to_vec()),
 			height: block_header.height,
```
ignopeverell: I'd add pretty much everything else there is in the header. The roots, timestamp, difficulty, etc. All can be useful.

jgannonjr:
Addressed your comments.  Ready for a look!

hendi:
Great work!

I think the hash of the current block should also be included.

jgannonjr:
@hendi thanks for catching

ignopeverell:
Awesome work, thanks a lot for that! Keep it coming! :-)

yeastplume:
Sorry, just some notes here; my fault for not reviewing more closely. 

The BlockHeaderInfo struct was intentionally kept lean for performance/network reasons during the wallet restoration function, which has to fetch each and every block header from a node and parse them... this will add up very quickly when we have hundreds of thousands of headers to go through to restore a wallet. (There's also some problem parsing the difficulty struct causing wallet restore to fail, but that issue is more to do with lack of automated testing around this than anything else, opened another issue around that)

Doesn't have to be done immediately, but I think we're going to need to re-introduce a minimal structure that has only what's needed for wallet reconstruction.

jgannonjr:
Ok, no problem, that all makes sense.  I'll revert the BlockHeaderInfo struct and create a new struct to hold the full header in the api response.

PR #418, opened 2017-12-03
Small documentation improvements

bokobza:
Fixed a few typos and made a few bits of text more accurate.

### DIFF
--- a/doc/grin4bitcoiners.md
+++ b/doc/grin4bitcoiners.md
@@ -8,7 +8,7 @@ There are 3 main properties of Grin transactions that make them private:
 2. There are no amounts.
 3. 2 transactions, one spending the other, can be merged in a block to form only one, removing all intermediary information.
 
-The 2 first properties mean that all transactions look similar. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
+The 2 first properties mean that all transactions are indistinguishable from one another. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
 
 Moreover, there are no more transactions in a block. A Grin block looks just like one giant transaction and all original association between inputs and outputs is lost.
 
--- a/doc/grin4bitcoiners.md
+++ b/doc/grin4bitcoiners.md
@@ -8,7 +8,7 @@ There are 3 main properties of Grin transactions that make them private:
 2. There are no amounts.
 3. 2 transactions, one spending the other, can be merged in a block to form only one, removing all intermediary information.
 
-The 2 first properties mean that all transactions look the same. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
+The 2 first properties mean that all transactions look similar. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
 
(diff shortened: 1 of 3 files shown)


doc/style.md
```diff
@@ -1,6 +1,6 @@
 # Grin Style Guide
 
-Grin uses [rustfmt](https://github.com/rust-lang-nursery/rustfmt) to maintain consist formatting.
+Grin uses [rustfmt](https://github.com/rust-lang-nursery/rustfmt) to maintain consistant formatting.
```
antiochp: consistent

doc/merkle.md
```diff
@@ -60,7 +58,7 @@ a sum-tree over the commitments of outputs, and the negatives of the commitments
 of inputs.
 
 Input references are hashes of old commitments. It is a consensus rule that
-there are never two identical unspent outputs.
+there never are two identical unspent outputs.
```
antiochp: These both feel kind of awkward to me
Maybe this should be reworked as "all unspent outputs must be unique".

doc/grin4bitcoiners.md
```diff
@@ -8,7 +8,7 @@ There are 3 main properties of Grin transactions that make them private:
 2. There are no amounts.
 3. 2 transactions, one spending the other, can be merged in a block to form only one, removing all intermediary information.
 
-The 2 first properties mean that all transactions look the same. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
+The 2 first properties mean that all transactions look similar. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
```
antiochp: I think this should be `the same`. They are not just similar - they all look exactly like random data.

bokobza: That's why I changed to `similar`. They look similar but not `the same`. The same for me means exactly the same.

The Cambridge dictionary says [here](https://dictionary.cambridge.org/grammar/british-grammar/adjectives/same-similar-identical)
> Same means that two or more things are exactly like one another.

> We use similar if two or more things are not entirely the same, or identical if two or more things are exactly the same.

doc/merkle.md
bokobza: Yeah that's a good idea :-)

doc/style.md
bokobza: :grimacing: oh no, sorry.

doc/grin4bitcoiners.md
antiochp: Not _identical_ in that they are all the same exact piece of random data - but they're not similar to random data, they _are_ random data.

antiochp: Maybe we want to say "indistinguishable" here and remove the ambiguity.
But we're getting close to cryptography terminology there that I'll defer to other people for.

bokobza: ah! I was just writing that.
I can leave `the same` if you want :-)

bokobza: Yeah that's what I understood when I changed `the same` to `similar`.
How about (last try I promise):
`The 2 first properties mean that all transactions are indistinguishable from one another. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).`

antiochp: :+1:

PR #419, opened 2017-12-03
Update BlockHeaderInfo Struct

mtanzi:
## Overview
I am not sure if is relevant although I believe that for consistency, when we return the block header we should return the same parameters in both
- `/v1/chain`
- `/v1/chain/utxos/atheight`

In case the naming was deliberately different for some reason - Please feel free to close the PR :) 

## Changes
- Added `total_difficulty` to `BlockHeaderInfo` so it can be returned by `/v1/chain/utxos/atheight`
- Inside the Tip struct I changed `last_block_pushed` to `hash` and `prev_block_to_last` to `previous` to keep the  names consistent between Apis

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -25,10 +25,10 @@ use util;
 pub struct Tip {
 	/// Height of the tip (max height of the fork)
 	pub height: u64,
-	// Last block pushed to the fork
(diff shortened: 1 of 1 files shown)


ignopeverell:
Looks good, thanks!

PR #427, opened 2017-12-04
[testnet1] simplify sync process further - one thread for both body and header sync

antiochp:
Several fixes that should make sync simpler, faster and more robust (hopefully).

* spawn single thread to poll for both headers and bodies (was 2 threads competing for locks on peers)
    * loop every 250ms
    * run the body sync if 1s has passed
    * run the header sync if 5s has passed
* do not call `reset_header_head` if `is_bad_block`
    * sync process via `sync_head` should take care of this after banning a peer
* ~experiment with requesting `peer_count * 2` blocks at a time, per body_sync iteration~
    * stick with `10` for now (simple to understand)
* always ask for headers from `most_work_peer` 
    * we should have multiple `most_work_peers` and we choose a random one
* replace several `peer.read()` with `peer.try_read()` within sync thread

### DIFF
--- a/grin/src/adapters.rs
+++ b/grin/src/adapters.rs
@@ -85,11 +85,11 @@ impl NetAdapter for NetToChainAdapter {
 			if e.is_bad_block() {
 				self.p2p_server.borrow().ban_peer(&addr);
 
-				// and if we're currently syncing, our header chain is now wrong, it
-				// needs to be reset
-				if self.is_syncing() {
-					let _ = self.chain.reset_header_head();
-				}
+				// header chain should be consistent with the sync head here
+				// we just banned the peer that sent a bad block so
+				// sync head should resolve itself if/when we find an alternative peer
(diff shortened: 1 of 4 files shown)


antiochp:
Testing this on `testnet1` and it appears to be running all the way through to completion.
I have not (yet) seen it hang after a `locator:` log msg.

antiochp:
Definitely seen at least one of these scroll rapidly past in the logs - 

```
Dec 04 17:46:18.915 DEBG sync: locator: [1e37494f, f00b6ea2, 80f03ca4, 3b6e96bc, 6b2b466d, 00589cdc, a5e73b44, 66296f2e, 4475f6ca, 23d65094, 325113e3, e36b595d, 63721c32, e17ea64f]
Dec 04 17:46:18.915 WARN sync: request_headers: failed to get read lock on peer
```

PR #428, opened 2017-12-05
simplify sync process further - one thread for both body and header sync

antiochp:
* simplify sync process further - one thread, try_read on peer for robustness
* add most_work_peers() for convenience, add most_work_peers count to "monitoring peers" log msg


antiochp:
This was merged to `testnet1` in #427, porting across to master to keep it up to date.
Looks like peer banning is not in master - we should get that moved across at some point as well.

antiochp:
Rebased on master - will merge if/when travis is happy.

PR #431, opened 2017-12-05
 dry up how we check if a header is on the current chain 

antiochp:
We were doing similar (but slightly different) things to check the header by hash and then header by height and comparing the hashes to see if the header was on the current chain.

This PR pulls this logic out into `is_on_current_chain()` to DRY this up.

Unrelated (probably should have been in a separate PR) - initial sync_head is set to `head` now and not manually setting to the header on the genesis block.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -25,7 +25,7 @@ use core::core::pmmr::{HashSum, NoSum};
 
 use core::core::{Block, BlockHeader, Output, TxKernel};
 use core::core::target::Difficulty;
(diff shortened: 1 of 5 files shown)


antiochp:
@ignopeverell @yeastplume - want to take a quick look at this and make sure I haven't messed up the logic anywhere?
chain tests are passing and I've tested this locally so reasonably sure there are no stupid mistakes here, but you never know...

ignopeverell:
Yes, it's cleaner and a common operation, so makes total sense. Thank you for that.

PR #433, opened 2017-12-06
base threshold for sync on difficulty of past 5 blocks

### LINKED ISSUE #401: Better test to re-trigger sync
We will get back into sync mode only if we find a peer with a head at a 1000 higher difficulty than ours:

https://github.com/mimblewimble/grin/blob/c265e937e9d6f95461d7499f5368750d52fad97b/grin/src/adapters.rs#L293

This constant is likely to be wrong and should vary with current network difficulty. The sum of the difficulty in the last 5 blocks may be a better test?

antiochp:
* Use `DifficultyIterator` to find difficulty for past 5 blocks
* Calculate sync threshold from sum of difficulties from these 5 blocks
* Add back an initial 30s sleep to sync - give us time to find some peers

Resolves #401.

### DIFF
--- a/grin/src/adapters.rs
+++ b/grin/src/adapters.rs
@@ -287,28 +287,48 @@ impl NetToChainAdapter {
 	/// just receiving blocks through gossip.
 	pub fn is_syncing(&self) -> bool {
 		let local_diff = self.total_difficulty();
-		let peers = self.p2p_server.borrow().connected_peers();
+		let peer = self.p2p_server.borrow().most_work_peer();
 
 		// if we're already syncing, we're caught up if no peer has a higher
 		// difficulty than us
 		if self.syncing.load(Ordering::Relaxed) {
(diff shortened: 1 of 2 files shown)


grin/src/adapters.rs
```diff
@@ -287,28 +287,48 @@ impl NetToChainAdapter {
...
-			});
-			if higher_diff_padded {
-				info!(LOGGER, "sync: late on the most worked chain, enabling sync");
-				self.syncing.store(true, Ordering::Relaxed);
+			if let Some(peer) = peer {
+				if let Ok(peer) = peer.try_read() {
+					let mut threshold = Difficulty::zero();
+					let last_diffs = self.chain
+						.difficulty_iter()
+						.take(5)
```
ignopeverell: Map `None` to 0 and then `sum`?

antiochp: Sounds good - will update.

PR #436, opened 2017-12-06
add logging in monitor_peers for healthy/banned/defunct counts

antiochp:
Seeing nodes that do not have a full set of 8 connected peers and wondering why.
If we less than 8 connected peers we should be looking for healthy peers from the db and attempting to connect to them.


antiochp:
This is pretty minor - merging.

PR #438, opened 2017-12-06
homogenize MimbleWimble writing

ishigo974:
The `MimbleWimble` term is always written like a camel case of two words, but here it wasn't.


antiochp:
:+1:

PR #440, opened 2017-12-06
[WIP] Abridged sync

ignopeverell:
See #403.


antiochp:
This introduces the concept of a full, non-archival node?
I'm wondering what happens now when we try and spend really old coinbase outputs? Is this going to fail validation (on some nodes) as we cannot successfully look the block up in the store (by height or hash)?

ignopeverell:
They would be in the UTXO set, so should be fine.

PR #442, opened 2017-12-07
Revert BlockHeaderInfo struct to contain only the fields needed for w…

### LINKED ISSUE #397: API endpoints to browse blocks

jgannonjr:
…allet reconstruction.  Create a new struct for use in the /blocks api.

address @yeastplume comments which further resolves #397.

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -275,8 +275,8 @@ impl Handler for ChainHandler {
 	}
 }
 
-// Gets block details given either a hex address or height.
-// GET /v1/block/<address>
+// Gets block details given either a hash or height.
+// GET /v1/block/<hash>
 // GET /v1/block/<height>
 pub struct BlockHandler {
 	pub chain: Arc<chain::Chain>,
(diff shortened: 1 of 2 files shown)


yeastplume:
That looks all fine, thanks for addressing this further. Do you want to go ahead and add the extra fields into the BlockHeaderPrintable structure within this PR?

We'll probably have to do a pass on naming of the API structures are some point to make them better reflect their intended purposes, but we can take that as a separate action.

yeastplume:
Whoops, they're there, just got confused by the diff out. Will merge now.

jgannonjr:
The fields were added already, the diff highlighting is just a bit weird so it isn't obvious (since I renamed the existing struct which already has the extra fields they don't show up as being added, just a rename of s/BlockHeaderInfo/BlockHeaderPrintable/).

I agree on the naming, this is my first time going through the code and it wasn't quite obvious what each struct meant, but I tried to follow existing naming patterns as much as possible.

PR #448, opened 2017-12-08
body_sync every 5s, but request 10*most_work_peers blocks at a time

### LINKED ISSUE #447: 1s interval is too fast for body_sync
We are not waiting long enough to allow time for peers to respond and us to actually process the blocks before going off and (re)requesting a new set of blocks.

We should increase the time we sleep between iterations (5s or 10s)?
But - we should ask for more blocks at a time (50 at a time, more?)
And - we should increase the orphan limit (increase from 20 to 50?)

So proposal - we request 50 blocks every 5s (currently 10 blocks every 1s).
And an orphan limit of 50 should cover any out of order block responses.

antiochp:
* body_sync every `5s`
* request `10*most_work_peers` each time
* increase `MAX_ORPHANS = 50` (to handle larger number of blocks at a time)
* header_sync every `10s`

Sync will be slower if the network is not in a particularly stable state (few peers advertising most work) but faster once network starts to converge (multiple peers advertising most work).

Resolves #447.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -33,7 +33,7 @@ use sumtree;
 use types::*;
 use util::LOGGER;
 
-const MAX_ORPHANS: usize = 20;
+const MAX_ORPHANS: usize = 50;
 
 /// Facade to the blockchain block processing pipeline and storage. Provides
 /// the current view of the UTXO set according to the chain state. Also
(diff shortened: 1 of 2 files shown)


ignopeverell:
Looks good, thanks!

Issue #439, opened 2017-12-06
Peer banning should be temporary


ignopeverell:
The role of banning a misbehaving peer is preventing repeated offenses, whether it's DoS or feeding us false information that can take some time to recover from. It mitigates the damaging effect of someone trying to be nefarious or just running buggy code. However we may not want to ban forever.

First, someone who's nefarious at first may not always be. Second, if you're running experimental code, being banned forever for a bug sucks. Third, at this point our code isn't bug-free either. So letting a banned peer connection again after a few hour may be the best approach.

antiochp:
Proposal - 
* add a `last_banned` timestamp to `PeerData`
* store `now_utc()` in `last_banned` when banning
* periodic process to "un-ban" banned peers after 3 hours (or so)

heunglee:
I have a question. I thought about a case that server A banned a peer B, and server C connects to server A and receives peer information. Then server A and server C ban B.  How is the server C going to process banned B when receiving B info from server A almost 3 hours later? I also thought about the real case like hacking wallets and DoS attack as well. 3 hours will be enough to take measures? What if there is no mechanism to ban the attacker permanently? 

If the time interval is set to 3 hours, the ban information is thought to be propagated to many peers. If a peer banned because of hardware issues and needs recovery sooner after many peers already received banned status of the peer, what will be proper measures?

I just scratched my head to see unexpected situations with 3 hours refresh in order to keep the perspectives of sound grin network in production.

sesam:
@heunglee bans are local to the node; not propagated. TCP conns are cheap. The p2p gossip network can be slowed down by IP stealing attacks on ARP or BGP, or targeted DoS attacks, but nothing to worry about for a long time yet.

mw/grin doesn't need to care which node a transaction (tx) originates from. With cryptocurrencies, a tx is valid, or not. Spreading lies (bug or evil)? Ban you for a while, but let you upgrade and try again in some hour or so.

Ask me more on gitter to learn more on how blockchain's are built :) and attack/defence scenarios, since I think that's a chat-friendly topic that many in the Lobby can benefit from discussing.

PR #449, opened 2017-12-08
[WIP] change set_banned and is_banned -> bool to make the ban time limited

sesam:
- first attempt at #439

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::cell::RefCell;
 use std::net::SocketAddr;
 use std::sync::{Arc, RwLock};
 
@@ -23,6 +24,7 @@ use core::core;
 use core::core::hash::{Hash, Hashed};
 use core::core::target::Difficulty;
 use handshake::Handshake;
+use time;
 use types::*;
 use util::LOGGER;
 
@@ -39,6 +41,7 @@ pub struct Peer {
 	pub info: PeerInfo,
 	proto: Box<Protocol>,
 	state: Arc<RwLock<State>>,
+	banned_until: RefCell<i64>,
 	// set of all hashes known to this peer (so no need to send)
 	tracking_adapter: TrackingAdapter,
 }
@@ -53,6 +56,7 @@ impl Peer {
 			info: info,
 			proto: proto,
 			state: Arc::new(RwLock::new(State::Connected)),
+			banned_until: RefCell::new(1),
 			tracking_adapter: TrackingAdapter::new(na),
 		}
 	}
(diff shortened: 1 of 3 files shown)


p2p/src/peer.rs
```diff
@@ -39,7 +39,7 @@ pub struct Peer {
 	pub info: PeerInfo,
 	proto: Box<Protocol>,
 	state: Arc<RwLock<State>>,
-	banned_until: Arc<RwLock<time::Tm>>,
+	banned_until: Arc<RwLock<i64>>,
```
sesam: Maybe better use a RefCell https://doc.rust-lang.org/core/cell/struct.RefCell.html

p2p/src/peer.rs
```diff
@@ -123,16 +126,30 @@ impl Peer {
...
-		*state == State::Banned
+		*state == State::Banned && self.recheck_ban_timeout()
 	}
 
 	/// Set this peer status to banned
 	pub fn set_banned(&self) {
 		let mut state = self.state.write().unwrap();
 		*state = State::Banned;
+		let mut banned_until = self.banned_until.write().unwrap();
+		*banned_until = (time::now_utc() + time::Duration::minutes(42)).to_timespec().sec;
```
antiochp: Lol at 42 - lets do 3 hours initially? If this turns out to be too long then we can shorten it.
But I think we want to ban for an extended period for now.

sesam: Maybe wrongly, I assumed we'd just ban and ban again for nodes that aren't updated. My recent restarts only banned good nodes, so maybe leave a low number at least until that issue is fixed?

For Mainnet I'm thinking a tiered slightly random approach, and repeat offenders get bumped into lower levels of oblivion, with a slower poisson process to get kicked upwards to get another chance. When DoS attacks start, I think IP-based banning won't be effective protection, and we'll need to score connected nodes by age, trying to keep a mix of old known and recently discovered nodes in the 8 spot connection pool.

ignopeverell: What do you mean by "banned good nodes"? Nodes are banned if they send a block that will never be valid for you. If they accepted that node, there is nothing they can do to help you.

p2p/src/peer.rs
sesam: (switched to RefCell and it seems I got persistence to work, yay)

p2p/src/peer.rs
sesam: as per gitter chat, it seems we have examples of banning good nods incorrectly. Maybe making that painful for all the newcomers and teaching them to do rm -rf .grin/peers regularly is the right call. And mably was a bad seed node, but he already started resync and will be good again in shorter time than 3 hours.

ignopeverell:
Sorry, looking at this until now but this approach is unlikely to work. The banned setup and time setting is fine, but then disconnected peers are regularly cleaned up from the peers list to avoid clogging up memory. So the unbanning timer will never fire. I would instead look at adding an additional check in `grin/src/seed.rs` when we try to get more peers from db.

ignopeverell:
Any chance to have this fixed? :-)

sesam:
#439 fixed by #631, so removing this one.

PR #460, opened 2017-12-10
Seed debug outputs cleanup

sesam:
Just code cleanup and cleaner debug!s. Probably worth to squash before merge.


antiochp:
Looks good to me.

PR #467, opened 2017-12-12
Improve errors that confuse new users

sesam:
The changes are made with the goal to have less "first line" support work in gitter/mail/github issues


ignopeverell:
Thanks, that's very helpful!

PR #477, opened 2017-12-13
introduce cpu_pool to read_msg

antiochp:
This PR attempts to allow `read_msg` and `write_msg` to run at a more granular level.
We should be able to write msg back to a peer as soon as we have read the msg and handled it.
We should not wait for 10 GetBlock msgs to be processed before we start returning blocks to the peer requesting them for example.

--------------

TODO -
- [x] look into passing a single cpu_pool into the various peer connections we set up
- [x] failing tests (related to DummyAdaptor)

------------

* introduce `cpu_pool` to `read_msg`
    * `handler.handle()` off main loop now
* add `buffered(1)` adaptor to the "infinite iterator" in `read_msg`
    * ensures `fold()` runs on every msg to be read and does not keep going until stream is empty
* add some `trace!` logging to see what's going on when reading/writing msgs

We are now seeing interleaved `read_msg/write_msg` in the logs.
We are no longer waiting for all the `GetBlock` handlers to run for a given peer before starting to return data to the peer.

```
Dec 13 09:44:40.657 TRCE read_msg: start
Dec 13 09:44:40.657 DEBG locate_headers: [545b137e, 222b4104, 634195cd, 2f0a8880, b83053b4]
Dec 13 09:44:40.657 DEBG locate_headers: common header: 545b137e
Dec 13 09:44:40.657 DEBG locate_headers: returning headers: 1
Dec 13 09:44:40.657 TRCE read_msg: done (via cpu_pool)
Dec 13 09:44:40.658 TRCE read_msg: count (per buffered fold): 21
Dec 13 09:44:40.658 TRCE write_msg: start
Dec 13 09:44:40.658 TRCE write_msg: done
Dec 13 09:44:40.659 TRCE read_msg: start
Dec 13 09:44:40.660 DEBG handle_payload: GetBlock 050179b4
Dec 13 09:44:40.660 TRCE read_msg: done (via cpu_pool)
Dec 13 09:44:40.661 TRCE read_msg: count (per buffered fold): 22
Dec 13 09:44:40.661 TRCE read_msg: start
Dec 13 09:44:40.661 TRCE write_msg: start
Dec 13 09:44:40.661 TRCE write_msg: done
Dec 13 09:44:40.662 DEBG handle_payload: GetBlock 634195cd
Dec 13 09:44:40.662 TRCE read_msg: done (via cpu_pool)
Dec 13 09:44:40.662 TRCE read_msg: count (per buffered fold): 23
Dec 13 09:44:40.663 TRCE read_msg: start
[...]

PR #466, opened 2017-12-11
[WIP] stateful syncer tracking in_flight blocks

antiochp:
Bringing back the state to track "in flight" blocks.
We have scenarios where we start to see timeouts during the `body_sync` with the syncer repeatedly asking for the same blocks from the same peer.
I suspect we're causing a lot of backlog with this and get into a situation where these are timing out because we have queued up so many block requests that we cannot get through them fast enough.

Tracking "in flight" block requests will let us avoid asking for the same blocks again and again.

~Battling with lifetimes though in rust right now...~

See #454 for more info.

Issue #454, opened 2017-12-10
connection timeouts in body_sync (requesting blocks)


antiochp:
Seeing a node stuck trying to ask for more blocks and repeatedly timing out.

On the surface this looks suspiciously like this was caused by us Increasing the # blocks being requested each body_sync iteration.

```
Dec 10 02:39:22.453 DEBG block_sync: 19597/19785 requesting blocks [1bbd740d, 31703d3d, 243f79e6, ba833426, 64142ed6, 5bf2e0a6, 016bf1d0, a07a4e67, ae19c4e5, afbc415c, 949950b7, 4333497d, 345d97d6, 0146e928, e6d24ebc, 3975450d, 77c6f308, d423be34, b018b00c, f60fa457] from 2 peers
Dec 10 02:39:22.454 DEBG Requesting block 1bbd740d from peer 37.6.168.203:23414.
Dec 10 02:39:22.454 DEBG Requesting block 31703d3d from peer 37.6.168.203:23414.
Dec 10 02:39:22.455 DEBG Requesting block 243f79e6 from peer 37.6.168.203:23414.
Dec 10 02:39:22.455 DEBG Requesting block ba833426 from peer 37.6.168.203:23414.
Dec 10 02:39:22.456 DEBG Requesting block 64142ed6 from peer 37.6.168.203:23414.
Dec 10 02:39:22.456 DEBG Requesting block 5bf2e0a6 from peer 37.6.168.203:23414.
Dec 10 02:39:22.457 DEBG Requesting block 016bf1d0 from peer 51.15.219.12:13414.
Dec 10 02:39:22.457 DEBG Requesting block a07a4e67 from peer 51.15.219.12:13414.
Dec 10 02:39:22.458 DEBG Requesting block ae19c4e5 from peer 37.6.168.203:23414.
Dec 10 02:39:22.458 DEBG Requesting block afbc415c from peer 37.6.168.203:23414.
Dec 10 02:39:22.458 DEBG Requesting block 949950b7 from peer 51.15.219.12:13414.
Dec 10 02:39:22.459 DEBG Requesting block 4333497d from peer 37.6.168.203:23414.
Dec 10 02:39:22.459 DEBG Requesting block 345d97d6 from peer 51.15.219.12:13414.
Dec 10 02:39:22.460 DEBG Requesting block 0146e928 from peer 37.6.168.203:23414.
[...]

PR #479, opened 2017-12-13
[testnet1] cpu_pool and buffered fold in read_msg

antiochp:
* use a `cpu_pool` in `read_msg` for running `handler.handle()`
* add `buffered` adapter to the infinite interator in read_msg (so we fold over one msg at a time)

Related - #454.
Related - #466.
Original (master) - #477 (reverted and will be reapplied once master is consistent with testnet1)


antiochp:
Will port this back across to master once various PRs have been merged to master.

PR #483, opened 2017-12-14
Travis speedups

sesam:
trying recommended things for speed ups


sesam:
Probably just luck (and North America asleep?) but the current ordering of Travis tasks seems faster.
Also leaving "sudo: false" in there for now. Merge?

antiochp:
:+1:

PR #484, opened 2017-12-14
[testnet1] De-Duplicate expected responses in TimeoutConnection

antiochp:
* de-dupe requests in TimeoutConnection (based on expected responses we are waiting for)
* do not request blocks that we already have in our orphan list during body_sync

TimeoutConnection already tracks "expected" responses
use this to deduplicate requests and do not ask a peer for the same thing again
(until either success or error via timeout)

Add `trace!` logging so we have some visibility into how big the "expected" list of responses is on a given connection.

This should reduce a lot of the network congestion (and backlog of requests a peer is handling) as peers attempt to request the same set of blocks repeatedly.

We were originally doing something similar in the previous `sync` process - where we tracked "in flight" block requests, only queueing up new block requests once we had received and processed previous requests.
I think its cleaner/simpler and more generic if we do this at the connection layer.

Additional benefit here - we can end up asking for the same block from a couple of different peers this way (one per connection) if the first peer takes a few seconds to respond.
So we get some redundancy here on slow connections that might slow or halt the sync process.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -219,6 +219,11 @@ impl Chain {
 		}
 	}
 
+	pub fn is_orphan(&self, hash: &Hash) -> bool {
+		let orphans = self.orphans.lock().unwrap();
+		orphans.iter().any(|&(_, ref x)| x.hash() == hash.clone())
+	}
+
 	/// Pop orphans out of the queue and check if we can now accept them.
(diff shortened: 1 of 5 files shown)


antiochp:
Merging onto testnet1. Will cherry-pick onto master shortly.

PR #488, opened 2017-12-14
[testnet1] broadcast to subset of peers (in case we are connected to >8 peers)

antiochp:
In `seed.rs` we have -
* PREFERRED_PEER_COUNT = 8
* MAX_PEER_COUNT = 25

So we "prefer" 8 peers but if lots of peers connect to us we can actually be connected to significantly more than this.
`clean_peers()` will reduce the number of connected peers if necessary to keep it under MAX_PEER_COUNT.

Rather than peers broadcasting blocks and txns to _all_ connected peers this PR changes this behavior to broadcast blocks and txns to PREFERRED_PEER_COUNT random connected peers.
My understanding is this is closer to a true "gossip protocol" (we randomly select some neighbors to communicate with) while giving us the benefits of having a more connected network over time (more likely to be connected to a "true" most_work_peer on the longest chain if we can have >8 connected peers).

Not 100% sure if we want to introduce this complexity where we decouple the concept of "connected peers" from "who we broadcast to".

Some example logs from one of my nodes that got me thinking about this.
These are *really good* numbers (have not seen connectivity like this before).

```
Dec 14 17:39:07.855 DEBG monitor_peers: 7 most_work_peers, 15 connected, 322 total known
Dec 14 17:39:07.855 DEBG monitor_peers: all 322 = 318 healthy + 0 banned + 4 defunct
Dec 14 17:39:37.855 DEBG monitor_peers: 6 most_work_peers, 13 connected, 322 total known
Dec 14 17:39:37.855 DEBG monitor_peers: all 322 = 318 healthy + 0 banned + 4 defunct
Dec 14 17:40:07.855 DEBG monitor_peers: 7 most_work_peers, 15 connected, 322 total known
Dec 14 17:40:07.856 DEBG monitor_peers: all 322 = 318 healthy + 0 banned + 4 defunct
```

I *think* we want to be broadcasting new blocks and txns to 8 of those connected peers, not the full list of 15 connected peers (and they in turn would do something similar).

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -73,8 +73,11 @@ impl Peers {
 		self.get_peer(addr).is_some()
 	}
 
+	/// Get vec of peers we are currently connected to.
 	pub fn connected_peers(&self) -> Vec<Arc<RwLock<Peer>>> {
-		self.peers.read().unwrap().values().map(|p| p.clone()).collect()
+		let mut res = self.peers.read().unwrap().values().cloned().collect::<Vec<_>>();
+		thread_rng().shuffle(&mut res);
+		res
 	}
 
 	/// Get a peer we're connected to by address.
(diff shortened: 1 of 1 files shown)


sesam:
Cons
 - high PREFERRED means more densely connected (=chatty) network graphs

Pros
 - more to choose from while network peers come and go a lot

Ideas
 - prioritise having 1 peer with exactly "my" user_agent
 - further diversity quota, maybe try to get both low and high ping peers

ignopeverell:
@sesam the "my" user agent would be temporary, it's not something we should do on mainnet.

@antiochp sounds good to me. I've actually been looking for applicable research on gossip and what would be the minimal safe number of peers to broadcast to in a given network topology (i.e. 30% of the network is useless, selfish or attackers). I haven't found anything good so far, most gossip research takes the "private cloud" assumption and do not factor in bad connectivity, churn, etc. Only the older DHT papers (Chord, Kad, ...) take these into consideration but that's unrelated to gossip.

TL; DR: go for it.

PR #497, opened 2017-12-15
better error messages + some cleanup

### LINKED ISSUE #474: wallet info/outputs shows different height with/without grin server running
`wallet info` and `wallet outputs` show lower "height" if grin server can't be asked for live numbers for height etc.

- The info/outputs could just refuse if server not found.
- Maybe add --offline that shows more limited info.
- When server IS reachable, info&outputs should explain, i.e. "Status of outputs verified by (URL) after block (hash) @ (height). Compare your blockhash@height with grinexplorer.net or grinscan.net to be sure."

Background:
> @lukevmorris 05:29
> I'm periodically outputting my block height as reported by grin wallet info and I notice that
> with the server on in another terminal window, the height is 15032 blocks
> with the server off, the height is 13713 blocks
> is this kind of behavior to be expected?

sesam:
* Changes stdout log level back down one step to INFO + moves some chatty log outputs from INFO to DEBG. Specifically sync will now by default only report height + hash every 500 blocks on stdout.
* Changes log file default from Trace to Debug, hopefully makeing people less worried about log file size. This also assumes we'll use Trace primarily for developer use, maybe to defend against regressions and spurious println! :)
* Improves error messages, based on problems discovered from supporting users on gitter
* Renames a function in code. Probably no devs will have conflicts with their local stashes/branches
* Typos

Fixes #474

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -369,12 +369,16 @@ fn update_head(b: &Block, ctx: &mut BlockContext) -> Result<Option<Tip>, Error> 
 				.map_err(|e| Error::StoreErr(e, "pipe save head".to_owned()))?;
 		}
 		ctx.head = tip.clone();
-		info!(
+		debug!(
 			LOGGER,
 			"Updated head to {} at {}.",
 			b.hash(),
 			b.header.height
 		);
+		if b.header.height % 500 == 0 {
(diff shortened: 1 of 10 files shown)


antiochp:
Looks good. :+1:

Issue #512, opened 2017-12-18
during a fork 2nd block with same difficulty is not broadcast


antiochp:
In a situation where we have 2 blocks mined at the same time, both with the same total_difficulty, 
nodes do not broadcast out the 2nd block they see.

https://grinexplorer.net/block/21805

<img width="1139" alt="screen shot 2017-12-18 at 10 31 34 am" src="https://user-images.githubusercontent.com/30642645/34113743-afa136ea-e3de-11e7-8cc0-c53bc601a25f.png">

`pipe.process_block()` only return the new chain tip if it gets updated
`chain.process_block()` only broadcasts the block out if we have a new chain tip

I _think_ this explains how/why we see forks that do not quickly get resolved.
The 2nd block will get suppressed, causing it to propagate across the network poorly.
If this then turns out to be the winning fork then everyone sees orphaned blocks later than this and they have no way to retrieve the missing suppressed block.

In `pipe::update_head()` we only update the head if total_difficulty is strictly greater.

~I propose we relax this slightly to allow for equal `total_difficulty` and "last one wins" semantics to ensure this new block gets broadcast out.~

~Edit: I think we want to maintain the current "strictly increasing" semantics. There's a risk we flip-flop between forks like this. And I think its more correct for a node to see the first block at a given total_difficulty as the correct one at that time.~

Edit: Initial attempt is going to keep this as simple as possible and treat both blocks as if they were now the winning chain tip (i.e. "last one wins"). Otherwise we're adding a lot of complexity here to cover this edge case with some special handling.

PR coming shortly.

ignopeverell:
Sorry to come in late but we should default on first-one-win, otherwise you create a lot of incentives to keep forking.

antiochp:
Discussed on gitter. Reworking the PR to make this more generic for "broadcast all valid blocks".

PR #514, opened 2017-12-18
[testnet1] relax diff check in update_head (was gt, now gte)

antiochp:
This will cover the situation where a node sees two forked blocks at same height with identical total_difficulty.
We do not now which of these two blocks will "win" the fork and it may take a couple of blocks to resolve the situation.
So - we want to handle both blocks and make sure they are both broadcast across the network.

See #512 for context/investigation.

TODO - 

- [ ] Also relax this for `update_header_head`? (Need to think this through...)
    I don't _think_ we need to - not needed during sync?

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -355,7 +355,7 @@ fn update_head(b: &Block, ctx: &mut BlockContext) -> Result<Option<Tip>, Error> 
 		if tip.total_difficulty == ctx.head.total_difficulty {
 			debug!(
 				LOGGER,
-				"pipe: update_head: total_difficulty matches {:?} {:?} at {:?} vs. {:?} {:?} at {:?}",
+				"pipe: update_head: total_difficulty matches {}, {} at {} vs. {}, {} at {}",
 				tip.last_block_h,
 				tip.total_difficulty,
 				tip.height,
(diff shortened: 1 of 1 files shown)


antiochp:

sesam:
UTACK on code.

`update_header_head` is only called from inside:

````
/// Process the block header.
/// This is only ever used during sync and uses a context based on sync_head.
pub fn sync_block_header(
	bh: &BlockHeader,
	mut sync_ctx: BlockContext,
	mut header_ctx: BlockContext,
	validate_header(&bh, &mut sync_ctx)?;
	add_block_header(bh, &mut sync_ctx)?;

	// TODO - confirm this is needed during sync process (I don't see how it is)
	// we do not touch the sumtrees when syncing headers
	// just taking the shared lock
	let _ = header_ctx.sumtrees.write().unwrap();

	// now update the header_head (if new header with most work) and the sync_head (always)
	update_header_head(bh, &mut header_ctx)?;
	update_sync_head(bh, &mut sync_ctx)
}
````

and there it's only about storing a Tip, not about gossiping?

antiochp:
Closing - we can fix #512 with a simpler "broadcast all valid blocks" approach.

PR #515, opened 2017-12-18
broadcast all accepted blocks, main chain or fork

antiochp:
* broadcast all accepted blocks, main chain or fork
* tweak logging to be more consistent

See #512 for context. 
Replaces #514 (wrong approach).


ignopeverell:
LGTM!

PR #523, opened 2017-12-20
Make total_difficulty the sum of network difficulty [testnet2]

### LINKED ISSUE #280: Cumulative difficulty should be based on network difficulty for each block, not hash
Currently, difficulty is accumulated by using the block header hash as if it was the incremental difficulty to be added. This may cause issues as people can get very lucky with a solution. Adding the network difficulty calculated for that block is much more robust. Thanks to @tromp for pointing this out.

sesam:
Do not port to Testnet1! This would hardfork Testnet1.

Change total_difficulty to become sum of network difficulty instead of whatever we get from what the miner happened to mine.

TODO: Fix tests
Fixes #280

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -213,17 +213,19 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 			return Err(Error::DifficultyTooLow);
 		}
 
+		let diff_iter = store::DifficultyIter::from(header.previous, ctx.store.clone());
+		let difficulty =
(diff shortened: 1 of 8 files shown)


ignopeverell:
Code looks good, want to rebase master and fix tests? May be good to add a couple tests covering this as well.

ignopeverell:
Pinging for rebase.

sesam:
cc @antiochp @yeastplume maybe you want to look this over? I'm confident about the grin-part of the code, but unsure about if I fixed tests like I should have.

PR #526, opened 2017-12-20
[testnet1] fixup error handling in wallet restore

antiochp:
`wallet restore` was failing (for some still unknown reason) when recovering amounts in some early outputs.
Also ran rustfmt on `wallet/restore.rs` as an experiment.

### DIFF
--- a/wallet/src/restore.rs
+++ b/wallet/src/restore.rs
@@ -16,22 +16,16 @@ use keychain::{Keychain, Identifier};
 use util::{LOGGER, from_hex};
 use util::secp::pedersen;
 use api;
-use core::core::{Output,SwitchCommitHash};
+use core::core::{Output, SwitchCommitHash};
 use core::core::transaction::{COINBASE_OUTPUT, DEFAULT_OUTPUT, SWITCH_COMMIT_HASH_SIZE};
 use types::{WalletConfig, WalletData, OutputData, OutputStatus, Error};
 use byteorder::{BigEndian, ByteOrder};
 
-pub fn get_chain_height(config: &WalletConfig)->
-	Result<u64, Error>{
-	let url = format!(
-		"{}/v1/chain",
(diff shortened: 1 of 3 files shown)


wallet/src/restore.rs
```diff
@@ -103,85 +112,158 @@ pub fn utxos_batch_block(config: &WalletConfig, start_height: u64, end_height:u6
...
 		}
-		let mut blocks = utxos_batch_block(config, h+1, end_batch)?;
+		let mut blocks = utxos_batch_block(config, h + 1, end_batch)?;
 		blocks.reverse();
-		for block in blocks {
-			let result_vec=find_utxos_with_key(config, keychain, &switch_commit_cache,
-				block, &mut key_iterations, &mut padding);
-			if result_vec.len() > 0 {
-				for output in result_vec.clone() {
-					let _ = WalletData::with_wallet(&config.data_file_dir, |wallet_data| {
```
antiochp: Moved the wallet lock out of the loop (now we get it once per batch of blocks).
We need to be running `wallet restore` without any other processes touching `wallet.dat` anyway.

wallet/src/restore.rs
```diff
@@ -103,85 +112,158 @@ pub fn utxos_batch_block(config: &WalletConfig, start_height: u64, end_height:u6
...
-				if *key_iterations > switch_commit_cache.len() {
-					*key_iterations = switch_commit_cache.len();
+				let res = retrieve_amount_and_coinbase_status(
+					config,
+					keychain,
+					key_id.clone(),
+					&output.commit,
+				);
+
+				if let Ok((amount, is_coinbase)) = res {
```
antiochp: This is the additional error handling.
We may want to investigate this further - but it seems to be affected by some early blocks where we cannot safely recover the amount.

antiochp:
Going to merge this.

Issue #399, opened 2017-11-29
Support for Aggregate (Schnorr) Signatures


ignopeverell:
Likely a meta-issue as this includes multiple moving parts:

* Merge the secp256k1 PRs for aggsig support (461 and 473 I believe) in our fork.
* Integrate it into our fork of Rust secp.
* Implement a building library for aggsig transactions.
* Integrate in the wallet.

yeastplume:
Just leaving these here, as I think these are the most up-to-date descriptions of how a Schnorr enabled transaction should work:

https://lists.launchpad.net/mimblewimble/msg00087.html

https://lists.launchpad.net/mimblewimble/msg00091.html

yeastplume:
Quite happy to take this on when the time is right.. but what's the status of the PR itself? 473 was closed in favour of 486, but It 473 currently looks like it needs a rebase. Think this is a couple of months off, anyhow.

ignopeverell:
It might need a little cleanup, may be good to check with @apoelstra.

apoelstra:
461 needs to be rebased on 486 and it still hasn't seen serious review, only performance review. There is also no good API for single-signer signatures, which is what grin needs. Please do not merge these PRs.

ignopeverell:
@apoelstra my concern with that is mostly additional delays when we could parallelize. Building up all the transaction building logic and interaction between wallet is going to take some time. So we're not talking about having any live code using for at least 3 months. And if the APIs secp APIs change, it will likely be a tiny adjustment on our side.

Does that make more sense in that context?

apoelstra:
Yes, ok, that's reasonable enough. Can you [mark the aggsig functions deprecated](https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md) (or some other more appropriate marking, deprecated is just the one that I know how to use) warning that the API is likely to change and that the crypto is not yet reviewed upstream?

yeastplume:
As a start, performed a merge of 468 and 486 into our local branch, described here: https://github.com/mimblewimble/secp256k1-zkp/issues/9. Will move on to trying to get a rust wrapper around them and see how we get on.

yeastplume:
All good so far for the rust wrapper, have it up and working here with a sample test 5/5 sig:

https://github.com/mimblewimble/rust-secp256k1-zkp/blob/master/src/aggsig.rs

[...]

PR #530, opened 2017-12-21
[WIP] Aggsig Transactions

yeastplume:
See https://github.com/mimblewimble/grin/issues/399 for context

https://github.com/mimblewimble/rust-secp256k1-zkp/blob/master/src/aggsig.rs for latest state of rust aggsig API (still very subject to change)

- [X] Integrate existing aggsig work into mimblewimble secp fork
- [X] creation of single-sig version of API based on existing work
- [X] Additions to single-sig API to support encoding different nonces within schnorr 'e' message (to support @tromp's additive signature scheme)
- [X] Additions to single-sig API to support simple addition of two signatures
- [X] (FIXED) The implementation I have doesn't verify the individual signatures AND the aggregated signature consistently, at the moment I can consistently validate one or the other, but not both. They should both work, and I'm currently working through the maths to figure out what the issue is.
- [X] (Framework done) Automated testing to exercise this (at the moment, there are no tests in the wallet whatsover.. integration tests need to be added, probably a separate file) in grin/tests
- [X] Updating transaction workflow between client and receiver as needed, with approach to keeping state between calls
- [X] Updating validation
- [x] Updating all automated tests

Note that stdout transactions are gone (as we don't have a strategy to handle them now that this is an exchange instead of a single send)

Also, I've added an optional config parameter to turn off waiting for peers on startup (which was slowing down automated testing immensely)

### DIFF
--- a/chain/tests/mine_simple_chain.rs
+++ b/chain/tests/mine_simple_chain.rs
@@ -335,7 +335,10 @@ fn prepare_fork_block_tx(kc: &Keychain, prev: &BlockHeader, chain: &Chain, diff:
 fn prepare_block_nosum(kc: &Keychain, prev: &BlockHeader, diff: u64, txs: Vec<&Transaction>) -> Block {
 	let key_id = kc.derive_key_id(diff as u32).unwrap();
 
-	let mut b = core::core::Block::new(prev, txs, kc, &key_id).unwrap();
+	let mut b = match core::core::Block::new(prev, txs, kc, &key_id) {
+		Err(e) => panic!("{:?}",e),
+		Ok(b) => b
+	};
 	b.header.timestamp = prev.timestamp + time::Duration::seconds(60);
 	b.header.total_difficulty = Difficulty::from_num(diff);
 	b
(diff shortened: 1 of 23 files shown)


yeastplume:
Updated with progress and more detailed tasks.

yeastplume:
Update: This works end-to-end now, and (as advertised) produces transaction signatures without either party revealing their blinding factor (and with non-repudiation for either party).

Not complete yet, as I haven't looked at fixing up any tests that were run as a result, and there are a couple more features to add and cleanup to do. Particularly, I still need to ensure that there are multiple aggsig contexts allowed to support multiple transactions hitting the same wallet at once, and also decide when/how the receiver should decide a transaction isn't going to complete and remove the output from their wallet. But given this is still a first pass, these things can be tweaked/changed as we go along.

@antiochp @ignopeverell Probably in the state where you can start having a look/commenting on the changes (as and when you have time/inclination)

yeastplume:
Tests passing and ready to merge, if no objections. I'm going to open up separate issues for the tasks remaining, as I'd like to get all of the up to now changes merged.

PR #531, opened 2017-12-21
Minimal client implementation

quentinlesceller:
This PR is a very minimal implementation of the ```grin client```/```grin client status```. 
For now, it only displays the protocol version and the user agent based on the content of  [p2p/src/msg.rs](https://github.com/mimblewimble/grin/blob/master/p2p/src/msg.rs). 
Later modifications will display chain eight using the REST API and uptime of the server.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,13 +5,14 @@ authors = ["Ignotus Peverell <igno.peverell@protonmail.com>"]
 exclude = ["**/*.grin", "**/*.grin2"]
 
 [workspace]
-members = ["api", "chain", "config", "core", "grin", "keychain", "p2p", "store", "util", "pool", "wallet"]
+members = ["api", "chain", "client", "config", "core", "grin", "keychain", "p2p", "store", "util", "pool", "wallet"]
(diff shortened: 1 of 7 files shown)


ignopeverell:
Nice, thanks! Main remark would be that I don't this needs to be its own crate. I think it can just be a helper module under `src/bin` for now.

sesam:
This is one thing that interests me; advantages of extracting into crates / not. A pro could be compilation speed, especially if it's fairly unrelated to the other code? Does the "crate wrapping" of code add some overhead too, say for the linker or debug symbol generation?

ignopeverell:
I'm more concerned with humans coming into this repository. I like to go to some implementation and be able to figure out where to go from there to find the code I'm interested in. The more root crates we have, the messier this is. And either way, the compiler can suck it up :-)

quentinlesceller:
Okay I will move it into a module under ```src/bin``` :-).

quentinlesceller:
Closing for https://github.com/mimblewimble/grin/pull/541.

Issue #532, opened 2017-12-21
save_pos_index iteration order vs output order


antiochp:
When saving the `pos_indices` we iterate over the `new_output_commits` and save the pos of each commit.
I don't think this is guaranteed to be in the same order as the canonical order of outputs (we're just iterating over the k,v pairs of the hashmap).

```
	fn save_pos_index(&self) -> Result<(), Error> {
		for (commit, pos) in &self.new_output_commits {
			self.commit_index.save_output_pos(commit, *pos)?;
		}
		for (excess, pos) in &self.new_kernel_excesses {
			self.commit_index.save_kernel_pos(excess, *pos)?;
		}
		Ok(())
	}
```

So I don't think (but I'm not 100% sure) this is guaranteed to rewind to the correct place.
The `last` element here may not be the last element that we save a `pos` for - 

```
		let out_pos_rew = match block.outputs.last() {
```

@ignopeverell am I mistaken here (likely) or is this a smoking gun?

From rust HashMap docs -  
> An iterator visiting all key-value pairs in arbitrary order. The iterator element type is (&'a K, &'a V).

antiochp:
Sample debug logs from #533 showing an "out of order" iteration over `new_output_commits` - 

```
Dec 21 12:27:36.914 DEBG sumtree: save_pos_index: outputs: 3, 
{
	Commitment(0971620eda40cc89513f553ca403d9fd755d8f6bb5bf83c7badb6d17a898981d15): 2664, 
	Commitment(08ebfb6dca466a3e48db1fcb60d864b1e1e28b5d408cdd41c786c42c3d4f99b4cb): 2661, 
	Commitment(08eb595204e5264d941d2fc453c00c898dfa92c7e061665a3a21d1d13b6034a44f): 2663
}
```

Here I believe we would rewind to pos `2663` when the correct pos would actually have been `2664`.

ignopeverell:
The order in which this is saved shouldn't matter, what matters is the position at which they're pushed in the MMR. And that's returned on push. Unless the positions of commitments and kernels in the block is not consistent between nodes but I thought we had that checked and fixed?

antiochp:
As discussed on gitter - this should not actually be a problem...
Reworked #533 to clean some things up and add logging, but not changing behavior.

PR #533, opened 2017-12-21
[testnet1] fix - sort tx inputs/outputs

antiochp:
Ok initial thoughts were incorrect (see #532). 

Took the opportunity to make the following minor changes though -

1) use `sort()` in `write_sorted` to keep things DRY and consistent
2) add some debug logging in `save_pos_index`
3) keep tx inputs/outputs sorted as we add inputs & outputs

TODO - 

- [x] add explicit validation/verification to transaction and block to ensure inputs/outputs/kernels are sorted correctly
- [x] check we validate a tx before putting it in the pool (I assume we do)

### DIFF
--- a/chain/src/sumtree.rs
+++ b/chain/src/sumtree.rs
@@ -319,7 +319,7 @@ impl<'a> Extension<'a> {
 			LOGGER,
 			"sumtree: save_pos_index: outputs: {}, {:?}",
 			self.new_output_commits.len(),
-			self.new_output_commits,
+			self.new_output_commits.values().collect::<Vec<_>>(),
 		);
 
 		for (commit, pos) in &self.new_output_commits {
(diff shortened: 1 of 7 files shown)


ignopeverell:
So there are some assumptions in that code:

* The order of kernels and commitments is consistent across all nodes.
* The last kernel and commitment in a block are the last ones inserted in the MMR (so highest position).

Unless I'm missing something, picking the max positions from the block should always return the last elements?

antiochp:
Discussed on gitter and the intermediate hashmap is fine - this is not affecting the order at all.
Still investigating, may still put some logging in via this PR.

antiochp:
This just makes existing consensu rules more explicit - does not change behavior.
But I think its worth puttiing in as it makes the code cleaner and a bit more readable.
Going to merge to testnet1.

ignopeverell:
That's great, thank you!

PR #538, opened 2017-12-22
log details of connection to wallet coinbase

### LINKED ISSUE #493: Add INFO logging for Connection Lost and Connection Re-established to wallet server
Steps:
1. `grin wallet -e listen`
1. `grin server -m run`
1. Ctrl-C wallet process
1. Eventually see the following in the server log:

```
Dec 14 13:25:34.759 ERRO Failed to get coinbase via wallet API (will retry)...
```

5. `grin wallet -e listen`

You can tell that the connection is reestablished by a lack of the above error output. But it might be good to log a confirmation that the server was able to reestablish a connection with the wallet server. Additionally it might be useful to see log output for the connection being lost.

pjoz:
minor PR to resolve #493 with additional logging of successful receipt of coinbase.

### DIFF
--- a/wallet/src/client.rs
+++ b/wallet/src/client.rs
@@ -38,6 +38,12 @@ pub fn create_coinbase(url: &str, block_fees: &BlockFees) -> Result<CbData, Erro
 				"Failed to get coinbase from {}. Run grin wallet listen",
 				url
 			);
+		} else {
+			info!(
+				LOGGER,
+				"Successfully received coinbase from {}",
+				url
+			)
 		}
 		res
 	})


ignopeverell:
Any update on modifying this so it only logs on error?

sesam:
ping @pjoz are you still able to work on this? Or would you prefer we find someone to take your start on this and run with it, to get a close?

dbancajas:
@ignopeverell @yeastplume  was wondering what is the correct approach on how to record the occurence of a previous error. Can I just declare a static variable inside coinbase and check that in order to make logging "conditional on a previous error"? or should it be in grin/src/miner.rs where "create_coinbase" is being called?

ignopeverell:
It's always hard to tell without seeing the result but I'd try the 2nd option first.

yeastplume:
It's always a bit awkward to make something static in Rust.. I might try adding a set of flags into the actual util.LOGGER struct that can be tweaked as needed (like table that holds name-value pairs that you can add a flag like ONE_TIME or something).. but that's just off the top of my head. Best suggestion is to just try something and we can review.

PR #541, opened 2017-12-22
Minimal client implementation

quentinlesceller:
Update of https://github.com/mimblewimble/grin/pull/531.
This PR is a very minimal implementation of the grin client/grin client status.
For now, it only displays the protocol version and the user agent based on the content of p2p/src/msg.rs.
Later modifications will display chain eight using the REST API and uptime of the server.

PR #543, opened 2017-12-22
Display chain status with grin client status

quentinlesceller:
Based on https://github.com/mimblewimble/grin/pull/541.
Display current chain height, total difficulty and last block pushed.
Will not work with milestone/testnet1 since the format of Tip has changed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,7 @@ grin_grin = { path = "./grin" }
 grin_config = { path = "./config" }
 grin_core = { path = "./core" }
 grin_pow = { path = "./pow"}
+grin_p2p = { path = "./p2p"}
 grin_util = { path = "./util"}
 secp256k1zkp = { git = "https://github.com/mimblewimble/rust-secp256k1-zkp" }
 blake2-rfc = "~0.2.17"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -15,6 +19,7 @@ grin_grin = { path = "./grin" }
 grin_config = { path = "./config" }
 grin_core = { path = "./core" }
 grin_pow = { path = "./pow"}
+grin_p2p = { path = "./p2p"}
 grin_util = { path = "./util"}
 secp256k1zkp = { git = "https://github.com/mimblewimble/rust-secp256k1-zkp" }
```
yeastplume: Just noticed this here.. while you're here, this shouldn't be importing secp256k1 directly here, (it should be using the version exported from util) Is this actually being used directly from src/grin.rs?

quentinlesceller: Indeed, removing it cause no problems as it doesn't look like to be used by src/grin.rs

yeastplume: Okay, thanks! (I know it's tangental to what you're doing) Would you mind making that change and committing (to avoid future confusion). Thanks!

yeastplume:
It all looks like a good start to me, but you have another open PR that seems to be doing the same thing. Is this one supposed to replace that one?

Cargo.toml
quentinlesceller: Sure!

quentinlesceller:
Yes, this one is an extended version of the previous. If this one is okay I can close the previous one.

yeastplume:
Okay, sure go ahead and close that one, and when this passes CI I'll merge.

Issue #554, opened 2017-12-28
Cannot compile Grin Master because of rocksdb 0.7.2 on MacOS 


quentinlesceller:
Seems like I cannot compile Grin Master on MacOS 10.13.2 because of a compilation problem in RocksDB 0.7.2.

Logs:
```
Compiling rocksdb v0.7.2
error[E0425]: cannot find value `rocksdb_recovery_mode_tolerate_corrupted_tail_records` in module `ffi`
  --> /Users/quentin/.cargo/registry/src/github.com-1ecc6299db9ec823/rocksdb-0.7.2/src/db.rs:58:41
   |
58 |     TolerateCorruptedTailRecords = ffi::rocksdb_recovery_mode_tolerate_corrupted_tail_records as isize,
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `ffi`

error[E0425]: cannot find value `rocksdb_recovery_mode_absolute_consistency` in module `ffi`
  --> /Users/quentin/.cargo/registry/src/github.com-1ecc6299db9ec823/rocksdb-0.7.2/src/db.rs:59:32
   |
59 |     AbsoluteConsistency = ffi::rocksdb_recovery_mode_absolute_consistency as isize,
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `ffi`

error[E0425]: cannot find value `rocksdb_recovery_mode_point_in_time` in module `ffi`
  --> /Users/quentin/.cargo/registry/src/github.com-1ecc6299db9ec823/rocksdb-0.7.2/src/db.rs:60:24
   |
60 |     PointInTime = ffi::rocksdb_recovery_mode_point_in_time as isize,
   |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `ffi`

error[E0425]: cannot find value `rocksdb_recovery_mode_skip_any_corrupted_record` in module `ffi`
  --> /Users/quentin/.cargo/registry/src/github.com-1ecc6299db9ec823/rocksdb-0.7.2/src/db.rs:61:35
   |
61 |     SkipAnyCorruptedRecord = ffi::rocksdb_recovery_mode_skip_any_corrupted_record as isize,
[...]

PR #555, opened 2017-12-28
Bump RocksDB to ^0.8.0

quentinlesceller:
Fix to https://github.com/mimblewimble/grin/issues/554.
However, I have no idea if this update breaks something in Grin storage.


antiochp:
If the `TEST_DIR=grin` tests are passing in TravisCI then reasonable to assume the db is still working.
:+1:

antiochp:
I'll merge once travis passes, then port it across to testnet1 branch.

antiochp:
Tested locally against existing db files - so I think we're good.

PR #564, opened 2017-12-31
Add GET peers/a.b.c.d handler in REST API

quentinlesceller:
This PR adds a handler for get ```peers/a.b.c.d```.
This method will look in the store for a peer and displays its information.


ignopeverell:
Thank you for picking up a TODO and cleaning it up!

Issue #559, opened 2017-12-29
discrepancies between UTXO set on two different nodes


antiochp:
Differences between the UTXO sets on two different nodes.
Both nodes at same height.
Both nodes showing identical numbers of inputs and outputs based on traversing the blocks in the chain.
But the UTXO set on each node is different.
Node1 - 4 spent outputs still showing up in UTXO set.
Node2 - 6 spent outputs still showing up in UTXO set, 10 unspent outputs missing from UTXO set.

```
Node 1
======

Spent outputs still in UTXO set
-------------------------------
{ 'commit': '08d973a9ac3a792ed0329f926abbbc091d0922ada0e77a2d3c86443182936d41db',
  'height': 4092,
  'height_spent': 4093,
  'output_type': 'Transaction'}
[ { 'commit': '08d973a9ac3a792ed0329f926abbbc091d0922ada0e77a2d3c86443182936d41db',
    'height': 4092}]
{ 'commit': '089f09f378755d60bbb51290c3c8d94d36fae3059e5db470534f0513a939bc7a57',
  'height': 4069,
  'height_spent': 4070,
  'output_type': 'Transaction'}
[ { 'commit': '089f09f378755d60bbb51290c3c8d94d36fae3059e5db470534f0513a939bc7a57',
    'height': 4069}]
{ 'commit': '082fd6bcfb345f1513c0933fa5948a26d20aea14ebd3d27ff0bce4ff08b636e937',
  'height': 4067,
  'height_spent': 4068,
  'output_type': 'Transaction'}
[ { 'commit': '082fd6bcfb345f1513c0933fa5948a26d20aea14ebd3d27ff0bce4ff08b636e937',
    'height': 4067}]
{ 'commit': '08781f222444deb6a672972b543b05bd27f927a2938e08789f744b5034dc24b914',
  'height': 4041,
  'height_spent': 4042,
  'output_type': 'Transaction'}
[ { 'commit': '08781f222444deb6a672972b543b05bd27f927a2938e08789f744b5034dc24b914',
    'height': 4041}]

--------------------
Inputs:  2010
Outputs:  6450
*** Duplicate (sets) of inputs:  0
*** Duplicate (sets) of outputs:  0
*** Spent yet still in UTXO set:  4
*** Unspent but missing from UTXO set:  0
--------------------

Blocks where these outputs were created - 
4041, 4067, 4069, 4092, 


Node 2
======

Spent outputs still in UTXO set
-------------------------------
[...]

PR #568, opened 2018-01-02
[DNM] Remove header by output commit index (very slow but consistent)

antiochp:
[Do Not Merge]

Experimental approach removing the index so we are always consistent with the blocks themselves.
Mainly so we can verify the problem is with the index itself (and not the UTXO set).

* remove the header by output commit index
* lookup headers by traversing the chain each time (very slow for early outputs)

See #559 for context.

TODO - 
- [ ] reintroduce the index but allow multiple entries per commitment (multiple blocks, different forks)
- [ ] when we read the index we need to - 
    - [ ] check what chain/fork we should be on? We can check heights for main chain?
    - [ ] Is this even possible? Do we have a way to know if a given block is on the "right" fork?
- [ ] Can this be done without needing something like the output pmmr (but for heights?)

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -62,7 +62,12 @@ struct UtxoHandler {
 }
 
 impl UtxoHandler {
-	fn get_utxo(&self, id: &str, include_rp: bool, include_switch: bool) -> Result<Output, Error> {
+	fn get_utxo(
+		&self,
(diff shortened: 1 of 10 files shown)


antiochp:
Superseded by #215 (still WIP but handling coinbase maturity/lock_heights vis switch commits).

PR #575, opened 2018-01-04
Add 'sent' to wallet info output (#288)

heunglee:
The wallet info shows all transfers of successful sending and receiving. Please note that the current enhancement does not include recording local coinbase transactions. The timestamps of "Sent or Received At" is automatically generated by the wallet when each transfer is being processed.

screen shot 2018-01-03 at 12 22 57 am
screen shot 2018-01-03 at 12 24 21 am

### DIFF
--- a/wallet/Cargo.toml
+++ b/wallet/Cargo.toml
@@ -25,6 +25,7 @@ tokio-retry="~0.1.0"
 router = "~0.5.1"
 prettytable-rs = "^0.6"
 term = "~0.4.6"
+time = "^0.1"
 grin_api = { path = "../api" }
 grin_core = { path = "../core" }
 grin_keychain = { path = "../keychain" }
--- a/wallet/src/info.rs
+++ b/wallet/src/info.rs
@@ -15,10 +15,11 @@
 use checker;
 use keychain::Keychain;
 use core::core::amount_to_hr_string;
-use types::{WalletConfig, WalletData, OutputStatus};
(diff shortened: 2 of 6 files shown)


yeastplume:
Looks good ... note that the format of the PartialTx structure is going to change significantly on master due to upcoming aggsig changes. I think perhaps the best thing to do here is go ahead and merge on the testnet branch to test the functionality in the field a bit, and then cherrypick/merge onto master once aggsig transactions are in place (hopefully within the next couple of weeks).

Could you fix the current conflicts then I'll merge the PR?

heunglee:
I am aware of wallet functionality and data storage being changed. This PR is thought to be an interim solution to track transfer history for efficient developments.

mably:
Doesn't show anything on my side after a wallet restore, is it normal?

heunglee:
@mably The current enhancement does not support the restoration of transfer history. All the history of transfers is stored in wallet.stats file. If you delete the file, the history cannot be restored currently even though the outputs of wellet.dat can be restored by blockchain. Some data of each transfer record are not stored in chain, such as destination and timestamp of transfer action. They are not restored if the storage is deleted. In addition, Aggsig feature will change the workflow of the transaction process in the wallet. So this enhancement is on usability test on testnet. So please back up wallet.stats file before you remove the whole files under wallet directory.

The further enhancement may be to store each transfer record in persistent data storage like rocksdb to prevent loss of history.

mably:
Ok, so it's more like some kind of wallet activity logs, I'm still reasoning like if it is was a good old Bitcoin-like coin :wink:

PR #576, opened 2018-01-04
pmmr should not depend on switch_commit_hash to identify utxo

antiochp:
The utxo pmmr represents the state of the utxo set for a given chain.
For a given chain a commitment should be sufficient to identify a utxo.

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -119,11 +119,7 @@ impl UtxoHandler {
 		let outputs = block
 			.outputs
 			.iter()
-			.filter(|c| {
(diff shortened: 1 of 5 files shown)


ignopeverell:
Assuming the tests pass and the conflicts are resolved, this looks good to me. Note that we can't merge this on testnet1 as it invalidates all roots calculated there.

yeastplume:
Fine for me as well, it's just reversing some of the work done in here.

https://github.com/mimblewimble/grin/pull/179

We may need to revisit switch commits entirely.

PR #577, opened 2018-01-04
Fix typo ''atheight'' instead of ''byheight'' in utxo handler

quentinlesceller:


antiochp:
We should make sure this doesn't break anything (testing around the api endpoints is pretty minimal).
I _think_ "atheight" api endpoint is used in the wallet code in a couple of places.

quentinlesceller:
Oups... Yes, after quick search, it looks atheight is used in ```walllet/src/restore.rs```. Fixed it.

antiochp:
:+1:

PR #579, opened 2018-01-05
Fix #321 Limit length of API utxo queries

### LINKED ISSUE #321: url length limit when refreshing wallet outputs (many query params)
```
GET /v1/chain/utxos?id=xxx&id=yyy&id=zzz
```

Not sure what the limit is here but we are going to hit it at some point now that we have a large chain on testnet.
Wallet should probably split this up into multiple smaller api calls.

quentinlesceller:
API utxo queries are now done in batch to avoid errors. See https://github.com/mimblewimble/grin/issues/321 for context.


antiochp:
:+1:

PR #581, opened 2018-01-06
Allow the wallet to be queried for balances and to send a transaction…

camalolo:
… via new handlers

Hello grin masters,

I was playing around with the idea of writing a gui wallet for grin, but I could not find any way to query the balances or send grin units to another wallet. I also am new at rust and I wanted to get familiar with it, and I would be grateful to get your remarks if any.

So I implemented two new handlers for the wallet listener :

/v1/info returns a json encoded struct containing the various balances as calculated in wallet/src/info.rs :

```
> curl http://127.0.0.1:15000/v1/info
{"unspent_total":"150.004500000","unspent_but_locked_total":"150.004500000","unconfirmed_total":"2720.003000000","locked_total":"100.000000000"}
```

/v1/send/transaction expects a json encoded struct containing the details about the transaction to send and returns a json encoded struct with a single boolean value 'confirmed' if there was no error :

```
> curl -H "Content-Type: application/json" -t":"10.000","minimum_confirmations":0,"dest":"http:\/\/127.0.0.1:35000","selection_strategy":"all"}' http://localhost:15000/v1/send/transaction
{"confirmed":true}
```

Finally I reorganized a little the sources in wallet/src so all handlers go to handlers.rs

Obviously this is very crude, and I am not sure you will think it's useful, but it would be an honor to contribute to grin and learn rust while doing that.

Thank you for your time

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,26 +36,46 @@ env:
 
 after_success: |
   pwd &&
-  echo "HERE" &&
   wget https://github.com/mimblewimble/kcov/archive/master.tar.gz &&
   tar xzf master.tar.gz &&
   cd kcov-master &&
-  mkdir build &&
-  cd build &&
-  cmake .. &&
-  make &&
-  sudo make install &&
+  mkdir build && cd build &&
+  cmake .. && make && mv src/kcov ../.. &&
   cd ../.. &&
   rm -rf kcov-master &&
-  cargo test --no-run &&
   cd ../
-  pwd &&
-  kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_core* && 
+  cd api && cargo test --no-run && cd .. && rm target/debug/*.d &&
+  ./kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_api* &&
+  echo "Finished coverage for grin_api"
+  cd core && cargo test --no-run && cd .. && rm target/debug/*.d &&
+  ./kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov target/debug/grin_core* &&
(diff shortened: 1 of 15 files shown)


yeastplume:
Hi @camalolo .. First thanks for submitting this, this is definitely something that I think will be useful in the future, particularly an API to allow other services/GUIs to access wallet functionality without going through the command line.

Unfortunately, right now a lot of the changes you have here are very soon to be obsoleted due to work on aggsigg (schnorr) which I'll hopefully be integrating in the next couple of weeks (see the changes in https://github.com/mimblewimble/grin/pull/530) In particular, stdout transactions are going away (at least for now,) as the transaction exchange is becoming much more interactive than the simple request/response that exists now.

If we merge this now, it will just create headaches when merging #530 for functionality that will no longer be relevant. Would you mind waiting until that PR is done/merged, then perhaps looking at this again, particularly the send API service as well as a possible handler API cleanup?

camalolo:
No problem, let's check this later !

I think I totally overlooked the security aspect, I should at least ask the password when accepting an info / transaction send request, and maybe it would make more sense to have a separate api port listening on 127.0.0.1 for the wallet operations that need a security check ?

Thank you very much for taking the time to check my pull request !

camalolo:
As an excercise I did what I suggested : I moved the wallet "operations" part to a separate listener that can only listen on 127.0.0.1 and I added a requirement for the passphrase to be sent along with other information. The keychain is derived dynamically at each request so that one cannot use the /info and /transaction/send api calls without knowing the passphrase.

Cheers

camalolo:
Closed this as I seem to have rendered this pull request difficult to read after rebasing the recent commits into my repository. I will create a new pull-request.

PR #583, opened 2018-01-06
maintain switch_commit_hashes in the output pmmr (not used yet)

antiochp:
Initial pass at maintaining `switch_commit_hash` in the `output_pmmr` (in addition to the existing `commitment`).
We ignore the `switch_commit_hash` for the purpose of hashing the `SumCommit` (so we can still compare hashes using just the `commitment`).

This is a required piece of work for getting #215 fully working.
We will use the `switch_commit_hash` in the pmmr to verify coinbase maturity via the `lock_height`.


ignopeverell:
Looks good to me.

Trevorulliac:
api.blockcypher.com/v1/btc/test

PR #588, opened 2018-01-08
Wallet operator (clean) (was #586)

camalolo:
I am terribly sorry, I messed up the git history of my pull request #586. Looks like I need to read more documentation about pull requests !


ignopeverell:
Don't worry about your commit history, they who never got confused by the git command line can cast the first stone...

Like @yeastplume I have some security concerns. But for now, given that you're limiting it to localhost, I think this is ok. I have a couple requests until aggsigs are merged that you could tackle:

* I'd remove TLS support. When we actually do TLS on the wallet, I'll want it a lot more tied down (i.e. TLS 1.2 only, known secure ciphers, some solution to generate a cert, etc.)
* You seem to have some indentation issues, some code doesn't line up.

camalolo:
Hello @ignopeverell , thank you for your feedback. I removed the TLS part via a git revert + a small commit to fix the tests, not sure if it's the best way to do this, if you prefer me to make another branch with only the other commits just let me know and I'll do that.

I saw you opened issue #591 about the current topic, maybe I can propose another pull request about the TLS thing when there's enough information about the direction to be taken, if I am able to grasp it :D Anyways just let me know if I can help.

Cheers

camalolo:
PS : Looking into the indentations problems you mentionned

ignopeverell:
Thanks! And don't worry about doing another branch, I usually squash the history on PRs.

camalolo:
Closing this for now, will work on those conflicts

PR #597, opened 2018-01-10
coinbase maturity and output lock height docs

antiochp:
TODO - 
- [ ] define the problem
- [ ] lay out the proposed solution

### DIFF
--- a/doc/coinbase_maturity.md
+++ b/doc/coinbase_maturity.md
@@ -1,0 +1,38 @@
+# Coinbase Maturity
+
+Coinbase (block reward + fees) outputs are "locked" and require 1,000 confirmations (blocks added to the current chain) before they mature sufficiently to be spendable. This is to avoid subsequent txs from being at risk of being reversed if a chain reorganization occurs.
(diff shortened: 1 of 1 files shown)


antiochp:
@ignopeverell @yeastplume this is ready for review - feedback please.
I _think_ this can be done without needing to use the switch commitment hash.

tl;dr We don't need to include the lock_height in the input (and then jump through hoops to ensure the miner is not lying about this).
We can achieve this by hashing the output (not just the commitment) when we store the hash in the output MMR (therefore including coinbase vs non-coinbase in the value being hashed).
And we can use the merkle proof to verify inclusion of the output in the given block (and therefore determine the lock_height based on the originating block).

And we can do this with only block_headers and the output MMR.
Short term - we can use full block data (and migrate to block header only once we have merkle proofs implemented).

Thoughts? Concerns?

ignopeverell:
Yes, I think that works.

yeastplume:
Makes sense, simple enough approach and looks like it will work. We want to be 100% certain we can jam the switch commit (or whatever it evolves into) into a bullet proof, but as far as I understand this will be possible, so all good.

PR #599, opened 2018-01-11
Update ```grin client status```

quentinlesceller:
Updated the ```grin client status```to display useful information using a new ```v1/status``` handler:
- Protocol version.
- User agent.
- Number of connected peers.
- Info on current Tip.

In a cleaner way.


ignopeverell:
Looks good!

Issue #396, opened 2017-11-29
[wallet] panic: 'attempt to subtract with overflow' when sending small quantities


AsrielBelacqua:
wallet listener log: http://termbin.com/zmmi

I was trying to send a tiny quantity (1e-8 grins). Wallet send output:

```
Nov 29 00:17:07.655 INFO Using configuration file at: /home/stelmaria/src/grin/target/debug/grin.toml
Nov 29 00:17:07.660 DEBG Using wallet seed file at: ./wallet.seed
Nov 29 00:17:07.892 DEBG Refreshing wallet outputs
Nov 29 00:17:09.668 INFO Acquiring wallet lock ...
Nov 29 00:17:09.670 DEBG Attempting to acquire wallet lock
Nov 29 00:17:09.834 INFO ... released wallet lock
Nov 29 00:17:09.974 INFO Acquiring wallet lock ...
Nov 29 00:17:09.974 DEBG Attempting to acquire wallet lock
Nov 29 00:17:10.155 INFO ... released wallet lock
Nov 29 00:17:14.198 DEBG Posting partial transaction to http://127.0.0.1:13415/v1/receive/transaction
Nov 29 00:17:14.236 ERRO Error sending transaction - status: 500 Internal Server Error
Nov 29 00:17:14.237 ERRO Communication with receiver failed. Aborting transaction
Nov 29 00:17:14.237 INFO Acquiring wallet lock ...
Nov 29 00:17:14.237 DEBG Attempting to acquire wallet lock
Nov 29 00:17:14.351 INFO cleaning up unused change output from wallet
Nov 29 00:17:14.391 INFO ... released wallet lock
Nov 29 00:17:14.392 ERRO Tx not sent: Hyper(Status)
```

antiochp:
Hmm - probably related to the fact we are (incorrectly) building the tx as "receiver pays the fee" when we should be "sender pays the fee" (related #274).
So we're likely trying to subtract a fee thats larger the total amount being sent.

mably:
I might have been hit by the same bug: https://hastebin.com/yujufatozu.rb

I was trying to list my wallet outputs, the process crashed.

heunglee:
The receiver tries to subtract fee from amount and the result becomes negative and overflow takes place. The problem is that fee is more than the amount. It is not matter of who subtracts fee.
It is necessary to review fee policy for micro grin transfers, such as adjustment of weight of fee for micro grin transfers.

[test case]
[...]

PR #603, opened 2018-01-12
[wallet] panic: 'attempt to subtract with overflow' when sending small quantities #396

heunglee:
PR for [wallet] panic: 'attempt to subtract with overflow' when sending small quantities #396


yeastplume:
Seems good, thanks. If you just want to fix up the tests for the testnet branch I can merge that as well.

PR #604, opened 2018-01-12
Milestone/testnet1 - [wallet] panic: 'attempt to subtract with overflow' when sending small quantities #396

heunglee:
Fix for [wallet] panic: 'attempt to subtract with overflow' when sending small quantities #396

A separate PR for milestone/testnet1 because of backward incompatibility of wallet source codes on master.


yeastplume:
Thanks for these fixes... all look fine, will merge when all CI tests pass

Issue #573, opened 2018-01-03
While syncing with the testnet1 network around block 9938 - thread '<unknown>' has overflowed its stack


Thireus:
I would like to report a stack overflow bug which happened while grin server was still syncing with the network around block 9938.

I followed the steps on https://github.com/ignopeverell/grin/blob/master/doc/build.md. Build grin on ARMv7 server with Debian 9.

For some reason after several hours of uptime, the grin server has crashed:

```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
Aborted
```

Last grin.log:

```
Jan 03 22:28:09.206 DEBG Sumtree extension done.
Jan 03 22:28:09.207 DEBG (Server ID: Port 13414) Mining at Cuckoo16 for at most 90 secs on block a7be852c at difficulty 500.
Jan 03 22:28:09.207 DEBG Starting new sumtree extension.
Jan 03 22:28:09.395 DEBG (Server ID: Port 13414) No solution found after 1 iterations, continuing...
Jan 03 22:28:09.396 DEBG setting pubkey in miner to pubkey from block_fees - BlockFees { fees: 0, height: 9938, key_id: Some(Identifier(b88c8862b83928880a37)) }
Jan 03 22:28:09.396 DEBG in miner loop...
Jan 03 22:28:09.397 TRCE key_id: Some(Identifier(b88c8862b83928880a37))
Jan 03 22:28:09.878 DEBG chain_store: get_output_pos: Commitment(097193c5eddd79daaf48f1007a66991998a82992b14bf27b9dd5aca0f12a1b87a9), Err(NotFoundErr)
Jan 03 22:28:09.879 DEBG sumtree: get_output_pos (idx): Commitment(097193c5eddd79daaf48f1007a66991998a82992b14bf27b9dd5aca0f12a1b87a9), Err(NotFoundErr)
Jan 03 22:28:09.879 TRCE Hashing with additional data
```

I'm not sure what else I can do to debug this issue. I've restarted the server in the meantime to continue syncing with the testnet1 network.

antiochp:
Hmm - I think I've seen this a couple of times as well. Doesn't happen all the time.
I believe there is a recursive call somewhere (maybe handling a really old fork before we ban a peer?) that recurses over the full block chain (or header chain) and the chain is too big to do this safely via recursion.

Will dig into this further if I get chance.

Thanks for the report!

Jiloc:
[...]

PR #605, opened 2018-01-12
Attempted fix for stack overflow #573

yeastplume:
Attempted fix for https://github.com/mimblewimble/grin/issues/573 ... just an attempt to rewrite the orphan-check logic non-recursively

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -203,7 +203,7 @@ impl Chain {
 	/// has been added to the longest chain, None if it's added to an (as of
 	/// now) orphan chain.
 	pub fn process_block(&self, b: Block, opts: Options)
-		-> Result<Option<Tip>, Error>
+		-> Result<(Option<Tip>, Option<Block>), Error>
 	{
 		let head = self.store
 			.head()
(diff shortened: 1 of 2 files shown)


yeastplume:
@antiochp This seems to fix the issue, all it does is make the orphan check upon successfully adding a block non-recursive. Could you give it a once-over just in case there's anything I'd inadvertently break with this? Also, the call in adapters.rs to check_orphans didn't exist at all in master (the master version of this at #606 fixes that). I assume that's just an oversight on master?

ignopeverell:
I would just make `process_block` call some `process_block_no_orphans` or something similar,which would be the current `process_block`, followed by the orphan check. I don't think there is a situation where a consumer of the chain would *not* want to check the orphans, so encapsulating it makes sense.

ignopeverell:
Disagree with the suggestion? It'd be good to merge this.

yeastplume:
Sorry, will have a think over the next day or two and finalise. Neither agree nor disagree.. just moved on to other things without internalizing the suggestion :D

yeastplume:
Think this and #606 are good now, if you're okay with their current form

PR #615, opened 2018-01-15
hash (features|commitment) in output mmr

antiochp:
This PR solves #559 by extending the output pmmr to include the output features in the hash.

The proposed solution is described here in more detail - #597 

Replaces #215 (which attempted to encode lock_height in the switch commit)

tl;dr
1. Getting rid of the indexes ensures we never see the "wrong" version of an output (due to a fork).
Specifically we never see (and use) the wrong lock_height (which can and will break consensus).
1. We now include output features in the output mmr. This ensures a miner cannot lie about coinbase vs non-coinbase (we can be sure we are checking lock_height as necessary)
1. We derive lock_height by looking at the full block (we do not rely on the old index) so we know lock_height is being correctly handed 

The combination of these changes should ensure we handle the spending of coinbase outputs correctly even under adverse conditions (chain forks and miners reusing wallet keys, creating duplicate commitments).

--------------------------------------

* `SumCommit` takes `OutputFeatures` & `Commitment`
  * the hash is now `(features|commit)`
  * this ensures a given output in the MMR is known to be a coinbase or not (and miner cannot lie)
* introduce `OutputIdentifier` so we don't need to pass a `SumCommit` around everywhere 
  * we now need an `OutputIdentifier` to check `is_unspent` via the MMR (by comparing the hash)
* to spend an output we also need to specify the block hash in the `Input`
  * we can determine `lock_height` based on height of block + 1,000
  * miner cannot lie about `lock_height` (also cannot lie about coinbase vs non-coinbase)
* removed "header by commit" index
  * sender needs to provide block hash
  * wallet maintains block hash of outputs
  * this ensures we never mis-attribute an output in the presence of a chain fork
* removed "output by commit" index
  * if we need the full output we need to go back to the full block (via block hash)
* get utxo api endpoint now just returns a commitment (output mmr data only)
* rework wallet refresh and restore to handle output mmr changes and need to maintain block hashes

TODO - 
- [x] _many_ tests failing (tests failing to build)
- [x] _many_ build warnings, lack of docs, unused imports etc.
- [x] cleanup verbose debug logging
- [x] make block hash optional on an input (only necessary for spending coinbase outputs)


antiochp:
@ignopeverell this should be in a reasonable place for an early review.
Tests are still failing... and there are a lot of build warnings... but the approach outlined in #597 is implemented (and works locally).
If we are happy with this and think this is worth continuing with then I'll continue to clean it up and get the tests reworked/fixed up.

chain/src/sumtree.rs
```diff
@@ -89,27 +88,74 @@ impl SumTrees {
...
+
+					// At this point we can be sure the input is spending the output
+					// it claims to be spending, and it is coinbase or non-coinbase.
+					// If we are spending a coinbase output then go find the block
+					// and check the coinbase maturity rule is being met.
+					if input.features.contains(COINBASE_OUTPUT) {
+						// edge case here - we may be spending an output with zero-confirmations
+						// and it just got included in a block.
+						// The wallet may not know about this new block yet (not yet refreshed).
+						// Is it safe to assume this the tx was included in the latest block?
```
ignopeverell: If I understand well, I don't think we need to worry about that. We explicitly disallow inputs spending outputs in the same block, as they should have been cut-through. This happens in input validation as we're doing it before doing the outputs.

chain/src/sumtree.rs
```diff
@@ -304,6 +350,41 @@ impl<'a> Extension<'a> {
...
+				if hash != hash_sum.hash {
+					return Err(Error::SumTreeErr(format!("output pmmr hash mismatch")));
+				}
+
+				// At this point we can be sure the input is spending the output
+				// it claims to be spending, and it is coinbase or non-coinbase.
+				// If we are spending a coinbase output then go find the block
+				// and check the coinbase maturity rule is being met.
+				if input.features.contains(COINBASE_OUTPUT) {
+					// edge case here - we may be spending an output with zero-confirmations
```
ignopeverell: Same as previous.

chain/tests/test_coinbase_maturity.rs
```diff
@@ -176,6 +183,23 @@ fn test_coinbase_maturity() {
 
 	let prev = chain.head_header().unwrap();
 
+<<<<<<< HEAD
```
ignopeverell: Looks like you're committing a merge.

ignopeverell:
Minor comments but looks good to me overall!

chain/src/sumtree.rs
antiochp: Yeah - good catch. Just revisiting this now and this is unnecessary.

chain/tests/test_coinbase_maturity.rs
antiochp: oof

antiochp:
@ignopeverell this should be good to go now if you want to give it a final review?
(3rd time lucky - but this feels like the "right" solution now). 😀

ignopeverell:
Looking good, ship it!

antiochp:
Soon as travis sorts out its infra... multiple builds stuck for a couple of hours. Just kicked off a new one.

antiochp:
Merging - travis is still working through its backlog...

Issue #613, opened 2018-01-13
Serialize Aggsigs directly instead of DER


yeastplume:
Change serialization of aggsigs to just be the raw 64 bytes instead of DER encoded, to save space and make kernel storage easier.

PR #618, opened 2018-01-15
Aggsig Serialization

yeastplume:
For #613, creates read/write serialization functions for ffi::Signature struct (which can now hold aggsigs as well,) and change TxKernel/Transaction structs to use it. Sigs will now be serialized as [u8;64]

Note I've also modified the ser_multiread macro to allow for calling `as_fixed_bytes(arg)` with an argument, though I didn't end up using it. Might be handle to keep there.

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -20,6 +20,7 @@ use p2p;
 use util::secp::pedersen;
 use rest::*;
 use util;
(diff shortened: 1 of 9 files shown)


yeastplume:
@ignopeverell If you just want to cast a quick eye over before merging

PR #625, opened 2018-01-17
rework WalletData and OutputData serialization 

antiochp:
* serialize block hashes cleanly in wallet.dat (as hex string, not a vec of byte values)
* persist vec of values (and not the full map) in wallet.dat
  * this lets us sort it so we can diff the file more easily

```
[
  {
    "root_key_id": "a028f7f87c10b785203e",
    "key_id": "21536a327cc884e164f2",
    "n_child": 8,
    "value": 60004000000,
    "status": "Unspent",
    "height": 6,
    "lock_height": 9,
    "is_coinbase": true,
    "block": "daa87aa69d0bdfac87f4494f9fec797a815dbba03d0312173efb95c64fdd7a1f"
  },
  ...
]
```

### DIFF
--- a/keychain/src/extkey.rs
+++ b/keychain/src/extkey.rs
@@ -74,7 +74,7 @@ impl error::Error for Error {
 	}
 }
 
-#[derive(Clone, PartialEq, Eq, Hash)]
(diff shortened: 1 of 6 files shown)


antiochp:
Merging this (master only) - no behavioral changes.

Issue #544, opened 2017-12-23
duplicate commitments and sumtree positions


antiochp:
Related - #540 

Just been thinking through what it means to allow duplicate commitments and how this might interact with sumtree positions.

We don't allow duplicate commitments in the UTXO set _but_ we do allow duplicates if one is spent before the second is created.

I _think_ we have a problem given the following scenario - 

Commitment `aaaa` as last output in block 5 at sumtree pos 50.
Commitment `aaaa` spent (as input) in block 7.
Commit `aaaa` as (duplicate commitment) output in block 9 at pos 72 (overwriting index that pointed to pos 50).
We then fork and rewind to block 5, we need to rewind to pos 50 but we no longer have pos 50 in the index.

If we only store the "last" pos of any given commitment.
And we permit duplicate commitments.
Then, we can get into trouble if we need to rewind to (or beyond?) pos of previous versions of the duplicate commitment.

ignopeverell:
Mmmh you're right. This is annoying, I need to think some more about that.

sesam:
So we need to track which commitments are used and/or set a max depth for this tracking, likely also limiting reorg depth the same way to avoid complicating the implementation. Or do you have some tricks up your sleeve?

antiochp:
@ignopeverell Can we track pos via a stack per commitment (rather than a single pos for each commitment)?
i.e. every time we add to the sumtree we push the pos onto the stack maintained for that commitment?
Then on reading the pos for a given commitment we get a stack of possible values back - the latest may be "out of bounds" and invalid, so we would pop and look at the next pos?

I haven't thought this through fully - just thinking out loud (apologies if this makes no sense for some obvious reason...)
Definitely haven't fully thought through what would happen during a rewind.

ignopeverell:
It makes sense, I was thinking of something along the same lines, just needs to be a simple array in store (instead of a single element). The additional complexity still bothers me a little but haven't found much better.

It also occurred to me we could simplify the `chain/src/sumtrees.rs` implementation by having a MMR wrapper that implements the same `get`, `prune`, etc. operations but taking commitments instead of positions. It should encapsulate the use of the index a little better.

antiochp:
Now that we've merged #615 and hopefully fixed the underlying issues causing #559 - 
I wonder if we can use a related approach to get rid of this "multiple sumtree positions" issue.

We now track the block hash in the wallet for each output.
The block hash is required if we spend a coinbase output.
[...]

PR #626, opened 2018-01-17
[wip] [experimental] include block hash in output identifier

antiochp:
This _potentially_ solves the issue in #544 where rewinding on a fork can resurrect a once spent output back into the utxo set (resulting in incorrect output mmr pos).

* include `block_hash` in `output_identifier` (so ensure they are unique)
* use hash of `output_identifier` to build key for output_pos index
  * `features|commitment|block_hash`

Note - 
* this breaks the utxo endpoint as we need the block hash to retrieve a utxo by commitment
* also do we want to include block_hash in the entry hash in the MMR itself (or just as part of how we maintain the output_pos index)? Does it make a difference?

Putting this up for discussion.
This is quite likely the _wrong_ approach to solving this but keen on getting some feedback and thoughts on how we are relying on commitments right now.
@ignopeverell

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -59,6 +59,9 @@ struct UtxoHandler {
 }
 
 impl UtxoHandler {
+	// TODO - these are now broken - we _require_ the block hash to find unspent outputs
+	// TODO - rework api endpoint as utxo/:block_hash/:out_commit so we know the block?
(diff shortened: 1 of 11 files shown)


antiochp:
This ties protocol to implementation too tightly.
Closing.

Issue #627, opened 2018-01-17
Remove the range proof MMR


ignopeverell:
It occurred to me today that committing to range proof hashes in a MMR doesn't really help us for anything except malleability. And that we didn't really care about range proof malleability.

Here is the whole argument (hopefully all true statements):

* Only the owner of the output can generate a new range proof.
* The range proof doesn't participate in uniquely identifying the output.
* If the output owner tries to change its range proof, nothing at the protocol level cares as long as it remains valid.
* The output commitments are committed in their MMR, we only care that a range proof exists and can be provided by a node for all of these.
* The range proof also signs the hashed switch commitment (at least once #207 is fixed), so allowing a malleated range proof could be a problem. But the switch commitment hash is committed to in the UTXO MMR, so it's not (and actually makes it even harder to malleate the range proof).

Am I missing something? By removing the range proof MMR we would gain 32 bytes in each header and log2(N)*32 bytes of state data for the MMR itself.

To be clear, I'm not suggesting removing range proofs or any of the validation or syncing logic. Just commiting to their hashes in a dedicated MMR.

/cc @yeastplume @antiochp @apoelstra

yeastplume:
I've tried to give this some thought as well.. trying to think of what can be malleateificatedified in a range proof and what attack vectors it could give. Thus far I can think of

-You can intercept and generate a new range proof with different keys, which won't validate for that output, so moot.

-If you know the keys, you can swap out generate a new range proof with the same keys, which will produce a different hash due to random elements in generating the proof (not 100% sure, but assuming). However, this shouldn't matter as the malleated proof will work just as well and doesn't break consensus

-You can modify the stored amount used for wallet restore.. which again doesn't change the validity of the proof.. again this can't be done without knowing the keys, and is very much optional whether the user wants to store it there or not (if they don't it just means they won't be able to restore their wallet unless they keep amounts around as well). Shouldn't matter

-You can intercept/change the switch commit after #207... (which probably isn't worth doing till bulletproofs). So this needs to stay in the UTXO sumtree. Validation shouldn't require looking at the switch commit anyhow, except post-quantamageddon. You could change this to be nonsense, but again, not without the private keys, and it's up to the user whether they even want to include it in the first place. 

[...]

PR #630, opened 2018-01-18
add switch_commit_hash back into sum_commit

antiochp:
Related - https://github.com/mimblewimble/grin/issues/627


core/src/core/transaction.rs
```diff
@@ -839,10 +858,11 @@ impl ops::Add for SumCommit {
...
 			Err(_) => Commitment::from_vec(vec![1; 33]),
 		};
-		SumCommit::new(
-			self.features | other.features,
-			&sum,
-		)
+		SumCommit {
+			features: self.features | other.features,
+			commit: sum,
+			switch_commit_hash: SwitchCommitHash::zero(),
```
ignopeverell: I think that'd be better as the hash of the 2 commit hashes, so that it can roll up the tree as well.

PR #631, opened 2018-01-18
 Fix #439 Temporary peer banning

### LINKED ISSUE #439: Peer banning should be temporary

quentinlesceller:
Fix #439 peer banning is now temporary.
After 3 hours the peer is unbanned.
Tested on testnet1.

### DIFF
--- a/grin/src/seed.rs
+++ b/grin/src/seed.rs
@@ -86,7 +86,8 @@ impl Seeder {
 		let ban_windows: i64 = 10800;
 
 		// now spawn a new future to regularly check if we need to acquire more peers
-		// and if so, gets them from db and unban the banned peers after the ban is expired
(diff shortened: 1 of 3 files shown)


ignopeverell:
Nice and simple, thanks! For future PRs, review would be easier if there was "noise" due to rustfmt, even though fixing formatting is always a good thing.

PR #634, opened 2018-01-18
summing two sum_commits rolls the switch commit hashes up

antiochp:


antiochp:
@ignopeverell I think this is right - mind taking a look just to confirm?

ignopeverell:
Looks right!

PR #635, opened 2018-01-18
Revert "move siphash out of pow and into util"

antiochp:
Reverts mimblewimble/grin#633 (this siphash impl is specific to cuckoo/pow).
We are going to use a more general purpose impl for short ids.

### DIFF
--- a/pow/src/cuckoo.rs
+++ b/pow/src/cuckoo.rs
@@ -23,7 +23,7 @@ use std::cmp;
 use blake2;
 
 use core::core::Proof;
-use util::siphash::siphash24;
+use siphash::siphash24;
 use MiningWorker;
 
 const MAXPATHLEN: usize = 8192;
--- a/pow/src/lib.rs
+++ b/pow/src/lib.rs
@@ -44,6 +44,7 @@ extern crate grin_util as util;
 
 extern crate cuckoo_miner;
 
+mod siphash;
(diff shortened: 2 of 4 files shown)


ignopeverell:
Uh, I think you accidently reverted #631 as well?

antiochp:
What is going on here? 
I don't understand why that other commit is in there in this PR...
The diffs don't appear to include anything other than what I intended to revert either (there is nothing related to #631 or it reversion in the diff here).

This is what I did to generate this PR - 
* click the "revert button" on my original PR
* merge it...

And clicking on https://github.com/mimblewimble/grin/pull/635/commits/a0793dd3765b73a91d11b07c83a01b1a2a9d681d above shows an empty commit...

So I _think_ we're ok (other than a nasty looking revert commit message).

quentinlesceller:
Yeah it looks like #631 is still there.

antiochp:
Code from #631 is definitely still in master - 
https://github.com/mimblewimble/grin/blob/master/grin/src/seed.rs#L86

This must have been a timing issue or something on github side of things...

ignopeverell:
Ok cool, sorry I should have double-checked the diff.

antiochp:
No worries - it was weird that github included this "extra" commit here...
Thanks for catching and flagging this!

Issue #522, opened 2017-12-19
Compact Blocks


antiochp:
Tracking subtasks here - 
- [x] short ids - #632
- [x] CompactBlock model with vecs of short ids for inputs|output|kernels - #637 
  - [x] Note: we want coinbase outputs/kernels to remain "full" (we never see these in advance) - #641 
  - [x] limit # coinbase outputs in a tx? #644
  - [x] only include kernel short_ids in comact blocks #668 
- [x] API support for retrieving a compact block based on height/hash (for debugging etc.) - #637
- [x] modify p2p block propagation to broadcast "header first" - #654
  - [x] ask for full block after receiving a header
  - [x] ask for compact block first
- [x] p2p support for requesting full block if we cannot hydrate compact block successfully (i.e. we are missing some inputs/outputs/kernels from the underlying txs)
- [x] tx pool (or otherwise) support for querying what we have available (and what we are missing) by short id(s) #645
- [x] functionality to take a compact block and hydrate it into a full block as necessary

---------

Rough notes, may be wildly inaccurate.

* Nodes already have the txs in their tx pools - compact blocks avoid sending this data around a 2nd time as part of the block.
* Compact blocks _may_ remove the need for "header first propagation" mimblewimble/grin/issues/257 (full blocks vs. compact blocks vs. block headers)
* non-existence of txs (in the pool?) is a complicating factor
* cut-through in the tx pool?

-----------

Notes on tx pool and input/output/kernel short_ids - 

We maintain a map of transactions by hash in the pool.
If a node receive a compact_block it needs to go to the tx pool and look for all inputs/outputs/kernels based on short_ids.
But we store them as txs, not granular as inputs/outputs/kernels.
So - add an additional map that indexes tx hashes by short_ids of all inputs/outputs/kernels?
This would allow us to retrieve all txs - and then we can build the block from these?

sesam:
In this github repo, searching /compact.*block/ mostly gave core/src/core/block.rs that has functions compactable_block() and compact(&self). Grin maillist also not so much, so probably this was debated on other maillists and/or reddit.

ignopeverell:
See here:

https://lists.launchpad.net/mimblewimble/msg00039.html

Check all the way to the end as there is a vulnerability in the initial scheme that I discovered later on (https://lists.launchpad.net/mimblewimble/msg00111.html). Note that any short identifier used for inputs/outputs/kernels need to be matched with its equivalent in the transaction pool.

[...]

PR #637, opened 2018-01-18
ShortId implementation (and CompactBlock)

### LINKED ISSUE #632: short ids for inputs|outputs|kernels
This is part of the overall "compact blocks" proposal - #522.

The plan is for something analogous to "short transaction ids" in bitcoin (adapted for the more granular inputs|outputs|kernels of Grin) - 
https://github.com/bitcoin/bips/blob/master/bip-0152.mediawiki#Short_transaction_IDs

Short ids will allow us to propagate compact blocks.
A node will then be able to determine if they have seen the necessary inputs|outputs|kernels (via their tx pool?) and go request any missing data from peers as necessary.
This will allow us to minimize the amount of redundant data (no need to include full range proofs in both the tx and the block over the p2p network etc.)

From bitcoin bip-0151 - 
```
Short transaction IDs are used to represent a transaction without sending a full 256-bit hash. They are calculated by:

single-SHA256 hashing the block header with the nonce appended (in little-endian)
Running SipHash-2-4 with the input being the transaction ID and the keys (k0/k1) set to the first two little-endian 64-bit integers from the above hash, respectively.
Dropping the 2 most significant bytes from the SipHash output to make it 6 bytes.
```

Grin Proposal -
```
* We already have a block hash (blake2 based) including the nonce - so use this.

* We already have siphash-2-4 in the codebase for pow - so use this 
  * move it out of pow crate and into hash.rs?

* We need a concept of an identifier for inputs|outputs|kernels
  * is this just the hash of the output etc.?

* Some utility to conveniently extract k0/k1 from the block hash

* A fn to run siphash-2-4 on output id, k0, k1 and return the 6 bytes we care about
  * block.short_id(&output)
  * output.short_id(&block) 
  * Hash::short_id(&output, &block)
  * other?

* short ids seems like a good candidate for some decent test coverage...
```

antiochp:
TODO - 
- [x] make this more reusable (a trait?) so we can use it for inputs/outputs/kernels easily
- [x] add some test coverage (basically hashes in -> ids out)
- [x] serialize/deserialize short_ids as hex strings (also for debugging purposes)
- [x] impl minimal `CompactBlock` (lists of short_ids)
- [x] expose compact block via block api endpoint (optional `?compact` query param)

Resolves #632

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -385,9 +385,13 @@ impl Handler for ChainHandler {
 	}
 }
 
-// Gets block details given either a hash or height.
-// GET /v1/block/<hash>
-// GET /v1/block/<height>
+/// Gets block details given either a hash or height.
+/// GET /v1/blocks/<hash>
+/// GET /v1/blocks/<height>
+///
+/// Optionally return results as "compact blocks" by passing "?compact" query param
+/// GET /v1/blocks/<hash>?compact
+///
 pub struct BlockHandler {
 	pub chain: Arc<chain::Chain>,
 }
@@ -396,6 +400,11 @@ impl BlockHandler {
 	fn get_block(&self, h: &Hash) -> Result<BlockPrintable, Error> {
 		let block = self.chain.clone().get_block(h).map_err(|_| Error::NotFound)?;
 		Ok(BlockPrintable::from_block(&block, self.chain.clone(), false))
+	}
+
+	fn get_compact_block(&self, h: &Hash) -> Result<CompactBlockPrintable, Error> {
+		let block = self.chain.clone().get_block(h).map_err(|_| Error::NotFound)?;
+		Ok(CompactBlockPrintable::from_compact_block(&block.as_compact_block()))
 	}
 
 	// Try to decode the string as a height or a hash.
@@ -426,8 +435,21 @@ impl Handler for BlockHandler {
 		}
 		let el = *path_elems.last().unwrap();
 		let h = try!(self.parse_input(el.to_string()));
-		let b = try!(self.get_block(&h));
-		json_response(&b)
+
+		let mut compact = false;
+		if let Ok(params) = req.get_ref::<UrlEncodedQuery>() {
+			if let Some(_) = params.get("compact") {
+				compact = true;
+			}
+		}
+
+		if compact {
+			let b = try!(self.get_compact_block(&h));
+			json_response(&b)
+		} else {
+			let b = try!(self.get_block(&h));
+			json_response(&b)
+		}
 	}
 }
 
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -367,6 +367,30 @@ impl BlockPrintable {
 	}
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct CompactBlockPrintable {
+	/// The block header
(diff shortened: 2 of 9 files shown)


antiochp:
Regular block - 
```
curl "127.0.0.1:13413/v1/blocks/1" 
{"header":{"hash":"691a65eb002234b362fa1f7c810660c3197fb549945dbae3f9a57a8cd2960725","version":1,"height":1,"previous":"b83053b4b2e29376d6850de4fdaa60366300520bc6957cc69bec7089fcb1bfd8","timestamp":"2018-01-18T14:23:53Z","utxo_root":"045e1ef3b1fcf7d94dea52338b45eaa10edcf36cd6997918f0e9a5e5338f071d","range_proof_root":"acdc0a4e5fbd72f278f8a72830d1ed3e76f43e962fe5b7535bfe1decbf27253f","kernel_root":"e98b0ed4b5a761e75ead769534e3c178211bc3a1ebda7ef6256517a996704774","nonce":376657813926956815,"difficulty":10,"total_difficulty":20},
"inputs":[],
"outputs":[{"output_type":"Coinbase","commit":"083eafae5d61a85ab07b12e1a51b3918d8e6de11fc6cde641d54af53608aa77b9f","switch_commit_hash":"85daaf11011dc11e52af84ebe78e2f2d19cbdc76","spent":true,"proof":null,"proof_hash":"ed6ba96009b86173bade6a9227ed60422916593fa32dd6d78b25b7a4eeef4946"}],
"kernels":[{"features":"COINBASE_KERNEL","fee":0,"lock_height":1,"excess":"09b5ed0bc1ad24cf17943f20cf87480a413691d7b5222951e61ecbd7df7c0b9a86","excess_sig":"0b2c702e80100506f43f97711dccbaee70794eb2d83adc00a96d057d307d36ee38122b62faab09e85fe28d4a96deb054f399f4eb315760d8e57228d6774b3fa6"}]}
```

Compact block - 
```
curl "127.0.0.1:13413/v1/blocks/1?compact"
{"header":{"hash":"691a65eb002234b362fa1f7c810660c3197fb549945dbae3f9a57a8cd2960725","version":1,"height":1,"previous":"b83053b4b2e29376d6850de4fdaa60366300520bc6957cc69bec7089fcb1bfd8","timestamp":"2018-01-18T14:23:53Z","utxo_root":"045e1ef3b1fcf7d94dea52338b45eaa10edcf36cd6997918f0e9a5e5338f071d","range_proof_root":"acdc0a4e5fbd72f278f8a72830d1ed3e76f43e962fe5b7535bfe1decbf27253f","kernel_root":"e98b0ed4b5a761e75ead769534e3c178211bc3a1ebda7ef6256517a996704774","nonce":376657813926956815,"difficulty":10,"total_difficulty":20},
"inputs":[],
"outputs":["86009369a550"],
"kernels":["8707f4f875f0"]}
```

antiochp:
We are not relying or depending on this for anything right now.
Still many tasks needing to be completed before we can actually start broadcasting compact blocks out (see #522).
Going to merge.

antiochp:
Just realized we want to keep coinbase output(s) and kernel(s) "full" for every block.
We will never propagate these in advance of the block (assuming it makes sense never to do this).
So nodes will not have seen these before they see the compact block.

PR #639, opened 2018-01-19
Mining/POW Updates

yeastplume:
Large-ish update to POW/Mining for Testnet2 only, with more changes than are reflected here done within cuckoo-miner. I'll list what's different, then the couple of things that everyone needs to be aware of before this is merged.

-Integration of mean cuda plugin into cuckoo-miner (it's now just called 'cuda_30' rather than mean 30)
-Update to siphash key generation to 256 bytes, to correspond with @tromp's recent changes
-Allow per-device configuration in cuckoo-miner, and changes to grin.toml and config to reflect this.
-More output for device status (whether it's errored or not)
-Fix to casting issue with timing output
-Fix for async-mode InvalidPow (hadn't updated the nonce index since the block header format last changed)
-Added genesis block for testnet2 (will have to be regenerated if the header format changes, obviously)

Now a couple of things that are more important to note/reflect upon

-I've changed the minimum difficulty to 1, and to get around the issue of the difficultly not adjusting due to rounding errors at low numbers, just made the difficulty adjustment return a ceil().. this was the fastest way I could think of to do this without rejigging everything or adding a lot of bloat.. in practice I don't think this will make much a difference (though it does make consensus a bit more finicky, perhaps we can just use this for now and think if we can address this better later on).

-I've started to tune MAX_PROOF_TARGET to be more in line with current GPU findings on cuckoo 30, which has to be tweaked in tandem with lowering the minimum difficulty. Basically, I'm trying to work on the assumption that a single GPU miner at 1 graph per second on a single-user chain should be finding a block at difficulty 1 on average once a minute. The previous value (with 0x0f in the highest byte) was proving far too slow, with most blocks coming back with difficulty 0. I've currently set MAX_PROOF_TARGET to max u64, and from my limited testing it more or less achieves this rate. This may have to be lowered slightly at some point for fine tuning.

-Because of this change, user testing at cuckoo16 causes far too many valid blocks, so I've lowered this for Testnet and Usertesting modes (again to try and keep the rate of valid blocks on a single machine as before).


ignopeverell:
That makes sense to me. At Cuckoo30 we could even even have the max target at 0xff in the higest bit. I think I'm ok with the flooring for now, hopefully we won't run into differences in floating point architectures, but I we'll have to revisit before testnet2.

Issue #590, opened 2018-01-08
hashed switch commitments (thread resurrection)


antiochp:
Concerns raised that current "switch commit" implementation does not fully give us the desired properties.

We likely need to discuss and agree on an approach here that allows us to - 
  * have switch commits that do what we desire them to do
  * still allow us to lock coinbase outputs (or find an alternative way to do this)
  * still allow us "easy" wallet restore from seed functionality

https://lists.launchpad.net/mimblewimble/msg00373.html
(msg from Tim Ruffing)

I need to resurrect this thread.

What's currently implemented in Grin [1] is 
  (vG + bH), BLAKE2(bJ), 
where the hash is truncated to 160 bits.

Hashed switch commitments are supposed to give you two main properties:
   1. Switch binding: even if you have unlimited computation power from
      tomorrow on, commitments created today are still binding if fully
      verified. This is the original property from the switch commitments
      paper [2] and it means that you can still spend your money after the
      switch to full verification (but you may lose privacy).
   2. Privacy for "old" outputs, which have been spent long ago, is
      retained even against quantum computers.
   3. Opt-out money, opt-in privacy: If your privacy is more important
      than your money, then you can decide not to reveal the pre-image of
      the hash. Then you cannot spend the money but you retain privacy
      even against post-quantum attackers.                  (This was my
      understanding, let's call it "3a". Andrew and maybe the others
      involved in the conversation had a different variant "3b" in mind:
      If you prefer privacy, then you commit to ((vG + bH), r) in the
      first place, where r is just random. This requires users to decide
      upfront if they want money or privacy but "They will have plenty of
      warning before any such fork, so  they can either move their coins
      away or move them to upgradable outputs, taking the privacy hit
      voluntarily and knowingly and only for their most recent history."

It turns out that the implemented variant just fulfills 1 and 3b [3].
Property 2 is trivially broken and it kind of scares me that I had
overlooked this:
Given the commitment ((vG + bH), BLAKE2(bJ)), guess v, compute b (by
solving dlog), and check if the second element is indeed BLAKE2(bJ)).

If we want all properties to hold, then the following construction
works:
  (vG + bH), BLAKE2(bJ, r),
where r is 256 bits and the hash is 256 bits as well.

Then 1 holds only computationally against quantum computers [4]. Even
in a quantum world, you need 2^128 operations to find a second preimage
in the hash. (Note that you have a quantum computer only after the
switch, you so cannot find a collision upfront.)

[...]

PR #648, opened 2018-01-22
hashed switch commitments - BLAKE2(bJ, r)

antiochp:
See https://github.com/mimblewimble/grin/issues/590 for context.

This PR uses the key derivation impl in the wallet extended key to generate switch commit hash keys.

The commitment and switch commit hash are of the form - 
```
(vG + bH), BLAKE2(bJ, r)
```

The switch commit hash key `r` is now derived from the extended key in the wallet.
This key uses a separate seed key and seperate chaincode to ensure there is nothing relating the secret key and the switch commit hash key for a given derived key.

Additionally - 
* switch commit hash is 32 bytes (was 20 bytes)
* the key `r` the key is also 32 bytes

Knowing `b` does not give you any information about `r` so you cannot trivially identify outputs.

Wallet recovery still works as we can deterministically derive both `r` and `b` for a given key derivation to identify our own outputs.

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -760,7 +760,11 @@ impl Block {
 	) -> Result<(Output, TxKernel), keychain::Error> {
 		let commit = keychain.commit(reward(fees), key_id)?;
 		let switch_commit = keychain.switch_commit(key_id)?;
-		let switch_commit_hash = SwitchCommitHash::from_switch_commit(switch_commit);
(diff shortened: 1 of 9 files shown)


antiochp:
Merging.
Will open an issue to track the discussed extkey cleanup.
We don't derive `n` from `n-1` and `chaincode` unused for derived keys etc.

Issue #434, opened 2017-12-06
Unit Tests for API endpoints


yeastplume:
The recent update to API endpoints here https://github.com/mimblewimble/grin/pull/416 appear to have broken wallet restore; that's completely forgivable. What isn't forgivable is that something like this should have been easily caught in automated testing, which I don't think we do at all for the API (unless I'm mistaken) or if we do anywhere, they're very out of date.

Unit tests should be put into place that cover the API endpoints. Should be simple enough, build up a small chain and a wallet in a cargo test, then put a few tests together to make sure all the endpoints return more or less what's expected.

ignopeverell:
Mmmh yeah, good catch. I've been meaning to take over the code coverage task too sometime, would make it easier to spot the glaring omissions.

heunglee:
I can help, but need time to figure out how to. Any instruction is appreciated if my help is welcomed.

edbo:
Hey guys, I actually have been looking at ways to get involved in some way in the project and was thinking potentially I could learn more rust and more about grin by helping with the testing.

I reckon we all probably agree that making sure the test frameworks are stable and repeatable will allow the project to pump ahead with fewer bumps. To that end I reckon it's important for any project to agree on what we mean by unit tests and what we want from the tests in general.

Adhering to the terminology here and quick view of the current codebase:

https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html

There are patches of unit tests in some places but 0 unit test in the API.

There are also no integration tests for the API.

I'd also argue there is room to add another layer of acceptance tests by which I mean an automated framework to spin up a few nodes (likely in docker containers for repeatability) and simulate some common use cases including querying the APIs for state.

Sorry for the ramble but I guess my view is first it might be worth documenting what the goals for testing are. For example unit test coverage > 90%, integration tests to cover all main use cases potential error cases and finally a small number of acceptance tests on top of the lot.

I'm happy to take a first stab at this this weekend at documenting this and even do some work to start measuring things like coverage if the approach is agreed.

sesam:
Thanks for adding the direct link on testing. Ask on Gitter if you guys run into any problems, setup or "devops" things along the way.

Automation that spins up several nodes is out there, but maybe not in this repo and/or not documented. You can open separate issues about that. More testing-relate issues are open, but this one is I think both quite easy and urgent, so a good place to start.

ignopeverell:
[...]

PR #653, opened 2018-01-24
Add unit tests for API Endpoint

quentinlesceller:
Fixes https://github.com/mimblewimble/grin/issues/434
Adds unit tests for the API Endpoint.
For now, mine a simple chain with a wallet and test each handler.

- [x] Test blocks handler
- [x] Test status handler
- [x]  Test peers handlers (including ban/unban) 
- [x] Test sumtree handler
- [x] Test chain utxo handler


quentinlesceller:
Update:
Trying to create a pool of 2 servers similarly to "simulate_seeding" in ```grin/tests/simultnet.rs``` in order to test the P2P handler.
 However, I noticed that in the simulate seeding test, after getting connected to each others the servers disconnected themselves with a connection error ```Client 127.0.0.1:31020 connection lost: Connection(Error { repr: Custom(Custom { kind: UnexpectedEof, error: StringError("early eof") }) })```.
After discussion with @yeastplume on Gitter looks like there is probably an issue here. 
I will try to fix it before finishing this PR.

yeastplume:
Hi @quentinlesceller,

Just before you go too much further here, we couldn't merge this as is because of the duplicated code copied out of the grin/grin testing directory. 

Since these are integration tests that require a running server to test, they could probably just as easily go into grin/grin (where the rest of the integration tests are) for the time being rather then putting them in grin/api and having to refactor the testing framework out elsewhere. If you look in grin/tests for example I've put a wallet.rs file to test the wallet against a running server.. you can probably create an api.rs file alongside it and place tests there

quentinlesceller:
Okay no problem will do.

antiochp:
Looks good!

PR #657, opened 2018-01-27
fix bug - p2p store was ignoring port numbers

antiochp:
The p2p store was only taking the ip into consideration when generating keys for the db.
It was ignoring port numbers.

So the following were actually overwriting each other in the store ("all peers") - 
* `127.0.0.1:13413`
* `127.0.0.1:13414`

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -314,6 +314,7 @@ impl Readable for GetPeerAddrs {
 
 /// Peer addresses we know of that are fresh enough, in response to
 /// GetPeerAddrs.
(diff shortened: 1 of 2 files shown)


ignopeverell:
I *think* this was originally voluntary, because we can't know the receiving port of someone connecting to us. But I think we also fixed that to consider the port sent to us in the handshake, so that's likely fine now.

antiochp:
Ahh right I remember something about that.
I think this is safe but let me take another look at this tomorrow or Monday before I merge it in.

PR #661, opened 2018-01-28
Fix 658 and compiler complaints

### LINKED ISSUE #658: store sumtree tests failing randomly
Just saw the following pass 15 times in a row -
`cargo test -p grin_store sumtree`

Then it failed randomly on the next run - 

```
running 4 tests
test sumtree_append ... ok
test sumtree_prune_compact ... FAILED
test sumtree_reload ... FAILED
test sumtree_rewind ... ok

failures:

---- sumtree_prune_compact stdout ----
	thread 'sumtree_prune_compact' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 2, message: "No such file or directory" } }', src/libcore/result.rs:906:4

---- sumtree_reload stdout ----
	thread 'sumtree_reload' panicked at 'assertion failed: `(left == right)`
  left: `32`,
 right: `16`', store/tests/sumtree.rs:132:4


failures:
    sumtree_prune_compact
    sumtree_reload

test result: FAILED. 2 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

Thanks @mfronczyk for originally reporting this in gitter.

sesam:
Fix tests concurrency issue for `cargo test -p grin_store sumtree` by ensuring these test have method-tagged folder names so they don't overwrite each others' files
Fixes #658

Cleanup compiler errors

### DIFF
--- a/pool/src/lib.rs
+++ b/pool/src/lib.rs
@@ -32,7 +32,8 @@ extern crate grin_keychain as keychain;
 extern crate grin_util as util;
 extern crate rand;
 extern crate serde;
-#[macro_use]
+#[allow(unused_imports)]
+#[macro_use] // Needed for Serialize/Deserialize. The compiler complaining here is a bug.
 extern crate serde_derive;
 #[macro_use]
 extern crate slog;
--- a/grin/tests/framework/mod.rs
+++ b/grin/tests/framework/mod.rs
@@ -174,7 +174,7 @@ impl LocalServerContainer {
 		wallet_config.check_node_api_http_addr = config.wallet_validating_node_url.clone();
 		wallet_config.data_file_dir = working_dir.clone();
 		Ok(
-			(LocalServerContainer {
+			LocalServerContainer {
 				config: config,
 				p2p_server_stats: None,
 				api_server: None,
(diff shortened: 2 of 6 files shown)


pool/src/lib.rs
```diff
@@ -32,7 +32,8 @@ extern crate grin_keychain as keychain;
 extern crate grin_util as util;
 extern crate rand;
 extern crate serde;
-#[macro_use]
+#[allow(unused_imports)]
+#[macro_use] // Needed for Serialize/Deserialize. The compiler complaining here is a bug.
```
antiochp: Can we do the `#[cfg(test)]` trick here?

sesam: No, Serialize/Deserialize is used in a --release build, and I couldn't come up with any combination of allow(unused_imports) and/or cfg[test] to silence the compiler complaints here -- that's why the commit message is "Tried but failed" and then hoping that not all of us (maybe repeatedly) will loose time fighting that particular warning for this particular lib. I think I tried fixing this once before, but found no commits to jog my memory...

ignopeverell:
Ok then. Looks good to me, thanks!

PR #662, opened 2018-01-28
added INFO when reconnecting to wallet

### LINKED ISSUE #493: Add INFO logging for Connection Lost and Connection Re-established to wallet server

dbancajas:
Fixes #493 
Closes #538 

Signed-off-by: dean <dbancajas@gmail.com>

### DIFF
--- a/wallet/src/client.rs
+++ b/wallet/src/client.rs
@@ -30,7 +30,7 @@ use util::LOGGER;
 /// Call the wallet API to create a coinbase output for the given block_fees.
 /// Will retry based on default "retry forever with backoff" behavior.
 pub fn create_coinbase(url: &str, block_fees: &BlockFees) -> Result<CbData, Error> {
-    let mut has_error = false;
+	let mut has_error = false;
 
 	retry_backoff_forever(|| {
 		let res = single_create_coinbase(&url, &block_fees);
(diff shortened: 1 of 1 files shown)


dbancajas:
closes #538

ignopeverell:
The indentation still looks a little off. Also the Rust style for if statements is `if condition {` on one line with no parentheses. Thanks!

dbancajas:
@ignopeverell  I used the rustfmt to properly format the code and pushed my changes.

wallet/src/client.rs
```diff
@@ -38,6 +40,10 @@ pub fn create_coinbase(url: &str, block_fees: &BlockFees) -> Result<CbData, Erro
 				"Failed to get coinbase from {}. Run grin wallet listen",
 				url
 			);
+			has_error = true;
+		}
+		if (has_error == true) {
```
yeastplume: Sorry, just a further nit-pick, parentheses aren't necessary in rust if conditions, so just
````
if has_error
````
should suffice here. I only say so because the way its written will add an unnecessary warning to the build output.

dbancajas: done. I also removed the {} since it is a single line if condition.

ignopeverell:
Thanks!

PR #663, opened 2018-01-29
rustfmt install instruction gotcha

sesam:
rustfmt stopped working for me. It turns out people started having trouble with this last year already https://github.com/rust-lang-nursery/rustfmt/issues/2304

### DIFF
--- a/doc/style.md
+++ b/doc/style.md
@@ -35,8 +35,18 @@ ln -s -f ../../.hooks/pre-commit
 
 To run rustfmt against a single file, this __new__ command works with latest rust and after having done `rustup component add rustfmt-preview` and by setting --write-mode it doesn't overwrite files.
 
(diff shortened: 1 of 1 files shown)


antiochp:
I suspect `--write-mode diff` avoids overwriting additional files by not overwriting _any_ files (including the file you actually want formatting)?

sesam:
Yes. I could've made the text more clear. Once someone :) gets around to testing this properly, I suspect we'll have to overhaul the whole stye.md and maybe also the hooks.

As a later step I think we should have a preinstalled precommit hook that does 1. Complain and nag people to read about code formatting [url..] , 2. do manual style check (diff, overwrite etc), 3. preferably activate the blocker commit-hook we have under .hooks

antiochp:
Looks good.

PR #664, opened 2018-01-30
Rewrite peer-to-peer logic to replace tokio with simple threading

ignopeverell:
Tokio and futures have proven very difficult to work with in our specific case (peer-to-peer interactions), making the code far more unwieldy and complex than it should. The result is hard to read, hard for any new contributor to get into and hard to change when new logic needs to be added. All of this for very little benefit as a peer in a network should never have to handle extremely high traffic levels (like a chat or HTTP server would, for example).

This PR rewrites most of the `p2p` crate as well as the seeding logic with simple threading. Each peer have a thread and the TCP server has its own as well. Resulting code is a lot easier to understand.

Only part left is fixing the tests in the `grin` crate.


sesam:
I like it. Removes 1000loc and some dependencies. About this arch; should
DNS seeds now be built with their own separate thread or is asynch+callback
enough? "Asking for a friend" with a stale PR.. :)

ignopeverell:
The seeding has its own thread so it can be blocking within that thread.

ignopeverell:
@antiochp @yeastplume might be a little big for a thorough review but it'd be nice to have your take on this, now that there's an actual PR.

yeastplume:
I won't pretend to have absorbed all of the changes but this already looks much, much easier to follow than the previous iteration. Very glad you've gone and done this and I think this is going to save a lot of headache in the future.

The rewrite will make testnet2 more **interesting** and **fun** but it will be very much worth it in the long run.

antiochp:
(Have not looked at the code yet so apologies in advance, but wanted to ask this before I forgot) - 

If multiple threads are now potentially operating on both the tx pool and the block processing pipeline - are we confident that we are handling this correctly such that these operations are still atomic?

Previously we have a single thread handling all these peer connections and these operations were one at a time.

(Now going to go look at the code).

p2p/src/rate_limit.rs
```diff
@@ -1,224 +0,0 @@
-// Copyright 2016 The Grin Developers
```
antiochp: Are we just getting rid of the rate limit stuff for now?
I don't think its a bad idea to simplify this for now (was never sure what would happen if we actually hit the rate limit mid message anyway).

p2p/src/protocol.rs
```diff
@@ -15,256 +15,115 @@
...
-					&mut data,
-					&MsgHeader::new(Type::Block, body_data.len() as u64),
-				));
-				data.append(&mut body_data);
-				if let Err(e) = sender.unbounded_send(data) {
-					debug!(LOGGER, "handle_payload: GetBlock, error sending: {:?}", e);
-				}
+				debug!(LOGGER, "handle_payload: Block {}", bh);
+
+				adapter.block_received(b, self.addr);
```
antiochp: This is where I'm wondering about multiple threads hitting the block processing pipeline in parallel. Two peer threads could receive the same block (or different blocks) simultaneously.

antiochp: Maybe worth considering having a cpupool/threadpool with a single thread that we can offload work to (and keep the peer threads for simply handling the network connections?)
If we haven't solved this another way?

ignopeverell: Correct, and it's solved by the sumtrees lock. I actually think it's a good thing because it allows us to start validating block headers in parallel.

p2p/src/rate_limit.rs
ignopeverell: Yes, I decided to leave it out for now. It seemed a bit premature and a limit in message count for each message type may serve us better if it comes to that.

ignopeverell:
I've tried to keep most of the code that was actually doing stuff alone so hopefully we won't have too many surprises. But it's multi-threaded code so...

p2p/src/protocol.rs
antiochp: :+1:

PR #667, opened 2018-01-30
Request compact blocks

antiochp:
Default to requesting and sending compact blocks.
On receiving a compact block we fail to hydrate (not yet implemented) so go request a full block to replace it.

TODO - 
- [x] miner should send a full block if the block is empty (no benefit to empty compact block)
- [x] if a peer requests a compact block and block is empty, send the full (empty) block
- [x] randomly decide to send block vs compact block for empty blocks (to help testing the various states the system can be in)

~actually look in the tx pool for inputs/outputs/kernels~ (tackle in separate PR)

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -531,10 +531,6 @@ pub struct CompactBlockPrintable {
 	pub out_full: Vec<OutputPrintable>,
 	/// Full kernels, specifically coinbase kernel(s)
 	pub kern_full: Vec<TxKernelPrintable>,
-	/// Inputs (hex short_ids)
-	pub in_ids: Vec<String>,
-	/// Outputs (hex short_ids)
-	pub out_ids: Vec<String>,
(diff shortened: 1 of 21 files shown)


antiochp:
@ignopeverell this wires up the compact block serialization/deserialization with the p2p layer.
After a node handles a block_header it requests a compact_block.
Hydration currently fails every time (not yet implemented) - and the node then requests the full block.

I'd like to get this in now (unless there are any objections) and then start incrementally adding functionality to leverage the tx pool to retrieve missing inputs/outputs/kernels.

This does add complexity to the various code paths during block propagation so added in some randomness so we sometimes send full blocks and sometime compact blocks if the block in question is "empty" (majority of the time during testing).

ignopeverell:
I'm already dreading the merge with my p2p rewrite branch but looks good! Feel free to merge.

PR #674, opened 2018-01-31
block.merge() is unused (and potentially misleading)

antiochp:
This `merge()` function is only used in a single test.

I'm proposing getting rid of it.

I think it is potentially misleading because you can't really merge two blocks like this - 
* what happens to the 2nd block header?
* fees are basically undefined - how do we know these are fees for multiple blocks and not just a big invalid block with larger than allowed fees?

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -554,33 +554,6 @@ impl Block {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ignopeverell:
Sounds good, that was just an early-on test.

PR #681, opened 2018-02-02
Transaction contain kernels. Transactions and blocks maintain a kernel offset (split key).

### LINKED ISSUE #671: kernel combination step that voldie missed
https://gitter.im/grin_community/dev?at=5a71e52098927d57455df4a9

tl;dr If a valid tx consists of - 
* inputs
* output
* kernel(s)

And they sum to zero to be a valid tx.
Then it is possible to simply try all permutations of inputs and outputs for a given kernel to find a set that sum to zero (particularly given that most txs only have a few inputs and outputs).

From @tromp on gitter - 

> i just realized that making each block look like one huge tx doesn't help privacy much
the kernels still reveal the corresponding inputs and outputs

> you can still find inputs and outputs since they have to add up to 0 with the kernel?!

> i guess i'm having issues with this paragraph of the MW whitepaper:
Now, creating transactions in this manner supports OWAS already. To show this,
suppose we have two transactions that have a surplus k1G and k2G, and the
attached signatures with these. Then you can combine the lists of inputs and
outputs of the two transactions, with both k1G and k2G to the mix, and
voilá! is again a valid transaction. From the combination, it is impossible to
say which outputs or inputs are from which original transaction.

> it seems there's only one way to split inputs and outputs so that these form valid txs with their kernel

> most tx have very few ins and outs, so a polytime algorithm could recover them
e.g. suppose all tx have 2 in & 2 out. then just try all (n choose 4) subsets

Resolved by #681.

antiochp:
This turned into a pretty big PR...
See #671 for some context.
To make a "split key" possible it was necessary to rework transactions to contain tx kernels (technically only a single kernel was necessary, but a vec gives us future flexibility).

__tl;dr__ Blocks _really_ do look like big transactions now (list of kernels + single summed offset).

### Transaction and TxKernel changes
* A tx now contains a vec of kernels 
  * build the kernel during tx creation (not when we build the block)
  * currently only support building txs with a single kernel
  * but there is a test that verifies a multi-kernel tx is still valid 
  * tx validation involves verifying all pre-existing kernels (rather than building a kernel)
* Remove fee from tx 
  * total tx fee calculated from kernel fees
* Remove sig from tx
  * each kernel maintains separate excess and an excess_sig
* Tx now maintains a kernel "offset" (we split the signing key)
  * this is sum of all offsets from kernels 
  * each kernel sig verifies independently
  * but we need all kernels and the offset from the tx to validate the full tx
  * pass the offset in the partial_tx during sender/receiver interaction

### Block changes
* A block header now maintains an offset (similar to a transaction)
  * this is sum of all tx offsets
  * so to validate a block we use inputs, outputs, kernels and offset
  * not possible to identify or reconstruct an individual tx from a kernel 

### Wallet sender/receiver changes
* Sender "splits" their secret key before creating the aggsig context
  * resulting offset is passed to receiver in the partial_tx
  * sender uses their split key to create their sig
  * receiver uses the offset to generate the final excess to match the final sig
  * tx kernel is built by receiver during construction of final tx 

TODO -
- [x] build a valid tx with a split key (and store offset on tx)
- [x] check we can verify the sig with an offset
- [x] aggregate the offsets when building a block (and store total_offset on block header)
- [x] check we can still validate a block/block_header if we have a total_offset
  - [x] can we still validate individual kernels at this point?)
  - [x] core tests all passing with "offset" txs
- [x] think through short_ids on kernels (think we are good here)
- [x] store tx_kernel(s) on a transaction
  - [x] build the kernel as we build the tx (not a block creation time)
  - [x] rework validation to take this into consideration
- [x] get it working via aggsig in tests
- [x] get it working via aggsig via sender/receiver
- [x] cleanup sender/receiver code

This should be ready for review.
Merge conflicts - working on these now.
May still have tests failing - working on these also.

But any feedback/comments appreciated!
@yeastplume @ignopeverell.

Resolves #671.


core/src/core/build.rs
```diff
@@ -211,6 +239,26 @@ mod test {
...
 	}
 
+	#[test]
+	fn blind_simple_tx_with_offset() {
+		let keychain = Keychain::from_random_seed().unwrap();
+		let key_id1 = keychain.derive_key_id(1).unwrap();
+		let key_id2 = keychain.derive_key_id(2).unwrap();
+		let key_id3 = keychain.derive_key_id(3).unwrap();
+
+		let (tx, k1, k2) = transaction_with_offset(
```
antiochp: This exercises the new "split the key, then sign the fee" functionality.

ignopeverell:
I think this should be rebased and merge. Looks very fine to me!

PR #694, opened 2018-02-08
Fix #274 Sender now pays the fee

### LINKED ISSUE #274: tx fee should be paid by sender not receiver
Currently when you send `100.0` grin (for example),
the recipient only receives `99.992` grin.
We should rework this so sender actually sends `100.008` to cover the fee (receiver receives full `100.0`).

Might be nice to output amounts when running `wallet send` - 
`100.0 + 0.008 (fees)` or something similar.

quentinlesceller:
Fix #274 and #396 as simply as possible.
Just check if the amount in the outputs is enough to pay with the fee otherwise try to get more outputs. Fails if there is not enough grins in the wallet.

### DIFF
--- a/wallet/src/sender.rs
+++ b/wallet/src/sender.rs
@@ -45,7 +45,7 @@ pub fn issue_send_tx(
 	// proof of concept - set lock_height on the tx
 	let lock_height = chain_tip.height;
 
-	let (tx, blind_sum, coins, change_key) = build_send_tx(
+	let (tx, blind_sum, coins, change_key, amount_with_fee) = build_send_tx(
 		config,
 		keychain,
 		amount,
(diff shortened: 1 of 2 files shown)


wallet/src/sender.rs
```diff
@@ -178,16 +178,59 @@ fn build_send_tx(
 		)
 	})?;
 
+	// Get the maximum number of outputs in the wallet
+	let max_outputs =  WalletData::read_wallet(&config.data_file_dir, |wallet_data| {
+		wallet_data.select_coins(
```
ignopeverell: Not related to this PR (which is good), but it seems `select_coins` could bear some improvement: 6 arguments for a method should probably get refactored in a struct with some sensible default values.

quentinlesceller: Agree, I will look for a way to make in more compact.

Issue #536, opened 2017-12-21
Server stop hook to avoid inconsistent state


ignopeverell:
Add a hook on SIGINT to shutdown cleanly, especially when we're in the process of saving a block.

yeastplume:
Let me introduce a complication.. I was actually thinking the exact same thing a while ago when doing some work on the mining thread. We don't see this day-to-day just yet, but when mining using the much slower cuckoo 30 on multiple GPUs, there can be a very noticeable delay between hitting ctrl-c and the underlying threads controlling CUDA operations etc finishing up, meaning there's the potential for a very long and very user-unfriendly delay there when trying to stop the server.

Whatever the solution is here, it needs to be some kind of service that different threads in different parts of the code can register to monitor, do their cleanup activities (and output whatever appropriate status messages if there's a delay) then only exit when all threads registered to listen for the handler have reported back.

sesam:
Also in #502
There is https://doc.rust-lang.org/1.12.0/libc/constant.SIGHUP.html (libc wrapper) and apart from that rust is said to not support unix signals; https://users.rust-lang.org/t/unix-signals-in-rust/733

tl;dr use libc for now

antiochp:
We probably want to differentiate between threads that can potentially write "corrupted" data and threads that can be safely (if not cleanly) shut down.
And yeah - rust has some big limitations in how it handles unix signals.

heunglee:
Since I always encounter this type of issue with API servers, I read this issue here for better understanding of limitation on rust. I experienced some nodejs stuff for graceful shutdown. When googling, I did not have good luck to find good one for rust.  The rust book has a piece of graceful shutdown  which you probably already read and barely remember (https://doc.rust-lang.org/book/second-edition/ch20-06-graceful-shutdown-and-cleanup.html). LIke @sesam mentioned, if ctrl-c is not captured well, then the way of graceful shutdown won't be very useful. Yet it is worth of trying to feel out the roadblocks for graceful shutdown on multi threads?

heunglee:
I have been some research on graceful shutdown for grin. 

For graceful shutdown, it may be better to experiment with the wallet listener first.
We can experiment with https://docs.rs/graceful/0.1.1/graceful/

Description of the crate "graceful" for those who have not had chance to get to know it before.
Gracefully shutdown.
[...]

PR #700, opened 2018-02-12
Clean server shutdown, generalizes usage of `Weak`

ignopeverell:
Introduces 2 main changes:

* A shared `AtomicBool` that signals server shutdown. All server
threads regularly check it to break out of their main loops when
it changes to `true`.
* Breaking of circular `Arc` references, which can never be
destroyed, by downgrading to `Weak` instead. Only the main server
keeps the `Arc` while all other components get the `Weak` variant.

Both of these are required for all long-living structs to be
cleanly destroyed. Note that in Rust this is fairly important as
most resource-freeing logic is associated with `drop`,
which is only called when said struct is free of scope or `Arc`
references.

Should address most of #536 (only need the stop hook to call
`Server` shutdown).

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use std::io::Read;
-use std::sync::{Arc, RwLock};
+use std::sync::{Arc, Weak, RwLock};
 use std::thread;
 
 use iron::prelude::*;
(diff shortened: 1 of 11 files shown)


antiochp:
Yeah this makes a lot of sense to me. Just something we need to be aware of. :+1:

Issue #695, opened 2018-02-09
Get the sums out of the sumtrees


ignopeverell:
After having implemented most of the regular and fast sync validation, it seems that the sum in the sum trees was a premature optimization. The following seems sufficient:

* 3 MMR trees
* 3 AppendOnlyFile to contain the actual data (output, rangeproof, kernel)
 * One rm log and skip list for outputs and range proofs

Clean storage up. Renaming everything is likely to be the messiest...

ignopeverell:
Fixed by #702

Issue #404, opened 2017-11-29
MW scaling: full pruning


ignopeverell:
Current node model is one of a full archival node. To realize the scaling benefits of a MimbleWimble type chain we need 2 additional features:

* Pruning of chain data when it's not relevant to the UTXO set or the recent chain state.
* Sync only based on the header chain, the MMR sumtrees (tracked in #403).

This issue is for the former.

Currently until MMR sumtree data is pruned to account for outputs spending, however all block data remains. As outputs are spent, we can also get rid of all related data, like commitments and most importantly range proofs.

antiochp:
The `.grin/chain` is actually growing pretty quickly (did not realize it was growing _this_ fast) - 

```
du -h .
4.0K	./plugins/plugins
528K	./plugins
224K	./.grin/peers
538M	./.grin/chain
2.2M	./.grin/sumtrees/kernel
31M	./.grin/sumtrees/rangeproof
33M	./.grin/sumtrees/utxo
66M	./.grin/sumtrees
604M	./.grin
624M	.
```

ignopeverell:
Yup. Liking them large rangeproofs? :-)

PR #702, opened 2018-02-12
Take the 'Sum' out of 'Sumtree'

yeastplume:
I think this is nearing a mergeable state. Here's the general approach I've taken:

* The corresponding AppendOnlyFiles for the data represented by each PMMR belong to the Backend Implementation.
* the append and get implementations of the Backend trait are extended to work on a (Hash, Option<T>) tuple. If the data for a leaf node exists, it will be placed into the corresponding flatfile.
* The PMMR backend now contains 2 AppendOnlyFiles, one to store hashes as a pmmr and one to store elements of type T
* There's no need for an additional remove log or prune log, data within them can be used to find the correct positions in the (pmmr-based) hashfile and the flat data file
* I've added a Trait 'PMMRable' with all of the traits needed to be included in a PMMR implementation.. size of T needs to be returned as well as the ability to sum hashes of the type.
* Had to add an 'OutputStorableType' to actually do the storage.. it's too messy to try to serialise and deserialise while specifying different fields so 'view' on an output seemed the least worst option.

At present this seems to work, all tests pass, and even though pruning isn't really enabled anywhere at present I've added to the pmmr tests to ensure corresponding data files are read correctly after pruning. 
I haven't undertaken any global renaming or refactoring of chain/sumtree.rs, as there are so many changes here incompatible with other branches that it's probably better to get this all in sooner rather than later.

*****
Pre-edit comment:

Took on https://github.com/mimblewimble/grin/issues/695 to get a tour of the current state of the core, @ignopeverell can you just have a quick look over core/core/pmmr.rs as is and make sure this amount of chop isn't too much? (i.e. I've removed a level of abstraction you might want to keep) This version is just working on a plain and simple core::Hash with the backend just working on a Writeable.


ignopeverell:
No that's a good start. You'll still have to store the commitment and switch data into a separate `Backend` that should likely be held by the current `chain/src/sumtree.rs` implementation. But that'll be a flat file, not a MMR.

I think once all of that is done, `chain/src/sumtree.rs` should also get refactored a bit and split.

yeastplume:
Another few updates there, with the sums removed from store and core and all tests in there compiling and running. A couple of further thoughts now I understand the requirements a bit better and fallen into the rabbit hole:

-Just so I understand, we need a new Backend implementation (AppendOnlyFileBackend or something)
that takes a T:Writeable that provides functions to append Ts in chunks and read them back from a given index while keeping track of its remove log and prune list... similar to the PMMRBackend that's in there now, except operating on an AppendOnlyFile directly instead of the PMMR.

-Current range proofs will be challenging to get into that format as their lengths vary.. this will be much easier with fixed-length bullet proofs in place. I can just use MAX_PROOF_SIZE for now as the length (plus the actual length,) which is sub-optimal, but once bullet proofs are done this should become a lot easier.

yeastplume:
@ignopeverell Okay, not complete, but I think this is ready for a bit of review and a couple of pointers if you wouldn't mind indulging (sorry). This has been a good education so far but I think I'm still missing something conceptually.

Travis issues aside, the unit tests all pass which means appending and rewinding seem to be working properly. However, the flat files themselves aren't being pruned in any way (even though the flatfile implementation supports it), and I think I need to educate myself a bit more on what happens when a PMMR gets pruned. I'm kind of working on the assumption that when a leaf goes into the tree at h=0, it's 'leaf index', i.e. just assuming a flat index from 1..n counting the nodes at the bottom of the tree (outputs), never changes.. but I think I'm missing something here.

So from what I gather now, what needs to happen is:
* Pruning spent outputs from the flat-files and ensuring their index in the file always matches the entries along the bottom row of the PMMR (which I understand how to do in the append and rewind case, but what happens if/when the leaves are re-orged)
* A bit rename-athon to remove the word 'sumtree' from all over the place (and in the comments) and perhaps replace with 'hashtree' where it's publicly visible and 'pmmrtree' when referring to the implementation

core/src/core/pmmr.rs
```diff
@@ -661,6 +512,30 @@ pub fn peaks(num: u64) -> Vec<u64> {
...
+			cur_leaf = bintree_rightmost(cur_leaf);
+		}
+		count = count + 1;
+		cur_leaf = bintree_jump_left_sibling(cur_leaf);
+		if cur_leaf == 1 {
+			break;
+		}
+	}
+	count
+}
```
ignopeverell: Isn't this O(N) where N is the number of leafs in the tree? Each operation is fast but on a big tree it won't be too good...

yeastplume: Yes, sorry I'd meant to mention this in my description.. it is temporary O(n) crap. However, the peaks() function we'd talked about earlier was returning strange values (or I'd quite possibly just misunderstood) and the method we'd talked about earlier wasn't doing what I was expecting, so I just implemented this quickly and moved on to other parts. This obviously needs to be changed.

ignopeverell:
To illustrate PMMR pruning let's assume we have the following tree (numbers are orders of insertion):

```
      7
    /   \
  3     6
 /  \   /  \
1  2  4  5
```

The `AppendOnlyFile` will store data in the insertion order as a flat stucture [1,2,3, ...]. Now say nodes 1 and 2 get removed. The remove list will just have [1, 2] with the height information associated with it for rewind. But all of that is not very compact (for such a short tree it's not obvious but you can imagine for a large tree where 90% of the stuff has been removed).

So compaction goes and rewrites the `AppendOnlyFile` to be just [3, 4, 5, ...] (we still need 3 to be able to recalculate hashes, however if 4 and 5 got deleted as well, we could get rid of 3 and 6). Then the remove log also kind of sucks because it scales with the number of deleted nodes. So instead, there's a `PruneList` that's more clever. Instead of storing [1, 2] it will just store [3] and is fairly efficient at saying that 2 positions need to be skipped if we wanted to access 4 in the `AppendOnlyFile`. So when `AppendOnlyFile` gets rewritten, we also take the `RemoveLog`, use it to populate the `PruneList` and then nuke it.

Is that helpful?

Caveat Lector: compaction in testnet1 is implemented and has a few tests but is unused. My plan was to plug it in and do the database cleanup with #404.

yeastplume:
Thanks for the explanation, I think I have a much clearer idea of how to go about this now and fully understand what's going on. I've modified what I had to implement the flatfiles as part of the PMMR backend, as it seems to be a better way to go about this.

yeastplume:
Okay, starting to get there I think. I'll edit the top comment with a description of what's been done.

yeastplume:
O(n) function is gone, all tests pass and ready for a review, I think. I still don't like the fact I needed yet another Output wrapper structure, but every time I tried to simplify by just serialising or hashing a base Output differently, it just leads to an even bigger mess.

chain/src/sumtree.rs
```diff
@@ -515,17 +501,18 @@ impl<'a> Extension<'a> {
...
 	/// and kernel sum trees.
 	pub fn roots(
 		&self,
 	) -> (
-		HashSum<SumCommit>,
-		HashSum<NoSum<RangeProof>>,
-		HashSum<NoSum<TxKernel>>,
+		Hash,
+		Hash,
+		Hash,
```
ignopeverell: Should we introduce a `Root` struct to group those 3 hashes? Before the type system would prevent us from getting mixed up as to which is which but now it will be more confusing. A struct with 3 properties could help?

chain/src/sumtree.rs
```diff
@@ -70,20 +70,17 @@ where
...
-///
-/// In addition of the sumtrees, this maintains the full list of kernel
-/// data so it can be easily packaged for sync or validation.
+
 pub struct SumTrees {
-	output_pmmr_h: PMMRHandle<SumCommit>,
-	rproof_pmmr_h: PMMRHandle<NoSum<RangeProof>>,
-	kernel_pmmr_h: PMMRHandle<NoSum<TxKernel>>,
-	kernel_file: AppendOnlyFile,
+	utxo_pmmr_h: PMMRHandle<OutputStoreable>,
```
ignopeverell: I'm wondering whether we should call all of these `txo_pmmr*` instead of `utxo_pmmr*`. The reason is that all transaction outputs get stored in it, not only the unspent. They do get pruned when they're spent, which can get confusing but the root is the hash of all TXOs. What do you think? Would that be more or less confusing?

yeastplume: There are a lot of things within the code that need renaming at the moment due to this change, so I think I'll open a separate issue where we can decide on these things and then undertake a careful search and replace.

yeastplume:
Thanks for review! I'll merge once latest tests have passed and put together a separate issue for the renaming that needs to happen.

chain/src/sumtree.rs
antiochp: :+1: on renaming `utxo_pmmr*` to `txo_pmmr*` (in a separate PR).
I get the confusion now that I have a better understanding of how pruning/compaction works.
Your point that "the root is the hash of all TXOs" just clicked for me (and nicely explains how the Merkle proofs still work even after pruning...)

PR #707, opened 2018-02-13
archive_mode is optional in grin.toml

antiochp:
Failing to start if this config was missing from grin.toml - 

```
thread 'main' panicked at 'Error parsing config file: Error parsing configuration file at 
/antiochp/grin/node1/grin.toml - missing field `archive_mode` for key `server`', 
/antiochp/grin/src/bin/grin.rs:83:2
```

* default to None in grin.toml (optional config, defaults to false when parsing config)
* add example config in grin.toml

### DIFF
--- a/grin.toml
+++ b/grin.toml
@@ -42,13 +42,15 @@ db_root = ".grin"
 #
 chain_type = "Testnet1"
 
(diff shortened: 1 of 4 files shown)


ignopeverell:
Ah, thank you fo that!

PR #710, opened 2018-02-14
[WIP] tx pool lookup for kernel short ids (compact block hydration)

### LINKED ISSUE #645: Tx Pool lookup by input/output/kernel short_ids (for compact blocks)
See #522 for "Compact Blocks".

This issue is tracking the tx pool functionality required for handling input/output/kernel lookup from the tx pool by short_id on receiving a compact block.

We are considering changing compact block format so that they only contain short_ids for the kernels themselves. 
See #668.
We would use these to lookup inputs and outputs indirectly from the tx pool

- [x] maintain index of txs (hashes?) in the pool by ~input/output~ kernel short_ids

On receiving a compact block -
- [x] re-index based on new short_ids (keyed by new incoming block hash)
- [x] identify known txs that make up the block (by their constituent kernels)
- [x] retrieve txs from tx pool and if all inputs/outputs/kernels present build the full block
- [x] fallback to request full block if we cannot hydrate the full block from tx pool data

antiochp:
TODO - 
- [ ] deduplication and sorting of results
- [ ] include the nonce (after #709 merged)
- [ ] error handling
- [ ] tests

Resolves #645.


antiochp:
@ignopeverell finding that `compact_blocks` and `short_ids` are using a mix of both `core::hash::Hash` and `std:hash::Hash` and the interaction between these two feels clunky, verging on incompatible.
Have we put any thought into possibly reworking `core::hash::Hash` to leverage `std::hash::Hash` more? (Right now they are orthogonal _in some ways_).

The specific problem I came up against was attempting to place a bunch of `TxKernels` into a `HashSet` to eliminate duplicates. But `TxKernel` does not implement the `std::hash::Hash` trait and has no way of being hashed to put it in the set (even though we have implemented our canonical hashing function for it in `core`).

PR #711, opened 2018-02-15
Bulletproofs

yeastplume:
This integrates bulletproofs into Grin via an update to our libsecp256k fork. A couple of notes:

* Refactored a direct secp call or two into Keychain to ensure changes are localised.
* No real need to rename anything or create any new rangeproof structures... a Bulletproof is simply an instance of a Rangeproof, so whether bulletproofs or 'classic' are being used is an implementation detail
* I've inserted a build flag into keychain/Cargo.toml to compile with either bullet proofs or traditional range proofs. I've done it this way so as to still allow development on the wallet if needed (and in case of early disasters,) but bulletproofs are enabled by default.
* Bulletproof unwinding and hence wallet recovery won't work if bulletproofs are enabled... will open a separate issue for this.

That aside, revel in how much smaller everything should become!

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -1062,9 +1062,13 @@ mod test {
 		let b = new_block(vec![], &keychain);
 		let mut vec = Vec::new();
 		ser::serialize(&mut vec, &b).expect("serialization failed");
+		let target_len = match Keychain::is_using_bullet_proofs() {
+			true => 1_256,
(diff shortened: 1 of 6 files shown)


yeastplume:
Travis finally decided to co-operate? I take it nobody has any particular issues with me merging this shortly?

ignopeverell:
Knock yourself out.

Issue #166, opened 2017-10-10
Better error handling with error-chain?


ignopeverell:
Rust error handling with Result is pretty good, relatively, but one big drawback is stacktraces from the error origination are not captured. Stacktraces will only be captured at the time of panic, which can be several levels above where the error occurred.

The error-chain crate [1] solves this by instrumenting errors with macros and encapsulates useful error patterns. Evaluate the use of error-crate and migrate current error handling to it if appropriate.

[1] https://github.com/rust-lang-nursery/error-chain

casey:
`failure` might also be good to check out as an error-chain alternative:

[github](https://github.com/withoutboats/failure)
[website](https://boats.gitlab.io/failure/)
[discussion](https://www.reddit.com/r/rust/comments/7b88qp/failure_a_new_error_management_story/)

It looks a little bit less opaque and more obvious than `error-chain`, in my opinion.

dongcarl:
@ignopeverell @casey I want to get started on this. I know that `failure` is getting more and more popular, let's decide on one and I'll go ahead with refactoring.

antiochp:
I'd say go for it with `failure` on a branch and we can take a look - is it possible to do this in a limited area of the codebase initially (just to see how it looks?)

dongcarl:
@antiochp Will do with a limited set of files that makes sense and reference this ticket.

sesam:
@dongcarl Is this also about doing less of .unwrap()?

According to https://users.rust-lang.org/t/to-unwrap-or-not-to-unwrap/10900/10 the right thing is to not .unwrap().
Is grin different? I'm thinking that maybe getting a panic! quickly because of an .unwrap() is  a benefit in grin sometimes.

ignopeverell:
Right, for non-recoverable situations, panicking is the best thing to do. I think it would still be better to rely on `expect` instead of `unwrap` to have at least a minimal message.

dongcarl:
Busy with other things right now, so if someone else has got more time, go right ahead!

Derzouli:
Hello guys, I'm new in these project, and I want to start this ticket. =)
I start working on it. Which limited set of files are related to this issue?

sesam:
@Derzouli check the links above (https://github.com/mimblewimble/grin/issues/166#issuecomment-342335105) and then pick some part of grin to try and apply `failure` to.

[...]

PR #712, opened 2018-02-16
[WIP] Error handling with failure

hashmap:
This PR adresses https://github.com/mimblewimble/grin/issues/166

We explored two approaches:
* Custom Fail type  https://boats.gitlab.io/failure/custom-fail.html
* Error and ErrorKind pair  https://boats.gitlab.io/failure/error-errorkind.html

This PR illustrate the first approach in wallet code, see a separate PR for the second one. This one is very close to the existing code with standard error, it 1:1 maps underlying error type to the wallet error variant.

 This is work in progress, we'd like to get some initial feedback before investing more efforts into it.

PR #713, opened 2018-02-16
Error handling with failure (using Error and ErrorKind)

hashmap:
This PR adresses https://github.com/mimblewimble/grin/issues/166

We explored two approaches:
* Custom Fail type  https://boats.gitlab.io/failure/custom-fail.html
* Error and ErrorKind pair  https://boats.gitlab.io/failure/error-errorkind.html

This PR illustrate the second approach in the wallet code, see a separate PR for the first one. This approach changes a lot in error handling but potentially is more flexible. It allows n:m mapping of underlying errors to the wallet errors.

This is work in progress, the code can't be merged or even compiled (2 errors outside wallet module) we'd like to get some initial feedback before investing more efforts into it.


ignopeverell:
I would favor this over #712 but would like to hear what @antiochp, @yeastplume and @quentinlesceller would have to say as well.

Sorry for the stupid question, I'm not familiar with failure (yet) and the docs weren't really explicit: how does one specify whether a failure captures its stacktrace?

hashmap:
@ignopeverell I agree, this approach is more promising but requires more efforts. The author of failure mentioned that it could be (more) automated in the future.

There is no such thing as stupid question. It follows the panic convention - if `RUST_BACKTRACE` is set a stacktrace will be generated https://boats.gitlab.io/failure/fail.html#backtraces

hashmap:
Forgot to mention that we had to implement s`std::error::Error` for errors in `keychain` and `transaction`. It's temporary solution until it's converted to failure.

wallet/src/client.rs
```diff
@@ -67,33 +69,32 @@ pub fn send_partial_tx(url: &str, partial_tx: &PartialTx) -> Result<PartialTx, E
 }
 
 fn single_send_partial_tx(url: &str, partial_tx: &PartialTx) -> Result<PartialTx, Error> {
-	let mut core = reactor::Core::new()?;
+    let mut core = reactor::Core::new().context(ErrorKind::Hyper)?;
```
antiochp: We have to be more explicit here? By specifying the `ErrorKind::Hyper` context rather than just using `?`.
I've only taken a quick look through this but I'm not sure I fully understand the benefits that Error/ErrorKind bring here?

wallet/src/types.rs
```diff
@@ -63,107 +63,102 @@ pub fn tx_fee(input_len: usize, output_len: usize, base_fee: Option<u64>) -> u64
...
+    pub fn kind(&self) -> ErrorKind {
+        *self.inner.get_context()
+    }
 }
 
-impl From<hyper::Error> for Error {
-	fn from(e: hyper::Error) -> Error {
-		Error::Hyper(e)
-	}
+impl From<ErrorKind> for Error {
```
antiochp: Oh I think I see the benefit here. Everything collapses into the single `Error` type and we just to be able to convert from ErrorKind to Error (once).

wallet/src/types.rs
```diff
@@ -506,11 +500,7 @@ impl WalletData {
 		wdat.write(data_file_path)?;
 
 		// delete the lock file
-		fs::remove_file(lock_file_path).map_err(|_| {
-			Error::WalletData(format!(
-				"Could not remove wallet lock file. Maybe insufficient rights?"
-			))
-		})?;
+		fs::remove_file(lock_file_path).context(ErrorKind::WalletData("Could not remove wallet lock file. Maybe insufficient rights?"))?;
```
antiochp: And `.context()` is significantly cleaner than `map_err()`.

hashmap:
Thanks for nice and helpful feedback. Merged with the current master. We will check couple of places where `.map_err()` is still used. Need  some help with verifying context in each case, could be incorrect.

hashmap:
It's ready for review.

antiochp:
@hashmap looks like there are some conflicts to clean up.
Lets get this merged in if @ignopeverell and @yeastplume are happy with this approach.

yeastplume:
seems to makes sense, very much in favour of centralising error handling code better and even though this looks a bit boilerplate heavy to set up, it looks like it will ultimately make error handling easier to follow.. will be much better informed after trying to use it properly for a while, but I’m okay giving this approach a try

antiochp:
Sounds like everyone is on board? I'm going to go ahead and merge.

heunglee:
While I am working on taking iron out of wallet lib, I noticed the changes on error handling. Using failure::Fail in wallet seems not fully embraced by Api library yet with respect to client/server error handling. At this moment, I map Api::Error into Http response:
Error::Argument -> Http::StatusCode::BadRequest
Error::Internal -> Http::StatusCode::InternalServerError
Error::NotFound -> Http::StatusCode::NotFound
 and vise versa on the client side.

Sine I am not familiar with failure, I am not sure how to map Fail(e) on coinbase handler in wallet into Http status code. e is of Api::Error. So for now, I will uncomment the part for later visit and just map e into http status code. Any opinion and thoughts?

hashmap:
@heunglee  I'd suggest to map ErrorKind's variants (https://github.com/mimblewimble/grin/blob/master/wallet/src/types.rs#L76) to HTTP status codes 
here is an example https://github.com/mimblewimble/grin/blob/master/grin/tests/framework/mod.rs#L339

PR #716, opened 2018-02-19
Merkle Proofs 

antiochp:
* Functionality to generate a Merkle proof for a given entry in the output MMR.
* Merkle proof consists of the following - 
  * hash of the node in the tree (leaf node representing an output)
  * hash of the overall root of the tree
  * list of hashes representing the peaks (matches the root when hashed together)
  * list of sibling hashes to construct branch from node to peak
  * list of left/right sibling positions to allow the branch to be reconstructed correctly
* A Merkle proof allows us to - 
  * prove inclusion of the node beneath its peak (via the branch of sibling hashes)
  * prove peaks combine to form overall root
  * prove inclusion of an (unspent) output for a given block_header by verifying the utxo_root matches the root of the Merkle proof
  * if an output is unspent at a given block header then it is _at least as old_ as that block

* API serializes/deserializes Merkle proofs as hex
* Wallet maintains Merkle proof (and corresponding block hash) for each unspent coinbase output
* Spending a coinbase output requires the block hash and Merkle proof to be provided in the input
  * these are used to verify coinbase maturity (without requiring full block data for verification)

TODO - 
- [x] ~Cannot use `rewind` to put the MMR back into a state where we can create an old Merkle proof (because of pruning?) - confirm we can only create a Merkle proof at the same time the block is originally created?~
    We can now successfully generate a Merkle proof via `sumtree.rewind()`.
    This _definitely_ works for a full archival node.
    This will _in principle_ work for a pruned non-archive node - but we need to confirm this.
- [x] cleanup test failures 

We uncovered a subtle bug in our `sumtree.rewind()` logic - #727 (FIXED).


antiochp:
@ignopeverell I keep getting tripped up with MMR rewind...
I think I have managed to (finally) convince myself that we cannot use `rewind()` to reliably put the output MMR into the correct state for a historical block, and we are unable to create a Merkle proof for the coinbase output for a historical block.
Outputs in the MMR may have been removed and rewind will not handle this correctly.

So am I right in thinking the Merkle proof can _only_ be created at the time the block itself is created (we cannot reliable regenerate the exact same utxo_root otherwise)?
And that the miner will need to create any necessary Merkle proofs when they successfully mine a block (and presumably store them somewhere to make them available to the wallet)?

ignopeverell:
I think you're correct. Note that you should be able to rewind to that position for a little while, even in presence of pruning. But when the local storage gets compacted then actual data you would need gets removed. So we could wait 10-20 blocks to avoid storing data that could be useless in the common 1-block reorg case, but beyond that you're taking a chance. Note that waiting a bit sounds like a premature optimization to me at this point, just mentioning it as an example.

antiochp:
@ignopeverell After fixing `rewind()` and ignoring the remove list (`read_from_file()` vs existing `read()`) we can now generate Merkle proofs reliably and correctly on a full archival node.
As discussed - we should have all the non-leaf nodes in the tree even after pruning/compacting.

Planning to investigate if this works as-is with a pruned tree or if we need to modify the tree compacting behavior to retain leaf siblings to keep Merkle proofs working post-compaction step on the output PMMR.

I think we're getting close to landing this PR (hopefully).

antiochp:
This is ready for review.
We will want to revisit this when we tackle pruning/compaction of the PMMR (we need to ensure we still have access to the hashes of all siblings, including a potentially removed sibling of the leaf node).

@ignopeverell @yeastplume

api/src/types.rs
```diff
@@ -277,13 +292,22 @@ impl OutputPrintable {
...
-		let mut state = serializer.serialize_struct("OutputPrintable", 6)?;
+		let mut state = serializer.serialize_struct("OutputPrintable", 7)?;
 		state.serialize_field("output_type", &self.output_type)?;
 		state.serialize_field("commit", &util::to_hex(self.commit.0.to_vec()))?;
 		state.serialize_field("switch_commit_hash", &self.switch_commit_hash.to_hex())?;
 		state.serialize_field("spent", &self.spent)?;
 		state.serialize_field("proof", &self.proof)?;
 		state.serialize_field("proof_hash", &self.proof_hash)?;
+
+		// TODO - rework this
```
ignopeverell: Still needs rework?

core/src/core/block.rs
```diff
@@ -660,6 +662,30 @@ impl Block {
 		Ok(())
 	}
 
+	// TODO - how do we verify Merkle Proof here?
```
ignopeverell: Sounds like you figured this out? Maybe explain move up the comment below that explains what's verified here vs what will be verified later?

core/src/core/transaction.rs
```diff
@@ -470,14 +480,38 @@ impl Transaction {
 		self.kernels.verify_sort_order()?;
 		Ok(())
 	}
+
+	// TODO - how do we verify Merkle Proof here?
```
ignopeverell: Same

wallet/src/sender.rs
```diff
@@ -331,9 +331,19 @@ fn inputs_and_change(
 	for coin in coins {
 		let key_id = keychain.derive_key_id(coin.n_child).context(ErrorKind::Keychain)?;
 		if coin.is_coinbase {
-			parts.push(build::coinbase_input(coin.value, coin.block.hash(), key_id));
+			let block = coin.block.clone();
+			let merkle_proof = coin.merkle_proof.clone();
+			let merkle_proof = merkle_proof.unwrap().merkle_proof();
+
+			// TODO - we need a Merkle Proof here to spend a coinbase output
```
ignopeverell: Looks like you have it now?

chain/src/sumtree.rs
```diff
@@ -106,22 +107,20 @@ impl SumTrees {
...
 					&mut self.utxo_pmmr_h.backend,
 					self.utxo_pmmr_h.last_pos,
 				);
 				if let Some((hash, _)) = output_pmmr.get(pos, false) {
-					println!("Getting output ID hash");
 					if hash == output_id.hash() {
-						Ok(())
+						Ok(hash)
 					} else {
-						println!("MISMATCH BECAUSE THE BLOODY THING MISMATCHES");
```
yeastplume: Whoops, meant to remove that :D

yeastplume:
Cannot find fault here.. really well done. Ensuring siblings still exist on compaction is a fairly trivial implementation detail, all we need to do is ensure the top peak in the prune list isn't actually excluded (just everything underneath it.

chain/src/sumtree.rs
antiochp: 😀

antiochp:
@yeastplume yeah I'm planning to go take a look at the compaction stuff later today I think.

@ignopeverell thanks for that - I'd kind of skipped that last pass through all those... working through them now.

antiochp:
Just waiting for final travis run then I'll merge.

PR #719, opened 2018-02-20
[WIP] Basic Dandelion 

### LINKED ISSUE #706: Dandelion relay and aggregation for transactions
This follows from #693 (and actually reading the paper [1] and the BIP [2]). From where we stand, implementing Dandelion seems somewhat straightforward, and adding a transaction aggregation phase specific to Grin, would offer an attractive anonymity boost.

The main tweaks to our current protocol would be:

* regularly elect 2 nodes for Dandelion relay
* add a Dandelion transaction relay message type
* aggregate a received valid Dandelion transaction with all other Dandelion transactions
* relay a valid Dandelion transaction as Dandelion 90% of the time, otherwise diffuse as we do now
* accept pool replacement for valid a Dandelion transaction with a strict (input, ouput, kernel) superset
* monitor our transactions so we can re-send if we never see them from the fluff phase

We currently do not support any transaction pool querying or transaction hash relays as bitcoin does, so we do not need to worry about specific pool policies or exclusion beyond the above. Please let me know if I forgot something important or if this would break something else

Paging @gfanti in case she'd like to follow this.

[1] https://arxiv.org/pdf/1701.04439.pdf
[2] https://github.com/gfanti/bips/blob/master/bip-dandelion.mediawiki

quentinlesceller:
Work in progress.
Adds a very simple Dandelion policy. Prerequisite to fix #706. Does not include transaction aggregation.

Including:

- add a stem transaction relay message type
- relay a valid stem transaction in stem phase 90% of the time, otherwise diffuse as we do now
- monitor our transactions so we can re-send if we never see them from the fluff phase.
- create a stem pool to aggregate transaction in the future.

TODO:

- [x] Add a stem memory pool
- [x] Simple stempool architecture
- [x] Add a stem memory pool config
- [x] Add embargo time for every transaction in stem pool.
- [x] Add a transaction monitor to send transaction in case not fluffed
- [x] Add the selection of Dandelion relay every time interval
- [x] Externalize the Dandelion settings
- [x] Deal properly with child transaction
- [x] Update ```search_best_output``` for stempool
- [x] Update and add transaction test


quentinlesceller:
Going to push down some of the logics of the dice roll in order to avoid having duplicates code.
>  you could push down some of the logic from the adapter in the stem pool (like the dice roll) and have the adapter only know about "some pool-like stuff"

quentinlesceller:
@yeastplume @antiochp @ignopeverell wanna take a look? Still need to fix and add some tests but the core is here.

pool/src/pool.rs
```diff
@@ -267,12 +308,39 @@ where
...
+			if stem && random <= self.config.dandelion_probability {
+				// Stem phase: transaction is added to the stem memory pool and broadcasted to a
+				// randomly selected node.
+				self.stempool.add_pool_transaction(
+					pool_entry,
+					blockchain_refs,
+					pool_refs,
+					new_unspents,
+				);
+				// TODO watch transaction with embargo timer
```
ignopeverell: Not todo anymore?

api/src/handlers.rs
```diff
@@ -517,8 +517,12 @@ where
 			tx.outputs.len()
 		);
 
+		//  Push into the stempool
 		let pool_arc = w(&self.tx_pool);
-		let res = pool_arc.write().unwrap().add_to_memory_pool(source, tx);
+		let res = pool_arc
```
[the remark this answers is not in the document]
quentinlesceller: Agree, adding that

quentinlesceller:
Currently, adding an already seen transaction in the stempool will result in an error ```PoolError::AlreadyInPool```. If that happens both nodes wait until their embargo timer times out to fluff the transaction.

pool/src/pool.rs
quentinlesceller: Yep removed.

api/src/handlers.rs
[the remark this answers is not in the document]
quentinlesceller: Added param ```fluff```for api handler.

quentinlesceller:
Still some issues with the behavior of the pool and the stempool to fix.

grin/src/dandelion_monitor.rs
```diff
@@ -0,0 +1,82 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+use time::{self, now_utc};
+use util::LOGGER;
+
+use pool::TransactionPool;
+use pool::PoolConfig;
+use pool::TxSource;
+use pool::BlockChain;
+
+pub fn monitor_transactions<T>(
```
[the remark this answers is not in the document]
quentinlesceller: Done

quentinlesceller:
Added TODO for Randomize the embargo time (between min and max value).

quentinlesceller:
>If a node receives a child transaction that depends on one or more currently-embargoed Dandelion transactions, then the transaction is also relayed in stem mode, and the embargo timer is set to the maximum of the embargo times of its parents. This helps ensure that parent transactions enter fluff mode before child transactions.
[ ] Need to do that

grin/src/dandelion_monitor.rs
```diff
@@ -0,0 +1,90 @@
...
+use pool::TransactionPool;
+use pool::PoolConfig;
+use pool::TxSource;
+use pool::BlockChain;
+
+/// A process to monitor transactions in the stempool.
+/// With Dandelion, transaction can be broadcasted in stem or fluff phase.
+/// When sent in stem phase, the transaction is relayed to only node: the dandelion relay. In
+/// order to maintain reliability a timer is started for each transaction sent in stem phase.
+/// This function will the stempool and test if the timer is expired.
```
kargakis: uncomplete sentence?

quentinlesceller: Oups sorry yes

quentinlesceller: Fixed thank you.

quentinlesceller:
@antiochp and @ignopeverell comments ?

quentinlesceller:
Just remember that the relay should be an outgoing peer. Will fix that.

grin/src/dandelion_monitor.rs
```diff
@@ -0,0 +1,85 @@
...
+		.spawn(move || {
+			loop {
+				let tx_pool = tx_pool.clone();
+				let stem_transactions = tx_pool.read().unwrap().stem_transactions.clone();
+				let time_stem_transactions = tx_pool.read().unwrap().time_stem_transactions.clone();
+
+				for tx_hash in stem_transactions.keys() {
+					let time_transaction = time_stem_transactions.get(tx_hash).unwrap();
+					let interval = now_utc().to_timespec().sec - time_transaction;
+					// Unban peer
```
ignopeverell: Leftover comment from a copy/paste?

quentinlesceller: Yep fixed.

p2p/src/peers.rs
```diff
@@ -71,6 +73,32 @@ impl Peers {
...
 	}
 
+	// Update the dandelion relay
+	pub fn update_dandelion_relay(&self) {
+		let peers = self.outgoing_connected_peers();
+		match thread_rng().choose(&peers) {
+			Some(peer) => {
+				// Clear the map and add new relay
+				self.dandelion_relay.write().unwrap().clear();
+				self.dandelion_relay
```
ignopeverell: This takes 2 write locks in a row, without unlock. Does that actually work? Either way I'd still recommend taking the lock only once and assigning it to a variable.

quentinlesceller: I *think* I fix it. Can you take a look?

Issue #705, opened 2018-02-12
Attempt to recover from isolated orphaning


ignopeverell:
Due to occasional bad network conditions, a node can miss a block. When the next one is received and detected orphaned, we should at least try a request for the previous block, assuming it's not already an orphan in itself and we're not syncing already.

mably:
Was forked again at height 73296:

https://gist.github.com/mably/9af1b92bbdcd27dba1a8ae1884ed8804

hashmap:
I'd like to tackle this.

ignopeverell:
Fixed by #724

PR #724, opened 2018-02-22
Attempt to recover from isolated orphaning

hashmap:
Due to occasional bad network conditions, a node can miss a block.
When the next one is received and detected orphaned, we should at least try a request for the
previous block, assuming it's not already an orphan in itself and we're not syncing already.

No additinal tests were implemented, tests like `simulate_full_sync` test this functionality because
sync starts later. It seems to be an issue with sync test coverage. As fix I'd suggest to start sync
as soon as we get a peer connection.

[#705]

### DIFF
--- a/grin/src/adapters.rs
+++ b/grin/src/adapters.rs
@@ -381,24 +381,22 @@ impl NetToChainAdapter {
 		let chain = w(&self.chain);
 	    match chain.process_block(b, self.chain_opts()) {
                 Ok(_) => true,
-                Err(e) => {
-			        debug!(LOGGER, "Block {} refused by chain: {:?}", bhash, e);
-                    match e {
-                        chain::Error::Orphan => {
-                            // make sure we did not miss the parent block
-                            if !self.currently_syncing.load(Ordering::Relaxed) && !chain.is_orphan(&prev_hash) {
-                                debug!(LOGGER, "adapter: process_block: received an orphan block, checking the parent: {:?}", prev_hash);
-                                self.request_block_by_hash(prev_hash, &addr)
-                            }
-                            true
-                        }
-                        _ if e.is_bad_data() => {
-                            debug!(LOGGER, "adapter: process_block_logic: {} is a bad block, resetting head", bhash);
-                            let _ = chain.reset_head();
-                            false
-                        }
-                        _ => true
+                Err(chain::Error::Orphan) => {
+                    // make sure we did not miss the parent block
+                    if !self.currently_syncing.load(Ordering::Relaxed) && !chain.is_orphan(&prev_hash) {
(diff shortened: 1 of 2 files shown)


grin/src/adapters.rs
```diff
@@ -376,20 +375,33 @@ impl NetToChainAdapter {
...
-				return false;
-			}
-		};
-		true
-	}
+	    match chain.process_block(b, self.chain_opts()) {
+                Ok(_) => true,
+                Err(e) => {
+			        debug!(LOGGER, "Block {} refused by chain: {:?}", bhash, e);
+                    match e {
```
ignopeverell: I would merge that match into the one above, the debug statement can just be moved into the `true` case as the other branches are already covered.

grin/src/adapters.rs
```diff
@@ -88,9 +88,8 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 			b.header.height,
 			addr,
 		);
-
-		self.process_block(b)
-	}
+        self.process_block(b, addr)
```
ignopeverell: Looks like your indentations are off.

grin/src/adapters.rs
```diff
@@ -398,16 +410,21 @@ impl NetToChainAdapter {
 	// TODO - currently only request block from a single peer
 	// consider additional peers for redundancy?
 	fn request_block(&self, bh: &BlockHeader, addr: &SocketAddr) {
-		if let None = wo(&self.peers).adapter.get_block(bh.hash()) {
+        self.request_block_by_hash(bh.hash(), addr)
+	}
+
+	fn request_block_by_hash(&self, h: Hash, addr: &SocketAddr) {
+		if let None = wo(&self.peers).adapter.get_block(h.clone()) {
```
ignopeverell: This loads a whole block, including parsing it, just to check if we already know about it. How about using `block_exists` instead of `get_block`?

grin/src/adapters.rs
hashmap: Tried to preserve the original log message, but you are right, we have specific ones in each case

grin/src/adapters.rs
hashmap: Not just here, I'm a bit lost, it looks like 8-spaces tabs are used (I used 4), also do we replace tabs with spaces? Should be fixed in this PR anyway

hashmap:
I refactored block request part, originally thought it should be a separate pr, but since it was mentioned during review I implemented it. I found that existing code was ignoring too many error cases, so new version is much bigger

hashmap:
Not sure why it failed, will trigger another travis build.

ignopeverell:
Travis can be unreliable... Thanks for the great work!

PR #725, opened 2018-02-23
Add Inbound/Outbound for connected peers

quentinlesceller:
This PR adds direction Inbound/Outbound for connected peers:
When the node initiate the connection, it's outbound, otherwise it's inbound.
This will help for Dandelion stem relay https://github.com/mimblewimble/grin/issues/706.
~~Also fix unnoticed typos in tests.~~

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -1065,7 +1065,7 @@ mod test {
 		};
 		assert_eq!(
 			vec.len(),
-			target_len
+			target_len,
 		);
 	}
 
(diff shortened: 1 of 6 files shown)


core/src/core/block.rs
```diff
@@ -1065,7 +1065,7 @@ mod test {
 		};
 		assert_eq!(
 			vec.len(),
-			target_len,
+			target_len
```
antiochp: I think our "style guide" has trailing commas in lists of args like this.
Is this happening in your editor? Or rustfmt?

quentinlesceller: Found out that my rustc was not updated and did not accept these trailing commas. Reverted.

PR #727, opened 2018-02-23
last output in a block is the last regular output (last coinbase output if no regular outputs)

### LINKED ISSUE #726: sumtree.rewind does not correctly account for coinbase pos
If we have a block containing the following outputs (sorted lexicographically) -
* [out<sub>1</sub>, cbout<sub>2</sub>]

When we call `sumtree.apply_block()` we process the contents of the block in the following order - 
* coinbase outputs
* inputs
* regular outputs

We process these two outputs "out of order", processing the coinbase output before the regular output.

If we then subsequently call `extension.rewind()` we find the `last` output and rewind to that pos.
This is based on the sort order of the outputs and does not account for coinbase outputs being processed first.

So if the pos are as follows - 
* cbout<sub>2</sub>, pos: 1 (coinbase processed first)
* out<sub>1</sub>, pos: 2

Rewinding to this block will incorrectly rewind to pos 1 (based on raw sort order) when it should actually be pos 2 (if we account for coinbase _and_ sort order).

antiochp:
It is not sufficient to simply look for the last output in sorted order.
The behavior in `apply_block()` is to process all coinbase outputs, then all regular outputs so we need to be consistent with this.

This is a pretty subtle issue but we have probably been seeing this with some percentage of blocks containing actual txs with regular outputs.

I was seeing this a *lot* on my branch when generating Merkle proofs for coinbase outputs as it was really obvious when an invalid Merkle proof was generated (vs. subtly rewinding to an off by one position in the MMR while processing a block on a fork).

Resolves #726.

### DIFF
--- a/chain/src/sumtree.rs
+++ b/chain/src/sumtree.rs
@@ -359,7 +359,7 @@ impl<'a> Extension<'a> {
 				// check hash from pmmr matches hash from input (or corresponding output)
 				// if not then the input is not being honest about
 				// what it is attempting to spend...
-				if output_id_hash != read_hash || 
+				if output_id_hash != read_hash ||
 					output_id_hash != read_elem.expect("no output at position").hash() {
 					return Err(Error::SumTreeErr(format!("output pmmr hash mismatch")));
 				}
@@ -668,7 +668,7 @@ where
 	file_store.get(leaf_index - 1)
 }
 
-fn _remove_element_at_pmmr_index<T>(file_store: &mut FlatFileStore<T>, pos: u64) 
+fn _remove_element_at_pmmr_index<T>(file_store: &mut FlatFileStore<T>, pos: u64)
 	-> Result<(), String>
 where
 	T: ser::Readable + ser::Writeable + Clone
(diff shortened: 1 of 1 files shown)


antiochp:
Pinging @ignopeverell and @yeastplume for review.

ignopeverell:
Yes, great catch! As you mention this is a bit subtle and could be easy to miss again.

Maybe forcing the coinbase outputs to be first (as well as the kernels for symmetry I guess), similarly to your `Vec` suggestion, could simplify this? The code that keeps things in order would have to be updated.

Thinking about it, I'm on the fence. On one hand we'd be influencing the block format to accomodate an implementation constraint, on the other coinbase-first is fairly common.

antiochp:
Yeah I actually started down that path, trying to get the sort impl to do this for us. But it feels very implicit - its really not obvious that this is what is happening. We're not just sorting them but actually partitioning the vec into two lists to be processed separately in apply_block().
I'm leaning toward merging this as-is right now but maybe breaking out the output and kernel index gathering into separate functions to make it a bit more explicit.

PR #730, opened 2018-02-26
Bulletproof messages

yeastplume:
WIP for integrating bulletproof messages, cleaning up any storage around the switch to using bulletproofs as well as embedding all values that need to go into the proof.

### DIFF
--- a/chain/src/sumtree.rs
+++ b/chain/src/sumtree.rs
@@ -639,7 +639,7 @@ impl<'a> Extension<'a> {
 					let commit = out.commit.clone();
 					match self.rproof_pmmr.get(n, true) {
 						Some((_, Some(rp))) => out.to_output(rp).verify_proof()?,
-						res => {
+						_res => {
 							return Err(Error::OutputNotFound);
 						}
 					}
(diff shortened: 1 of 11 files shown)


yeastplume:
This is ready for merging, with caveats that still need to be addressed here https://github.com/mimblewimble/grin/issues/734. A full description of what's changed:

* The preliminary rust libraries for bulletproof message hiding and unwind are being used in grin now, and proofs in grin have their messages being populated with the amount and switch commit hash. (This isn't final, see caveats in issue above)
* I've added a compilation flag to rust-libsecp..etc... to use the size of a single bullet proof (674 bytes) as PROOF_MAX_SIZE, and enabling it via our build to ensure we're only storing the much size-reduced proofs everywhere.
* Removed conditional compilation selecting rangeproof classic or bulletproof as there's no need to litter the code with conditions if BPs are the way forward. 

Not all that much here (the bulk of the work is in libsecp256k), so if no objections over the next few hours will just merge.

ignopeverell:
Looks like the tests are genuinely failing (for once).

yeastplume:
TravisCI being useful for once.. forgot to commit Cargo.toml with shiny new feature flag enabled.

PR #732, opened 2018-02-26
Add initial Dandelion doc

quentinlesceller:
Add initial documentation for Dandelion in Grin. Similar to http://github.com/gfanti/bips/blob/master/bip-dandelion.mediawiki but adapted to Grin. Missing the Aggregation part for now.

### DIFF
--- a/doc/internal/dandelion.md
+++ b/doc/internal/dandelion.md
@@ -1,0 +1,68 @@
+Dandelion in Grin: Privacy-Preserving Transaction Propagation
+==================
+This document describes the implementation of Dandelion in Grin.
+## Introduction
+
+Dandelion is a new transaction broadcasting mechanism that reduces the risk of eavesdroppers linking transactions to the source IP. Moreover, it allows Grin transactions to be aggregated (removing input-output pairs) before being broadasted to the entire network giving an additional privacy perk.
+
+Dandelion was introduced in [1] by G. Fanti et al. and presented at ACM Sigmetrics 2017. On June 2017 a BIP was proposed introducing a more practical and robust variant of Dandelion called Dandelion++ [2].  This document is an adapation of this BIP for Grin.
+
+## Mechanism
+
+Dandelion transaction propagation proceeds in two phases: first the “stem” phase, and then “fluff” phase. During the stem phase, each node relays the transaction to a *single* peer. After a random number of hops along the stem, the transaction enters the fluff phase, which behaves just like ordinary flooding/diffusion. Even when an attacker can identify the location of the fluff phase, it is much more difficult to identify the source of the stem.
+
+Illustration:
+<pre>
+                                                   ┌-> F ...
+                                           ┌-> D --┤
+                                           |       └-> G ...
+  A --[stem]--> B --[stem]--> C --[fluff]--┤
+                                           |       ┌-> H ...
+                                           └-> E --┤
+                                                   └-> I ...
+</pre>
+
+TODO Add transaction aggregation in stem mode
+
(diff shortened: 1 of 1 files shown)


antiochp:
@quentinlesceller this looks great! Going to take a read through of this shortly.

antiochp:
@quentinlesceller Some thoughts on the aggregation functionality - 

Assuming a node will always aggregate all txs seen during the stem phase,  I *think* it is sufficient to just maintain a single aggregated tx in the stem pool.
* The stem pool tx must be valid.
* We should not aggregate and accept a tx into the stem pool unless we could also add it to regular tx pool given current state of tx pool.

Flow - 
1. Receive stem tx
1. Check if any existing tx in stem pool is valid (check against current tx pool)
    * If not valid - do we just evict from stem pool? (And drop it?)
1. Aggregate incoming stem tx with any existing stem pool tx
1. Check aggregated tx is valid (check against current tx pool)
    * If valid then store aggregated tx in stem pool
    * If not valid then discard and revert stem pool to previous tx
1. Flip the biased coin
    * 90% stem phase
        * relay single aggregated tx to single stem peer 
    * 10% fluff 
        * add single aggregated tx to tx pool
        * reset/clear stem pool 
        * broadcast aggregated tx to peers as normal

If a node wants to leverage the Dandelion stem phase for tx cut-through (not just aggregation) then the node must send the individual stem txs in the correct order for them to be accepted (and cut-through) sequentially. Each individual tx must result in the stem pool being in a valid state.

If at any time the stem pool tx is invalid then we should reject the tx on the basis of the current tx pool rejecting it.

Thoughts on this?

@ignopeverell Is this in line with how you are thinking about tx aggregation?

I think if we approach aggregation roughly like this then it may be simpler to implement the Dandelion stem phase *with* aggregation initially. This way we will not need to maintain multiple txs in the stem pool and the logic around valid vs. invalid stem txs is potentially a lot simpler.

quentinlesceller:
@antiochp Thank you for your ideas on aggregation. I agree with this flow.
I think, we must also define the aggregation time windows for node; the patience of each node.
Just like the embargo in Dandelion, each node should maintain a random timer (around a ```patience``` value). Once a node receives a stem transaction, the timer start, allowing him to aggregate stem tx receivd in this time windows. At the end of the timer, the coin flip happens.
We should agree on a time windows fast enough not to slow the tx propagation too much but slow enough to allow significant aggregation.
I imagine that most of the stem tx cut through will happen at the first hops. However, it might be problematic that the receiver submit the transaction to the stempool.

antiochp:
Yes I think that makes sense. Each node in the stem phase will introduce a delay during which time aggregation can occur.

There is also the additional timer on each node where we fall back to fluff and propagate if we do not see the tx appear in the tx pool or a block. This is going to add some complexity because we will need to account for aggregation that may have happened is subsequent stem relay hops. We may never see _this_ tx in the tx pool (but we can look for the set of kernels making up the tx).

For a given node in the stem there may be many nodes that have randomly selected it as the next stem relay, so even nodes far along a given stem may be aggregating txs so I don't think the first hops will be any different to later hops in terms of aggregation. (At least that is my understanding).

tromp:
regardless of patience, if one tx has a lockheight of 1 day ahead, and another has a lockheight 1 week ahead, then it would be wise to keep them separate, so the former can be propagated with less delay (and less aggregation) than the latter...

antiochp:
@tromp - I see what you're saying there - we should avoid aggregating txs if it would result in delaying a tx due to extending the lockheight.

PR #736, opened 2018-02-27
Log the reason of peer ban

### LINKED ISSUE #455: Add more INFO (or DEBUG?) outputs on when/why bans happen
Me and mably got stuck at 70% of best known total_difficulty, probably by banning the few good grin nodes at some point during initial sync. grin should be more chatty about why and how bans happen, since even with #439 we'd not want to ban good nodes during initial sync.

hashmap:
This PR fixed #455 by adding a few debug log messages


antiochp:
I think travis is acting up - restarted the one failed job (appeared unrelated).

Issue #735, opened 2018-02-27
Store PMMR file position index to mitigate data corruption


yeastplume:
Augment PMMR storage data to also contain a metadata file that contains the byte-sizes of the PMMR as of the last N blocks. This may have several benefits, but is mostly to aid recovery in the event of corruption

* Metadata file will be written to as the absolute last action when committing a block.
* On start up and PMMR loading, the system can truncate to the last written size. If that results in invalid data, system can attempt to read earlier indices to find a block/position pairing that results in valid data.

yeastplume:
Will address https://github.com/mimblewimble/grin/issues/371

ignopeverell:
Just a thought. The chain head is already written last and is what the chain, on restart, uses as a basis to know where it's at. So maybe add the PMMR sizes/positions to that would make sense. The downside is that there's only one head, so it wouldn't work to store it for the last N blocks. But the head has the latest block hash as well as the one before last.

antiochp:
That might be a nice simple solution and would avoid another file to manage.
Do we trust rocksdb in this scenario? I assume we do, we don't see issues with writes to the db, the issue is non-atomicity between writing the pmmr file and the associated rocksdb indices.

ignopeverell:
Right, it would be more a 2-phase-commit, XA type of thing. And we don't really care about either of those things as long as we get to write one last authoritative record at the end.

yeastplume:
Yes, have just been looking at it and storing it in rocksdb seems to be far simpler and makes sense. We can start off storing positions for the last block and last block-1, and if that turns out to be inadequate we can look at further options.

Will implement this and then try and create a test or two that corrupts data and recover.

yeastplume:
Done via #739

PR #739, opened 2018-03-01
PMMR Block to File positions metadata storage

yeastplume:
For https://github.com/mimblewimble/grin/issues/735

Will fill more detail here later, but the general approach is to collect all of the file positions when a block is commit to the dataset and the files are synced, then store this metadata within the DB by block hash. This should allow us to build a list of rollback positions for the PMMR files, however for a first out we'll just limit storage to the current block hash and the previous block.


antiochp:
Question - do we need this for every block or just the last known good one? 
I'm not sure if we gain much by tracking this for every block (but I guess this would support reverting to any historical block).

chain/src/pipe.rs
```diff
@@ -108,7 +108,49 @@ pub fn process_block(b: &Block, mut ctx: BlockContext) -> Result<Option<Tip>, Er
...
+		},
+		Err(e) => Err(e),
+	}
+}
+
+fn sync_file_metadata(b: &Block, sumtrees: &sumtree::SumTrees, store: Arc<ChainStore>) -> Result<(), Error> {
+	// We want to store written file data for the last 2 blocks, so we potentially
+	// have a couple of options to roll back to
+
+	// TODO: Delete all old blocks before doing this
```
ignopeverell: I think we can handle the deletion at the same time we do compaction. So we'd delete the old blocks, delete the metedata record, and compact the MMR regularly.

chain/src/pipe.rs
```diff
@@ -108,7 +108,49 @@ pub fn process_block(b: &Block, mut ctx: BlockContext) -> Result<Option<Tip>, Er
...
+	// Delete the metadata of the oldest block written
+	let oldest_block_hash = store
+		.get_previous_pmmr_file_block();
+	if let Ok(h) = oldest_block_hash {
+		store
+			.delete_block_pmmr_file_metadata(&h)
+			.map_err(|e| Error::StoreErr(e, "deleting oldest stored pmmr block".to_owned()))?;
+	}
+
+	// Swap previous block index
```
ignopeverell: Instead of the 2 below wouldn't it be simpler (and more atomic) to save that directly in the chain head?

yeastplume:
@antiochp Originally we'd talked about storing the last 'n' blocks, where n is a smallish number, so we'd have options to roll back to in the event of corruption, but I think the last block and the previous one to that is enough to start with. I've implemented the data storage to store file positions by hash so we can theoretically add more historical block positions if needed.

chain/src/pipe.rs
yeastplume: Yes, so long as compaction happens fairly regularly.. just worried about leaving stale pairings around in the DB in the event of forks, and I don't think we actually have any store functions implemented yet to delete all records by their prefix (sure it's trivial, just haven't looked up the docs). Compaction's probably a good place to do this

chain/src/pipe.rs
yeastplume: Absolutely, good point, I can just add a field to the Tip struct and save it then, then we just have chain head and previous block position to deal with for the time being.

yeastplume:
updated now to just use the Tip fields for the current and previous blocks.. these should be okay in a fork, because block hash/PMMR pairings are just added every time a block is successfully processed therefore the paring should always be valid for the latest version of a block (until compaction at least, which should remove all of the pairings)

chain/src/chain.rs
```diff
@@ -176,6 +187,7 @@ impl Chain {
 					genesis.header.nonce,
 					genesis.header.pow,
 				);
+				let _ = pipe::save_pmmr_metadata(&tip, &sumtrees, store.clone());
```
antiochp: Probably needs a `?` here in case of failure?

PR #743, opened 2018-03-02
rustfmt now runs and reformats files (need to opt out explicitly)

antiochp:
*** DO NOT MERGE UNTIL THE WEEKEND *** RUN RUSTFMT GLOBALLY FIRST ***

rustfmt is now available as preview (not just nightly) - https://blog.rust-lang.org/2018/02/15/Rust-1.24.html

This PR is to open up our semi-regular (and no doubt contentious) discussion around "do we want to make this opt-out rather than opt-in which we have currently".

Technically this is still opt-in as you need to enable the pre_commit hook.

But once enabled - the hook will do the following - 

* check any files being committed
* for any files that need reformatting - 
    * actually reformat them
    * notify the user of the list of affected files
    * (does not commit them, leaves them unstaged)
* the user will then need to make a conscious decision - 
    * `git add` and `git commit` to clean this up (easiest choice to make)
    * or revert any local changes to avoid reformatting (potentially appropriate in some circumstances)

Rustfmt docs are here - 
https://github.com/rust-lang-nursery/rustfmt

To (re)install rustfmt (via rustup component add) - 
* remove any existing rustfmt install
* `rustup component add rustfmt-preview`

---------------
Example output when file(s) are reformatted - 

screen shot 2018-03-02 at 12 57 54 pm

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Copyright 2017 The Grin Developers
+# Copyright 2018 The Grin Developers
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -14,9 +14,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cargo +nightly fmt -- --version &>/dev/null
+rustfmt --version &>/dev/null
 if [ $? != 0 ]; then
-	printf "[pre_commit] \033[0;31merror\033[0m: \"cargo +nightly fmt\" not available?\n"
+	printf "[pre_commit] \033[0;31merror\033[0m: \"rustfmt\" not available. \n"
+	printf "[pre_commit] \033[0;31merror\033[0m: rustfmt can be installed via - \n"
+	printf "[pre_commit] $ rustup component add rustfmt-preview \n"
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


antiochp:
Pinging @ignopeverell @yeastplume

sesam:
Having rustfmt nag for each and every file during regular development is a productivity killer, but I would be very happy if we can run rustfmt once and for all files at some point, sooner rather than later.

Surely better before Testnet2 is opened up for public testing.

Once https://github.com/mimblewimble/grin/pull/716/ (touching 23 files) and https://github.com/mimblewimble/grin/pull/719/ (touching 13 files) have been merged maybe there will be an opportunity.

Or rustfmt everything except for the files those two PRs above are touching? If @quentinlesceller and @yeastplume just merge the rustfmt reformat mega-commit, and as long as it doesn't touch any files they are working on, it could actually pass quite smoothly.

This is something I feel strongly enough about to offer to get it done :)

antiochp:
@sesam - I think we may see a (short) period of time when we're being nagged. But once files are cleaned up they should _stay_ cleaned up. 
Hopefully everyone gets into the habit of running rustfmt locally before committing to avoid the nagging messages.

ignopeverell:
I'd be ok with:

1. Doing another global rustfmt soon, during a period of lower activity (maybe this w/e)?
2. Presuring for the use of the commit hook for all future PRs.

Sounds reasonable?

antiochp:
@ignopeverell sounds good to me if everyone is on board with this.

quentinlesceller:
Perfect for me too.

yeastplume:
Sure, just let me rebase and merge my outstanding PR (will do this evening)

antiochp:
If anyone does have any branches outstanding, running rustfmt across both master and the branch itself will (in theory) get rid of any formatting related conflicts.

sesam:
3way merge should (in theory...) handle this if rustfmt is applied equally
on both sides.

PR #746, opened 2018-03-03
Compaction of pruned data for chain data structures

ignopeverell:
Facilities to trigger the compaction of all chain data that has previously been pruned: output and kernel MMRs, position indexes, full blocks.

For now, the actual triggering has to be done proactively by invoking a specific API endpoint. Once this has been tested with live data and we have a better idea of how resource-intensive the operation is, we can easily add a scheduler to trigger it automatically a couple times a day or at startup for example.


ignopeverell:
@yeastplume @antiochp I think this is ready for review. Just needs merging with master (and associated fmt).

yeastplume:
Looks good overall and good to have a basic test in there to exercise it. You're still missing the PMMR file metadata in your branch, though, and I think as part of the compaction step we need to delete all keys/values that start with 
```
const BLOCK_PMMR_FILE_METADATA_PREFIX: u8 = 'p' as u8;
```

ignopeverell:
Good point, I'll make sure to add that after I rebase.

Issue #723, opened 2018-02-22
Internal Renaming due to Storage Changes


yeastplume:
After https://github.com/mimblewimble/grin/pull/702, PMMRTrees aren't sumtrees anymore, they're PMMR trees that store hashes, and the tree's node indexing optionally references data in flat storage files.. many variables and structure names which currently reference sumtrees should be renamed for clarity. Before this can happen, we need to decide on what names we're going to use for things.

* I'm thinking Sumtrees -> Hashtrees, however there's a data storage element that we might want to reflect in this. Also, should the term PMMR be exposed or would we only leave that in points in the code where it's an implementation detail? Would PMMRTrees be more appropriate? 
* The structure in chain/sumtree.rs currently called SumTrees that now holds all of the PMMRTrees.. can we refer to this as overall TXOSet? It contains utxos, rangeproofs and kernels, all of which are 'Transaction OBjects' (unless there's a more specific meaning for this term that I don't know)
* the object within that called utxo_pmmr_h should probably just be output_pmmr (for better clarity) but I think I changed this originally so my fault.

There are probably a couple more if anyone can think of any. If we can solidify this here I can go make the changes (which more annoying than you'd realise due to the number of references in the code and comments).

ignopeverell:
I'm good with Hashtrees. I was thinking of leaving PMMR as an implementation detail.

TXOSet is a little ambiguous, as it usually means transaction Ouputs. How about TxHashSet? I agree for output_pmmr.

PR #747, opened 2018-03-05
Remove Sumtree References and disambiguate some naming

yeastplume:
In accordance with https://github.com/mimblewimble/grin/issues/723, probably a good time to do this as there don't seem to be any large branches outstanding. This is what I've attempted to do (I need to go over each renaming instance again to check if appropriate.

-Rename the struct that held the sumtrees, formerly called 'chain::SumTrees' is now TxHashSet, and all references to it renamed as such.
-Changed references to 'utxo' to Output in the cases where it meant to refer to Outputs, and TxHashSet where it was referring to the TxHashSet (as UTXO was being used ambiguously to refer to either outputs or the entire set, I thought it best that we try to limit the use of the term UTXO)
-Left PMMR references as is (I think the structs built around them are specific enough to PMMRs rather than the generic concept of a hash set to want to rename them)
-All copyright references now consistently read 2018


yeastplume:
Think it's all okay now, tests compile and run and after going though the changes I don't see anything that looks awkward. 

Should be a fun review for everyone!

ignopeverell:
I think I'll trust you and all our tests on that one...

yeastplume:
Hope you had as much fun as I had reviewing. Apologies for outstanding PRs, hopefully this should merge without conflicts

PR #748, opened 2018-03-05
Include index as part of pmmr element hash

yeastplume:
As discussed in https://github.com/mimblewimble/grin/issues/735

Added a hash_with_index() function to a trait implemented in the PMMRable type (just to keep the index hashing localised). Updated tests.


yeastplume:
Merging this, and opening an issue about Travis CI

PR #753, opened 2018-03-06
check_compact retains leaves and roots until parents are pruned

### LINKED ISSUE #741: PMMR backend check_compact needs to work with Merkle proofs
To correctly construct a Merkle proof we need to maintain the hash for each pos in the MMR that is a sibling of the branch or path from the leaf node up to its peak.

We need to be able to reconstruct a Merkle proof for a "historical" block by `rewinding` the MMR to the last pos of that block.

See -https://github.com/mimblewimble/grin/blob/b3104fe2f6c5bea8dd93c11aebd75aff7f8b4f28/doc/merkle_proof/merkle_proof.png

In the example above we want to construct a Merkle proof for pos 8.
We need enough information to reconstruct the intermediate hashes for pos 10, 14, 15, connecting the leaf at pos 8 to the peak at pos 15.

To do this we use the following "sibling" hashes - 
  * pos 9 (leaf node sibling of 8)
  * pos 13 (root of a "pruned" subtree)
  * pos 7 (root of another "pruned subtree)
  * pos 15 (a peak at the time that block existed, may not be a peak pre-rewind)

`check_compact()` currently flushes the remove log. This behavior needs to be modified to - 

* maintain leaf siblings in the remove log and underlying file so we have the hash at this pos
    * we *can* prune/remove/compact these once _both_ leaves have been removed
* maintain "roots" of subtrees that get pruned in addition to peaks themselves (maybe already do this?) 

Related - #716 Merkle Proofs (used to verify coinbase maturity when spending coinbase outputs).

antiochp:
This PR is an exploratory attempt to tweak the implementation of prune/compact/shifting to handle keeping the necessary leaves and subtree roots around after compaction, so we still have access to the necessary hashes to build a Merkle proof.

----------

To allow Merkle proofs to be built from a pruned/compacted MMR - 

* keep pruned leaves "removed" but not yet "compacted" until sibling (and parent) also removed
* keep roots of pruned subtrees "removed" but not yet "compacted"

The prune list storage will remain unchanged.
But we will interpret it differently - a pruned pos will remain in the remove log after compaction, and will compact everything beneath that pos, not that pos itself.

~So after compacting we will not simply truncate the remove log, as we will need to keep all the "pruned but not yet removed pos" in there (i.e. everything in the prune list stays in the remove log).~

Alternatively - maybe we can keep on truncating the rm_log and instead treat entries in rm_log _or_ prune list as "removed" (but not yet compacted).

* descendents of an entry in prune list: "removed" _and_ "compacted" (data no longer in the file)
* entries in the prune list itself: "removed" but not "compacted" (data still in underlying file)
* entries in the rm_log: "removed" but not "compacted" (data still in underlying file)

TODO - 
- [x] tweak PruneList `get_shift()` and `get_leaf_shift()` to account for this
- [x] do not compact things we want to keep around (roots and individual leaves)
- [x] investigate compacting an already compacted mmr
- [x] investigate data file compaction woes (suggest we tackle this separately)

Resolves #741.


antiochp:
This is getting closer - uncovered some issues when testing `check_compact` with a `horizon`.

@ignopeverell let me know if this roughly on the right track (or if it even makes sense what I am attempting to do here...)?

ignopeverell:
This seems more complicated than it should be. Pardon my asking but why would this be needed at all given that we have a 2 days horizon? In case a miner's wallet data gets lost?

antiochp:
The issue is more pervasive than that I think.

Even for a really minimal MMR containing 2 leaves (3 nodes) - 
* pos 1 regular output
* pos 2 coinbase output
* pos 3 parent/peak in the MMR

If the output at pos 1 is spent more than 2 days ago and the 2 day horizon causes pos 1 to get compacted, we have no way of generating the necessary Merkle proof for pos 2 (we need the hash for pos 1).

The coinbase output is not directly impacted by the 2 day horizon, but its spent sibling is and we have no control over the timing of this. It could have been spent months earlier.

core/src/core/pmmr.rs
```diff
@@ -588,15 +590,30 @@ impl PruneList {
...
+							let height = bintree_postorder_height(*n);
+							if height == 0 {
+								// a leaf node can be pruned but will not be compacted
+								// so we do not shift for a leaf node
+								0
+							} else {
+								// height 1, 3 nodes, offset 2 = 1 + 1
+								// height 2, 7 nodes, offset 6 = 3 + 3
+								// height 3, 15 nodes, offset 14 = 7 + 7
+								2 * ((1 << height) - 1)
```
ignopeverell: I believe this returns zero for a height of zero, consider removing the `if`?

store/src/pmmr.rs
```diff
@@ -322,82 +312,102 @@ where
...
 }
+
+/// Filter remove list to exclude roots.
+/// We want to keep roots around so we have hashes for Merkle proofs.
+fn removed_excl_roots(removed: Vec<u64>) -> Vec<u64> {
+	removed
+		.iter()
+		.filter(|&pos| {
+			let (parent_pos, _, _) = family(*pos);
+			removed.contains(&parent_pos)
```
ignopeverell: I think `removed` is ordered, so `binary_search` will behave a lot better than `contains` here. As the `Vec` could grow large, it's better to keep an eye on O(n^2) operations.

antiochp: Yes it is - good catch on this. Will update.

core/src/core/pmmr.rs
antiochp: Oh nice. That feels better than having that (redundant) special case... 😀

antiochp:
Something is wrong with the way we are compacting the flat data file. I _think_ we're having trouble mapping between pos and leaf pos once we start removing nodes.
Need to investigate further.
Commented this out for now (we compact the hash file based on prune list, but data file is untouched).

This gets us to a node that works after being compacted _once_.
It will successfully accept and validate blocks propagated from another node.

But - something very wrong happens when we attempt to compact a 2nd time...
I think the file offsets are wrong on an already compacted node if we attempt to prune based on the prune list.

Will pick this up tomorrow and see if I can identify the issue with the 2nd compaction.

yeastplume:
It's entirely possible there's an off-by-one or something within the flat file position mapping or the like that's only showing up after pruning. Did you manage to put together a test that reproduces this? I can have a look as well.

antiochp:
Fixed a variety of problems.
The data file compaction is currently commented out. I want to advocate tackling this separately.
But this PR is now correctly compacting the hash file and appears to be running successfully locally.

api/src/handlers.rs
```diff
@@ -118,18 +118,35 @@ impl OutputHandler {
...
-			.filter(|output| commitments.is_empty() || commitments.contains(&output.commit))
-			.map(|output| {
-				OutputPrintable::from_output(output, w(&self.chain), &block, include_proof)
-			})
-			.collect();
-		BlockOutputs {
-			header: BlockHeaderInfo::from_header(&header),
-			outputs: outputs,
+
+		// TODO - possible to compact away blocks we care about
```
antiochp: If we compact blocks from the index before we have refreshed a coinbase output then we lose the ability to spend it.
But wallet code is handling this correctly now and does not blow up.
We just don't select these outputs as spendable.

To tackle in a separate PR - retrieve block and merkle proof as soon as we mine a block (don't wait to manually refresh the wallet).

chain/src/chain.rs
```diff
@@ -550,11 +550,28 @@ impl Chain {
...
-		let mut sumtrees = self.txhashset.write().unwrap();
-		sumtrees.compact()?;
+		{
+			let mut txhashes = self.txhashset.write().unwrap();
+
+			// compact the txhashset
+			txhashes.compact()?;
+
+			// print out useful debug info after compaction
+			txhashset::extending(&mut txhashes, |extension| {
```
antiochp: Added `dump()` here for the output MMR after every call to the the `/compact` API endpoint so we can see what it did.

chain/src/txhashset.rs
```diff
@@ -211,15 +211,24 @@ impl TxHashSet {
 
 	/// Compact the MMR data files and flush the rm logs
 	pub fn compact(&mut self) -> Result<(), Error> {
-		let horizon = global::cut_through_horizon();
 		let commit_index = self.commit_index.clone();
+		let head = commit_index.head()?;
+		let current_height = head.height;
+
+		// horizon for compacting is based on current_height
+		let horizon = (current_height as u32).saturating_sub(global::cut_through_horizon());
```
antiochp: We were using the `cut_through_horizon` effectively as a hardcoded block height.
We need to look at the current height and offset by `cut_through_horizon` blocks in the past.

core/src/core/pmmr.rs
```diff
@@ -371,10 +374,12 @@ where
 		// creation of another parent.
 		while bintree_postorder_height(pos + 1) > height {
 			let left_sibling = bintree_jump_left_sibling(pos);
-			let left_elem = self.backend
-				.get(left_sibling, false)
-				.expect("missing left sibling in tree, should not have been pruned");
-			current_hash = left_elem.0 + current_hash;
+
+			let left_hash = self.backend
+				.get_from_file(left_sibling)
```
antiochp: We need to bypass the rm_log here and go straight to the underlying file.
We may have "removed" the left_sibling but we still have access to the hash in the file.

core/src/core/pmmr.rs
```diff
@@ -421,14 +426,17 @@ where
 		// loop going up the tree, from node to parent, as long as we stay inside
 		// the tree.
 		let mut to_prune = vec![];
+
 		let mut current = position;
-		while current + 1 < self.last_pos {
+		while current + 1 <= self.last_pos {
 			let (parent, sibling, _) = family(current);
+
+			to_prune.push(current);
```
antiochp: Logic changed because we "remove" roots, but the hash is still available in the underlying file.
Push `current` to the prune list _before_ we check for the parent in the tree.

store/src/pmmr.rs
```diff
@@ -138,19 +138,27 @@ where
 		if self.rm_log.includes(position) {
 			return None;
 		}
+
 		// ... or in the prune list
-		let prune_shift = match self.pruned_nodes.get_leaf_shift(position) {
-			Some(shift) => shift,
-			None => return None,
-		};
+		// let prune_shift = match self.pruned_nodes.get_leaf_shift(position) {
```
antiochp: Data file pruning/compaction commented out for now.
Tackle this in a separate PR once we are comfortable that hash file compaction is working.

store/src/pmmr.rs
```diff
@@ -138,19 +138,27 @@ where
...
+		// 	None => return None,
+		// };
 
 		let hash_val = self.get_from_file(position);
 		if !include_data {
 			return hash_val.map(|hash| (hash, None));
 		}
 
+		// if this is not a leaf then we have no data
+		if !pmmr::is_leaf(position) {
```
antiochp: We were incorrectly returning "next leaf" data for non-leaf nodes because `n_leaves()` got smarter in an earlier PR.

store/src/pmmr.rs
```diff
@@ -138,19 +138,27 @@ where
...
+			return hash_val.map(|hash| (hash, None));
+		}
+
 		// Optionally read flatfile storage to get data element
```
antiochp: Data file compaction commented out for now - so shift is always 0 for now.
Will tackle in separate PR as discussed above.

store/src/pmmr.rs
```diff
@@ -324,82 +329,118 @@ where
...
-			}
-		});
-		self.data_file
-			.save_prune(tmp_prune_file_data.clone(), to_rm, record_len, prune_cb)?;
-
-		// 3. update the prune list and save it in place
-		for &(rm_pos, idx) in &self.rm_log.removed[..] {
-			if idx < cutoff_index {
-				self.pruned_nodes.add(rm_pos);
+		// 2. Save compact copy of the data file, skipping removed leaves.
```
antiochp: Data file compaction commented out for now.

store/src/pmmr.rs
```diff
@@ -324,82 +329,118 @@ where
...
 		)?;
 		self.hash_file = AppendOnlyFile::open(format!("{}/{}", self.data_dir, PMMR_HASH_FILE), 0)?;
 
-		// 5. and the same with the data file
-		fs::rename(
-			tmp_prune_file_data.clone(),
-			format!("{}/{}", self.data_dir, PMMR_DATA_FILE),
-		)?;
-		self.data_file = AppendOnlyFile::open(format!("{}/{}", self.data_dir, PMMR_DATA_FILE), 0)?;
+		// 5. Rename the compact copy of the data file and reopen it.
```
antiochp: Data file compaction commented out for now.

store/src/types.rs
```diff
@@ -162,41 +163,46 @@ impl AppendOnlyFile {
...
-					writer.write_all(&buf[buf_start..prune_at])?;
-				} else {
-					prune_cb(&buf[buf_start..prune_at]);
-				}
-				buf_start = prune_at + (prune_len as usize);
-				if prune_offs.len() > prune_pos + 1 {
-					prune_pos += 1;
-				} else {
-					break;
+		if prune_offs.is_empty() {
```
antiochp: We have legit cases where we do not prune anything and we need to handle this safely.

chain/src/txhashset.rs
ignopeverell: Oopsie, good catch.

core/src/core/pmmr.rs
```diff
@@ -290,20 +290,24 @@ where
...
 		let peaks_pos = peaks(self.last_pos);
-		let peaks: Vec<Option<(Hash, Option<T>)>> = peaks_pos
+		let peaks: Vec<Option<Hash>> = peaks_pos
 			.into_iter()
-			.map(|pi| self.backend.get(pi, false))
+			.map(|pi| {
+				// here we want to get from underlying hash file
+				// as the pos *may* have been "removed"
+				self.backend.get_from_file(pi)
+			})
```
ignopeverell: I don't follow this. A peak should never be pruned.

core/src/core/pmmr.rs
ignopeverell: That may be the same as my misunderstanding with the peaks, I don't see what tree construction would lead to this. Maybe we're missing a test in `prune` instead and overpruning?

ignopeverell:
Thanks for the debugging and the fixes, this is great. There's one bit I do not understand, where you need to check straight to the file because peaks could be pruned. Normally this line makes it impossible for pruning to go after a peak:

https://github.com/mimblewimble/grin/blob/bbbbdc455d8cebd868d7c1f090a183602c7382c8/core/src/core/pmmr.rs#L427

Am I missing something?

antiochp:
@ignopeverell I've been struggling to work out exactly what's going on here.
I think the changes in this PR are tweaking slightly how we iterate here.

The original behavior was to prune roots of subtrees (both in the underlying file and the prune list) and stop at the peak leaving it unpruned.

The modified behavior is to prune (via the prune list) roots of subtrees but leave them in the underlying file. And in doing so we now want to prune the peaks (via the prune list), but leave it in the underlying file.

So I think this is now traversing one level higher in the tree to get to the peak and the peak was inadvertently being compacted out of the file.
It wasn't clear to me if I had changed the behavior here or not - so I think this was a regression due to this PR with some unexpected side effects.

Note that we are now pushing the `current` to the prune list before this last_pos check (we were originally doing this after) so the peak ends up in the prune list.

antiochp:
Actually this might explain another strange edge case I think I was seeing.
In the original behavior - what did we do when a peak was also a leaf (in the case of a new output causing a new mountain (with a single node in it) to be appended to the MMR after a larger mountain?

I think we had an underlying issue there where the new peak/leaf could then be immediately spent; it was a peak so we could not prune it, but it was also a leaf so needed to be pruned.
This would not normally happen in practice the horizon would prevent this; but with the horizon being hardcoded I think we may have been seeing this within the first 20 blocks if the stars aligned and we got the timing really spot on when we hit the `/compact` endpoint.

The new behavior normalizes the way we treat peaks/roots/nodes - a peak is just like any other subtree root. It gets pruned via the prune list but does not get compacted away from the file _even_ if also a leaf.
Once a pruned node becomes a non-root we then also compact it away from the underlying file.
Peaks are always roots until a new peak is added above it so will never be compacted away until the mountain range itself creates a new higher mountain and the list of peaks is regenerated.

ignopeverell:
> what did we do when a peak was also a leaf

That was the reason for inserting coinbase outputs first in a block. Because they can't be spent for a while, so it guarantees that we'll never have pruned leaf peaks.

Thanks for the explanation on new behavior.

antiochp:
I think this is solid enough now to merge.
If I can get the tests to pass on travis and nobody has any objections I'm going to merge this today.

I'll open a separate issue to investigate the pruning/compaction of the data file.
And another issue to track putting `total_kernel_offset` in the block headers.

yeastplume:
No objections whatsoever, thanks for all of the testing and debugging of all of this.

PR #756, opened 2018-03-07
Basic TUI Integration

yeastplume:
In an effort to start making Grin a little more user friendly, I'm attempting to integrate a basic Text User Interface to present summary information adequate for most users (as opposed to just having to watch logs scroll by). To this end, I've started to integrate Cursive, a rust library built on ncurses:

https://github.com/gyscos/Cursive

Right now this is still experimental, testing how well this will work with our current logging and error handling (because ncurses captures stdout) as well as making a few augmentations to the code to allow for better real time stats.

The idea is to start simple, but insert the framework in such a way that it can be expanded upon as we go. Will start off just displaying easily parsed, useful information, but later on we could possibly allow users to modify configuration and parameters via the UI, implement a TUI wallet using the same approach, or display more advanced status on the chain and/or network.


ignopeverell:
Just noting, for future reference, the use of "user-friendly" and "text user interface" in the same sentence ;-)

yeastplume:
Sorry, where you see 'User-Friendly' it actually means 'mildly less confounding than scrolling logs'.

yeastplume:
This thing seems to be twitching nicely.. think I've worked out the UI integration, now just need to populate it in a sensible fashion. (Please note this isn't an endorsement of a logo choice, I just picked one that ASCIIfied nicely for an example).

grin-synching-1

grin-sync-2

yeastplume:
Right, this is going to be an ongoing effort obviously, but I think there's enough there to start sharing this around.

There's a new option in grin.toml called run_tui which will launch grin with the new UI if enabled. I've set this to false by default for now, so nothing should change for anyone unless they explicitly enable this.

Also note I removed the call to the internal cuckoo miner within grin/src/miner.rs as it should really go away shortly and was just needlessly complicating stat collection. Updated grin.toml appropriately

If no objections I can merge?

ignopeverell:
Push it!

yeastplume:
Pushed it real good

Issue #758, opened 2018-03-08
Kernel position index, is it worth it?


ignopeverell:
We seem to care (or I cared I guess) about duplicate kernels a lot. The kernel position index costs us an additional 40 bytes per kernel without accounting for data structure overhead and it's only useful to prevent miners from doing something stupid with their coinbase outputs. Which is mostly already covered by checked dups in the UTXO set.

Yes, we need some for rewind but that's only for the "last" one in a block and we can get rid of those after horizon. So how about getting rid of that check?

ignopeverell:
No one seemed to think it was really worth it. Removed in #759.

PR #759, opened 2018-03-08
Cleanup positional indexes for rewind, introducing block markers

### LINKED ISSUE #544: duplicate commitments and sumtree positions

ignopeverell:
Get rid of the kernel position index and introduces a new block marker index that keeps, for each block, the respective positions in the output and kernel MMRs. This is now sufficient for rewind.

A great side-effect is that we don't need to worry anymore about output and kernel ordering used for the PMMRs. The last indexes are just the last indexes.

Remaining: check whether markers could replace the pmmr file metadata.

Fixes #544 and #758


yeastplume:
Looks like a good cleanup. I think in general, it's desirable to store positions at a point in the code where we're as sure as we can be they're correct, then verify when we need them, as opposed to just derive them.

In the case of the actual file positions, we can derive them in theory but the PMMRFileMetadata was meant to store hard file positions to give us options to roll back to in the case of unexpected corruption (and perhaps make it easier to step through data, debug forks, etc without having to derive positions from whole chain state) Perhaps it's a premature optimization and markers will suffice, but still not sure

ignopeverell:
I guess the main question would be what failure scenario wouldn't be recoverable from positions but would be from file metadata?

PR #760, opened 2018-03-08
Rm log rewind on startup

yeastplume:
This calls rewind on a PMMR rm log on startup, using the current block height. Note I've just added a field to the PMMRFileMetadata structure and populate the height on startup to avoid passing another parameter around.

I'm not convinced that we don't need to store this by file position, but I'll comment about this a bit more in #759

### DIFF
--- a/store/src/types.rs
+++ b/store/src/types.rs
@@ -28,7 +28,6 @@ use libc::{ftruncate as ftruncate64, off_t as off64_t};
 
 use core::ser;
 
-
 /// Noop
 pub fn prune_noop(_pruned_data: &[u8]) {}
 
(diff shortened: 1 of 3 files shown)


store/src/types.rs
```diff
@@ -229,14 +230,18 @@ pub struct RemoveLog {
...
-		Ok(RemoveLog {
+		let mut rl = RemoveLog {
 			path: path,
 			removed: removed,
 			removed_tmp: vec![],
 			removed_bak: vec![],
-		})
+		};
+		if rewind_to_index > 0 {
+			rl.rewind(rewind_to_index)?;
```
ignopeverell: Shouldn't this flush to?

yeastplume: Yes. Fixed.

ignopeverell:
Push it too!

PR #761, opened 2018-03-08
Use header hash for by height index

aedesymen:
I would like some feedback on this.

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -146,11 +146,19 @@ impl ChainStore for ChainKVStore {
 
 	fn get_header_by_height(&self, height: u64) -> Result<BlockHeader, Error> {
 		option_to_not_found(self.db.get_ser(&u64_to_key(HEADER_HEIGHT_PREFIX, height)))
+			.and_then(|hash| self.get_block_header(&hash))
 	}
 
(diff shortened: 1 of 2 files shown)


chain/src/store.rs
```diff
@@ -146,11 +146,19 @@ impl ChainStore for ChainKVStore {
...
 		option_to_not_found(self.db.get_ser(&u64_to_key(HEADER_HEIGHT_PREFIX, height)))
+			.and_then(|hash| self.get_block_header(&hash))
 	}
 
 	fn save_header_height(&self, bh: &BlockHeader) -> Result<(), Error> {
-		self.db
-			.put_ser(&u64_to_key(HEADER_HEIGHT_PREFIX, bh.height), bh)
+		let batch = self.db
+			.batch()
+			.put_ser(
```
ignopeverell: The semantic of `save_header_height` is just to make the header accessible by height, in addition to hash. So I don't think you need to resave the header under its hash, it should already be there (and a separate call to `save_header` should be added otherwise).

ignopeverell:
Looking good overall, with a minor comment.

aedesymen:
@ignopeverell I fixed the semantics of save_header_height, but I don't know why the build failed

ignopeverell:
Just Travis acting up, I restarted the build, all good now. Thanks for the great PR!

PR #765, opened 2018-03-12
[WIP] Needs more TUI

yeastplume:
Just leaving this open now for further visibility. Updated the UI to allow switching between feature areas and added a rudimentary peer view:

screenshot from 2018-03-12 16-19-21


antiochp:
We should consider what it would take to move the current wallet functionality over to the TUI - it would be awesome to have info/outputs/send in there.

yeastplume:
Absolutely, yes.. now that the framework is in place I don't think it would be a big deal at all.. definitely have this planned, and will open an issue.

antiochp:
I can imagine a future where we can (at least for more focused testing) choose specific outputs to spend etc. (if we are keeping a wishlist)

ignopeverell:
Wouldn't a web UI make more sense for the wallet? Node operations are fairly technical in nature, but that's less true of wallets.

yeastplume:
Possible, yes but then you have to start thinking about tiers and javascript and bower and browser compatibility and ports... blah... the point of the TUI is it’s close to the existing code and can be run anywhere, even if you’re just grinning on a vps via ssh. 

I can see the value in doing a browser version in future, but can also see the value in doing a TUI version closer to the metal

ignopeverell:
Fair enough.

CarstenHouweling:
I've been thinking about a web UI for a while now and i would love to set something up.

For a standalone web UI, loading your wallet.dat file would be enough to show your coin info correct? What else would be a an option to add to a standalone web UI? 

Or would it be better to have it communicate directly with the running wallet? Making it dependent on running the grin wallet?

sesam:
Actually making it independent of the local wallet is much more
interesting. and hard.

ignopeverell:
How about filing a different issue for discussion?

PR #769, opened 2018-03-13
store "total kernel offset" on block header

antiochp:
Store `total_kernel_offset` (total accumulated offset) on BlockHeader
rather than the per-block `kernel_offset`.
This allows us to -
* trivially access the total accumulated kernel offset when validating the full txhashset.
* derive the per-block offset from the total kernel offset on a given block header and the total kernel offset of the previous block header

Note: This introduces a dependency on the previous block header when validating a block.
Interested in hearing feedback on this - is this a worthwhile tradeoff?

Related - #753 

If we are happy with this - I wonder if we want to do something similar with the `difficulty` vs `total_difficulty` on a block header?
If we can derive the `difficulty` from `total_difficulty` (and the previous block header) - maybe it makes sense to save 8 bytes per block header and only store the `total_difficulty`?

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -331,8 +331,11 @@ fn validate_block(
 	ctx: &mut BlockContext,
 	ext: &mut txhashset::Extension,
 ) -> Result<(), Error> {
-	// main isolated block validation, checks all commitment sums and sigs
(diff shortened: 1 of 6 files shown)


ignopeverell:
Yes, we should remove `difficulty`, it used to be calculated differently but now it's redundant.

PR #777, opened 2018-03-14
remove difficulty field from block_header (derive from total_difficulty)

### LINKED ISSUE #771: remove difficulty from block_header (redundant as we have total_difficulty)
See #769 for context.

We currently store both -
* difficulty
* total_difficulty

We can (I think) easily derive difficulty from total_difficulty (given the previous block header).

So we would save 8 bytes per block header if we removed the difficulty field.

antiochp:
Trying this out to see how disruptive removing `difficulty` from `block_header` might be.

TODO - 
- [x] grin_pow tests are not building
- [x] cleanup TODOs in the code
- [x] rework difficulty iterator so we hit the db less
- [x] rework the difficulty validation process in `pipe::validate_header()`
    -  [x] Check this header's difficulty is strictly greater than previous first.

Resolves #771.
Related #769 (kernel_offset vs. total_kernel_offset on block_header).

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -39,7 +39,7 @@ pub struct BlockContext {
 	/// The head
 	pub head: Tip,
 	/// The POW verification function
-	pub pow_verifier: fn(&BlockHeader, Option<BlockHeader>, u32) -> bool,
+	pub pow_verifier: fn(&BlockHeader, u32) -> bool,
 	/// MMR sum tree states
 	pub txhashset: Arc<RwLock<txhashset::TxHashSet>>,
 }
@@ -245,16 +245,9 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		return Err(Error::InvalidBlockTime);
 	}
 
-	// TODO - confirm genesis height is 1
-	let prev_header = if header.height > 1 {
-		Some(ctx.store.get_block_header(&header.previous)?)
-	} else {
-		None
-	};
-
 	if !ctx.opts.contains(Options::SKIP_POW) {
 		let n = global::sizeshift() as u32;
-		if !(ctx.pow_verifier)(header, prev_header, n) {
+		if !(ctx.pow_verifier)(header, n) {
 			error!(
 				LOGGER,
 				"pipe: validate_header failed for cuckoo shift size {}", n
@@ -279,6 +272,7 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		)),
 	}?;
 
+	// make sure this header has a height exactly one higher than the previous header
 	if header.height != prev.height + 1 {
 		return Err(Error::InvalidBlockHeight);
 	}
@@ -290,16 +284,26 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		return Err(Error::InvalidBlockTime);
 	}
 
+	// at this point we have a previous block header
+	// we know the height increased by one
+	// so now we can check the total_difficulty increase is also valid
+	// check the pow hash shows a difficulty at least as large
+	// as the target difficulty
+	if !ctx.opts.contains(Options::SKIP_POW) {
+		let target_difficulty = header.total_difficulty.clone() - prev.total_difficulty.clone();
+		if header.pow.clone().to_difficulty() < target_difficulty {
+			return Err(Error::DifficultyTooLow);
+		}
+	}
+
 	// verify the proof of work and related parameters
 	if !ctx.opts.contains(Options::SKIP_POW) {
-
 		// TODO - think this through...
 		// we have the following -
 		// total_difficulty (from this header)
 		// total_difficulty (from previous header)
-		// consensus::next_difficulty() (what the network says the diff should be based on previous header)
-		// anything else?
-
+		// consensus::next_difficulty() (what the network says the diff should be based
+		// on previous header) anything else?
 
 		let current_difficulty = header.total_difficulty.clone() - prev.total_difficulty.clone();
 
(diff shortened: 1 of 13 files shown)


chain/src/pipe.rs
```diff
@@ -245,9 +245,16 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		return Err(Error::InvalidBlockTime);
 	}
 
+	// TODO - confirm genesis height is 1
+	let prev_header = if header.height > 1 {
```
ignopeverell: Genesis height is 0.

ignopeverell: Also genesis doesn't go through validation, so I don't think this is needed.

chain/src/pipe.rs
```diff
@@ -245,9 +245,16 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
...
+		Some(ctx.store.get_block_header(&header.previous)?)
+	} else {
+		None
+	};
+
 	if !ctx.opts.contains(Options::SKIP_POW) {
 		let n = global::sizeshift() as u32;
-		if !(ctx.pow_verifier)(header, n) {
+		if !(ctx.pow_verifier)(header, prev_header, n) {
```
ignopeverell: I think it's better to have PoW verification not depend on any previous block and be self-sufficient. Then pull up the actual `pow > difficulty` test here. It also means you can validate the PoW before loading the previous block above.

chain/src/pipe.rs
```diff
@@ -283,43 +290,53 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
...
+
+		// TODO - think this through...
+		// we have the following -
+		// total_difficulty (from this header)
+		// total_difficulty (from previous header)
+		// consensus::next_difficulty() (what the network says the diff should be based on previous header)
+		// anything else?
+
+
+		let current_difficulty = header.total_difficulty.clone() - prev.total_difficulty.clone();
```
ignopeverell: Check this header's difficulty is strictly greater than previous first.

grin/src/miner.rs
```diff
@@ -209,7 +209,9 @@ impl Miner {
 					proof_diff.into_num(),
 					difficulty.into_num()
 				);
-				if proof_diff >= b.header.difficulty {
+				if proof_diff >=
```
ignopeverell: Should be strictly greater.

chain/src/store.rs
```diff
@@ -276,7 +276,15 @@ impl Iterator for DifficultyIter {
 			Err(_) => None,
 			Ok(bh) => {
 				self.next = bh.previous;
-				Some(Ok((bh.timestamp.to_timespec().sec as u64, bh.difficulty)))
+
+				// TODO - error handling, base case for the loop, maybe look at height?
+				let difficulty = if let Ok(prev_header) = self.store.get_block_header(&bh.previous) {
```
ignopeverell: You should be able to use the previous you just loaded in the next iteration?

grin/src/miner.rs
```diff
@@ -327,9 +329,11 @@ impl Miner {
 					"Found cuckoo solution for nonce {} of difficulty {} (difficulty target {})",
 					b.header.nonce,
 					proof_diff.into_num(),
-					b.header.difficulty.into_num()
+					b.header.total_difficulty.into_num()
 				);
-				if proof_diff >= b.header.difficulty {
+				if proof_diff >=
```
ignopeverell: Strictly greater too.

grin/src/miner.rs
```diff
@@ -438,7 +442,9 @@ impl Miner {
 			let pow_hash = b.hash();
 			if let Ok(proof) = miner.mine(&pow_hash[..]) {
 				let proof_diff = proof.clone().to_difficulty();
-				if proof_diff >= b.header.difficulty {
+				if proof_diff >=
```
ignopeverell: And... again.

pow/src/lib.rs
```diff
@@ -76,23 +76,22 @@ pub trait MiningWorker {
...
 /// Validates the proof of work of a given header, and that the proof of work
 /// satisfies the requirements of the header.
-pub fn verify_size(bh: &BlockHeader, cuckoo_sz: u32) -> bool {
-	// make sure the pow hash shows a difficulty at least as large as the target
-	// difficulty
-	if bh.difficulty > bh.pow.clone().to_difficulty() {
+pub fn verify_size(bh: &BlockHeader, prev_bh: Option<BlockHeader>, cuckoo_sz: u32) -> bool {
+	// make sure the pow hash shows a difficulty at least as large
+	// as the target difficulty
+	let difficulty = if let Some(prev_bh) = prev_bh {
```
ignopeverell: That's the tests I would move up to `pipe.rs`, leaving `verify` to only check the validity of the Cucko Cycle.

chain/src/pipe.rs
antiochp: I think I follow what you're saying. Let me rework this and get back to you.

pow/src/lib.rs
antiochp: Ok - yeah that makes sense.

chain/src/store.rs
antiochp: But its the _next_ iteration - we haven't loaded it yet?

tromp:
I'm fine with its removal.
One could argue that it's convenient to have a PoW check self-contained within a single header,
but deriving it from consecutive headers is only a minor inconvenience.

antiochp:
@ignopeverell This should be good to go now.

yeastplume:
Just note this will break async mining mode as the header format needs to be updated, I can address this at the same time as doing https://github.com/mimblewimble/grin/issues/780 is we decide to go ahead with it.

antiochp:
Ahh good catch - we don't have any type safety around that? 
I guess the other end just deserializes the bytes.

antiochp:
@yeastplume is this a blocker for merging this PR in?

yeastplume:
it’s fine, just merge it in and I’ll do the fixes next

PR #779, opened 2018-03-14
A minimal stratum server implementation for grin

bladedoyle:
added stratum-style server 
integrated with configuration file
provided reference implementation of a stratum-miner client ( will update it to use PluginMiner in the next pass )


grin/src/miner.rs
```diff
@@ -137,14 +139,107 @@ impl Miner {
...
+				info!(LOGGER, "Sent block");
+			}
+			Err(e) => {
+				warn!(LOGGER, "Error in connection with stratum client: {}", e);
+				*stream_err = true;
+				return;
+			}
+		}
+	} // End send_block
+
```
yeastplume: If possible, I think it would be best to put as much of the stratum-server related code as possible into a separate file to keep them distinct. miner.rs is very messy already (for reasons I'll outline in another issue) and needs cleaning up.

grin/src/miner.rs
```diff
@@ -527,22 +621,182 @@ impl Miner {
...
+								b.header.height
+							);
+
+							// Push the new block "b" to the connected client
+							self.stratum_send_block(&mut stream, &b, &mut stream_err);
+						}
+
+						// Get a solved block header - if any are waiting
+						// Here, we are expecting real solutions at the full difficulty.
+						match self.stratum_get_solution(&mut stream, &mut stream_err) {
```
yeastplume: I know it's WIP, but we definitely need to be getting back solutions of any difficulty, not just the full, so that a client can prove work.

grin/src/miner.rs
```diff
@@ -527,22 +621,182 @@ impl Miner {
...
+							            * same block */
+						} // checking for solution
+		// sleep before restarting loop
+						thread::sleep(Duration::from_millis(500));
+					} // loop forever
+				} // end OK() match stream
+			} // match stream
+		} // for stream incoming
+	} // fn run_stratum_server_loop()
+
```
yeastplume: Would try to split up the functionality here a bit into separate functions, but understand it's still evolving.

grin.toml
```diff
@@ -105,17 +102,21 @@ cuckoo_miner_async_mode = false
 
 #cuckoo_miner_plugin_dir = "target/debug/plugins"
 
+#flag whether stratum server is enabled
+enable_stratum_server = true
```
yeastplume: This is 'advanced functionality', we just need to make sure this is disabled by default

src/bin/stratum-miner.rs
```diff
@@ -0,0 +1,343 @@
+// Copyright 2018 The Grin Developers
```
yeastplume: I actually would really like to see this as as a binary for the (cuckoo-miner) [https://github.com/mimblewimble/cuckoo-miner] module rather than into Grin... this would allow people just interested in mining to just download, understand and run a much thinner mining client, as well as give us an 'official' and supported pooled mining client, something that most coins don't have. It would mean replicating some of the configuration into cuckoo-miner, but I think it would be worth it.

src/bin/stratum-miner.rs
```diff
@@ -0,0 +1,343 @@
...
+    debug_output_id,
+    cuckoo_size,
+    b.header.difficulty,
+    b.header.height,
+    b.header.nonce
+  );
+
+	loop {
+		// Do mining for a bit
+		for _ in 1..42 {
```
yeastplume: This should all be using cuckoo-miner in async mode, which will just run on another thread and report when solutions are found, and you can send cancel messages to clients when a block has been found or time has expired.

grin/src/server.rs
```diff
@@ -223,10 +227,31 @@ impl Server {
...
+					thread::sleep(secs_5);
+				}
+				miner.run_stratum_server_loop(config.clone(), cuckoo_size as u32, proof_size);
 			});
 	}
 
+
 	/// The chain head
 	pub fn head(&self) -> chain::Tip {
 		self.chain.head().unwrap()
```
yeastplume: Would definitely need some integration tests to exercise this once the functionality's a bit further along.

yeastplume:
Hi @bladedoyle, thanks very much for submitting this and it all looks to be moving in the right direction. I'd be really happy to have some pooled mining functionality built in as well as a mining client that supports pooling functionality. Usually you have to go scouring nasty places on the internet to find appropriate pooled mining clients, so I'd really like to be in a position to have 'official ones'

Some comments above (and some are nit-picks as this is a WIP,) but overall:

* I'm a bit concerned that inserting a stratum mining server directly into Grin (yet another port and listener) is a bit 'and the kitchen-sink-y', but if the code is properly compartmentalised and none of the other developers have an issue with it, then I think it should be okay. 
* I would definitely move the client binary directly into the cuckoo miner module, for reasons stated above (it will make everything much more flexible)

Other than that, I won't get into too much more detail because WIP.

Thanks again, and please feel free to contact me with questions at any time, especially around using cuckoo-miner (and if there are any changes required there to support this)

src/bin/stratum-miner.rs
antiochp: I like that idea. :+1:

bladedoyle:
Thanks for the feedback, @yeastplume.

The lightest-weight solution would be to add 2 API to the existing server - for "get_block" and "submit_solution".  The down-side is that a client would need to poll over http for new blocks, which I think would not work well especially for a short block interval.  The other extreme would be to implement a full featured stratum server within grin.
My solution proposes a middle ground - a tcp listener for a single client connection and methods to push data in both directions.  Not a full "stratum" implementation.  I was thinking there would always be "pool" or "proxy" service running between the external miner(s) and this grin "stratum-like server".  I intend to write a (or adapt an existing) pool as the next step.  That pool would implement a more complete stratum protocol (including authorization, job difficulty adjustment, standard error codes).  That pool would be a separate project from grin itself.

I agree that the miner binary does not belong directly in the grin project.  Moving it to the cuckoo miner project sounds right.  

Let me take another pass and then request a lower level review.

ignopeverell:
I wouldn't be opposed to a minimal stratum-like server embedded in grin, enabled with appropriate configuration.

No matter what, we need a way to get a block header that's ready for mining out of the grin process. Even outside of pool mining, we can't expect people to run a grin process on each mining rig. Bitcoin Core and friends have `getblock` and `getblocktemplate`, Ethereum has a `getWork` RPC, but very few miners use it directly, they usually have a stratum bridge in between. So our way to get a block header ready for mining might as well directly be stratum.

bladedoyle:
@ignopeverell - 

Bitcoin, Monero, and friends implement "getblocktemplate" and "submitblock" as RPC api.

I dont like polling for new blocks because its inefficient in both directions:  The pool needs to call getblocktemplate RPC frequently or waste time mining old blocks, and the server needs to process each call and keep sending block templates on each request even if the pool already has the latest (though I guess the getblocktemplate rpc call could take a height parameter and at least avoid sending the same block data on each call).

I would prefer a persistent tcp connection between the pool and grin server allowing new blocks to be pushed immediately.

yeastplume:
@bladedoyle Thanks for the update and doc. This sounds great, and I really look forward to seeing your results on this. As always please don't hesitate to contact me with any questions and keep us posted on your progress

ignopeverell:
What's the status on this?

bladedoyle:
Good question. 

I decided that it would be nice to have the same stratum API for all components.  Most important to be able to connect a miner directly to grin without needing a pool or proxy, but also because its cleaner to document and maintain, etclll   I am currently working on a minimal implementation of that interface, using JSONRpc.  I also spent a little time studying the pluginminer and trying to understand exactly which part(s) of the block header is needed by the async miner.

grin.toml
bladedoyle: Agreed, will make sure its disabled by default in final checkin

sesam:
Thanks @bladedoyle for this. If there's anything you need like info or a helping hand, or just brainstorming ideas, we're here for you, as well as on https://gitter.im/grin_community/dev

grin/src/miner.rs
```diff
@@ -527,22 +510,45 @@ impl Miner {
...
-					"setting pubkey in miner to pubkey from block_fees - {:?}",
-					block_fees
+					"setting pubkey in miner to pubkey from block_fees - {:?}", block_fees
 				);
 				key_id = block_fees.key_id();
 			}
 		}
 	}
 
+        // Ensure a block is built and returned
```
ignopeverell: I think this should be externalized from the miner into its own thing as it's a generally useful facility. Just add the chain and pool (needed by `build_block`) as an argument and put it into a `mine_block.rs` or something similar.

grin/src/stratumserver.rs
```diff
@@ -0,0 +1,518 @@
...
+		}
+	}
+} // impl Worker
+
+// ----------------------------------------
+// Grin Stratum Server
+
+pub struct StratumServer {
+	id: String,
+	miner: Miner,
```
ignopeverell: The stratum server shouldn't need the miner, just the pool, the chain and the utility to build a block. We should be able at some point, if we decide to, to remove the embedded miner and keep the stratum server.

grin/src/stratumserver.rs
```diff
@@ -0,0 +1,518 @@
+// Copyright 2018 The Grin Developers
```
ignopeverell: Not 100% sure about this but maybe we should have a separate mining crate. We could have the stratum server, the existing miner, and the block building facility in there. @yeastplume what do you think?

grin/tests/simulnet.rs
```diff
@@ -77,6 +84,148 @@ fn basic_genesis_mine() {
 	pool.run_all_servers();
 }
 
+// Create a grin server, and a stratum server.  Simulate a few JSONRpc requests
+// and verify the results.  Validate disconnected workers and broadcasting new
+// jobs.
+#[test]
+fn basic_stratum_server() {
```
ignopeverell: Can you break this into its own test file? The simulnet one is getting a little abused already.

grin/src/miner.rs
```diff
@@ -115,8 +114,8 @@ impl ser::Writer for HeaderPartWriter {
 
 pub struct Miner {
 	config: MinerConfig,
-	chain: Arc<chain::Chain>,
-	tx_pool: Arc<RwLock<pool::TransactionPool<PoolToChainAdapter>>>,
+	pub chain: Arc<chain::Chain>,
```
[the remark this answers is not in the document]
bladedoyle: Sure.
I merged my changes with testnet2 and created a new pull request that now includes these changes:
https://github.com/mimblewimble/grin/pull/937

grin/src/miner.rs
bladedoyle: I think I understood this request, and made an attempt at it.
Again, please see the changes in the new testnet2 pull request:
https://github.com/mimblewimble/grin/pull/937

grin/tests/simulnet.rs
bladedoyle: Ya, I made this change as part of https://github.com/mimblewimble/grin/pull/937

bladedoyle:
I merged these changes into testnet2 and created a new pull request to continue this work:
https://github.com/mimblewimble/grin/pull/937

grin/src/stratumserver.rs
yeastplume: Perhaps we can extend the `pow` crate that exists at the moment to contain all of this? At the moment, what's actually contained in `pow` is looking a little thin to justify its own crate... however 'pow' is a bit conceptually distinct from 'mining' so I'm not sure.

Issue #770, opened 2018-03-13
we are not currently pruning/compacting the data file


antiochp:
This functionality was commented out and disabled as part of #753.
We _are_ compacting the hash file but the data file is left un-compacted.
Now that #753 is merged in we should investigate and get this working again.

https://github.com/mimblewimble/grin/blob/65633c7611c07fc2467e352dab5367789f0a05bd/store/src/pmmr.rs#L375-L396

https://github.com/mimblewimble/grin/blob/65633c7611c07fc2467e352dab5367789f0a05bd/store/src/pmmr.rs#L417-L423

yeastplume:
looking into this ATM

PR #784, opened 2018-03-15
PMMR data compaction

yeastplume:
Hopefully this fixes https://github.com/mimblewimble/grin/issues/770 and compacts data files properly again. The tests pass now, data is where it's expected to be and the files are the right length. The main thing was really just to use n_leaves for getting the data as opposed to just leaf shift, so that leaf indices are mapped to flat-file positions.

@antiochp do you see any hidden issues with this as a result of all your recent investigations?

### DIFF
--- a/store/src/pmmr.rs
+++ b/store/src/pmmr.rs
@@ -342,8 +342,7 @@ where
 
 		// Paths for tmp hash and data files.
 		let tmp_prune_file_hash = format!("{}/{}.hashprune", self.data_dir, PMMR_HASH_FILE);
-		let tmp_prune_file_data = format!("{}/{}.dataprune", self.data_dir,
-			PMMR_DATA_FILE);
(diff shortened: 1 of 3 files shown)


antiochp:
No that looks good. That makes a lot of sense based on my investigation earlier.

I'm checking this out locally and giving it a try.

store/src/pmmr.rs
```diff
@@ -373,27 +373,29 @@ where
...
+
+			let off_to_rm = leaf_pos_to_rm
+				.iter()
+				.map(|pos| {
+					let shift = self.pruned_nodes.get_leaf_shift(*pos);
+					(pmmr::n_leaves(pos - shift.unwrap()) - 1) * record_len
+				})
+				.collect::<Vec<_>>();
+
+			println!(
```
antiochp: println

antiochp:
:+1: looks good from here.
I can compact both the mining node and the non-mining node locally and everything is running smoothly.

yeastplume:
Excellent, thanks for testing! Will merge once all tests pass.

PR #786, opened 2018-03-15
Fix test dirs to isolate the tests

antiochp:


antiochp:
Merging as we've seen this causes failures a couple of time today on other PRs.

PR #787, opened 2018-03-15
TUI/Cursive updates

yeastplume:
The author @gyscos has graciously and responsively made a few fixes to cursive based on feedback, but it makes his master branch a bit of a moving target. Temporarily tying to my own fork to avoid surprise build issues.


yeastplume:
Sick of endlessly restarting the travis build.. if this breaks something I'll fix it :angry:

PR #788, opened 2018-03-15
hash with the pos for non-leaf nodes

### LINKED ISSUE #778: tree hashing - hash_with_index for non-leaf nodes of the MMR
Currently we include the `pos` when hashing leaf nodes. But we do not use the `pos` for non-leaf nodes.

For leaf nodes we construct the hash via - 
```
let mut current_hash = elmt.hash_with_index(elmt_pos);
```

For the non-leaf nodes we construct the hash via - 
```
current_hash = left_hash + current_hash
```

I _think_ we want to consider including the `pos` when generating the hash for non-leaf nodes as well.

Something roughly like - 
```
current_hash = (left_hash, current_hash).hash_with_index(pos);
```

__But__ - I feel like I don't have a good handle on the implications of not doing this (vs. doing this, vs. doing something else). 

Some links here on prior work on "tree hashing modes" and "standard" ways of doing this -
https://www.grin-forum.org/t/tree-hashing-mode/270

antiochp:
When hashing two nodes together in the MMR to generate the hash for the parent use `hash_with_index()` and pass in the `parent_pos`.

This makes the hashing consistent across leaf and non-leaf nodes.

Add `hash_with_index()` to `(Hash, Hash)` tuple for convenience (via `PMMRIndexHashable` trait).

We now need the pos for each path element in the Merkle proof, so rework it to store these. We no longer need the left_right flag for each path element as we can derive this based on the pos.

Resolves #778.

This was a bigger piece of work than originally assumed as I had not realized the impact on the Merkle proofs. But the end result is better I think.

### DIFF
--- a/core/src/core/pmmr.rs
+++ b/core/src/core/pmmr.rs
@@ -139,7 +139,6 @@ impl Readable for MerkleProof {
 			peaks.push(Hash::read(reader)?);
 		}
 
-		// Should we use a tuple here for (path, path_pos, left_right)?
 		let mut path = Vec::with_capacity(path_len as usize);
 		for _ in 0..path_len {
 			let hash = Hash::read(reader)?;
--- a/core/src/core/pmmr.rs
+++ b/core/src/core/pmmr.rs
@@ -885,6 +885,7 @@ pub fn family(pos: u64) -> (u64, u64) {
 	}
 }
 
+/// Is the node at this pos the "left" sibling of its parent?
 pub fn is_left_sibling(pos: u64) -> bool {
 	let (_, sibling_pos) = family(pos);
 	sibling_pos > pos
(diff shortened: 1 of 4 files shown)


antiochp:
Going to merge. 
If it turns out we need to iterate on this based on #778 then we can revisit.
But this gets us to the point where every node (except the root itself) is hashed with the pos in the MMR (and tidies up what we store in the Merkle proof).

One thought (if/when we do revisit this) -
Pretty sure we can avoid storing the pos for each node in the path in the Merkle proof.
For a given MMR size the pos are all known ahead of time.
I _think_ if we know the pos of the node we care about and the (unpruned) size of the MMR then we have everything we need to determine the pos of each sibling in the path.

We would presumably do something like - 
* for a given node pos - 
  * find the parent and the sibling pos (which we know how to do, via `family()`)
  * repeat

PR #793, opened 2018-03-16
Improve bounding block timestamp

### LINKED ISSUE #783: Grin uses deprecated `time` crate
For example in BlockHeader:

```
	/// Timestamp at which the block was built.
	pub timestamp: time::Tm,
```

Which may be ok per se, unfortunately I found a bug during fuzz tests (at least on OSX):
```
thread '<unnamed>' panicked at 'gmtime_r failed: Value too large to be stored in data type (os error 84)', /Users/alex/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.1.39/src/sys.rs:273:17
...
 #13 0x10ae67fdf in time::sys::inner::time_to_utc_tm::haccd130fa42a2309 sys.rs:273
    #14 0x10ae3fd85 in time::at_utc::h2475696bedb92f2a lib.rs:404
    #15 0x10a52c43a in _$LT$grin_core..core..block..BlockHeader$u20$as$u20$grin_core..ser..Readable$GT$::read::_$u7b$$u7b$closure$u7d$$u7d$::h7685abd8874970d6 block.rs:192
...
```

This code reproduces this bug:

```
extern crate time;

fn main() {
	let _t =  time::at_utc(time::Timespec {
				sec: std::i64::MAX,
				nsec: 0,
	});
}
```
It looks gross, type system is clearly misused here. Btw libc docs sound strange:
``` 
In the GNU C library and other POSIX-compliant implementations, time_t is equivalent to long int. 
...
In the GNU C Library, time_t is equivalent to long int. In other systems, time_t might be either an integer or floating-point type. 
```

Bottom line - should we migrate to `chrono`?

hashmap:
Fixes #783.
* 2^60 still fails on OSX, experimentally found safe max 2^55
* Handle negative values

I tried to use .abs(), unfortuantely fuzz test kills it with
`panicked at 'attempt to negate with overflow', /Users/travis/build/rust-lang/rust/src/libcore/num/mod.rs:1146:17`

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -181,7 +181,7 @@ impl Readable for BlockHeader {
 		let total_kernel_offset = BlindingFactor::read(reader)?;
 		let pow = Proof::read(reader)?;
 
-		if timestamp > (1 << 60) {
+		if timestamp > (1 << 55) ||  timestamp < -(1 << 55) {
 			return Err(ser::Error::CorruptedData);
 		}
 


ignopeverell:
Thanks for trying out different values.

Issue #780, opened 2018-03-15
PoW hash suggestions


tromp:
currently the nonce and pow are in the middle of the header.
it would look neater to have them at the end, since they must causally follow the other fields.
this also allow a slight simplification in miner/verifier code in that it can create the siphash key by
hashing a prefix of the full header, rather than having to elide a field from the middle.

note that there are two kinds of header hashes:
the pre-pow hash used as siphash key to define the cuckoo graph, and
the post-pow hash that must satisfy the difficulty target.

currently, the post-pow hash is a hash of the whole header including the pow.
i would prefer to have it be just the pow hash (or what i call cyclehash in my solvers),
since the cycle already commits to the header.
so then the post-pow hash would be blake2(cuckoo_cycle(blake2(header))),
which can be compared to bitcoin's   sha256(                         sha256(header))
it's like we have another hash function sandwiched in between two blake2 iterations.

yeastplume:
I'm fine with this and have no problem implementing if everyone is in agreement.. it would actually remove awkwardness from miner.rs and within cuckoo-miner (both will need to be updated).

I'd wait until @antiochp is finished his current work on removing diff from the header.

ignopeverell:
Agreed about having them at the end. I think it was originally that way actually, but then stuff got removed and stuff got added.

I think I agree with the post-pow hash as well.

yeastplume:
Done

PR #794, opened 2018-03-16
Move nonce/pow to end of block header, change post-pow hash

yeastplume:
Changes as per https://github.com/mimblewimble/grin/issues/780

 - [x] Block header ordering updated
 - [x] Update call to cuckoo miner async mode
 - [x] Update post-pow hash to blake2(cuckoo_cycle(blake2(pre-pow))),
 - [x] Check/update all tests

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -105,22 +105,22 @@ pub struct BlockHeader {
 	pub previous: Hash,
 	/// Timestamp at which the block was built.
 	pub timestamp: time::Tm,
+	/// Total accumulated difficulty since genesis block
+	pub total_difficulty: Difficulty,
 	/// Merklish root of all the commitments in the TxHashSet
 	pub output_root: Hash,
 	/// Merklish root of all range proofs in the TxHashSet
 	pub range_proof_root: Hash,
 	/// Merklish root of all transaction kernels in the TxHashSet
 	pub kernel_root: Hash,
-	/// Nonce increment used to mine this block.
-	pub nonce: u64,
-	/// Proof of work data.
-	pub pow: Proof,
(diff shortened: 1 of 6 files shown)


yeastplume:
Tests pass locally, now to refresh Travis until it feels like accepting them.

Ready for review anyhow.. it's always a bit awkward to try to serialise objects in different ways depending on the context, but I think a specific pre_pow hash function for the block header makes sense. 

Also note that since we're essentially using the header pow hash as the block hash, we need a way to populate the pow randomly during tests that don't care about the POW to prevent duplicate blocks. I considered changing the default BH struct to use a random POW, but I don't think that's as clean.

core/src/core/block.rs
```diff
@@ -135,33 +135,21 @@ impl Default for BlockHeader {
...
-			[write_fixed_bytes, &self.kernel_root]
-		);
-
-		try!(writer.write_u64(self.nonce));
-		try!(self.total_difficulty.write(writer));
-		try!(self.total_kernel_offset.write(writer));
-
-		if writer.serialization_mode() != ser::SerializationMode::Hash {
+		if writer.serialization_mode() == ser::SerializationMode::Hash {
 			try!(self.pow.write(writer));
```
antiochp: From eavesdropping on prior discussions it sounds the `pow` effectively encodes everything we care about - so we can just hash this (which is presumably faster)?

So effectively `block.hash() == header.hash() == pow.hash()`? Which has a nice feel to it.

yeastplume: Yes ... all seems to slot nicely in place with this approach, and everything commits to exactly as much as it needs to and no more

PR #800, opened 2018-03-16
Fix limits for Transaction

hashmap:
In prev pr I set them too high, also typos in comments were fixed.


ignopeverell:
I think we can raise the number of inputs to a couple thousands at least (but not outputs). I can do that directly, very minor.

PR #805, opened 2018-03-18
TUI Difficulty stats for past few blocks

yeastplume:
Aims:
* Add another sub-tab in mining to show the average time and difficulty of last WINDOW blocks
* Show last WINDOW block times in a table
* Also try to ensure panics from other threads aren't written to stderr when the server panics so as not to overwrite the TUI view


yeastplume:
Should be done once Travis passes:

screenshot from 2018-03-19 15-26-31

ignopeverell:
Snazzy.

PR #808, opened 2018-03-19
rewind to header as part of txhashset validation

antiochp:
Otherwise we risk including a new block and the roots do not match.

I saw `compact()` fail a couple of times during `validation()` and I think (but not 100% sure) we processed a new block between the first 2 lines (so the extension was actually at a different pos to the header we are validating against) - 
https://github.com/mimblewimble/grin/blob/7c3994e5219ea8ca5ba1b744f650ff18df26fe3b/chain/src/chain.rs#L416-L418

### DIFF
--- a/chain/src/txhashset.rs
+++ b/chain/src/txhashset.rs
@@ -550,8 +550,13 @@ impl<'a> Extension<'a> {
 		}
 	}
 
-	/// Validate the current txhashset state against a block header
-	pub fn validate(&self, header: &BlockHeader) -> Result<(), Error> {
+	/// Validate the txhashset state against the provided block header.
+	/// Rewinds to that pos for the header first so we see a consistent
+	/// view of the world.
+	pub fn validate(&mut self, header: &BlockHeader) -> Result<(), Error> {
+		// first rewind to the provided header
+		&self.rewind(header)?;
+
 		// validate all hashes and sums within the trees
 		if let Err(e) = self.output_pmmr.validate() {
 			return Err(Error::InvalidTxHashSet(e));
--- a/store/src/types.rs
+++ b/store/src/types.rs
@@ -250,21 +250,21 @@ impl RemoveLog {
 		Ok(rl)
 	}
 
-	/// Truncate and empties the remove log.
-	pub fn rewind(&mut self, last_offs: u32) -> io::Result<()> {
+	/// Rewinds the remove log back to the provided index.
+	/// We keep everything in the rm_log from that index and earlier.
+	/// In practice the index is a block height, so we rewind back to that block
+	/// keeping everything in the rm_log up to and including that block.
+	pub fn rewind(&mut self, idx: u32) -> io::Result<()> {
 		// simplifying assumption: we always remove older than what's in tmp
 		self.removed_tmp = vec![];
 		// backing it up before truncating
 		self.removed_bak = self.removed.clone();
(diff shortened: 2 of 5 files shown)


antiochp:
Think I found a bug in how we rewind the rm_log.

If we are at block at height 5 and we rewind to the pos for that block then nothing should happen, the rewind should be a no-op.
But - it looks like rewinding the rm_log actually rewinds the rm_log to the _previous_ block height index here - 
https://github.com/mimblewimble/grin/blob/7c3994e5219ea8ca5ba1b744f650ff18df26fe3b/store/src/types.rs#L265

This is causing `mine_simple_chain` tests to fail if we attempt to rewind during txhashset validation.

antiochp:
That should tighten things up quite a bit.
If `rm_log.rewind()` was susceptible to incorrectly dealing with things removed in the block being rewinded to then this was probably causing some subtle problems when dealing with forks (and probably not all the time because our blocks tend to be empty).

chain/src/txhashset.rs
```diff
@@ -550,8 +550,13 @@ impl<'a> Extension<'a> {
...
 	}
 
-	/// Validate the current txhashset state against a block header
-	pub fn validate(&self, header: &BlockHeader) -> Result<(), Error> {
+	/// Validate the txhashset state against the provided block header.
+	/// Rewinds to that pos for the header first so we see a consistent
+	/// view of the world.
+	pub fn validate(&mut self, header: &BlockHeader) -> Result<(), Error> {
+		// first rewind to the provided header
+		&self.rewind(header)?;
```
ignopeverell: Good catch. As `validate` is mostly used during a fast sync, this was actually done here for that specific case:

https://github.com/mimblewimble/grin/blob/master/chain/src/chain.rs#L521

Speaking of which, doesn't this break `validate` for fast sync then? Because there shouldn't be any block marker for headers yet. We could have `txhashset_write` in `Chain` save a block marker prior to validation then?

store/src/types.rs
```diff
@@ -250,21 +250,21 @@ impl RemoveLog {
...
+		if idx == 0 {
 			self.removed = vec![];
 		} else {
-			self.removed = self.removed
-				.iter()
-				.filter(|&&(_, idx)| idx < last_offs)
-				.map(|x| *x)
-				.collect();
+			// retain rm_log entries up to and including those at the provided index
+			self.removed.retain(|&(_, x)| x <= idx);
```
ignopeverell: Nice!

chain/src/txhashset.rs
antiochp: Hmm good point - didn't consider fast sync in this...
Let me take a look and see what's involved in saving the block marker earlier.

store/src/types.rs
antiochp: The more I work with Rust the more I like it. 😀 
It has some very clean high level functions in the std libs.

antiochp: (You should have seen the amount of debug logging going on before I realized the rm_log looked different before/after here...)

antiochp:
Added a call to `save_block_marker()` early on in `txhashset_write` so we know what to rewind to later on.
No longer need to call `rewind_pos()` explicitly, the call to `rewind()` in `validate` does the job now.

Going to merge - all tests are passing and things seem good running locally.

PR #813, opened 2018-03-19
Generate Cargo.lock and remove it from .gitignore.

clefru:
As grin is not a library but a binary, the [Cargo FAQ](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries) suggests putting Cargo.lock into version control. Also in the interest of security (as one day we will store the world's riches in grin wallets), we should not depend on arbitrary un-checksummed other repos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 .grin*
 node*
 target
-Cargo.lock
 **.iml
 grin.log
 wallet.seed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,0 +1,2361 @@
+[[package]]
+name = "adler32"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "aho-corasick"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "aho-corasick"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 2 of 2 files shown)


ignopeverell:
Soon. The plan was to do without `cargo.lock` until testnet2 is out, which should hopefully be this week. Before then, we still had too many regular changes in dependencies.

clefru:
Understood & thanks for the quick response.

Wouldn't it be possible to check in a change to Cargo.lock with the change to Cargo.toml? It would clutter the diff a bit, yes, but in theory not a big deal. My motivation is that my distro [makes me jump through hops](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/rust/fetchcargo.nix#L11) in packaging without Cargo.lock in git, but that's my own problem.

ignopeverell:
I see. It's definitely possible, up to recently it would have just added friction to what should otherwise be considered unreleased software (some dependency changes actually impact quite a few entries in `Cargo.lock`). Can you give it a few more days?

clefru:
> Can you give it a few more days?

Of course & thanks for your work!

sesam:
This is a good idea! We already had a first (?) [example of a user just checking out master and getting bit by an unexpected disruptive upgrade of the "slog" crate](https://gitter.im/grin_community/Lobby?at=5ab7850827c509a7749fb503)

The guilt lies on slog for breaking the SemVer rules that cargo crates are supposed to adhere to - quite possibly by mistake.

But if we would be shipping a versioned Cargo.lock then a single person can handle cargo update + checking in after verifying that build and tests still pass.

ignopeverell:
Preparing a PR with version updates and `Cargo.lock` for `testnet2`. I'd prefer a single PR for that transition. And thanks for the proposal in the first place!

PR #815, opened 2018-03-19
Change wallet api_listen_port format

quentinlesceller:
Small PR to standardize the port format.


ignopeverell:
Okay, whatever Travis says about tests, this PR is obviously not the problem, so merging.

PR #823, opened 2018-03-20
split out verify_rangeproofs from sum_outputs

antiochp:
* split out verify_rangeproofs from sum_outputs
* added some timing to the debug msgs - 

```
Mar 20 14:00:25.036 DEBG Validated, summed (and offset) 4120 kernels, pmmr size 8237, took 4s
Mar 20 14:00:25.688 DEBG Summed 4117 Outputs, pmmr size 8245, took 0s
Mar 20 14:01:37.413 DEBG Verified 4117 Rangeproofs, pmmr size 8245, took 71s
```

Summing and offsetting the kernels is reasonably fast (4s for ~4,000 kernels).
Summing the outputs is fast.
Verifying the rangeproofs for each of these outputs is _slow_ (71s for ~4,000 rangeproofs).

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -415,6 +415,7 @@ impl Chain {
 	pub fn validate(&self) -> Result<(), Error> {
 		let header = self.store.head_header()?;
 		let mut txhashset = self.txhashset.write().unwrap();
+		txhashset.force_rollback();
 		txhashset::extending(&mut txhashset, |extension| extension.validate(&header))
 	}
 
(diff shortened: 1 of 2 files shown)


chain/src/chain.rs
```diff
@@ -415,6 +415,7 @@ impl Chain {
 	pub fn validate(&self) -> Result<(), Error> {
 		let header = self.store.head_header()?;
 		let mut txhashset = self.txhashset.write().unwrap();
+		txhashset.force_rollback();
```
antiochp: This slipped in from another PR.

ignopeverell:
We may have to look closer at batch verification for those range proofs.

antiochp:
I wonder if its worth passing a `fast=true|false` flag into validate() to either validate all the rangeproofs, which we want to do for a fast sync vs. possibly skipping the rangeproof validation, say on check_compact, where we have presumably already validated all the rangeproofs but we still want to quickly re-validate the chain state?

ignopeverell:
Seems like it would make sense.

PR #825, opened 2018-03-20
Dandelion relay update

quentinlesceller:
- Dandelion relay update after monitor peer
- Debug logger instead of error logger when cannot find a peer relay.


ignopeverell:
Can you update something to kick a build in? Travis has been a lot better since the last improvements.

quentinlesceller:
Sure!

quentinlesceller:
Looks like something went wrong with the ````wallet test```

```
thread 'basic_wallet_transactions' panicked at 'Coinbase not confirming in time': grin/tests/wallet.rs:105
```

PR #830, opened 2018-03-20
Sync flag fix

### LINKED ISSUE #827: Stop mining when sync is occurring
The mining thread needs to be stopped or suspended when sync is active, as is fairly evident from initial pre-testnet2 ad-hoc testing. 

This is very evident on intial sync now when a node is configured for mining, has recieved all headers but hasn't started receiving blocks (or the fast-sync package).. in this case the sync flag is set to 'not syncing', the miner is kicked into action and confusingly tries to start putting blocks on top of what's being read.

There may be another issue in that the sync flag is being set to 'not-synching' too early in the fast sync process, but that should become evident while doing this.

ignopeverell:
Sets the sync flag properly when changing to txhashset download mode to fast-sync to full state. Fixes #827.

@antiochp looking good to you (I think you're the most familiar with this part of the code)?

Ignore all the commits listed on the PR and just check the diff, I think I need to do a full history reset on my personal master branch.


antiochp:
Looks good to me. :+1:

PR #837, opened 2018-03-21
[WIP] Wallet Restore

yeastplume:
This at least gets wallet restore working again (there were serialization issues with the rangeproof meaning it would fail earlier) but it's still much, much slower than it was on testnet1. However, the key derivation was far simpler in that iteration as the switch proof hash was just a function of b instead of b + another derived key.

The slowdown is all due to the iterating needed `derived_child_key`:

```
-   97.88%     0.02%  grin     grin                 [.] grin_keychain::keychain::Keychain::derived_child_key                                                                                                                                                                               ▒
   - 97.86% grin_keychain::keychain::Keychain::derived_child_key                                                                                                                                                                                                                           ▒
      - 97.01% grin_keychain::extkey::ExtendedKey::derive                                                                                                                                                                                                                                  ▒
         - 80.64% grin_keychain::extkey::Identifier::from_secret_key                                                                                                                                                                                                                       ▒
            - 74.92% secp256k1zkp::key::PublicKey::from_secret_key                                                                                                                                                                                                                         ▒
               - 74.88% secp256k1_ec_pubkey_create                                                                                                                                                                                                                                         ▒
                  + 62.80% secp256k1_ecmult_gen                                                                                                                                                                                                                                            ▒
                  + 11.76% secp256k1_ge_set_gej                                                                                                                                                                                                                                            ▒
            + 5.68% grin_keychain::extkey::Identifier::from_pubkey                                                                                                                                                                                                                         ▒
         + 14.86% blake2_rfc::blake2b::blake2b                                                                                                                                                                                                                                             ▒
      + 0.61% <std::collections::hash::map::HashMap<K, V, S>>::contains_key   
```

We could potentially make this work much, much faster by generating key in pairs, so if a child key for the blinding factor is derived from child_index n, then the key for the switch commit hash is generated from index n+1.

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -22,7 +22,6 @@ use chain;
 use p2p;
 use util;
 use util::secp::pedersen;
-use util::secp::constants::MAX_PROOF_SIZE;
 use serde;
 use serde::ser::SerializeStruct;
 use serde::de::MapAccess;
@@ -233,7 +232,7 @@ pub struct OutputPrintable {
 	/// Whether the output has been spent
 	pub spent: bool,
 	/// Rangeproof (as hex string)
-	pub proof: Option<pedersen::RangeProof>,
+	pub proof: Option<String>,
 	/// Rangeproof hash (as hex string)
 	pub proof_hash: String,
 
@@ -260,7 +259,7 @@ impl OutputPrintable {
 		let spent = chain.is_unspent(&out_id).is_err();
 
 		let proof = if include_proof {
-			Some(output.proof)
+			Some(util::to_hex(output.proof.proof.to_vec()))
 		} else {
 			None
 		};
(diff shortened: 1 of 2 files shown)


yeastplume:
Hmm... okay I think I've spent the day confused since I was assuming we were using the form `blake2(bJ,r)` for the switch commits, however, from what I can gather what's ending up being produced is `blake2(bJ,b)`, and b ends up being produced via a very slow reverse lookup from the identifier in `derived_child_key`

antiochp:
We are using the same key (via the same key identifier) for both the commitment and the switch_commit_hash. I don't think things would work if the key was different.

For example here - 

https://github.com/mimblewimble/grin/blob/e3601b6c8e9d71f7ed06560bda199d3c6edfec65/core/src/core/build.rs#L104-L107

yeastplume:
Closing in favour of removing switch commits altogether

antiochp:
😢

PR #841, opened 2018-03-21
Removed all switch commitment usages, including restore

ignopeverell:
After some discussion with @antiochp, @yeastplume and @tromp, we decided switch commitments weren't worth the cost of maintaining them and their drawbacks. Removing them.


ignopeverell:
For reference, switch commitments were found to:

* add a lot of complexity and assumptions
* take additional space for little benefit right now
* allow the inclusion of arbitrary data, potentially for the worst
* provide little to no advantage in case of quantamageddon (as range proofs are still a weakness)

apoelstra:
What does "as range proofs are still a weakness" mean? Non-soundness of rangeproofs are part of the security model of switch commitments.

Also what additional assumptions are required when making outputs 32 bytes larger by including data that has literally no consensus rules on it?

How do wallets identify their own outputs now?

apoelstra:
After chatting with @yeastplume on IRC, I realize that we can actually use rangeproof rewinding for wallets to recognize their own outputs, which even avoids the "gap" problem of just scanning for pre-generated keys.

With that in mind, it's true that the benefit of switch commitments for MW are not spectacular.

PR #843, opened 2018-03-22
Restore wallet restore

yeastplume:
An attempt to re-insert wallet restore (or at least ensure the data is in place for it) based on gitter conversations. Note at present, this is modified from what Andrew suggested Basically:

On output creation:
* Store *value|value|value|blinding_factor[0..8]* (first 8 bytes of blinding factor) in the first 32 bytes of the bullet-proof message (which gets masked against alpha), and all zeroes in the second 32 bytes.
* When generating the proof, use the blinding factor as the key, and `hash(commitment|root_key_id)` as the nonce.
To check if an output is yours:
* Attempt to retrieve the bullet-proof message using a random value as the key and `hash(commitment|root_key_id)` as nonce.. this should get you back the first 32 bytes of the *value|value|value|bf[0..8] message.
* if there's a *value|value|value* match (the same value 3 times), then iterate through some keys to try to find a match with the retrieved first 8 bytes of the blinding factor. Once you have a match confirm the key is correct by unwinding the proof with (key_candidate, hash(committment|root_key_id). If the second 32 bytes is all zeroes, then the commit is a definite match.

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -409,13 +409,8 @@ impl<'de> serde::de::Deserialize<'de> for OutputPrintable {
 			}
 		}
 
-		const FIELDS: &'static [&'static str] = &[
-			"output_type",
-			"commit",
-			"spent",
-			"proof",
-			"proof_hash",
-		];
+		const FIELDS: &'static [&'static str] =
+			&["output_type", "commit", "spent", "proof", "proof_hash"];
 		deserializer.deserialize_struct("OutputPrintable", FIELDS, OutputPrintableVisitor)
 	}
 }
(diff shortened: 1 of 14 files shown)


antiochp:
Forgive my ignorance but can someone explain how using "whatever as the key" will allow (some, the first few bytes?) of the message to be rewound?

anargle:
I think this is enough to replace switch commitments in their entirety, including the story for quantum resistance. By hiding the value in the range proof, the range proof itself becomes a commitment to the value. Moreover, since this just comes from xoring with the result of a hash (from the HMAC random number generator), even a quantum computer should not be able to identify what value was committed. In the event of a quantumaggedon, you can just require that all inputs have a quantum-resistant zero-knowledge proof of an RNG seed (in this case, a `root_key_id`) which reveals the value and blinding factor of the input.

yeastplume:
Going to merge this.. it's not consensus breaking and just amounts to adding some optional data in the bulletproof message.

PR #847, opened 2018-03-22
Rewind force rollback

### LINKED ISSUE #812: investigate rewind on every extension (even if no-op)
Related - #808 .

- [x] We should `force_rollback()` when calling `chain.validate()` (we `rewind()` but it is readonly and throwaway)
- [x] audit for other places where we `rewind()` to make sure we explicitly force rollback where necessary
- [x] ~investigate _always_ calling calling rewind() when starting a new txhashset extension (I don't think there are any reasons why we want to simply assume the head is the latest block, particularly with the one thread per peer networking)~

It should be safe to always `rewind()` to the latest block (and be a no-op 99% of the time).
In the cases where `rewind()` would have an effect because a new block was processed - we actually do want to rewind to keep things consistent.

So I propose we always rewind when starting a new extension.
Maybe the cleanest way to do this is to require a block header to be provided when starting a new extension?

And we should take care to consistently force a rollback on any extension that is "readonly".

antiochp:
Added `rewind_readonly()` as companion to `rewind()`.
Cleaned up callers of these to use one or the other as appropriate.

Resolves #812.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -421,7 +421,6 @@ impl Chain {
 		// the block header to ensure the view is consistent.
 		// Force rollback first as this is a "read-only" extension.
 		txhashset::extending(&mut txhashset, |extension| {
-			extension.force_rollback();
 			extension.validate(&header, skip_rproofs)
 		})
 	}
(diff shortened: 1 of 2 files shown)


ignopeverell:
Great function naming choice. Going to merge this as the `grin_grin` tests will be unhappy anyway for any PR that doesn't merge my test fix.

PR #851, opened 2018-03-23
Add versioning and build information

yeastplume:
Turns out there's a handy crate https://docs.rs/built/0.2.3/built/index.html that only needs to be added as a build time dependency, then spits out a file full of constants with useful build info that the rest of the code can pick up and compile in. This adds this crate, spits out the info to the debug log and displays the rust crate version+other info in the TUI

screenshot from 2018-03-23 10-18-14


antiochp:
:+1:

Issue #804, opened 2018-03-18
More compact MMR serialization (txhashset during fast sync)


antiochp:
#797 got me thinking about how we zip up the MMRs and send them over as part of fast sync.

If we know the (unpruned) size of the MMR then we know all the pos in advance.

So if we have the following - 
* unpruned size of the MMR
* roots of all pruned subtrees (and corresponding pos)
* existing data file
* existing rm_log
* existing prune_list

Then I think we have enough information to reconstruct the full MMR all the way up to the peaks?

So one possibility to reduce the amount of data being sent over the wire during fast sync may be to just send the above over. And the receiving node can then hash nodes together as necessary to reconstruct the full MMR.

Without putting any actual numbers on it it feels like this would be less data than sending the entire MMR across.
The tradeoff would be the increased cost of hashing on the receiving side.

For example - if we send `pos 1` and `pos 2` over then the receiving node can reconstruct the hash for `pos 3`.
Edit: The data elements for `pos 1` and `pos 2` will be in the data file, so we can create these hashes as well. I think we just need the pos for each element in the data file (and the roots of the pruned subtrees that won't be in the data file).

tromp:
If I understand them correctly, then a pruned MMR corresponds to a sequence of data points, each of which is either
an unpruned leaf, labeled as such, or
a pruned merkle subtree hash, labeled as such with its size (not height, since rightmost branch can have non-full binary trees).

ignopeverell:
@antiochp seems that could be a good optimization. Re-hashing is necessary anyway, you can't trust the hashes sent to you, so they all need to be checked.

antiochp:
Some back of the envelope numbers for how much data would be involved.

```
2^20 outputs (single peak, height 20, for a simple example)
1,048,576 leaves
2,097,151 hashes in hash file

full unpruned hash file: 2,097,151 * 32 bytes ~= 67 MB


Scenario 1:
No pruned nodes, full hash file.
Prune list is empty.

Instead of sending the data file and the hash file across we could send just the data file
and simply reconstruct the hash file from the full unpruned data file.
We would save the full 67 MB by not sending the hash file.


Scenario 2:
Hash file 50% pruned (full left hand side).
Prune list would consist of a single pos (and we would need a single hash for that node).

Pruned hash file would be approx 33MB.
Instead of sending that across we could send the single hash (and the pruned data file).
We would save 33MB by not sending the pruned hash file.


Scenario 3:
[...]

PR #853, opened 2018-03-23
[WIP] read/write prune list with associated hashes (for fast sync)

antiochp:
See #804 for context/discussion.

This is part of the "more compact" txhashset representation that we will send over the wire as part of the fast sync.

We need the following to fully reconstruct a txhashset on the receiving end -
1. prune list (with corresponding hashes for each pos)
1. data file (with corresponding pos for each element, maybe better way to do this?)
1. rm_log

We do not need -
1. full hash file (we can rebuild this from data file and "pruned" hashes)

This PR covers (1) and is relatively straight forward implementation wise.

From the vec of `(pos, hash)` we _should_ be able to -
* build the actual prune list (just the vec of pos)
* insert the hashes at the necessary pos back into the hash file during rebuild

### DIFF
--- a/store/src/pmmr.rs
+++ b/store/src/pmmr.rs
@@ -322,6 +322,29 @@ where
 		}
 	}
 
+	pub fn experimental_write_for_fast_sync(&self) -> io::Result<()> {
+		let tmp_prune_file = format!("{}/{}.tmp", self.data_dir, "fast_sync_prune");
+
+		let mut pruned_pos_hash = vec![];
+
+		// collect the hashes for all the pos in the prune list
+		for pos in &self.pruned_nodes.pruned_nodes {
+			let hash = self.get_from_file(*pos).expect("missing hash");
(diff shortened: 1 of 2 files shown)


ignopeverell:
Before we have any real perf/size numbers, this raises more questions (at least for me). This makes me wonder:

1. How much time does it add for the issuing node write another file?
2. What's the space gain on the resulting zip?

antiochp:
Yeah both good questions - let me think a bit more about this.
Could well be not worth the additional effort/cost for the savings we may see.

sesam:
Idea: fast sync is only available for 100 blocks (say), so the file write would only happen once every 100 blocks and is cachable.

antiochp:
@ignopeverell put some rough numbers down in #804 for amount of data.
Need to actually run some tests to see how long it would take to write the file etc.

antiochp:
Closing this PR for now - still tracking the issue at #804.

Issue #849, opened 2018-03-22
Investigate gitlab CI/CD (now available with github integration)


antiochp:
https://about.gitlab.com/2018/03/22/gitlab-10-6-released/

> While GitLab works best when you use it all end-to-end, we also seek to play well with others. In this spirit, we’ve added CI/CD integration with GitHub, and more generally the ability to integrate CI/CD with other external repos like BitBucket as well. We see four primary audiences that this functionality is designed for. 

> If you have a public, open source project on GitHub you can now take advantage of free CI/CD on GitLab.com. As part of our commitment to open source, we offer all public projects our highest tier features (Gold) for free. While other CI/CD vendors limit you to running a handful of concurrent jobs, GitLab.com gives open source projects hundreds of concurrent jobs with 50,000 free CI pipeline minutes.

sesam:
I'm working through https://gist.github.com/MoOx/64010ce5847276159018794352b8c987 and https://docs.gitlab.com/ee/ci/yaml/#cache

My .gitlab-ci.yml is here https://gitlab.com/mimblewimble/grin-ci/blob/master/.gitlab-ci.yml

Build status https://gitlab.com/mimblewimble/grin-ci/-/jobs/

Needs more packages.

sesam:
After #860 or simlar is merged, maybe add a link (an image badge) into README.md near the top.

PR #860, opened 2018-03-24
adds gitlab CI config

sesam:
Basic impl. for #849 using test --all

See an example of `cargo test --all` passing:
https://gitlab.com/mimblewimble/grin-ci/-/jobs/58984836

Gitlab syncs from github master branch automatically, so this would allow us to optimistically merge into master if Travis is acting up, and get a second opinion from Gitlab.


ignopeverell:
Time for CI races!

PR #870, opened 2018-03-26
Testnet2 test fixes

yeastplume:
Just fixing up tests in advance of T2 launch


yeastplume:
Tests good now. Tag time?

PR #879, opened 2018-03-27
Make peer connect attempts asynchronous

yeastplume:
As experienced on Testnet2 just now, grin didn't seem to be reconnecting to peers properly on startup (or doing so inconsistently)

The reason appears to be that the peer connection code is synchronous and goes one by one down the list, so if the first in the list takes a while to timeout for any reason, the sync wait ends, decides there are no peers and continues solo. Symptom was an inconsistent view on startup.

This puts each peer connection attempt onto its own thread, so all known peers up to max are connected to instantly on startup.

### DIFF
--- a/grin/src/seed.rs
+++ b/grin/src/seed.rs
@@ -213,18 +213,18 @@ fn listen_for_addrs(
 				.spawn(move || {
 					let connect_peer = p2p_c.connect(&addr);
 					match connect_peer {
-					Ok(p) => {
-						trace!(LOGGER, "connect_and_req: ok. attempting send_peer_request");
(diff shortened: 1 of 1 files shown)


yeastplume:
the peers are already on separate threads as far as I understand, this is (I hope) just the TCP connection call I've multi-threaded here. Will wait for @ignopeverell to consider before bunging in

antiochp:
Ok yeah that makes sense.

ignopeverell:
Yup, that sounds right. Each peer get handed to a thread *once connected*, to manage the communication. But establishing connections are done in a different thread. Thanks for the fix!

PR #880, opened 2018-03-27
Externalize Grin User Agent

### LINKED ISSUE #875: Externalize Grin User Agent
The Grin User Agent should be externalized.
The version should be the same that the one in the root ```Cargo.toml```.
https://github.com/mimblewimble/grin/blob/fcfe7bc6a418b8ed79e0f8bf633171688ebcd2af/p2p/src/msg.rs#L35

hashmap:
Take the version from the root Cargo.toml
Fixes #875

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -31,8 +31,8 @@ use types::*;
 /// Current latest version of the protocol
 pub const PROTOCOL_VERSION: u32 = 1;
 
-/// Grin's user agent with current version (TODO externalize)
-pub const USER_AGENT: &'static str = "MW/Grin 0.1";
+/// Grin's user agent with current version
+pub const USER_AGENT: &'static str = concat!("MW/Grin ", env!("CARGO_PKG_VERSION"));
 
 /// Magic number expected in the header of every message
 const MAGIC: [u8; 2] = [0x1e, 0xc5];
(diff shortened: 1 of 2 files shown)


p2p/tests/peer_handshake.rs
```diff
@@ -76,6 +76,8 @@ fn peer_handshake() {
 		net_adapter,
 	).unwrap();
 
+	assert!(peer.info.user_agent.ends_with(env!("CARGO_PKG_VERSION")));
```
yeastplume: We definitely don't want to do this.. this would exclude all peers with even a slightly different version than ours.. peers are generally okay to have different version numbers from our so long as their fork is compatible and they're not misbehaving.

p2p/src/msg.rs
```diff
@@ -31,8 +31,8 @@ use types::*;
 /// Current latest version of the protocol
 pub const PROTOCOL_VERSION: u32 = 1;
 
-/// Grin's user agent with current version (TODO externalize)
-pub const USER_AGENT: &'static str = "MW/Grin 0.1";
+/// Grin's user agent with current version
+pub const USER_AGENT: &'static str = concat!("MW/Grin ", env!("CARGO_PKG_VERSION"));
```
yeastplume: Yep, much preferable to what's there now.

p2p/tests/peer_handshake.rs
hashmap: @yeastplume this in an integration test, it just checks that `user_agent` has been properly generated in our code. did I miss smth?

yeastplume: No.. apologies that was a brainfart, all good.

antiochp:
:+1:

PR #881, opened 2018-03-27
Change log level for found cuckoo solution event

hashmap:
Trace seems to be to low, there is a  user request to make it more visible
https://gitter.im/grin_community/Lobby?at=5aba24967c3a01610d7c211f

### DIFF
--- a/grin/src/miner.rs
+++ b/grin/src/miner.rs
@@ -172,7 +172,7 @@ impl Miner {
 			if let Some(s) = job_handle.get_solution() {
 				let proof = Proof::new(s.solution_nonces.to_vec());
 				let proof_diff = proof.clone().to_difficulty();
-				trace!(
+				debug!(
 					LOGGER,
 					"Found cuckoo solution! nonce {} gave difficulty {} (block diff {})",
 					s.get_nonce_as_u64(),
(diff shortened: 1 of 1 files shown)


yeastplume:
Definitely don't think this is info level worthy, though I can live with putting this at debug level for now.

hashmap:
Oops, I wanted debug level as was suggested in gitter, my fault, fixing

sesam:
See also https://github.com/yourowncrypto/grin/blame/b7bc8f41e19b92381fbb9972d316758c894596c3/grin/src/miner.rs#L564

that's an info! saying valid PoW found, and on which block it was applied. The trace! line getting changed here is maybe earlier in the process, and so with some more chance of being lost. Seeing a "sucessful block mined" type of message, only to loose it again, might be suboptimal.

hashmap:
Yeah, in fact I stumbled upon this line before making change, probably it's why i typed `info` instead of  `debug`.

sesam:
If you have time, maybe you can figure out some way to make this even clearer. If people are primarily wishing for a way to notice when they get some grins (so they can celebrate :tada:), one idea is to also adjust the other info! output to make is maximally userfriendly, and maybe add a link to, say, "https://grinexplorer.net/block/" + b.hash()

PR #883, opened 2018-03-27
Make peer parameters user-configurable

yeastplume:
Make certain parameters around peers user-configurable. Done in such a way that they remain optional in grin.toml.

Any comments on why anything here shouldn't be exposed is welcome.


antiochp:
These are all good to have in grin.toml I think.
:+1:

quentinlesceller:
Agree 👍

PR #898, opened 2018-03-28
Merkle proofs fee

### LINKED ISSUE #873: Additional fees for Merkle proofs
Right now they're free. Note that it doesn't necessarily matter to spend mining rewards are miners can always include their transactions for free, but in the general case they need to pay for the extra bandwidth usage.

hashmap:
This PR fixes #873

I tried 3 approaches:

1. Implemented `weight()` on MerkleProof as a function of size of ` peaks` and `path`. Tx's `weight()`  takes `Vec<&MerkleProof>` as an extra parameter

2. Tried to simplify and replaced vector of proofs with just a total number of peaks

3. Simplified even more, just pass number of MerkleProofs in a tx.

Let me know if I oversimplified it.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -458,12 +458,25 @@ impl Transaction {
 
 	/// Calculate transaction weight
 	pub fn tx_weight(&self) -> u32 {
-		Transaction::weight(self.inputs.len(), self.outputs.len())
+		Transaction::weight(
+			self.inputs.len(),
+			self.outputs.len(),
+			self.input_proofs_count(),
+		)
+	}
+
+	/// Collect input's Merkle proofs
+	pub fn input_proofs_count(&self) -> usize {
+		self.inputs
+			.iter()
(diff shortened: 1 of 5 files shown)


core/src/core/transaction.rs
```diff
@@ -456,6 +456,33 @@ impl Transaction {
...
+		self.inputs
+			.iter()
+			.filter(|i| i.merkle_proof.is_some())
+			.count()
+	}
+
+	/// Calculate transaction weight from transaction details
+	pub fn weight(input_len: usize, output_len: usize, proof_len: usize) -> u32 {
+		let mut tx_weight =
+			-1 * (input_len as i32) + (4 * output_len as i32) + (proof_len as i32) + 1;
```
antiochp: I think a proof should count for _at least_ the equivalent of an output.
But I'd be good with merging this in now to get the structure of this in place, we can always tweak/adjust weights later.

antiochp:
Merging this - going to open an issue to determine the desired weight of these proofs.

PR #899, opened 2018-03-28
Testnet2 small fixes

sesam:
- minor logging cleanup
- add hint for gcc-5 for building CUDA plugin
- add N_BLOCKS comment for >=6GB GPU cards

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -169,10 +169,10 @@ pub fn sync_block_header(
 pub fn process_block_header(bh: &BlockHeader, mut ctx: BlockContext) -> Result<(), Error> {
 	debug!(
 		LOGGER,
-		"pipe: process_block_header: {} at {}",
-		bh.hash(),
-		bh.height
-	);
+		"pipe: process_block_header at {} [{}]",
+		bh.height,
+		bh.hash()
+	); // keep this
 
 	check_header_known(bh.hash(), &mut ctx)?;
 	validate_header(&bh, &mut ctx)
--- a/chain/src/txhashset.rs
+++ b/chain/src/txhashset.rs
@@ -502,7 +502,7 @@ impl<'a> Extension<'a> {
 	pub fn rewind(&mut self, block_header: &BlockHeader) -> Result<(), Error> {
 		let hash = block_header.hash();
 		let height = block_header.height;
-		debug!(LOGGER, "Rewind to header {} at {}", hash, height);
+		debug!(LOGGER, "Rewind to header at {} [{}]", height, hash); // keep this
 
 		// rewind each MMR
 		let (out_pos_rew, kern_pos_rew) = self.commit_index.get_block_marker(&hash)?;
--- a/grin.toml
+++ b/grin.toml
@@ -222,11 +222,17 @@ NUM_THREADS = 1
 #[mining.miner_plugin_config.device_parameters.0]
 #NUM_THREADS = 1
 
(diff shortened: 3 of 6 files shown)


chain/src/pipe.rs
```diff
@@ -167,7 +167,7 @@ pub fn sync_block_header(
 /// We validate the header but we do not store it or update header head based on this.
 /// We will update these once we get the block back after requesting it.
 pub fn process_block_header(bh: &BlockHeader, mut ctx: BlockContext) -> Result<(), Error> {
-	debug!(
+	trace!(
```
ignopeverell: Not this one, it's very useful for debugging.

chain/src/txhashset.rs
```diff
@@ -502,7 +502,7 @@ impl<'a> Extension<'a> {
 	pub fn rewind(&mut self, block_header: &BlockHeader) -> Result<(), Error> {
 		let hash = block_header.hash();
 		let height = block_header.height;
-		debug!(LOGGER, "Rewind to header {} at {}", hash, height);
```
ignopeverell: And not this one either, same thing :-) And it's not very frequent anyway.

grin/src/miner.rs
```diff
@@ -221,7 +221,7 @@ impl Miner {
 						}
 					}
 				}
-				info!(LOGGER, "Mining at {} graphs per second", sps_total);
+				info!(LOGGER, "Mining! Total Cuckoo power is {} graphs per second", sps_total);
```
ignopeverell: "power" sounds weird. You want to show Cuckoo on the log line?

sesam: Yes and yes. Goal is to say "Cuckoo30" to make it hard to miss what algo and gps we're getting (idea triggered by recent test framework bugfix, and some Lobby confusion on benchmarks during in last 5-10 days).

Better wording incoming.

grin.toml
```diff
@@ -222,11 +222,17 @@ NUM_THREADS = 1
 #[mining.miner_plugin_config.device_parameters.0]
 #NUM_THREADS = 1
 
-#CUDA Miner (Included here for integration only, Not ready for use)
+#CUDA Miner
+#
+# Note! 4+ GB GPU cards only!
+#   If you have 6GB GPU RAM, set N_BLOCKS = 128
+#   Docs: https://github.com/tromp/cuckoo/blob/master/GPU.md
```
yeastplume: Just came across this.. could you add another line here explicitly stating if you have 6GB or less (or you have 6GB and the GPU is also driving the display,) to lower N_BLOCKS?

sesam: After cuckoo merge [talked about in Lobby](https://gitter.im/grin_community/Lobby?at=5abcac182b9dfdbc3a344ec7) do you know what are the defaults for N_TRIMS and N_BLOCKS? Or I can try to figure it out :)

The commented out lines in grin.toml are:
````
#N_TRIMS = 240
#N_BLOCKS = 128
````
These should probably show the default, then any changes to the default values would be explained separately.

And maybe a line
`# Mining and driving your computer display on the same card? Then maybe you'll need to lower N_TRIMS`

tromp: N_TRIMS should be adjusted to 256.
N_BLOCKS should be 128 IF you have 6+GB and have enough to spare for your video needs plus other possible GPU applications you like to run.
Otherwise set it to 64, which is 5% less performant on 1080ti.

Issue #912, opened 2018-03-29
New node failing to sync


yeastplume:
New node (basically my dev environment with .grin deleted each time) appears to halt and stop synching every time... and not consistently on the same block. Logs from a couple of runs, first time at 278 with mining enabled and second time at 410 with mining disabled:

[grin.log](https://github.com/mimblewimble/grin/files/1859842/grin.log)

[grin-stuck-1.log](https://github.com/mimblewimble/grin/files/1859843/grin-stuck-1.log)

yeastplume:
Note stopping and restarting seems to get it moving again in each case, which is a good sign indicating something silly instead of some consensus issue

tromp:
My last server run also got stuck, at height 2830. See the grin.log at https://github.com/tromp/tromp.github.io/blob/master/grin.log

It first gets 5235138d at 2831 while it's at height 2828.
Grepping for initial occurences of 5235138d:
```
Mar 29 12:56:08.135 DEBG handle_payload: CompactBlock: 5235138d
Mar 29 12:56:08.135 DEBG Received compact_block 5235138d at 2831 from 198.245.50.26:13414, going to process.
Mar 29 12:56:08.137 DEBG block: hydrate_from: 5235138d, 1 txs
Mar 29 12:56:19.024 DEBG handle_payload: CompactBlock: 5235138d
Mar 29 12:56:19.024 DEBG Received compact_block 5235138d at 2831 from 78.46.193.41:13414, going to process.
Mar 29 12:56:19.026 DEBG block: hydrate_from: 5235138d, 1 txs
Mar 29 12:56:19.053 DEBG handle_payload: CompactBlock: 5235138d
Mar 29 12:56:19.053 DEBG Received compact_block 5235138d at 2831 from 78.46.193.41:13414, going to process.
Mar 29 12:56:19.055 DEBG block: hydrate_from: 5235138d, 1 txs
Mar 29 12:56:35.448 DEBG Received block header 5235138d at 2831 from 139.59.105.169:13414, going to process.
Mar 29 12:56:35.448 DEBG pipe: process_block_header: 5235138d at 2831
Mar 29 12:56:35.448 DEBG Block header 5235138d refused by chain: Orphan
```

Why does it take 27 seconds to go from the first "going to process" to the actual processing?

Later it decides it needs this block after all:
```
Mar 29 12:57:32.389 DEBG Requesting block 5235138d from peer 195.201.95.62:13414.
[...]

PR #916, opened 2018-03-29
[WIP] Fix and cleanup of fast sync triggering logic

ignopeverell:
Fixes the triggering of fast sync, which used to consider the header head instead of where other peers are at. Fast sync now triggers and seem to work properly but the next block fails with an `InvalidRoot`. Debugging that as well.


ignopeverell:
That last commit also fixes #912

PR #917, opened 2018-03-30
TUI Average view

yeastplume:
Just take difficulty / duration average over last 59 blocks so it remains accurate.

### DIFF
--- a/grin/src/server.rs
+++ b/grin/src/server.rs
@@ -322,8 +322,8 @@ impl Server {
 			DiffStats {
 				height: tip_height as u64,
 				last_blocks: diff_entries,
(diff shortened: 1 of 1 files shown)


tromp:
Why is taking average over 60 blocks not accurate?

yeastplume:
:D Because I don't have the 61st block in the past right at that point in the code so I can't get the duration between that and the first block in the window without a bit of re-write and handling/testing the genesis case. Will get to that soon but just want a quick fix to ensure the displayed average isn't the sum of 59 elements divided by 60.

tromp:
Ok, I see.

PR #924, opened 2018-04-03
use std lib function in pmmr for all_ones() and most_significant_pos()

RCasatta:
Use standard library function instead of manually calculating the value of ones and the most significant pos.

### DIFF
--- a/core/src/core/pmmr.rs
+++ b/core/src/core/pmmr.rs
@@ -987,28 +987,12 @@ fn bintree_jump_left(num: u64) -> u64 {
 
 // Check if the binary representation of a number is all ones.
 fn all_ones(num: u64) -> bool {
-	if num == 0 {
-		return false;
-	}
-	let mut bit = 1;
-	while num >= bit {
-		if num & bit == 0 {
-			return false;
-		}
-		bit = bit << 1;
-	}
-	true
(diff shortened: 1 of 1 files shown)


yeastplume:
Thanks for submitting this! Though we have problems with travis CI, the test failures in /chain are genuine this time around. Could you have a look?

RCasatta:
At the moment I can't reproduce the travis failure on my machine, both `chain` and `grin` dirs are passing the tests. I am using a slightly older version, I am trying again after upgrading `rustc`

RCasatta:
It's very strange, sometimes I've got the error, sometimes I havent, need more time to investigate

core/src/core/pmmr.rs
```diff
@@ -987,28 +987,12 @@ fn bintree_jump_left(num: u64) -> u64 {
...
-	}
-	let mut bit = 1;
-	while num >= bit {
-		if num & bit == 0 {
-			return false;
-		}
-		bit = bit << 1;
-	}
-	true
+	num.leading_zeros() + num.count_ones() == 64
```
hashmap: looks like `0` will produce `true`

RCasatta: you are very right

hashmap:
LGTM

PR #930, opened 2018-04-04
Full sync ask blocks only from other archival nodes

### LINKED ISSUE #929: Full sync should only ask blocks from other archival nodes
I believe it requires adding a feature bit for archival nodes, don't think we have that right now.

quentinlesceller:
Fix #929.

~~Just need to add the ```FULL_HIST``` internal check after discussion with @ignopeverell.~~ Done.

Not as elegant as I wanted it to be though.

Before starting the p2p server, in case of archive mode, we try to retrieve the full block 1. If it fails, we remove the ```FULL_HIST``` capability from the ```capabilities```.

### DIFF
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -255,7 +255,7 @@ where
 		// Track the case where a parent of a transaction is in stempool
 		let mut parent_in_stempool = false;
 		// The timer attached to this transaction
-		let mut timer: i64 = <i64>::max_value();
+		let mut timer: i64 = 0;
 
 		// The next issue is to identify all unspent outputs that
 		// this transaction will consume and make sure they exist in the set.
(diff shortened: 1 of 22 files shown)


ignopeverell:
Looks like the `p2p` crate doesn't compile.

quentinlesceller:
Weird, everything is working on my machine (and Travis?).

ignopeverell:
Mmh not sure where I saw that, must have been confused sorry.

pool/src/pool.rs
```diff
@@ -252,6 +252,11 @@ where
 		let stem_propagation = random <= self.config.dandelion_probability;
 		let mut will_stem = stem && stem_propagation;
 
+		// Track the case where a parent of a transaction is in stempool
```
ignopeverell: Did you mean to commit that stuff too?

ignopeverell:
Mmh not sure where I saw that, must have been confused sorry.

quentinlesceller:
@ignopeverell Looks like I did a ```git checkout``` in the wrong branch :). Just merged branch from upstream.

Issue #935, opened 2018-04-05
Support for remote miners


bladedoyle:
Grin needs to support remote miners, and mining proxies/pools.

This issue proposes:
* a minimal stratum server running in the Grin core process
* using JSONRpc
* when enabled the stratum server listens on port 13416
* accepts multiple clients
* supports methods:  login, submit, keepalive, getjobtemplate, status
* broadcasts new jobs

Please see the first few comments for some discussion history

bladedoyle:
yeastplume commented 23 days ago 
Hi @bladedoyle, thanks very much for submitting this and it all looks to be moving in the right direction. I'd be really happy to have some pooled mining functionality built in as well as a mining client that supports pooling functionality. Usually you have to go scouring nasty places on the internet to find appropriate pooled mining clients, so I'd really like to be in a position to have 'official ones'

Some comments above (and some are nit-picks as this is a WIP,) but overall:

    I'm a bit concerned that inserting a stratum mining server directly into Grin (yet another port and listener) is a bit 'and the kitchen-sink-y', but if the code is properly compartmentalised and none of the other developers have an issue with it, then I think it should be okay.
    I would definitely move the client binary directly into the cuckoo miner module, for reasons stated above (it will make everything much more flexible)

Other than that, I won't get into too much more detail because WIP.

Thanks again, and please feel free to contact me with questions at any time, especially around using cuckoo-miner (and if there are any changes required there to support this)

bladedoyle:
bladedoyle commented 22 days ago

Thanks for the feedback, @yeastplume.

The lightest-weight solution would be to add 2 API to the existing server - for "get_block" and "submit_solution". The down-side is that a client would need to poll over http for new blocks, which I think would not work well especially for a short block interval. The other extreme would be to implement a full featured stratum server within grin.
My solution proposes a middle ground - a tcp listener for a single client connection and methods to push data in both directions. Not a full "stratum" implementation. I was thinking there would always be "pool" or "proxy" service running between the external miner(s) and this grin "stratum-like server". I intend to write a (or adapt an existing) pool as the next step. That pool would implement a more complete stratum protocol (including authorization, job difficulty adjustment, standard error codes). That pool would be a separate project from grin itself.

I agree that the miner binary does not belong directly in the grin project. Moving it to the cuckoo miner project sounds right.

Let me take another pass and then request a lower level review.

bladedoyle:
ignopeverell commented 22 days ago 
[...]

PR #937, opened 2018-04-06
minimal stratum server

bladedoyle:
Originally implemented on testnet1 source:  https://github.com/mimblewimble/grin/pull/779

For support of mining pools and remote miners: https://github.com/mimblewimble/grin/issues/935

Merged with testnet2


bladedoyle:
Continued from testnet1 branch: https://github.com/mimblewimble/grin/pull/779

bladedoyle:
Sure, give me a few days to refactor mining and stratum code into a "miner" create one level up.  

Im eager to get these changes merged so I would prefer to handle miner decoupling, etc.. as a spearate pull request.

yeastplume:
Yep the miner decoupling will evolve over time, and I can do some work on it as well.

bladedoyle:
@yeastplume @ignopeverell 
Can you provide some guidance on how to deal with the "adaptors.rs" module currently in grin_grin.  Parts of adaptors.rs is needed by the new "miner" crate and also in "servers.rs".  Maybe I need to move it up a level so it can be used by both without creating a circular dependency?  It does not seem to fit into any existing crate (does it?), so maybe it needs to be a crate by itself?

Ref:  
* https://github.com/bladedoyle/grin/blob/7c3e0359e30bd729471ba6677b4391402261d59d/grin/src/stratumserver.rs#L200
* https://github.com/bladedoyle/grin/blob/7c3e0359e30bd729471ba6677b4391402261d59d/grin/src/server.rs#L52

sesam:
@bladedoyle realistically the options are
1. new crate
2. squeeze into existing crate which might get messy
3. (unless I've misundersttod things here) allow circular crate deps (I think we have some already) and use weak references that get upgraded on use, see macros.rs `wo` and `w`

yeastplume:
Let me try and refactor a bit.. I think the best thing to do might be to rename the `grin` crate to `servers`, and have the individual servers in their own subdirectories

yeastplume:
See https://github.com/mimblewimble/grin/pull/955.. if everyone else is okay with the approach there then we'll have a logical place to put the stratum miner.

yeastplume:
Okay, though the merge history has got a bit mangled, I think the merge should be fine so I'll go ahead and merge so we can move on to the next phases

Issue #218, opened 2017-10-31
Lookup list of seed servers from a DNS record


ignopeverell:
For now we simply download a file on github that has a list of server names. This was the simplest to get started but we need something less dependent on github. Add the ability to list all A records from a DNS name to feed seed connections.

sesam:
Testnet uses WebStatic = http://grin-tech.org/seeds.txt which is no more or less secure than DNS. If grin-tech.org gets a reddit hug this might fail.

We have /v1/peers/all which is perhaps too chatty.

Suggestion:
1. add /v1/peers/listening (listing only API roots of "full" nodes)
2. change WebStatic to a list of long running API roots of whoever is communicates with the project, stays upgraded, and agrees to be listed, so we remove one level of indirection. Anyone who wants can then list both HTTPS and HTTP. Fast start could pick a single HTTP peer to know block height (with low peer timeout, maybe even lower than current "< 5s"), then prioritize getting HTTPS peers.

## If we really need DNS
HTTPS works where DNS won't. On the other hand, there are Hotspot wifi points that might let through some DNS or other low level traffic unharmed, if we want to get crafty later on to escape any naive attempts at forcing a chain split.

This use case is often solved with TXT records. If we add a dependency on the [trust_dns crate](http://trust-dns.org/target/doc/trust_dns/rr/rdata/txt/struct.TXT.html) we'd get TXT records.

Using A would require as many DNS requests as there are lines in the seed list, though if a single seed is enough then round-robin could be a nice side-effect. On the other hand, 1 of 3 of the grin projects current seeds was offline yesterday, and making a single DNS request is better than 3.

CNAME could be abused to ascii armor a short list of (IPv4, port) tuples, but that's hard to decode and not human readable.

## Privacy issues
Then there's the issue of privacy; any client requesting the seed list tags themselves as a grin user. Before that becomes a concern, maybe grin can tag on to some less trackable discovery methods, i.e. use some existing E2E chat solutions to get a better anonymity set and plausible den. of having crypto.

Possibly tagging on to some file that M2M communications keep hitting. /favicon.ico probably not; but robots.txt / humans.txt / abuse.txt allows for free form comments.

sesam:
Also, we'd like to be able to filter the list of full peers by genesis block hash, so a Testnet1 peer only gets Testnet1 blocks. This might again cause privacy issues, making it easier to locate grin developers.

ignopeverell:
[...]

PR #940, opened 2018-04-06
Add support for DNS Seed

quentinlesceller:
Fix https://github.com/mimblewimble/grin/issues/218.
This PR add support for DNS Seed in Grin.
Currently only one DNS which resolve to the current list of seeds at http://grin-tech.org/seeds.txt.
It will need more DNS to be the default connection method.


ignopeverell:
Maybe also update your seed to have multiple A records, one for each existing seed? For example see:

```
$ dig seed.bitcoin.sipa.be
```

kargakis:
I can confirm I see multiple A records 
```
;; ANSWER SECTION:
seed.grin.lesceller.com. 3600	IN	A	192.241.160.172
seed.grin.lesceller.com. 3600	IN	A	109.74.202.16
seed.grin.lesceller.com. 3600	IN	A	198.245.50.26
seed.grin.lesceller.com. 3600	IN	A	46.4.91.48
```

PR #942, opened 2018-04-07
Add DNS Seed and make DNSSeed default

quentinlesceller:
Thanks to @hashmap, we now have three DNS seeds (grin-seed.owncrypto.de).
This PR also make DNS seeding the default seeding method.


ignopeverell:
Can we add contact email in front of each seed?

quentinlesceller:
Sure give me a sec

PR #945, opened 2018-04-08
introduce extending_readonly to simpify txhashset extensions

antiochp:
we used to have a variety of combinations of -
  * rewind
  * rewind_readonly
  * force_rollback
  * cancel_rollback

We now have  - 
  * txhashset::extending()
  * txhashset::extending_readonly()
  
We still use force_rollback as necessary for the cases where a writeable extension fails and we actually want an explicit rollback.

This way we are not messing about with remembering to call `force_rollback()` when creating a Merkle proof, or remembering to call `cancel_rollback()` during fast sync because we know we need to cancel the `force_rollback()` called as part of validation...

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -435,7 +435,7 @@ impl Chain {
 		// against the latest block header.
 		// We will rewind the extension internally to the pos for
 		// the block header to ensure the view is consistent.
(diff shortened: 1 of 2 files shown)


ignopeverell:
LGTM!

PR #949, opened 2018-04-09
Error handling using failure in API

hashmap:
This PR addresses #166
Error handling in wallet was ported to failure in https://github.com/mimblewimble/grin/pull/713
Using the same error model makes wallet code simpler and may simplify migration to Hyper.

* rustfmt
* replace `try!` with `?`


yeastplume:
Think this looks fine, no reason to leave it outstanding. Thanks!

PR #950, opened 2018-04-09
[WIP] Core PMMR and API updates to support wallet restore

yeastplume:
Given we can't rely on full blocks being present on all nodes, for functions such as wallet restore we need to way to return the contents of the UTXO set to a requesting caller in reasonably-sized chunks. The most logical way seems to be via returning UTXOs by their PMMR insertion index, so if there are N utxos, the caller can say 'get me utxos 0..99' .. scan them.. then say 'now get 100..199' etc...

However, because of pruning, it will be more practical to say 'get me the first 100 unpruned outputs starting at position 0, and then have the caller return the actual index of the last returned output for the next query, which the caller can repeat until it's up to the most recent output.

Going to add this in a few parts, with this PR adding the appropriate functions to PMMRs and the API, then do the wallet work separately.

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -127,7 +127,12 @@ impl OutputHandler {
 				.iter()
 				.filter(|output| commitments.is_empty() || commitments.contains(&output.commit))
 				.map(|output| {
-					OutputPrintable::from_output(output, w(&self.chain), Some(&header), include_proof)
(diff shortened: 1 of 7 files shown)


yeastplume:
API calls work now.. wallet restore works to a point, will merge what's there now and address that in another PR.

PR #955, opened 2018-04-11
Split `grin` crate into separate modules

yeastplume:
As has come up in: https://github.com/mimblewimble/grin/pull/937, it's desirable to have some way of compartmentalising server types, (as we're now trying to add a stratum mining server) Rather than creating a whole load of crates, and because there are some elements that all servers will need (such as adapters, stats, types, etc), I think it makes more sense just to split up the `grin` crate into separate modules.

This renames the `grin` crate to `servers`, and splits up the existing functionality into common/grin/mining mods, which should allow for the new mining server to be slotted in in a logical way.

@ignopeverell @antiochp @quentinlesceller let me know if this makes sense to you


quentinlesceller:
@yeastplume makes sense to me. I also think it makes more sense renaming the ```grin``` crate to ```servers```.
Overall 👍

bladedoyle:
Ya, this looks great!

yeastplume:
Okay, will merge to let @bladedoyle continue ... hopefully it shouldn't be too difficult to merge your changes on top of this new structure

Issue #819, opened 2018-03-20
Dandelion Transaction Aggregation


quentinlesceller:
This follows from https://github.com/mimblewimble/grin/issues/706 and https://github.com/mimblewimble/grin/pull/719.
Now that we have the Dandelion logic in place we can aggregate the transaction during the stem phase.
This mechanism will likely need the anti aggregation mechanism mentioned here https://github.com/mimblewimble/grin/issues/799.
Relevant documentation https://github.com/mimblewimble/grin/tree/master/doc/dandelion

sesam:
We [talked about it](https://gitter.im/grin_community/dev?at=5ae8e8826d98e53e0445088c) being necessary to prevent too low fee transactions from being aggregated. Otherwise it opens for an attack; by creating heavy transactions with way too low fee, so the aggregate is never mined, in extreme meaning that only non-aggregated transactions are ever mined.

quentinlesceller:
@antiochp @ignopeverell  I've been thinking a lot about stem transaction aggregation in the past few days and I come up with a new to do it. However, I'm not entirely sure that it makes sense, hence I want your opinion on this.
Currently (in #1023), I have a an aggregator that will every 30 seconds (for example)  aggregate all the transactions in the stempool and broadcast them as stem to its Dandelion relay.
When a node receives this stem transaction, he will flip a coin and see if he want to "fluff it" or "stem it".
This way transaction are aggregated along the way like a snowball.
However, I see a lot of problem with this idea:
1. Low fee transaction getting in the way (mentioned above by @sesam).
2. Recovering a previous state if a previous node decide to fluff the transaction.
3. Other reconciliation problems.

Then, I had an idea where the transaction are not aggregated in the stempool but relayed individually just like now. But, when a node flips a coin and fluff it, it will aggregate all the transaction in its stempool and send it to the whole network.
This way we can avoid:
1 - Low fee transaction, by checking that the fee is decent.
2 - Problems of recovering a previous state since all the Dandelion relays keeps all the transaction and can just use the previous developed ```deaggregate``` function.
3 - One "problem" that I see is it not as efficient for privacy since we do not do cut though along the way but at the end.
Also in that case we don't need an aggregator.
Thoughts ?

tromp:
[...]

PR #966, opened 2018-04-16
Add aggregate function for multi-kernel transaction

quentinlesceller:
This PR adds an ```aggregate``` function into ```core/transaction.rs```. This function will aggregate multiple transactions with cut-through.
While unused at the moment, this function will be useful for potential transaction aggregation in the stempool.
See https://github.com/mimblewimble/grin/issues/819.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -468,8 +468,9 @@ pub fn aggregate(transactions: Vec<Transaction>) -> Result<Transaction, Error> {
 	let mut outputs: Vec<Output> = vec![];
 	let mut kernels: Vec<TxKernel> = vec![];
 
-    // we will sum these together at the end to give us the overall offset for the transaction
-    let mut kernel_offsets = vec![];
+	// we will sum these together at the end to give us the overall offset for the
+	// transaction
+	let mut kernel_offsets = vec![];
 
 	for mut transaction in transactions {
 		// we will summ these later to give a single aggregate offset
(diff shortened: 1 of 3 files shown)


tromp:
The following test cases should be added, because we may want to change the behaviour of these edge cases, like for example indicate an error.
1. aggregating a coinbase output with a spend of that coinbase.
2. aggregating the three txs {a->b, b->c, b->d}
currently 1. produces a tx with overlap between inputs and outputs, while 2 produces in invalid tx a -> c,d

quentinlesceller:
Thanks for the review. 
1. Currently Coinbase outputs are filtered out as you can see [here](https://github.com/mimblewimble/grin/pull/966/files#diff-404c77cf79ebda67cdbd58c1c99daa8cR511). So the aggregation will work but will not do anything in that case.
2. As of now, the aggregation function will build a multi-kernel transaction with such transactions. However, this transaction will be rejected by the pool. In the future, the aggregation function will work in conjunction with the stempool (in that case one of the last 2 transaction will get rejected). I will add a test to ensure that it is the case.

core/src/core/transaction.rs
```diff
@@ -461,6 +462,75 @@ impl Transaction {
...
+		.cloned()
+		.collect::<Vec<_>>();
+
+	let new_outputs = outputs
+		.iter()
+		.filter(|out| !to_cut_through.contains(&out.commitment()))
+		.cloned()
+		.collect::<Vec<_>>();
+
+	let tx = Transaction::new(new_inputs, new_outputs, kernels);
```
antiochp: We need to sort the `inputs|outputs|kernels` here to ensure the tx is valid.
And we should cover this in the test for completeness.

quentinlesceller: I just added the sorting and the test in ```core/mod.rs```.

PR #967, opened 2018-04-17
minor code reorg, add mining job difficulty, enhance miner status response

bladedoyle:
minor code reorg: move HeaderPrePowWriter to mine_block module, 
add mining job difficulty value to the stratum job template,
enhance miner status rpc response (not to be confused with server stats)


yeastplume:
Looks good, just a note that I'm working on getting together the standalone mining client here: https://github.com/mimblewimble/grin-miner, which will hopefully soon become the 'official' mining method

Issue #968, opened 2018-04-17
Do we need to still validate Odd Fee?


ziyuanliu:
If I'm understanding the arch properly, isn't the OddFee restriction due to burning? (which has been recently removed here: #767 

https://github.com/mimblewimble/grin/blob/cc9ffcc1ab784be22e8fce2ae5b32e5c4b8bc500/core/src/core/transaction.rs#L423-L425

antiochp:
I believe this is no longer necessary.
Want to open a PR to remove this check (and cleanup the OddFee error itself)?
There may or may not be test coverage around this that would need cleaning up also.

antiochp:
Closing - resolved in #975.

PR #975, opened 2018-04-18
Remove oddfee error references + txn validation check

ziyuanliu:
[#968] From what I can tell, this is pretty modular from all burn_txn


pow/src/types.rs
```diff
@@ -56,6 +56,7 @@ pub struct MinerConfig {
 	/// Base address to the HTTP wallet receiver
 	pub wallet_listener_url: String,
 
+	/// TODO: should be removed or renamed, due to the lack of burning?
```
antiochp: Burning fees and burning rewards are different (confusing) concepts.
Burning a reward is used in testing - instead of "sending" the reward to a wallet we use a random key and simply make it unspendeable aka burnt.

PR #979, opened 2018-04-18
[WIP] Miner removal

yeastplume:
Removing mining from the main grin executable in favour of the new embedded stratum server. Once this is complete, all mining for grin should be through the stratum server and the new standalone mining client at https://github.com/mimblewimble/grin-miner

Makes the grin core a lot cleaner and moves mining completely out-of-process, which makes sense as it's the part most likely to crash hard with all of the bleeding-edgeness and parameter tweaking, etc.

Also rename the current MiningConfig to StratumServerConfig as all of the plugin parameters have been shifted to grin-miner

Todos:
 - [x] Remove 'pow' crate and put cuckoo verification elsewhere
 - [x] Change existing mining loop into test-only version instead (as that's all it's needed for now), and fix up tests to use it
 - [x] Replace TUI mining information with Stratum info
 - [x] Internal test miner to allow dev testing at cuckoo 16
 - [x] Documentation and mining guide updates (though it's mostly the same as before, just interfacing to grin via another process)
 - [x] Any other test fixes
 - [ ] Provide facility for genesis block creation (probably within grin-miner, and probably leave this for another day)

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -9,12 +9,15 @@ What's working so far?
 * Linux x86_64 and MacOS [grin + mining + development]
 * Not Windows 10 yet [grin kind-of builds. No mining yet. Help wanted!]
 
+## Mining in Grin
+
+Please note that all mining functions for Grin have moved into a separate, standalone package called
+[grin_miner](https://github.com/mimblewimble/grin-miner). Once your Grin code node is up and running,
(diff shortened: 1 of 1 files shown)


yeastplume:
More or less works, with some rough edges that'll be smoothed out over time. I'm thinking of letting this fly very shortly if there are no hard objections

PR #983, opened 2018-04-19
fix various issues with rewinding data file

antiochp:
Fixes - 
* we were not accounting for pruned data when rewinding the data file... 
  * surprised this was working at all
* make MMR backend more robust when MMR is empty
  * do not attempt to memory map empty file
* make MMR more robust when reading past the end 
  * return empty vec

### DIFF
--- a/store/src/pmmr.rs
+++ b/store/src/pmmr.rs
@@ -189,14 +189,16 @@ where
 			.rewind(index)
 			.map_err(|e| format!("Could not truncate remove log: {}", e))?;
 
+		// Hash file
 		let shift = self.pruned_nodes.get_shift(position).unwrap_or(0);
 		let record_len = 32;
 		let file_pos = (position - shift) * (record_len as u64);
 		self.hash_file.rewind(file_pos);
 
 		// Data file
-		let flatfile_pos = pmmr::n_leaves(position) - 1;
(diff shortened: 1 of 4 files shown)


store/src/pmmr.rs
```diff
@@ -189,14 +189,16 @@ where
...
+		// Hash file
 		let shift = self.pruned_nodes.get_shift(position).unwrap_or(0);
 		let record_len = 32;
 		let file_pos = (position - shift) * (record_len as u64);
 		self.hash_file.rewind(file_pos);
 
 		// Data file
-		let flatfile_pos = pmmr::n_leaves(position) - 1;
-		let file_pos = (flatfile_pos as usize + 1) * T::len();
+		let leaf_shift = self.pruned_nodes.get_leaf_shift(position).unwrap_or(0);
```
antiochp: We were not accounting for pruned data here.
So reading any data after we pruned would have potentially returned the _wrong_ data (wrong output commitment etc.)

Kind of surprised we did not see this happen (although it may be pretty subtle and hard to actually identify).

antiochp:
Going to merge this - should make rewind in the presence of pruned data more correct and will eliminate various consensus problems with bad MMR data after a fork.

Issue #799, opened 2018-03-16
Tx Anti-Aggregation for regular (non-Dandelion txs)


antiochp:
tldr - 
* Dandelion provides a mechanism for tx aggregation and cut-through.
* We need to support multi-kernel txs for this.
* We need an anti-aggregation mechanism is place for regular non-Dandelion txs (to minimize aggregation abuse on the network).

------

When we support Dandelion we will have the ability for both tx aggregation and tx cut-through at the "stem" phase of Dandelion tx propagation.

Once txs enter the "fluff" phase (or skip Dandelion stem phase entirely for speed) - we want to prevent any further tx aggregation.
If txs are aggregated during normal tx propagation we run a significant risk of conflicts and open the network up to a variety of denial-of-service concerns.

For example - nefarious node starts aggregating random txs in different combinations, preventing other nodes from seeing a consistent view of the current tx pool.
Sends the aggregated tx `(tx1, tx2)` to some peers and a conflicting `(tx1, tx3)` to other peers etc.

The problem - 
If we support Dandelion then we must by definition support txs with multiple kernels.
If we support multi-kernel txs then we have no way of differentiating between the following - 
* cut-through txs
* legitimately aggregated txs (via Dandelion)
* any set of txs aggregated by any node at any point in time, anywhere on the network

This is _good_ for privacy, but _bad_ for the reasons given above.

Proposed solution - 
(This has been discussed a few times previously but putting it down in a specific issue here for reference).
1) allow nodes to aggregate txs (we cannot prevent this)
2) but - if we see evidence of an aggregated tx - 
  1) attempt to de-aggregate it if we have also seen any of the constituent txs
  1) store them locally in our pool as de-aggregated as much as possible
  1) propagate them in this de-aggregated state  as much as possible

Basically if a node sees the following (based on tx kernel sets) -
* `(tx1, tx2)`
* `tx1` (already in pool, accepted prior to aggregated tx)

Then - we can recreate `tx2` by "subtracting" `tx1` from `(tx1, tx2)` (inputs, outputs _and_ kernels).
We then act like we never saw `(tx1, tx2)` but we add `tx2` to our pool and propagate `tx2` out to our peers.

I think this will mean well behaved nodes on the network will actively work to un-aggregate, thus reducing or even eliminating any effect any badly behaved node would have on the network.

[...]

PR #984, opened 2018-04-19
Anti-aggregation mechanism for multi-kernel transaction

quentinlesceller:
Fix https://github.com/mimblewimble/grin/issues/799.

This PR adds an anti-aggregation mechanism in Grin.
The following function where implemented:
- ```aggregate_with_cut_through``` aggregate multiple tx in one big tx with cut_through
- ```aggregate``` which aggregates without cut-through (used internally for deaggragation)
- ```PartialEq``` for transaction for faster testing
- ```deaggregate``` based on a set of transaction, this function is capable to deaggregate a multi kernel transaction as much as possible.
-  ```deaggragate_and_add_to_memorypool```.

Whenever we receive a multi kernel transaction, we first try to deaggregate it.
  - If we are able to deaggregate it, we add the subset in our memory pool.
  - If we are not able to deaggregate it, we add the transaction as is in our memory pool.

One step closer to stempool aggregation :).

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -234,6 +234,17 @@ pub struct Transaction {
 	pub offset: BlindingFactor,
 }
 
+/// PartialEq
+impl PartialEq for Transaction {
+    fn eq(&self, tx: &Transaction) -> bool {
+        if self.inputs == tx.inputs && self.outputs == tx.outputs && self.kernels == tx.kernels && self.offset == tx.offset {
+			true
+		} else {
+			false
+		}
+    }
+}
+
 /// Implementation of Writeable for a fully blinded transaction, defines how to
 /// write the transaction as binary.
 impl Writeable for Transaction {
@@ -482,8 +493,8 @@ impl Transaction {
 	}
 }
 
-/// Aggregate a vec of transactions into a multi-kernel transaction
-pub fn aggregate(transactions: Vec<Transaction>) -> Result<Transaction, Error> {
+/// Aggregate a vec of transactions into a multi-kernel transaction with cut_through
+pub fn aggregate_with_cut_through(transactions: Vec<Transaction>) -> Result<Transaction, Error> {
 	let mut inputs: Vec<Input> = vec![];
 	let mut outputs: Vec<Output> = vec![];
 	let mut kernels: Vec<TxKernel> = vec![];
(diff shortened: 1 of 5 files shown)


pool/src/pool.rs
```diff
@@ -409,6 +409,74 @@ where
...
+	pub fn deaggregate_and_add_to_memory_pool(
+		&mut self,
+		tx_source: TxSource,
+		tx: transaction::Transaction,
+		stem: bool,
+	) -> Result<(), PoolError> {
+		match self.deaggregate_transaction(tx.clone()) {
+			Ok(deaggragated_tx) => self.add_to_memory_pool(tx_source, deaggragated_tx, stem),
+			Err(e) => {
+				debug!(
```
sesam: maybe make this a warning! level log?

quentinlesceller: Yeah, we can expect this to happen quite often once we have the Dandelion transaction aggregation, so ```DEBUG``` is fine for now.

core/src/core/transaction.rs
```diff
@@ -234,6 +234,19 @@ pub struct Transaction {
 	pub offset: BlindingFactor,
 }
 
+/// PartialEq
```
antiochp: Can we just `derive` PartialEq on Transaction (we're just comparing all the fields)?

core/src/core/transaction.rs
```diff
@@ -556,6 +570,121 @@ pub fn aggregate(transactions: Vec<Transaction>) -> Result<Transaction, Error> {
 	Ok(tx.with_offset(total_kernel_offset))
 }
 
+/// Aggregate a vec of transactions into a multi-kernel transaction
+pub fn aggregate(transactions: Vec<Transaction>) -> Result<Transaction, Error> {
```
antiochp: What's the situation where we want to aggregate but we _don't_ want to apply cut-through?
Can we not just always apply cut-through?

pool/src/pool.rs
```diff
@@ -409,6 +409,74 @@ where
...
+		tx: transaction::Transaction,
+	) -> Result<Transaction, PoolError> {
+		// While the inputs outputs can be cut-through the kernel will stay intact
+		// In order to deaggregate tx we look for tx with the same kernel
+		let mut found_txs: Vec<Transaction> = vec![];
+		let mut total_kernels_found = 0;
+		// Since its a multi-kernel transaction we loop over the kernels
+		for kernel in tx.kernels.clone() {
+			// Look for tx with the same kernels
+			for (_, tx) in &self.transactions {
```
antiochp: We're looking for txs in the pool that are "subsets" of the incoming tx (based on kernel sets).

It might be simpler (and more intuitive) to put the incoming tx kernels into a set, then for each tx in the pool do the same and look at the the intersection of the two sets.

antiochp: And maybe have this as a separate fn on the pool - so we can retrieve all the txs in the pool that are candidates for deaggregating the incoming tx.

core/src/core/transaction.rs
quentinlesceller: That's what I did at the beginning and for some reasons the comparison were not the same.

core/src/core/transaction.rs
quentinlesceller: This function is used internally in the deaggregate function. 
For example, we receive (tx1,tx2,tx3) and we have (tx1)(tx2) in our pool. It allows us to quickly aggregate the last two txs into (tx1,tx2) for faster comparison.

pool/src/pool.rs
quentinlesceller: Okay got it updating.

core/src/core/transaction.rs
ignopeverell: I don't think you need the `if` though...

pool/src/pool.rs
```diff
@@ -409,6 +409,81 @@ where
...
+		// Check each transaction in the pool
+		for (_, tx) in &self.transactions {
+			let candidates_kernels_set: HashSet<TxKernel> =
+				tx.kernels.iter().cloned().collect::<HashSet<_>>();
+
+			let kernels_set_intersection: HashSet<&TxKernel> =
+				kernels_set.intersection(&candidates_kernels_set).collect();
+
+			// Consider the transaction only if all the kernels match
+			if kernels_set_intersection.len() == tx.kernels.len() {
```
antiochp: Looks good with the set intersection - one more question though... is it sufficient to compare the `len()` here?

Can we use a stricter check here and look for one being the subset of the other?
https://doc.rust-lang.org/beta/std/collections/struct.HashSet.html#method.is_subset

PR #985, opened 2018-04-19
Simplify block marker metadata

antiochp:
Rather than attempting to keep a combination of the PMMR files and the metadata in sync to provide an approximation of atomicity, I think we can do this in a far simpler way.

On node chain init we can validate the MMR roots against the last known chain head.
If the roots are valid then we are good to go.
If the roots are not valid (for _whatever_ reason) we can simply revert to an older block (and use the existing `rewind()` functionality) and validate the roots for this older header.
We can continue iterating back until we find a "good" block that validates against the PMMR roots.

We never need to open a txhashset with anything but the full files. We can let `rewind()` take care of rewinding to the correct positions as necessary.

Proposing the following changes - 
* we do not need to track PMMR metadata (we can use block markers for this)
* we do not need provide pos and height to open a txhashset (we can rewind to a block marker)
* on node startup we rewind the txhashset based on the chain head
  * and for completeness we validate the full txhashset (skipping expensive rangeproof verification)
* take it a step further - 
  * if validating fails on chain init, go back a block and try again until we find a "good" block

The txhashset validation adds a slight delay on startup but this means we catch consensus problems earlier.
~A node refuses to start up if the MMR backend files are corrupted for *any* reason.~
We iterate back through prev blocks until we find a "good" one that validates fully. 
We can then start the node successfully based on the txhashset for that block.

Update: doing some "extensive" local testing here (stopping and restarting a node manually...)
Got this in the logs - 

```
Apr 19 17:24:52.500 INFO Starting server, genesis block: 3346f63c
Apr 19 17:24:52.598 DEBG chain: init: rewinding and validating before we start... 3cbf9f4c at 3239
Apr 19 17:24:52.598 DEBG Rewind to header 3239 @ 3cbf9f4c
Apr 19 17:24:52.598 DEBG Rewind txhashset to 3239, BlockMarker { output_pos: 10130, kernel_pos: 8303 }
Apr 19 17:24:52.922 ERRO Corrupted storage, could not read an entry from hash store: IOErr(Custom { kind: UnexpectedEof, error: StringError("failed to fill whole buffer") })
Apr 19 17:24:52.922 ERRO Corrupted storage, could not read an entry from hash store: IOErr(Custom { kind: UnexpectedEof, error: StringError("failed to fill whole buffer") })
Apr 19 17:24:52.923 DEBG Error returned, discarding txhashset extension.
Apr 19 17:24:52.923 DEBG chain: init: rewinding and validating before we start... 57bb9806 at 3238
Apr 19 17:24:52.924 DEBG Rewind to header 3238 @ 57bb9806
Apr 19 17:24:52.924 DEBG Rewind txhashset to 3238, BlockMarker { output_pos: 10129, kernel_pos: 8301 }
Apr 19 17:24:56.739 DEBG Validated, summed (and offset) 4154 kernels, pmmr size 8301, took 3s
Apr 19 17:24:56.875 DEBG Summed 3365 Outputs, pmmr size 10129, took 0s
Apr 19 17:24:56.876 DEBG Committing txhashset extension. sizes (10129, 10129, 8301)
Apr 19 17:24:56.939 DEBG Chain init: 17512 @ 3238 [57bb9806]
```

This indicates we managed to corrupt the PMMR backend files on the previous shutdown.
We failed to validate the txhashset at `header 3239 @ 3cbf9f4c`.
We then reverted to the previous block and successfully validated the txhashset at `header 3238 @ 57bb9806` and then started the node at that point.
🚀 

TODO - 
- [x] fix the tests
- [x] is the full chain validation too slow on startup? (even skipping rangeproof verification)
  - [x] can we get away with just validating the roots? (Is this enough, probably?)
- [x] test various scenarios manually in full archival mode
- [x] test various scenarios manually in fast sync mode
- [x] cleanup commented out code

Just putting this here for reference -
```
truncate -s -1 .grin/txhashset/output/pmmr_hash.bin
```
This will "corrupt" the pmmr hash file by removing a single byte from the end...


antiochp:
Sort of related to this - if we stored the total `output count` and `kernel count` on each block header,
we could derive the `output pos` and `kernel pos` for each block.
I think @tromp was advocating for this elsewhere (Merkle proof related?)

This would allow us to get rid of the block_marker index.

More specifically it would also allow us to take an arbitrary txhashset and a full set of block headers and recreate chain state for any block header height (assuming block headers are on the same fork as the txhashset) _without_ requiring a block_marker.

i.e. fast sync could be more flexible -
* pull a full set of block headers 
* pull _any_ (within reason) txhashset (daily archive?)
* choose a horizon locally _our choice_ and verify txhashset vs. header
* start pulling full blocks

ignopeverell:
I like the incremental rewind, this is a good idea. Need to think some more about your 2nd idea to get rid of the block marker, but how about tracking it in a different issue to start?

PR #987, opened 2018-04-21
add request method name to the result to allow matching onthe client side

bladedoyle:
NOTE:  This needs to be merged at the same time as the "rpc_response" branch pull request in the grin-miner project.  It changes the rpc response api.


yeastplume:
Looks good.. the client side isn't being used widely at the moment, so we just have to make sure everything works together when we tell everyone to switch to grin-miner (hopefully soon).

PR #992, opened 2018-04-22
Optimize logging

dpc:
`Async<Discard>` means "drain that will send each logging
record to an async thread, that will discard it". This means
logging record is sent to another thread for no reason.

This change fixes that problem.


yeastplume:
Okay, makes sense. Thanks for the fix!

PR #993, opened 2018-04-22
Make `pre-commit` more portable

dpc:
Right now it's failing on NixOS, and this is fixing it.


antiochp:
:+1: thanks for this!

PR #996, opened 2018-04-22
[WIP] add block header cache to store

antiochp:
The `DifficultyIter` is expensive to use over when validating block headers. as we need to 
retrieve a large number of block headers from the store each time.
We store block headers by hash in RocksDB and these are immutable - so it should be safe to simply cache them locally in a HashMap.

TODO - 
- [ ] run some perf numbers and document them here (validating a block header is approx twice as fast now)
- [ ] what is the simplest eviction policy we can get away with here?


ignopeverell:
Looks good. I expect we'll need a few more of these as we haven't been particularly careful about optimizing data access. But having immutable data hashes makes it easy. We'll just have to keep track of those to avoid too much bloat. We might want to eventually add something to evict old data as well (LRU will stay full forever) but not a huge deal right now.

Btw the other perf thing I've been thinking of we might hit eventually is hashing. Granted, blake2 is fast, but we get hashes over and over without memoization.

Issue #973, opened 2018-04-18
do not fallback to requesting full block if we receive a compact block during syncing


antiochp:
We fallback to requesting a full block if we fail to successfully hydrate a compact block.
This will _always_ happen if we receive a compact block while we are syncing.

We should be able to check the store for the previous full block (we have recent full blocks).
If we don't have the previous full block then there is no point requesting the current full block.

We should only ask the store for the previous full block if we fail to hydrate the compact block.

i1skn:
Hello, I would like to work on this one, if nobody already started.

sesam:
@i1skn great! Probably only @antiochp is working on something nearby right now.
See also https://gitter.im/grin_community/dev if you run into any troubles.

i1skn:
@antiochp I have a question. I see here:
https://github.com/mimblewimble/grin/blob/master/servers/src/common/adapters.rs#L137-L149
that we are already check for the previous header existence in the chain and only than request the full block.   

I could not find here:
https://github.com/mimblewimble/grin/blob/master/core/src/core/block.rs#L646-L653
anything relying to the *full* previous block.   

Can you help me to understand why having a *full* previous block is important at this stage? Thanks!

sesam:
I think the reason is to avoid filling up orphan pool (limited to max 200)
with blocks we don't yet need

i1skn:
@sesam thanks for the answer. I have a clarification to ask, in case we do not have a *full* previous block and current compact block is invalid - we should just stop processing the current, right?

sesam:
yes. Gitter is often faster, today is an exception :)

quentinlesceller:
Should be fixed by #1002.

PR #1002, opened 2018-04-24
Do not request a full block for invalid compact block during syncing

i1skn:
During sync, if we fail to hydrate a compact block, we should ignore if it's invalid.   
Currently, we try to get a full block, if a compact one is invalid.

Should solve https://github.com/mimblewimble/grin/issues/973

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -157,12 +157,20 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 					debug!(LOGGER, "adapter: successfully hydrated block from tx pool!");
 					self.process_block(block, addr)
 				} else {
-					debug!(
-						LOGGER,
-						"adapter: block invalid after hydration, requesting full block"
-					);
-					self.request_block(&cb.header, &addr);
-					true
+					if self.sync_state.status() == SyncStatus::NoSync {
+						debug!(
+							LOGGER,
+							"adapter: block invalid after hydration, requesting full block"
+						);
+						self.request_block(&cb.header, &addr);
+						true
+					} else {
+						debug!(
+							LOGGER,
+							"adapter: block invalid after hydration, ignoring it, cause still syncing"
+						);
+						true
+					}
 				}
 			} else {
 				debug!(


servers/src/common/adapters.rs
```diff
@@ -134,8 +134,8 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 				.upgrade()
 				.expect("failed to upgrade weak ref to chain");
 
-			if let Ok(prev_header) = chain.get_block_header(&cb.header.previous) {
-				if let Ok(()) = block.validate(&prev_header) {
+			if let Ok(prev_block) = chain.get_block(&cb.header.previous) {
+				if let Ok(()) = block.validate(&prev_block.header) {
```
ignopeverell: I think that means anyone could send us an invalid compact blocks to force us to load full blocks from the db, a relatively expensive operation. Why not completely refusing compact blocks during sync? We have no use for them anyway.

servers/src/common/adapters.rs
```diff
@@ -141,15 +141,14 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 				} else {
 					debug!(
 						LOGGER,
-						"adapter: block invalid after hydration, requesting full block"
+						"adapter: block invalid after hydration, ignoring it"
 					);
-					self.request_block(&cb.header, &addr);
```
antiochp: I don't think we can simply ignore these.
We need some conditional way of knowing if we are syncing or not.
* If we are syncing then we _do not_ want to handle propagated compact blocks and we can just ignore the compact block (this can be earlier in the processing, not just when we try and hydrate it)
* If we are not syncing then we _do_ want to handle the propagated compact block and we _do_ want to go request the full block (otherwise we will run the risk of never seeing the latest propagated block and getting out of sync)

i1skn: Can you guide me where or how I can distinguish these two states?

ignopeverell: Sorry for taking so much time, the adapter has a sync state struct that you can directly check the current syncing status on.

i1skn: Thanks, I will look into it and will update the PR.

i1skn:
@antiochp  @ignopeverell This is ready to review again

Issue #1005, opened 2018-04-25
Body sync is disabled if fast sync is enabled


hashmap:
However we don't run fast sync periodically anymore https://github.com/mimblewimble/grin/blob/master/servers/src/grin/sync.rs#L426 so it makes sync really slow. 

Not sure if it's a real world issue, but I'm working on #697 where I need to rewind the tip down beyond the horizon. In my test a node can't catch up after that, because gossip sync basically just populate orphan pool. It may explain https://github.com/mimblewimble/grin/issues/977 however my test is more extreme.

antiochp:
Body sync _should_ kick back in once we've downloaded and validated the txhashet.

Once our headers catch up to the highest known peer we disable fast sync (which doesn't necessarily feel right?) -
https://github.com/mimblewimble/grin/blob/8227ce941a440da958146698f0f695ef2c69379e/servers/src/grin/sync.rs#L109-L110

But agreed about it being slow - I think #977 is causing this as a full orphan pool effectively throttles how many blocks we request at a time.

I think the overall sync process needs to be a bit more stateful with some clearer state changes as we progress through it.

* check to see if we need to sync
* get the headers (all of them)
* optionally get the txhashset 
  * only do this once (until this is significantly faster we only want this on once on node init)
  * validate it and set current block_marker
* start pulling blocks based on block_marker (or from 0 if full sync)

I'm not sure we really gain much by getting blocks and headers in parallel right now (for full sync) and I think it makes the sync process harder to understand and reason about.
And fast sync vs. full sync are not really very different (except full sync just skips the txhashset step).

hashmap:
I suspect there is an issue with this condition  when we calculate `fast_sync_enabled`. I'm not completely sure, but it may be the case that while we were processing a txhashset, which takes time, a peer produced more blocks, so the gap is more than the horizon value, so we enable fast sync, later ignore it, but just block body sync forever.

I agree with your proposal reg more detailed tracking - we started discussing it in https://github.com/mimblewimble/grin/issues/957

hashmap:
I just build from master branch and removed `.grin`, it seems that fast sync finished and then my node stuck at height 39418.
```
[arch@castle grin-node]$ grep "txhashset data" grin.log
Apr 25 22:50:17.650 INFO Received valid txhashset data for ca1f3f32.
[arch@castle grin-node]$ grep body_sync grin.log
[arch@castle grin-node]$
[arch@castle grin-node]$ tail grin.log
[...]

PR #1006, opened 2018-04-25
Enable body sync if fast sync is enabled

hashmap:
We don't run fast sync periodically anymore but disable body sync forewer.
I'm not sure how we should fix this issue #1005 this fix is the simplest solution, let's discuss options.


hashmap:
I agree that this is not a proper solution, closing the pr. However I added some info into the issue, for me body sync doesn't start at all.

PR #1007, opened 2018-04-26
Handle valid reorgs older than horizon in fast sync

### LINKED ISSUE #697: Test valid reorgs older than horizon
The node should get back into fast-sync mode and resync.

hashmap:
This pr fixes #697. If fast sync requests sync with a longer fork a txhashset will be asked at the height when the fork was created.
Also if a new block is a tip of a new (longest) fork then a new txhashset will be requested at the height when the fork was created.

Test is commented out, it's time consuming (60+ seconds) because a lot of `sleep` to create good forks.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -494,30 +494,31 @@ pub fn rewind_and_apply_fork(
 	// keeping the hashes of blocks along the fork
 	let mut current = b.header.previous;
 	let mut hashes = vec![];
+	let mut curr_header = store.get_block_header(&current)?;
 	loop {
-		let curr_header = store.get_block_header(&current)?;
-
 		if let Ok(_) = store.is_on_current_chain(&curr_header) {
 			break;
 		} else {
 			hashes.insert(0, (curr_header.height, curr_header.hash()));
 			current = curr_header.previous;
 		}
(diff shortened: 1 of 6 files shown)


hashmap:
@ignopeverell I assume you are referring to https://github.com/yourowncrypto/grin/blob/be437932fb8cdc2316e8557ad6b44e1641ce4206/servers/src/common/adapters.rs#L421 
My reasoning was the following - we got a block which is a part of a fork which goes beyond horizon, so we can't properly validate it (or can we?). This situation kind of invalidate the current txhashset so we ask a new (deeper) one.

Doing this check when we have a new head makes sense, but what we should do in `process_block` in this case - validate it using data we have?

ignopeverell:
We can only validate its header chain fairly cheaply. With just that block we don't know if that header chain is valid. The job of detecting valid header chains with more work and check them is done by sync. So I do think the least bad thing to do is drop that block and let sync do its job (if applicable).

garyyu:
@hashmap  I see this PR is quite old, last update was 5 months ago.
Should we close this one?  Or need more discussion?

hashmap:
@garyyu I will update it soon, thanks

hashmap:
will reopen when updated

PR #1008, opened 2018-04-26
Wallet restore fixes

yeastplume:
This *should* let wallet restore work (regardless of whether it was fast synched), with the limitation that coinbase transactions are re-locked for 1000 blocks after the restore. This is because the merkle proof is gone when identifying outputs from the UTXO set and needs to be regenerated. The simplest thing to do is generate one for the current PMMR state and re-lock.

We should be able to optimise this later, and even restore differently if we're restoring against a full archival node, but in the meantime this should at least get basic wallet restore working.


antiochp:
:+1:

PR #1009, opened 2018-04-26
Prioritize and allow no change transaction

### LINKED ISSUE #981: Wallet is stuck when creating a transaction with 'smallest' strategy
For testing purposes I have been sending transactions around with a couple of wallets. I have noticed that at some point, a wallet gets stuck indefinitely when trying to create a transaction when using the 'smallest' strategy:
```
$ grin wallet send -f -d DEST -s smallest 0.1
Apr 18 21:27:18.643 INFO This is Grin version 0.2.0 (git testnet2-64-g323480a), built for x86_64-unknown-linux-gnu by rustc 1.25.0 (84203cac6 2018-03-25).
Apr 18 21:27:18.643 DEBG Built with profile "debug", features "" on Tue, 17 Apr 2018 14:55:39 GMT.
Apr 18 21:27:18.643 DEBG Using wallet seed file at: ./wallet.seed
Apr 18 21:27:18.722 DEBG Refreshing wallet outputs
Apr 18 21:27:19.004 INFO Acquiring wallet lock ...
Apr 18 21:27:19.101 INFO ... released wallet lock
```

After that last line it just waits forever. On the server side it looks like this (I have removed the middle part)
```
Apr 18 21:29:33.126 DEBG outputs_by_ids: ["09cc058b8598e07877b3784d933932d715528f3c0c755c0b3c298d6b85962cf51d", "09ea353622cd5500071bdb00457d6727c63f4df7d1070132da678244abe5725d5c", ..., "0878ad63cbe5af18833084a156f09b0523a7acc6193694ecc71b23f46d8d8493f4", "0907aebe3d636c1cb35107010ac9489ca1328d0f1f0f68551eb9450e9d97fac68b"]
```
The server appears to keep function normally after that. There is no error of any kind on the wallet or server side.

I asked @quentinlesceller to verify independently and he was able to reproduce the situation with the specific seed. He also noticed that the wallet gets "unstuck" by using the default strategy.
If you want to see it yourself, you can restore from the seed `0ed9be2bfccddb04f0bc938196a48b697ac2bda8332d19a42b41fc680ae7e593`. Please use the same command as I did, otherwise the wallet might get unstuck and I have to try to create another seed with the same error.

quentinlesceller:
Fix #981 

Allows building transaction with no change outputs when possible. Sender might overpay (a little) the fee  in some cases.

Details:
- Prioritize no change output transaction.
- Fix receiver wrong fee in aggsig context (based on real fee and not calculated fee).

### DIFF
--- a/wallet/src/sender.rs
+++ b/wallet/src/sender.rs
@@ -317,7 +317,6 @@ fn build_send_tx(
 	parts.push(build::with_lock_height(lock_height));
 
 	let (tx, blind) = build::partial_transaction(parts, &keychain).context(ErrorKind::Keychain)?;
-	println!("Sender tx: {:?}", tx);
 
 	Ok((tx, blind, coins, change_key, amount_with_fee))
 }
--- a/wallet/src/sender.rs
+++ b/wallet/src/sender.rs
@@ -108,7 +108,7 @@ pub fn issue_send_tx(
 				Some(change) => {
 					info!(LOGGER, "cleaning up unused change output from wallet");
 					wallet_data.delete_output(&change);
-				},
+				}
 				None => info!(LOGGER, "No change output to clean from wallet"),
 			}
 		})
@@ -129,7 +129,7 @@ pub fn issue_send_tx(
 				Some(change) => {
 					info!(LOGGER, "cleaning up unused change output from wallet");
 					wallet_data.delete_output(&change);
-				},
+				}
 				None => info!(LOGGER, "No change output to clean from wallet"),
 			}
 		}).unwrap();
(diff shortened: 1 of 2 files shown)


wallet/src/sender.rs
```diff
@@ -299,6 +317,7 @@ fn build_send_tx(
 	parts.push(build::with_lock_height(lock_height));
 
 	let (tx, blind) = build::partial_transaction(parts, &keychain).context(ErrorKind::Keychain)?;
+	println!("Sender tx: {:?}", tx);
```
antiochp: Minor: `println!`

quentinlesceller: Ahah you busted me! Removing that.

PR #1015, opened 2018-04-27
Improved (more compact) Merkle Proofs

antiochp:
Reworking Merkle proofs so we only store a `path` and the underlying `mmr_size`.
The path includes the full sibling path and the necessary hashes derived from the peaks.
We need the `mmr_size` as we cannot rely on this being available when we come to verify the proof.
The `mmr_size` is used to determine the pos of the various entries in the path.

For siblings in the path we hash the node and the sibling together using the index of the parent.
For peak derived hashes we hash with the `mmr_size` (as a substitute for the non-existent pos).

Everything should hash up to the `root` itself.

TODO - 
- [x] get `verify()` working for `ImproveMerkleProof`
- [x] think through how we can do this in a backward-compatible way (if its even possible or worth doing)
- [x] rename `ImprovedMerkleProof` -> `MerkleProof`
- [x] how to check validity of txs (when adding to tx pool etc.) as we need MMR pos etc.
- [x] get chain tests passing
- [x] fixup wallet code (Merkle proof PartialOrd?)
- [x] cleanup and fn docs
- [x] cleanup `println!` all over Merkle Proofs...
- [x] test locally (sending txs etc.)
- [x] move merkle_proofs into their own file


antiochp:
This I think is ready for review.

* Simplified `MerkleProof` impl - just a `path` and the `mmr_size`
* We now verify Merkle proofs in one place and we require the `root` and the `node_pos` to do so.

/cc @ignopeverell @tromp

core/src/core/pmmr.rs
```diff
@@ -92,82 +92,49 @@ where
...
+/// Merkle proof errors.
+#[derive(Clone, Debug, PartialEq)]
+pub enum MerkleProofError {
+	/// Merkle proof root hash does not match when attempting to verify.
+	RootMismatch,
+}
+
+/// A Merkle proof that proves a particular element exists in the MMR.
+#[derive(Debug, Eq, PartialEq, Clone, PartialOrd, Ord)]
 pub struct MerkleProof {
```
ignopeverell: Should we take this as an opportunity to move Merkle proofs in their own file? I think it'd be better to not clutter `pmmr.rs` too much.

core/src/core/pmmr.rs
```diff
@@ -309,6 +297,52 @@ where
 			.collect()
 	}
 
+	fn peak_path(&self, peak_pos: u64) -> Vec<Hash> {
+		let rhs = self.bag_the_rhs(peak_pos);
+		println!("rhs bagged - {:?}", rhs);
```
ignopeverell: Still lotsa println :-)

antiochp:
@tromp - want to take a look over this before I merge onto `consensus_breaking` branch?

core/src/core/merkle_proof.rs
```diff
@@ -0,0 +1,159 @@
...
+		// handle special case of only a single entry in the MMR
+		// (no siblings to hash together)
+		if self.path.len() == 0 {
+			if root == node_hash {
+				return Ok(());
+			} else {
+				return Err(MerkleProofError::RootMismatch);
+			}
+		} else if self.mmr_size == 1 {
+			if self.path.len() == 1 && root == node_hash && vec![root] == self.path {
```
tromp: An MMR of size 1 has a single leaf, which is also the root, and no path of siblings. So testing for a path of length 1 makes no sense to me.

core/src/core/merkle_proof.rs
```diff
@@ -0,0 +1,159 @@
...
+				return Ok(());
+			} else {
+				return Err(MerkleProofError::RootMismatch);
+			}
+		}
+
+		let sibling = self.path.remove(0);
+		let (parent_pos, sibling_pos) = pmmr::family(node_pos);
+
+		let peaks_pos = pmmr::peaks(self.mmr_size);
```
tromp: determining peaks at every step of a merkle path is rather inefficient, as are all the family calls. the computed node positions above the peaks also don't make much sense, but it does look like it should produce the right result. i plan to rewrite all this in the near future anyway:-)

core/src/core/pmmr.rs
```diff
@@ -322,48 +200,41 @@ where
...
-
 		let node = self.get_hash(pos)
 			.ok_or(format!("no element at pos {}", pos))?;
 
 		let mmr_size = self.unpruned_size();
 
+		if mmr_size == 1 {
+			return Ok(MerkleProof {
+				mmr_size,
+				path: vec![node],
```
tromp: shouldn't this be an empty path of siblings?!

core/src/core/pmmr.rs
```diff
@@ -322,48 +200,41 @@ where
...
+			});
+		}
+
 		let family_branch = family_branch(pos, self.last_pos);
 
-		let path = family_branch
+		let mut path = family_branch
 			.iter()
-			.map(|x| (self.get_from_file(x.1).unwrap_or(Hash::default()), x.1))
+			.filter_map(|x| self.get_from_file(x.1))
```
tromp: why filter_map instead of map ?

core/src/core/pmmr.rs
```diff
@@ -940,6 +811,9 @@ pub fn bintree_postorder_height(num: u64) -> u64 {
 /// We know the positions of all leaves based on the postorder height of an MMR of any size
 /// (somewhat unintuitively but this is how the PMMR is "append only").
 pub fn is_leaf(pos: u64) -> bool {
+	if pos == 0 {
```
tromp: this ugly border test case can be removed again when we change to 0-based positions in the future:-)

core/tests/merkle_proof.rs
```diff
@@ -0,0 +1,192 @@
...
+
+	let mut ba = VecBackend::new();
+	let mut pmmr = PMMR::new(&mut ba);
+
+	pmmr.push(elems[0]).unwrap();
+	let pos_0 = elems[0].hash_with_index(0);
+	assert_eq!(pmmr.get_hash(1).unwrap(), pos_0);
+
+	let proof = pmmr.merkle_proof(1).unwrap();
+	assert_eq!(proof.path, [pos_0]);
```
tromp: again, shouldn't this be an empty path?

core/src/core/pmmr.rs
antiochp: So we don't need to do the extra `unwrap_or()` call to handle `None` values.
If there is nothing in the file at this pos then we filter it out and don't need to map it to anything.
This should never happen in practice, but the code is cleaner this way.
Either way we end up building an invalid Merkle proof if the file is actually missing hashes at the necessary pos.

core/src/core/merkle_proof.rs
antiochp: Yeah good catch - we can actually get rid of this entire `else if` block now that an MMR of size 1 produces a path of length 0.

antiochp:
Addressed feedback, will merge once all tests pass.

PR #1019, opened 2018-04-30
demonstration of a double spend (corrupted rm_log)

antiochp:
Something has been bugging me for a couple of days around how we validate individual blocks vs. validating the full txhashset (say in a fast sync).

Outside of the "pruning" behavior of the various MMR data structures there is nothing in the regular block processing pipeline that verifies we are not attempting a "double spend".
And pruning behavior relies exclusively on the rm_log state.

This PR adds a test case that "corrupts" the rm_log, allowing a previously spent output to the "double spent" in the next block. I think this demonstrates some fragility around how we rely exclusively on the rm_log data structure in the output MMR to identify and prevent double spends.

If we were to ever get in the situation where the rm_log was truncated or otherwise corrupted we would not be able to identify a double spend, nor would we be able to prevent the block chain from building on an invalid block containing a double spend.

If (for _any_ reason) the rm_log was to get truncated the node would allow an "invalid" block to be processed and validated and we would happily apply this invalid block the txhashset - without ever realizing it was invalid.

This feels a little surprising to me that we would not actually catch the "double spend" during normal block validation.
It feels _very_ fragile to be relying on the rm_log as the authoritative data structure to prevent "double spends" like this.

During full txhashset validation we check (we catch a double spend here) - 
```
outputs - total_supply == total_kernel_excess
```

When validating a single block - 
```
outputs - inputs == block_kernel_excess
```

So two different blocks can spend the same output but each block individually will successfully validate. And with a corrupted rm_log both blocks will be successfully added to the chain.

### DIFF
--- a/chain/tests/mine_simple_chain.rs
+++ b/chain/tests/mine_simple_chain.rs
@@ -31,10 +31,8 @@ use core::core::target::Difficulty;
 use core::consensus;
 use core::global;
 use core::global::ChainTypes;
-
-use keychain::Keychain;
-
 use core::pow;
+use keychain::Keychain;
 
 fn clean_output_dir(dir_name: &str) {
 	let _ = fs::remove_dir_all(dir_name);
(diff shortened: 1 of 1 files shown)


antiochp:
Hey - @ignopeverell @yeastplume @tromp. Thoughts on this?

antiochp:
Strawman proposal:
1) we keep running totals of the output and kernel sums for each block
2) validating an individual block actually involves summing the inputs|outputs|kernels with the running totals for the previous block and checking the total kernel excess is correct (i.e. we basically validate the full txhashset without needing to sum everything from scratch each time)

antiochp:
We're also already storing the "kernel_offset" in each block header. 
And I _think_ we can modify this to be the "kernel_sum_with_offset" so we won't need an additional sum for the kernels, just one for the outputs.

See #1020 for kernel offsets on blocks (currently blinding_factors but being reworked to store them as commitments).

antiochp:
Closing - tracking a proposed enhancement at #1021.

PR #1020, opened 2018-04-30
make kernel offsets commitments

antiochp:
We track "kernel offsets" of transactions and blocks.

We have been storing these as `blinding_factors` but I think the code is probably cleaner if we store these as `commitments`.

It never felt clean to be serializing/deserializing a `blinding_factor` on a tx (or a block).

This adds a single byte to each transaction and each block (a commitment is 33 bytes, a blinding_factor 32 bytes).
But the tradeoff is less code (and less time) converting between blinding_factors, secret_keys and commitments.

TODO
- [x] refactor common code related to `zero_commit` default values
- [x] cleanup wallet code (to use commitments)
- [x] clean up serialization code (no need to ser/deser blinding_factors?)

NOTE: this is consensus breaking - changes the serialization format of both transactions and blocks...

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -498,7 +498,7 @@ impl BlockHeaderPrintable {
 			kernel_root: util::to_hex(h.kernel_root.to_vec()),
 			nonce: h.nonce,
 			total_difficulty: h.total_difficulty.into_num(),
-			total_kernel_offset: h.total_kernel_offset.to_hex(),
+			total_kernel_offset: util::to_hex(h.total_kernel_offset.0.to_vec()),
 		}
 	}
 }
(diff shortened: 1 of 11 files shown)


antiochp:
@ignopeverell @yeastplume gitter is down currently, so communicating via PRs on the off chance you see this...

I just created a `consensus_breaking` branch in github for `mimblewimble/grin`
I'm going to merge this branch into there, mainly because I want somewhere to play around with committing to output and kernel sums in block headers.
I guess I could do this in my fork but it felt like a good time to attempt a non-testnet2 branch on the main repo.

Lets discuss later - I'm totally fine with bringing these changes back into my forked repo and working there.

ignopeverell:
That feels a lot better. We shouldn't have anything that looks like a secret go over the network. Even if the offset is throwaway, I can see some silly wallet implementation trying to reuse keys...

antiochp:
Definitely - I'm cringing at some of the code I wrote a couple of months ago... 😑

Issue #1026, opened 2018-05-01
Grin's Cuckoo Implementation


Edgecombe:
Grin's Cuckoo implementation uses an EDGEMASK that is half of the reference EDGEMASK for cuckoo cycle. This causes great weakness.

I wrote a solver to see how bad it was. I could not get my full solver up to 2^30 graph size, but at 2^13 graph size I was finding 200,000+ solutions per graph. This number was increasing as I increased the problem size. Drawing the line, I suspect the 2^30 graph size has more than 1 billion valid unique solutions in it. This means most of the time is spent on blake2b to validate the solutions in the best implementation.

The PoW for grin needs to be fixed. My recommendation is to remove the non-standard cuckoo-cycle changes - they appear to both be broken (I believe the easiness factor is also broken).

tromp:
the mask is applied correctly. the number of nodes is 1<<sizeshift,
and the number of edges is half of that. each edge has an odd and an even endpoint.
the endpoints can be generated with this parity bit either explicit (as in cuckoo.rs) or implicit
(as in my repo).

tromp:
If you are finding 200k + solutions, then all but a handful should fail the verification check. Could there be another issue with your code?

Edgecombe:
> the mask is applied correctly. the number of nodes is 1<<sizeshift, and the number of edges is half of that. 

https://github.com/mimblewimble/grin/blob/master/core/src/pow/cuckoo.rs#L58

size is (1 << sizeshift) = 30 bits

https://github.com/mimblewimble/grin/blob/master/core/src/pow/cuckoo.rs#L94

easiness is (ease * size / 100) ~= 29.6 bits for ease=75

https://github.com/mimblewimble/grin/blob/master/core/src/pow/cuckoo.rs#L99

number of edges is easiness ~= 29.6 bits for ease=75

https://github.com/mimblewimble/grin/blob/master/core/src/pow/cuckoo.rs#L69

mask is size / 2 - 1 = 29 bits

https://github.com/mimblewimble/grin/blob/master/core/src/pow/cuckoo.rs#L103

https://github.com/mimblewimble/grin/blob/master/core/src/pow/cuckoo.rs#L77

nodes can either be even, or odd. number of nodes = 30 bits.

> the mask is applied correctly. the number of nodes is 1<<sizeshift, and the number of edges is half of that. 

[...]

PR #1027, opened 2018-05-01
fix broken pow

Edgecombe:
the reduced EDGEMASK creates algorithmic weaknesses in cuckoo, must be removed


Edgecombe:
see https://github.com/mimblewimble/grin/issues/1026

ignopeverell:
Those changes break the test vector. They've been generated using Tromp's reference implementation and we should stay compatible with that at the very least.

yeastplume:
The internal miner in core/src/pow is only there for unit/integration testing testing at smaller cuckoo sizes. All mining is done via cuckoo miner here: https://github.com/mimblewimble/cuckoo-miner, which uses @tromp's solvers, all of which have all been very well exercised and tested at this stage, and correctly validate within Grin. 

This being said, the mask is being divided by 2 in grin whereas it's not in @tromp's cuckoo implementation, although any discrepancies would only affect verification in Grin as things stand. It does appear the algorithms are implemented differently with slightly different parameters (tromp's uses edgebits=29 whereas grin uses sizeshift=30, etc,) and I'd need to dig into details of both implementations to figure out why (@tromp would have a better idea offhand). But given these changes break unit tests and we don't have any verification problems so far, I need more evidence that there's actually a problem before I start digging into it.

tromp:
As I commented on #1026, the mask is applied correctly. the number of nodes is 1<<sizeshift,
and the number of edges is half of that. each edge has an odd and an even endpoint.
the endpoints can be generated with this parity bit either explicit (as in cuckoo.rs) or implicit
(as in my repo).

Your change produces edge endpoints that are in the wrong partition half the time.

Edgecombe:
> But given that validation doesn't appear to be broken, and that these changes break unit tests, I need more evidence that there's actually a problem before I start digging into it.

dividing the mask by 2 removes a bit. since it is an even bit, all cycles that exist in the full mask version will also exist in the half mask version. verifier will correctly verify tromps sovlers solutions because the set of solutions for reduced mask space are a strict superset

tromp:
> verifier will correctly verify tromps sovlers solutions because the set of solutions for reduced mask space are a strict superset

No; the verify() code implicitly assumes that the u nodes are even and the v nodes are odd, which allowed me to remove the assumed parity bit from sipnode() in https://github.com/tromp/cuckoo/commit/f61b841b772e9b621c50e1e3bfdd920a72854f6a#diff-84b5042f21cd24450893db48f7a8dd32

Issue #1021, opened 2018-05-01
validate blocks via output_sum and kernel_sum (store them in rocksdb)


antiochp:
We do not validate the overall input|output|kernel sums when we validate a single block (via block propagation say).
Block validation just checks that inputs|outputs|kernels are internally consistent within the block.
This does not prevent a "double spend" from occurring as each block where a spend occurs is itself internally consistent.

See #1019 for a contrived example (where we corrupt the rm_log to effect a "double spend").

The double spend is only identified and rejected if we validate the full txhashset.

Proposal: 
* for each block store `output_sum` and `kernel_sum` in rocksdb 
    * note: running totals of the chain so far, not sums for the individual block
* block validation -
    * `output_sum - total_supply == kernel_sum + kernel_offset`
* block validation (with prev block): 
    * `prev_output_sum + new outputs == output_sum`
    * `prev_kernel_sum + new kernels == kernel_sum`
    * `prev_output_sum + new outputs - new inputs == kernel_sum + kernel_offset`   

In this way we can validate "proof of unspentness" (i.e. prevent double spend attempts) at the block validation level, _before_ we start applying the block to our local txhashset MMRs.

Putting this here so I don't forget it - 
> The output_sum commits to an amount equal to the total supply.
But _in addition_ it actually represents _where_ this value comes from - the commitment includes information about all the private keys of all the unspent outputs that make up this aggregate amount.
Hence - a double spend will invalidate this.

Related - #1020. Storing the total kernel_offset as a commitment makes the code a lot cleaner and removes the need to store and pass around the offset as a blinding_factor.
This fits in well with the "everything is just a commitment" tenet of Grin.

  * output_sum (commitment)
  * kernel_sum (commitment)
  * kernel_offset (commitment)

We do need to store the `kernel_sum` and the `kernel_offset` separately as we need the `kernel_offset` when validating the txhashset (we get the `kernel_sum` by summing the MMR contents).

But when validating individual blocks (or block headers) we will use `kernel_sum + kernel_offset`.

~Needless to say - all this is `consensus breaking` and cannot be added (easily) to testnet2.~

-------------

TASKS -
 
- [x] store `kernel_offset` on block header as commitment (see #1020)
- [x] store `output_sum` and `kernel_sum` running totals, for each block, stored in rocksdb
- [x] modify block validation to check `output_sum` against `kernel_sum`
- [x] modify block validation to perform necessary checks against previous block

ignopeverell:
[...]

PR #1031, opened 2018-05-03
[WIP] block validation via prev block_sum from rocksdb

antiochp:
See #1021 for details.

* store the following in a `block_sum` in RocksDB for each block hash
  * output_sum (commitment)
  * kernel_sum (commitment)
* validating a block now does the following - 
  * sum the inputs and outputs in the block and sum with the prev block output_sum
  * sum the kernels and apply the kernel_offset
  * check the sums match (i.e. apply the block to prev state and check everything sums correctly)
* reworked both block validation and txhashsert validaton to be more consistent
* txhashset write during fast sync now stores the initial `block_sum` for subsequent block validation

TODO - 
- [ ] fix tests in core
- [ ] cleanup code and get rid of warnings
- [ ] test locally in various scenarios
  - [ ] fast sync
  - [ ] archive mode
  - [ ] with txs


ignopeverell:
As discussed on Gitter and #1021 I think this is a good change to block validation. Couple remarks and questions:

1. Maybe delete old `block_sum`s on compaction? Doesn't seem we need them to stick around.
2. Could this be applied on master with a much smaller changeset for `consensus_breaking`? It'd be great to keep the diff size on `consensus_breaking` as much as possible.

antiochp:
(1) is already accounted for, happening here - 
https://github.com/mimblewimble/grin/pull/1031/files#diff-dc8555c8638c6e8bf7723ae776ce254bR652

(2) is a good idea. I'll work on splitting this into 2 PRs across the two branches.

ignopeverell:
Ah missed that part for 1), thanks!

PR #1033, opened 2018-05-03
Add a .editorconfig file

casey:
See editorconfig.org for plugins and details.

The only setting of note is to use hard tabs in rust source files. This is pretty nice for contributors who use soft tabs by default.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+# top-most .editorconfig file
+root = true
+
+# use hard tabs for rust source files
+[*.rs]
+indent_style = tab


antiochp:
Seems like a good idea to me.
I just installed editorconfig on atom to see what its like (never used it before).

antiochp:
@ignopeverell @yeastplume @quentinlesceller any reason not to merge this?

ignopeverell:
Not from me! Let's merge in a day if no one objects.

quentinlesceller:
No objection from me there. Let's merge this.

antiochp:
Giving our cherrypicking tool a go -
/cherrypick consensus_breaking

PR #1035, opened 2018-05-03
Wallet+Keychain refactoring

yeastplume:
Leaving this here early for review, starting to refactor the keychain and wallet to make sure concerns are properly separated, which should make it easier to refine the design.

The aim is to move everything except key management into the new libwallet mod from the keychain for now.. Some changes to aggsig are also needed to ensure their context is kept around (the earlier revision was relying on the keychain to provide a bit of persistence)

One issue I'm not sure how to address, a lot of the tests in core perform transaction building that relies on functions I've moved out into the wallet lib (as it would seem they properly belong there). We'll either have to move these tests elsewhere or perform some crate-management surgery.

----

Think this is ready for review now.. note that some tests that were previously unit tests now need to be considered integration tests, as they rely on transaction building functionality that's now in the wallet lib. The pool in particular required making quite a few more things public than I'd like, but the pool hasn't been touched in a while (while all other crates have changed and morphed quite a bit) and will probably need some attention at some stage.


yeastplume:
I've done some testing now and going to merge this.. I know it's a big massive thing to sit down and review at present, so if there are aspects of it that turn out less than ideal, I can revert them on an individual basis.

PR #1036, opened 2018-05-03
forgot to use the split key correctly

antiochp:
We introduced this in #1020.
Incorrectly using `skey` when we need to extract the secret key from `kernel_blind`.


antiochp:
Merging as we cannot create txs successfully on consensus_breaking branch without this fix.

PR #1038, opened 2018-05-06
[DO NOT MERGE] test

kargakis:


ignopeverell:
Nifty. We've been using [DNM] as well for prefix, is that configurable?

PR #1044, opened 2018-05-07
merge master into consensus_breaking

antiochp:
Merging `master` into `consensus_breaking` branch so we minimize divergence.


kargakis:
You can always use `/cherrypick [branch]` in individual PRs (even before the PR merges; see https://github.com/mimblewimble/grin/pull/1038#issuecomment-386879514) to avoid manual cherrypicks. Works only when there are no conflicts.

antiochp:
@kargakis yeah thanks - I keep forgetting to give that a try...

Issue #1037, opened 2018-05-05
Minimal Transaction Pool


antiochp:
Three different entities all basically look and act the same in Grin - 
* transactions: inputs, outputs, kernel(s), kernel_offset
* blocks: inputs, outputs, kernels, kernel_offset
* the txhashset: inputs (aka supply), outputs, kernels, kernel_offset

In each case we can treat the entity as a set of inputs, a set of outputs and a corresponding set of kernels.
In each case we can validate the commitments sum correctly with - 
```
sum_outputs - sum_inputs = sum_kernels + kernel_offset
```

In #1021 we are proposing a modification to the block validation process to validate against the full chain state by keeping "running totals" of the output_sum and kernel_sum at each block height.
We can then validate the sums for a new block; summing them together with the sum totals of the current txhashset.

To validate a full chain at height _h_ - 
```
output_sum(h) - total_supply(h) = 
kernel_sum(h) + kernel_offset(h)
```

To validate a new block against existing chain at height _h_ -
```
output_sum(h) + sum(block_outputs) - sum(block_inputs) = 
kernel_sum(h) + sum(block_kernels) + kernel_offset(h+1)
```

Interestingly we can do something _very_ similar for an individual tx, by validating the sums against the existing txhashset, again at height _h_ - 
```
output_sum(h) + sum(tx_outputs) - sum(tx_inputs) = 
kernel_sum(h) + sum(tx_kernels) + kernel_offset(h) + tx_kernel_offset
```

This validates that the given transaction, if summed with the current chain state would result in a chain state that was also valid.
In effect we have validated that the tx could be added in a new block.

This is one of the main responsibilities of the transaction pool.
And I think we can drastically simplify the internl implementation of the transaction pool by leveraging this "validate the sums against the existing chain state" approach.
I think we can even go as far as representing the tx pool as a simple Vec of transactions.

If the tx pool consists of `[tx1, tx2, tx3]` and these are valid, such that -
`txhashset + [tx1 + tx2 + tx3]` is valid then we can add a new `tx4` to the tx pool if
`txhashset + [tx1 + tx2 + tx3] + tx4` is also valid.

i.e. we can validate the sums with something along the lines of - 
```
output_sum(h) + sum(tx_pool_outputs + tx_outputs) - sum(tx_pool_inputs + tx_inputs) = 
[...]

PR #1049, opened 2018-05-07
[WIP] txhashset extension now implements Committed

antiochp:
We use the `Committed` trait for blocks and transactions to give a consistent way to sum commitments accounting for overage (fees or rewards).
This PR is an attempt to get `txhashset_extension` to also implement the `Committed` trait.

* Committed now returns vecs of commitments (not inputs|ouputs)
* `overage` reworked as a fn param to be usable by txhashset
  * txhashset now uses `overage` to represent the total_supply
* added `sum_kernel_excesses()` to Committed for convenience
* `verify_rangeproofs()` and `kernel.verify()` no explicitly called as necessary

This makes `blocks` and `transactions` and the full `txhashset extension` _very_ similar and the validation process is now relatively consistent across all three entities.
They are all containers holding vecs of `inputs|outputs|kernels` with a `kernel_offset`.


antiochp:
I think this could be taken a step further with the full `validate_sums()` or `validate_kernels()` pulled up into the `Committed` trait itself.

This way we only implement it once and it is reusable across blocks, transactions and the full txhashset_extension.

Note: as alluded to in #1037 I think we can leverage this to give us 80% of the functionality of the existing transaction pool.
So this will (in theory) by reusable across blocks, transactions, the full txhashset_extenson and the tx_pool impl (and presumably the Dandelion stem_pool).

ignopeverell:
I think the idea of generalizing `Committed` to be more widely applicable. However it would be good to do it without increasing too materially the amount of cloning that takes place. I believe that as-is, this PR would clound all input and output commitments within a block?

antiochp:
The existing impl does this (in `Committed`) - 
https://github.com/mimblewimble/grin/blob/ad6e7814de279d6428f351ae0cca4c7648ab45bb/core/src/core/mod.rs#L56-L57

The new impl moved this into `block` - 
https://github.com/antiochp/grin/blob/565547a4f2708714fbb7b5f2e43a96e16dedcf61/core/src/core/block.rs#L374-L376

So I'm not 100% sure this increases the amount of cloning going on?
Or are you meaning somewhere else?

I suspect it is just more visible now?
(Or very likely I'm not seeing something obvious here).

ignopeverell:
You're right, my bad, I missed that it was already done in block.

antiochp:
That said - maybe there is a way of doing this without the cloning going on.
But I'm coming up against the "borrow checker" playing around with this locally.

sesam:
👍 And I think any opt-level above the default 0 for dev buiilds allows the compiler to exorcise away (nearly) all superfluous cloning. Otherwise it's a nice [helpwanted] for someone to fix later on :)

PR #1054, opened 2018-05-09
Automatically trigger compaction every 2000 blocks

ignopeverell:
Does so in a separate thread as it can take some time. Also remove validation pre-compaction as it's hopefully unecessary now.


antiochp:
:+1: on removing the (hopefully redundant) pre-compaction validation step.

ignopeverell:
Tested it with the transition to block 62,000 and all seems fine.

PR #1061, opened 2018-05-11
Move aggsig transaction building functions into separate lib

yeastplume:
Continuing with wallet refactoring, the next logical step appears to be moving the transaction building functions from where they are into a separate library. Each function in the transaction building lib should just return the results of the previous stage (as well as whatever required context was created), and it's up to the caller to decide what to do with it. When this part is done, it should be easier to create an 'offline' version of a wallet transaction.

Note I've added the serialise trait to core::core Transaction, TxKernel and OutputFeatures, to allow for easier json serializing of a transaction.


yeastplume:
Still underway but going well.. I've added some transaction helper testing functions that operate directly on a chain instance (as opposed to having to back and forth via an http endpoint) that should make it far easier to test transactions and library functions... Once I have some tests in place and am convinced the current refactoring state won't break anything, I'll merge this and move on to the next phases of refactoring.

yeastplume:
Still plenty to do, but it's coming into shape... going to test this now against testnet2 and merge before moving on to the next phase. Just to document where my head is at with this, the code is now organised as follows:

* `grin/keychain` is just key management + commit and blinding factor creation (both of which I think should be moved out of there eventually)
grin/wallet is now structured as follows:
* `libwallet` should contain all of the transaction building functions, and should eventually have zero awareness of the wallet implementation. Its functions should work on keychains and key Identifiers (or lists of them). This is mostly true now, with the exception that a dependency on wallet::types still needs to be removed from transaction building. Callers to this lib should be responsible for how and when they want to save and manage their outputs.
* `grinwallet` contains common and oft-used functions specific to the current wallet implementation.. many chunks of code can still be moved from src/wallet/*.rs into there.

* `tests`:
-contains whatever tests were there previously in `libwallet.rs`.
-has new functionality in `common` that allow for building up wallets and building transactions directly against a chain instance, making it much easier to write predictable tests (as opposed to 'mine for a bit' between every operation).
-`transaction.rs` contains a single test that build up wallets and performs a single transaction, as is the best example of how using lib functions work in the current iteration. You can see a lot of code dealing with managing outputs within the wallet has been moved out of the transaction building functions into there, and this will start to look a bit cleaner as some of this functionality is factored away into 'grinwallet'.

Note all of these mods may be temporary depending where we eventually want to put all the code, they're just the easiest way to organise them for now.

That's the current state.. next steps are more refactoring of both libwallet and grinwallet to ensure proper division of concerns in libwallet and cleaning up some redundancy in grinwallet. Once that is done, then I think we'll be in a much better position to figure out how to possibly factor out functionality into common traits (i.e. a keychain must implement this and a wallet implementation must provide this,) and then figure out how to best structure them.

PR #1064, opened 2018-05-12
refactor on master, pass in offset as blinding_factor

antiochp:
Refactor - make `sum_kernel_excesses` take a `blinding_factor`.
So we generate the kernel_offset commitment in one central place.

Related - #1062.


ignopeverell:
Makes sense!

antiochp:
Merging.

PR #1023, opened 2018-05-01
[WIP] Dandelion transaction aggregation

quentinlesceller:
Do not merge yet.

Fix https://github.com/mimblewimble/grin/issues/693 and https://github.com/mimblewimble/grin/issues/819.

This PR introduces the aggregation of stem transactions before being included in a block.
The node currently maintains a pool of stem transactions called stempool.
Every x seconds (the value of the ```dandelion_patience```), the aggregator aggregates the stem transactions into a multi-kernel transaction with cut-through and broadcast it to its Dandelion relay as stem transaction. Hence the only way for a stem transaction to be broadcasted is through this aggregator. That way we do not have to maintain an aggregated_stem_tx_pool. 
To do:
- [x] Aggregator
- [ ] Stempool and pool reconciliation (attempt to deaggregate mk stem tx)
- [ ] Change the Dandelion relay time according to the other timer
- [ ] Tests

PR #1067, opened 2018-05-15
Minimal Transaction Pool

antiochp:
This is still very early - but wanted to push this up to get feedback before going any further on this.

The existing tx pool impl is pretty large (~2000 lines of code) and relatively complex.
This PR proposes an alternative solution - leveraging our `txhashset` to check if we can add a new tx to the pool (as an alternative to building and maintaining the "directed acyclic graph" of inputs/outputs in the existing impl).

 Basically - if aggregate new tx and the existing tx pool together we can check via a readonly txhashset extension is we can successfully add this single aggregated tx to the current chain state via the extension.

This covers all the usual checks -
  * attempting to spend non-existent outputs
  * attempting to "double spend" outputs
  * coinbase maturity checks

__But__ - we can leverage all our existing txhashset code to do all these checks, in the _exact same way_ we do when applying new blocks to the chain.

Possible downsides - 
  * Is this slower? How much slower?
  * Reconciling new blocks against the pool - we lose some of the benefits of the DAG here, when calculating which txs are still compatible with the more recent chain state and which txs must be evicted from the pool.

Not yet implemented - 
  - [x] this PR rips out (temporarily, work is progress) the stempool and Dandelion functionality - we need to reintroduce this (but I think we can follow the same pattern here, leveraging the txhashset extension validation to check for tx validity).

@quentinlesceller keen to hear your thoughts on this, specifically around the stem_pool behavior.
I think we can add txs to the stempool if we can aggregate the txs in the stempool with the txs in the regular pool and have this single aggregated tx successfully validate agains the txhashset.

* If `tx + pool + txhashset` is valid then we can add `tx` to the regular pool.
* If `tx + stem_pool + pool + txhashset` is valid then we can add `tx` to the stem_pool.

i.e. at any given time we know that we can take _all_ (or a subset) txs from the pool, put them all in a block and have that block validate successfully against the chain state (the txhashset).

This standardizes validation even further in Grin - all the following behave in the same way and we can validate them all consistently - 

  * an individual tx
  * a set of aggregated (and possibly cut-through) txs
  * the tx pool
  * the stem pool
  * an individual block
  * a set of blocks
  * full chain state (txhashset)

See #1037 for some early thoughts around this.

@MoaningMyrtle You wrote a lot of the initial pool implementation. 
Any strong objections to considering simplifying and consolidating the logic here?


pool/src/pool.rs
```diff
@@ -111,816 +70,112 @@ where
...
+	// If we can validate the aggregated tx against the current chain state
+	// then we can safely add the tx to the pool.
+	pub fn add_to_pool(&mut self, tx: &Transaction) -> Result<(), PoolError> {
+		let mut txs = self.transactions
+			.iter()
+			.map(|x| *x.clone())
+			.collect::<Vec<_>>();
+		txs.push(tx.clone());
+		let agg_tx =
+			transaction::aggregate_with_cut_through(txs).map_err(|_| PoolError::GenericPoolError)?;
```
quentinlesceller: Missing ```self.transactions.push(tx.clone())```? Don't want to bug you since it's a WIP ^^but figure I would write it anyway.

pool/src/transaction_pool.rs
```diff
@@ -0,0 +1,142 @@
...
+
+/// A minimal (EXPERIMENTAL) transaction pool implementation
+pub struct TransactionPool<T> {
+	/// Pool Config
+	pub config: PoolConfig,
+
+	/// Our transaction pool.
+	pub pool: Pool<T>,
+	/// TODO - Our stem transaction pool (just another instance of a pool?)
+	pub stem_pool: Pool<T>,
```
quentinlesceller: We can use the same pool instance or simplify it by creating another instance ```Stempool``` (which implement the ```Pool```) which is basically the same but with ```time_stem_transactions``` (to track the transactions). 
Also ```stempool``` and not ```stem_pool``` :) https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2018-May/015956.html.

ignopeverell:
Isn't this missing the spending of unconfirmed transactions? (may very well still be WIP, just thought I'd mention it)

antiochp:
@ignopeverell So actually no - what's happening is we are aggregating up all the txs in the pool and then attempting to validate the _aggregated tx_ against the txhashset.
Tx aggregation and cut-through should be handled in the usual way.

__Pros:__ There is only a single codepath for aggregating txs and validating unconfirmed txs. There is no risk of block/tx handling having subtle differences to the tx pool impl.
__Cons:__ Performance cost of aggregating a large tx for the tx pool and applying it to the txhashset.

Hope this makes my intent here a little clearer.
Thoughts?

ignopeverell:
Yes, sorry, I should have done more than skimming before commenting.

antiochp:
This is working locally now.
Still need to do some cleanup and get the pool tests working against the new impl.

@quentinlesceller Am I right in thinking we do not yet have the `patience` impl in place? i.e. we don't actually wait any time between receiving a stem tx and forwarding it to the next stem peer?
And this effectively means we will never actually aggregate any txs during the Dandelion stem phase?

antiochp:
@quentinlesceller ahh - just saw #1023 and realized this is where `patience` gets implemented.
In this PR I have moved the logic for handling "fluff a stem tx 10% of the time" into `dandelion_monitor` so we have a single place for handling timers and randomization.

tl;dr - we always add a stem tx to the local stempool.
We then let the dandelion_monitor decide if/when to propagate to the next Dandelion relay or if we want to fluff it (locally) then broadcast to the network.

I have not yet put the patience timer in place but this should be pretty trivial to add.

servers/src/grin/dandelion_monitor.rs
```diff
@@ -45,37 +47,110 @@ pub fn monitor_transactions<T>(
...
+				let mut fluff_stempool = false;
+				{
+					let mut tx_pool = tx_pool.write().unwrap();
+					let mut rng = rand::thread_rng();
+
+					for mut entry in tx_pool.stempool.entries.iter_mut() {
+						//
+						// TODO patience timer config
+						//
+						let cutoff = now - 10;
```
quentinlesceller: Are you planning, to make the patience based on each transaction ?
From what I understood, it should be a repetitive process every ```dandelion_patience``` seconds.
Just to be clear the ```cutoff``` is here because you run the process every 10 seconds instead of every second ? (Also why?)

pool/src/transaction_pool.rs
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		// for a basic transaction (1 input, 2 outputs) -
+		// (-1 * 1) + (4 * 2) + 1 = 8
+		// 8 * 10 = 80
+		if self.config.accept_fee_base > 0 {
+			let threshold = (tx.tx_weight() as u64) * self.config.accept_fee_base;
+			if tx.fee() < threshold {
+				return Err(PoolError::LowFeeTransaction(threshold));
+			}
```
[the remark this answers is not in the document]
antiochp: I think this simply prevents any "low fee" tx from ever being accepted by the tx pool.

antiochp:
See #1091 for a bug uncovered during this work.
There is test coverage here for this now.

quentinlesceller:
I will take a look later today.

servers/src/grin/dandelion_monitor.rs
```diff
@@ -41,120 +40,212 @@ pub fn monitor_transactions<T>(
...
 	let _ = thread::Builder::new()
 		.name("dandelion".to_string())
 		.spawn(move || {
 			loop {
+				if stop.load(Ordering::Relaxed) {
+					break;
+				}
+
+				// this is the patience timer, we loop every n secs.
+				let patience_secs = 10;
```
quentinlesceller: Maybe externalize this value ?

antiochp: Yeah - let me take a look at your PR and I can pull the config changes into this, unless you want me to merge this as-is and then you can open a PR to make the "patience timer" stuff more robust and configurable?

quentinlesceller: Having it all in this PR would be nice. But I can do it if you want, no preference of my side. 
Also speaking of that, the current values are 10 seconds of patience and 5 minutes of embargo. Which led to approximately 1m30s for a stem transaction to get in the mempool. Are you conformable with that? Should we reevaluate them?

pool/src/transaction_pool.rs
```diff
@@ -0,0 +1,175 @@
...
+		Ok(())
+	}
+
+	/// Add the given tx to the pool, directing it to either the stempool or
+	/// txpool based on stem flag provided.
+	pub fn add_to_pool(
+		&mut self,
+		src: TxSource,
+		tx: Transaction,
+		stem: bool,
```
hashmap: Unrelated, I saw the same flag in the previous version, but this boolean flag doesn't look good, perhaps we should refactor later on to make sure that method has one responsibility

pool/src/transaction_pool.rs
```diff
@@ -0,0 +1,175 @@
...
+			self.add_to_stempool(entry)?;
+		} else {
+			self.add_to_txpool(entry)?;
+		}
+		Ok(())
+	}
+
+	/// Reconcile the transaction pool (both txpool and stempool) against the
+	/// provided block.
+	pub fn reconcile_block(&mut self, block: &Block) -> Result<(), PoolError> {
```
hashmap: If I'm not mistaken the previous version was returning evicted transactions, do we still have such functionality?

antiochp: We do not - I simplified this for now because we were not using the return value (the evicted txs) for anything.
It should be easy enough to add this back in if desired in the future.

pool/src/transaction_pool.rs
antiochp: Yeah we can definitely revisit this. I went back and forth on this with the flag vs two separate functions. Landed on this for now to keep it closer to existing impl.

pool/src/transaction_pool.rs
```diff
@@ -0,0 +1,175 @@
...
+			stempool: Pool::new(chain.clone(), format!("stempool")),
+			blockchain: chain,
+			adapter: adapter,
+		}
+	}
+
+	fn add_to_stempool(&mut self, entry: PoolEntry) -> Result<(), PoolError> {
+		// Add tx to stempool (passing in all txs from txpool to validate against).
+		self.stempool
+			.add_to_pool(entry.clone(), self.txpool.all_transactions())?;
```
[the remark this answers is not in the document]
antiochp: I'm going to open an issue to revisit the tx pool and do another pass looking at cloning and ownership.
This impl works but there is definitely excessive cloning going on.

chain/src/txhashset.rs
```diff
@@ -574,28 +646,40 @@ impl<'a> Extension<'a> {
...
+			LOGGER,
+			"Rewind txhashset to output {}, kernel {}, height {}",
+			output_pos,
+			kernel_pos,
+			height
+		);
 
 		self.output_pmmr
-			.rewind(marker.output_pos, height as u32)
+			.rewind(output_pos, height as u32)
```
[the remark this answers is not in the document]
antiochp: I'm not actually 100% certain - we maintain height as a `u32` in a few places. But I don't think this is completely consistent across the codebase.

antiochp:
Merging. Hold on to your hats...

PR #1071, opened 2018-05-16
Remove redundant hashing of header before proof of work

### LINKED ISSUE #1066: redundant blake2b hashing on pre-pow header
header pow validates as Cuckoo::new(&bh.pre_pow_hash()[..], cuckoo_sz).verify(bh.pow.clone(), consensus::EASINESS as u64),

passing the blake2b hash of the pre-pow header as first argument, but Cuckoo::new considers its first argument a header rather than the siphash keys, and hashes it (again!) with blake2b.

ignopeverell:
We used to hash the hash of the header, needlessly. This also makes a few function signatures more explicit. Note that this is consensus breaking and should be accompanied with a similar change on the cuckoo-pow repository:
    
 https://github.com/mimblewimble/cuckoo-miner/commit/6fcecf516a111424d8bd5542fe20250e4f02cd8f
    
Fixes #1066


ignopeverell:
And please ignore all the commit history, for some reason git didn't fast-forward. The diff is clean though and it should be squashed.

yeastplume:
Bit of a confused commit history, but looks good to me. I’ll leave the cuckoo miner changes until close to t3, as the test miner should suffice for our testing

PR #1072, opened 2018-05-17
[WIP] Wallet refactor - part 3

yeastplume:
Continuing to refactor and rework the aggsig workflow to better separate concerns and create more usable library functions. Will give a fuller explanation of what's going on here as I get closer to commit time. For now the most interesting thing is the introduction of a transaction 'Slate', a term which I'm coining right now to mean the semi-public data that gets passed around between all parties involved in a transaction, all of whom add their parts to it until the data is complete. Once the slate data is complete, anyone should be able to build and post the final transaction from it.

The slate replaces the PartialTx struct (which wasn't really fully appropriate to aggregated transactions, the slate is designed with them in mind,) and will reside in the libwallet layer with helper functions to de/ser itself. Then callers should be able to pass the slate around by whatever means necessary to complete the transaction.


wallet/tests/transaction.rs
```diff
@@ -57,8 +57,125 @@ fn setup(test_dir: &str, chain_dir: &str) -> Chain {
 	).unwrap()
 }
 
+/// Build and test new version of sending API
+#[test]
+fn build_transaction_2() {
```
yeastplume: See this test fn for minimal transaction building workflow

ignopeverell: Looking nice!

ignopeverell:
A few remarks/questions:

* It might be worth materializing the `(WalletConfig, Keychain)` pair as a utility structure (perhaps just `Wallet`?) as the seem to very commonly go together and it seems cumbersome to pass them around one by one in most functions.
* The aggsig `ContextManager` could be added to that, but I'm unsure about its utility? Couldn't the consumers just hang on to the `Context` for their transactions? Perhaps there are usage flows I missed or am not considering here.

yeastplume:
Thanks! Agree on (WalletConfig, Keychain) pairing somehow as they almost always appear together... but will leave that for a bit later.. also the ContextManager is likely to become more and more pared down as I rework the aggsig, and could quite possibly go away or put somewhere else.

The next thing I'm working on is trying to really reduce the transaction API to its essence, and I think the notion of 'I'm the receiver' and 'I'm the sender' can be removed it in favour of "I'm participating and here are my inputs, outputs, contributions to the excess and signature," (and the kernel offset is introduced at the end) this should make the flow far more flexible and allow for better support of multi-party, multi-sig etc, and give participants an option of who actually does the posting. This should also enable the shortest possible route for a transaction (sender->reciever->Sender->post), for our carrier pigeon model.

wallet/src/libwallet/transaction.rs
```diff
@@ -18,37 +18,464 @@ use rand::thread_rng;
...
+			fee: 0,
+			height: 0,
+			lock_height: 0,
+			participant_data: vec![],
+		}
+	}
+
+	/// Adds selected inputs and outputs to the slate's transaction
+	/// Returns blinding factor
+	pub fn add_transaction_elements(
```
yeastplume: This is getting closer now to what I think it should look like... all parties involved add their inputs/outputs to the transaction, then all parties have to perform round 1, then all parties have to perform round 2 (using the provided functions). This workflow should scale easily to support multi-party/multi-sig scenarios.

core/src/core/transaction.rs
```diff
@@ -400,8 +393,9 @@ impl Transaction {
 
 	/// To verify transaction kernels we check that -
 	///  * all kernels have an even fee
-	///  * sum of input/output commitments matches sum of kernel commitments after applying offset
-	///  * each kernel sig is valid (i.e. tx commitments sum to zero, given above is true)
+	/// * sum of input/output commitments matches sum of kernel commitments
+	/// after applying offset * each kernel sig is valid (i.e. tx commitments
```
mably: Looks like a new line is missing here.

core/src/core/transaction.rs
```diff
@@ -480,10 +474,11 @@ impl Transaction {
...
 	/// We can verify the Merkle proof (for coinbase inputs) here in isolation.
-	/// But we cannot check the following as we need data from the index and the PMMR.
-	/// So we must be sure to check these at the appropriate point during block validation.
-	///   * node is in the correct pos in the PMMR
-	///   * block is the correct one (based on output_root from block_header via the index)
+	/// But we cannot check the following as we need data from the index and
+	/// the PMMR. So we must be sure to check these at the appropriate point
+	/// during block validation.   * node is in the correct pos in the PMMR
+	/// * block is the correct one (based on output_root from block_header
+	/// via the index)
```
mably: Missing new line.

core/src/core/transaction.rs
```diff
@@ -795,29 +792,33 @@ impl Input {
...
+	/// that block (may be older).
 	///
 	/// We can verify maturity of the output being spent by -
 	///
-	/// * verifying the Merkle Proof produces the correct root for the given hash (from MMR)
-	/// * verifying the root matches the output_root in the block_header
-	/// * verifying the hash matches the node hash in the Merkle Proof
-	/// * finally verify maturity rules based on height of the block header
+	/// * verifying the Merkle Proof produces the correct root for the given
+	/// hash (from MMR) * verifying the root matches the output_root in the
```
mably: Missing new lines.

yeastplume:
Still more refining to do, but given the number of changes and that it happens to be at a point where all tests are passing, I'm going to merge this now.

Issue #820, opened 2018-03-20
Store and communicate reasons for peer banning


yeastplume:
* Store reason why peer is banned along with rest of the peer data
* Figure out a way for peers to find out they've been banned and why
* **Bonus** show in TUI along with other peer information

casey:
This looks fun. If nobody else is working on it I'd like to grab it.

ignopeverell:
Go for it!

casey:
I ran into an issue ;_;

Some assumptions on my part:

1. The right way to solve this is to add a string field to p2p::store::PeerData, something like `last_ban_reason: String`.

2. Unless I'm mistaken, the serialization/deserialization in `core::ser` doesn't allow for schema evolution. I.E. if you want to add or remove fields to something that you're serializing, the serialization routines will then be incompatible with data serialized with the old routines, causing errors when reading data out of the store. This is in contrast to serialization formats like JSON or Protocol Buffers, where can detect when an object has extra or missing fields and do something sensible.

So, I could implement this by adding `last_ban_reason` to `PeerData`, but that would force users to clear their store lest they retain incompatible data. This is fine if it's a one-off event, but it seems likely that this will be a fairly regular occurrence, and it would be nice if updating to new versions didn't sometimes require users to clear their store.

Another option is to detect invalid data in the store and silently delete it, but this seems highly undesirable since it may mask legitimate errors.

My solution would probably be to switch to something extensible as the serialization format for things in the local store (like protocol buffers, flatbuffers, cap'n proto, json, or messagepack), which would then allow us to evolve schemas as necessary. If it's a compact, binary format it might be nice to also use it for P2P messages transmitted over the wire, but that's probably a separate discussion.

Thoughts?

casey:
I've been swamped with other work, so I haven't been able to give this the attention it deserves. If someone else is interested in grabbing it, they should definitely feel free.

quentinlesceller:
Okay got it.

PR #1085, opened 2018-05-23
Store and communicate ban reason

quentinlesceller:
Fix https://github.com/mimblewimble/grin/issues/820.
Store the ban reason and notify the remote peer of the ban reason.


quentinlesceller:
Tests are passing locally. No idea why Travis fails.

PR #1094, opened 2018-05-27
Migrate main node store to LMDB

ignopeverell:
In preparation to using LMDB as a wallet database, migrate the
node db. There's no point in having 2 key-value stores.
In addition LMDB provides a few advantages as a node db, namely a
much faster build (compared to RocksDb), lesser dependencies and
transactions.


ignopeverell:
Applying the `consensus_breaking` label even though it's strictly not, because it would cause enough migration mayhem that it might as well be, and should be tracked with the other consensus breaking PRs.

ignopeverell:
Included in and replaced by #1151

PR #1096, opened 2018-05-28
[WIP] Factoring out wallet traits and continued wallet library work

yeastplume:
The aim of this step is to:

* Rename what was previously 'libwallet' to 'libtransaction', as it deals with transaction building only.
* Rename what was 'grinwallet' to 'libwallet'.. libwallet will build on top of lib transaction, and contain types and traits needed by a wallet implementation, as well as helper functions that will operate on these types.
* Factor out traits from the current wallet implementation (WalletData) into a 'backend' that wallet implementations will need to provide
* Change all of the functionality in libwallet to operate upon these traits instead of a the current concrete 'WalletData'
* Provide a unified 'Wallet' container (not sure what to name it) that holds references to a keychain and backend together, and use this in all libwallet functions.


yeastplume:
I welcome your suggestion. Please enjoy a slightly reduced average line length.

yeastplume:
rabbitholing far more than expected but getting there

yeastplume:
At a working checkpoint now so just going to merge. Wallet Backend trait (might want to rename this) has been factored out, and next stage is to shift some more functions into libwallet as they’re no longer specific to the FileWallet implementation.

Getting there anyhow. More refinements coming, but it should be possible to slot in an LMDB backend just by implementing the WalletBackend trait now. It’s probably difficult to review the changes here but if anyone wants to look at the totality of how the wallet is starting to look and has suggestions I’m open to them.

PR #1112, opened 2018-05-31
[consensus breaking] apply block outputs first (simplification) and add verify_cut_through()

antiochp:
We were iterating over outputs twice, once to get the coinbase outputs (and applying them first).
Then again to get non-coinbase outputs and applying them _after_ we applied the inputs.

I'm pretty sure we can simplify this and just apply outputs first.
For completeness we now `verify_cut_though()` during both `block` and `transaction` validation (to ensure we are not spending our own outputs).


chain/src/txhashset.rs
```diff
@@ -508,30 +508,17 @@ impl<'a> Extension<'a> {
 	/// applied in order of the provided Vec. If pruning is enabled, inputs also
 	/// prune MMR data.
 	pub fn apply_block(&mut self, b: &Block) -> Result<(), Error> {
-		// first applying coinbase outputs. due to the construction of PMMRs the
-		// last element, when its a leaf, can never be pruned as it has no parent
-		// yet and it will be needed to calculate that hash. to work around this,
-		// we insert coinbase outputs first to add at least one output of padding
+		// A block is not valid if it has not been fully cut-through.
+		// So we can safely apply outputs first (we will not spend these in the same
+		// block).
```
ignopeverell: The reason for this wasn't quite about spending in the same block but may still not apply anymore. Say a block gets applied, with all its outputs. Then in the next block, the _last_ output from the previous gets spent. Trying to extract the peaks from that used to fail because that last pruned output was often involved in a peak, being rightmost. However now we go straight to the data file, without checking pruning, which I believe fixed that issue.

antiochp: Ahh that makes sense. I sort of understood that this was not specifically about spending from the same block...

antiochp:
Merging into `testnet3`.

PR #1113, opened 2018-05-31
Libwallet refactoring - Library functions + ErrorTypes

yeastplume:
* Moves checker functions into libwallet and renamed to updater
* Doing a pass on the Error handling, trying to ensure all parts have their own error types and are handled consistently throughout

One issue I'm having.. if you look at libwallet/types/WalletBackend at the read_wallet trait, I'm not sure what kind of error is best to return here. Implementors want to be able to throw their own errors, but libwallet has no business knowing about them. It could return a generic failure::Error, but this doesn't play very nice with the ErrorKind pattern we have going. If anyone has any ideas here they're be welcome.


hashmap:
@yeastplume what do you think about returning `Fail` trait? https://boats.gitlab.io/failure/fail.html

yeastplume:
@hashmap thanks for the suggestion. I've tried returning a Box<Fail> there, but then run into trouble trying to convert it to the right error because of our use of the ErrorKind Pattern (Fail doesn't implement Clone, Eq and PartialEQ so I can't use it as a parameter to a message).

Need to have more of a think about this.. I think perhaps this ErrorKind pattern is making things more cumbersome than they need to be.

yeastplume:
Think I have it all sorted now Error handling wise. For the time being, all the layers are returning their respective error types, but anywhere that expects to return any kind of underlying error can just return failure::Error... these can also be downcast further up to check for specific conditions (with an example in 
src/bin/grin.rs around line 595.

I've actually just put a status update on where I am on the forum, so I won't repeat it here. The state of the code is still far from perfect, but I think it's far enough along we can start building on it and refining the libs as we go. If anyone wants to have a proper review of it and has suggestions, now's a good time!

https://www.grin-forum.org/t/yeastplume-progress-update-thread-may-sept-2018/361/5

PR #1114, opened 2018-05-31
[consensus breaking] use hash_with_index(pos - 1) when checking the hash in the MMR

### LINKED ISSUE #1106: We should not allow duplicate commitments in the output MMR
For testnet2 we allow duplicate commitments via a __bug__ in `apply_output()` in the txhashset extension.
We call `hash()` when we should be calling `hash_with_index(pos - 1)`.

https://github.com/mimblewimble/grin/blob/b6c31ebc789be8eec01396dd2a21adada0126cb9/chain/src/txhashset.rs#L512-L525

antiochp:
Resolves #1106.


antiochp:
Merging into `testnet3`.

PR #1115, opened 2018-05-31
Simplify DandelionConfig

quentinlesceller:
Use only one ```DandelionConfig```.
@antiochp Let me know if I did something wrong here.

### DIFF
--- a/grin.toml
+++ b/grin.toml
@@ -73,7 +73,7 @@ run_test_miner = false
 #test miner wallet URL (burns if this doesn't exist)
 #test_miner_wallet_url = "http://127.0.0.1:13415"
 
-#[server.dandelion_config]
+[server.dandelion_config]
 #dandelion relay time (choose new relay peer every n secs)
 #relay_secs = 600
 
--- a/grin.toml
+++ b/grin.toml
@@ -73,7 +73,7 @@ run_test_miner = false
 #test miner wallet URL (burns if this doesn't exist)
 #test_miner_wallet_url = "http://127.0.0.1:13415"
 
-[server.dandelion_config]
+#[server.dandelion_config]
 #dandelion relay time (choose new relay peer every n secs)
 #relay_secs = 600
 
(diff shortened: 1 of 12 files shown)


antiochp:
No. This looks good.
Originally I thought the config needed passing down into both the p2p and pool components.
And the multiple configs was to avoid any cross dependencies.

grin.toml
```diff
@@ -73,7 +73,7 @@ run_test_miner = false
 #test miner wallet URL (burns if this doesn't exist)
 #test_miner_wallet_url = "http://127.0.0.1:13415"
 
-[server.dandelion_config]
+#[server.dandelion_config]
```
antiochp: Is this necessary to be commented out?
Feels like it will bite someone if they attempt to change a value below and don't uncomment this.

pool/src/types.rs
```diff
@@ -24,21 +24,51 @@ use core::core::transaction::Transaction;
...
+			relay_secs: default_dandelion_relay_secs(),
+			embargo_secs: default_dandelion_embargo_secs(),
+			patience_secs: default_dandelion_patience_secs(),
+			stem_probability: default_dandelion_stem_probability(),
+		}
+	}
+}
+
+fn default_dandelion_relay_secs() -> u64 {
+	600
```
antiochp: Can we pull these out into `consts` above so they are easier to find?
One more level of redirection but these fns are only really needed for serde (and should refer to the constants).

grin.toml
quentinlesceller: I don't know why but if it's uncommented it will ask for the other values when doing ```grin server run```. I will take a look though.

pool/src/types.rs
quentinlesceller: Yes reverting.

grin.toml
antiochp: Actually should be allowed in the toml file itself.
So either our code or the serde parsing code that is causing problems.

> Empty tables are allowed and simply have no key/value pairs within them.

https://github.com/toml-lang/toml#table

yeastplume: I'd leave the [server.dandelion_config] section in there uncommented and required, but then allow for the values themselves to be optional.

PR #1122, opened 2018-06-01
Rustify core/src/core

hashmap:
Small refactoring of one folder, if it makes sense I could extend the scope.
* Remove some cloning (real and just verbosity in the code)
* Naming conventions like to/into*
* copy_from_slice instead of manual copy
* Some Clippy's suggestions
* Some refactoring

I found that we don't use field init shorthand syntax, so I didn't touch this part, was it discussed before?

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -568,7 +568,7 @@ impl BlockHandler {
 			))?;
 		}
 		let vec = util::from_hex(input).unwrap();
-		Ok(Hash::from_vec(vec))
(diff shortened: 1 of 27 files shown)


ignopeverell:
I don't think field init shorthand syntax was available early on, but I've been using it more lately. So I'm not against it.

yeastplume:
Yes, definitely keep this up. It's easy to just resort to going `.clone()` when you're in the middle of a borrow checker fight and just want it to work, but going through the code methodically and reducing these instances is definitely a worthwhile endeavour.

PR #1123, opened 2018-06-01
fix the intermittent error in pool tests

### LINKED ISSUE #1121: pool tests failing intermittently
We fail occasionally/intermittently for things like this - 
```
failures:

---- test_transaction_pool_block_reconciliation stdout ----
	thread 'test_transaction_pool_block_reconciliation' panicked at 'index 160 out of range for slice of length 128', libcore/slice/mod.rs:785:5
```

As discussed on gitter, it looks like the height we are rewinding to is incorrect in `apply_raw_tx()`.
We were rewinding to `height`, when it seems like we need to be rewinding to `height - 1` (to rewind to the previous state correctly).

__But__ - even after making the change to rewind to `height - 1` we still fail for these `out of range for slice` errors intermittently.

Still investigating...

antiochp:
Resolves #1121.

With the "bug" that allows duplicate commitments in testnet2 we have some weird edge cases when we rewind outputs.
The new_output_commits local index is keyed by commitment (to pos) so can be overwritten by a duplicate output.

We were accounting for the case where the output MMR backend had an empty buffer.

But we were not accounting for the case where we would attempt to read bytes from the buffer based on a "stale" pos from new_output_commits local index.

This index is non-authoritative (similar to the db index) - so if we cannot read bytes from the buffer based on this pos, return empty vec.

### DIFF
--- a/pool/tests/block_reconciliation.rs
+++ b/pool/tests/block_reconciliation.rs
@@ -183,6 +183,10 @@ fn test_transaction_pool_block_reconciliation() {
 			}).unwrap();
 		}
 
+		let tip = Tip::from_block(&block.header);
+		chain.store.save_block_header(&block.header).unwrap();
+		chain.store.save_head(&tip).unwrap();
+
 		block
 	};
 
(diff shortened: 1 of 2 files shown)


antiochp:
Turns out the height logic was actually correct (if confusing) - https://github.com/mimblewimble/grin/pull/1123

PR #1126, opened 2018-06-01
Trying out random PoW by default

ignopeverell:
Avoid unintentional collisions which could account for some occasional test failures.


antiochp:
I think this change along with the change in #1128 should cover all the cases where we create new headers in the tests.

ignopeverell:
Not needed anymore.

PR #1131, opened 2018-06-02
move verify_kernel_sums into committed trait

antiochp:
* move `Committed` trait out into `committed.rs`
* add `verify_kernel_sums()` to committed trait
* add `overage()` and `total_overage()` to block header for convenience
* rework block, transaction and txhashset validation to all use `verify_kernel_sums()` from trait (they were all using their own slightly different variations of the same code)


antiochp:
No behavior changes with this - its just moving code around.
Makes transaction, blocks and txhashsets even closer now in terms of validation.

Merging.

kargakis:
Are kernels and kernel offsets documented anywhere?

antiochp:
Kernels yes - https://github.com/mimblewimble/grin/blob/master/doc/intro.md
Kernel offsets - I don't think so. Maybe I'll put something together over the next few days.

kargakis:
@antiochp thanks - I opened https://github.com/mimblewimble/grin/issues/1135 to track

PR #1132, opened 2018-06-04
Support shares

bladedoyle:
* Support logging miner shares including difficulty
* Support minimum mining difficulty configuration
* fix duplicate copies of the stratum server config

### DIFF
--- a/grin.toml
+++ b/grin.toml
@@ -124,6 +124,9 @@ stratum_server_addr = "127.0.0.1:13416"
 #header before stopping and re-collecting transactions from the pool
 attempt_time_per_block = 90
 
+#The minimum acceptable share difficulty to request from miners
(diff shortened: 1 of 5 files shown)


yeastplume:
That looks good for now, thanks. At some stage, we might want to consider accepting varying length cuckoo solutions for each hash as opposed to difficulty (which will give each client much more visibility). That's not currently in place in the underlying cuckoo-miner module, but since @tromp's plugins return all solutions instead of just the 42 cycle it shouldn't be to difficult to implement.

tromp:
We may also consider expressing the minimum share difficulty as a fraction of the bock difficulty,
e.g. a fraction 1/1000 will lead to 1000 shares reported on avg per block.

PR #1134, opened 2018-06-04
Replace try! for ?

quentinlesceller:
Based on #1122. Replace try for ?.


antiochp:
I like this - been doing this ad-hoc occasionally.
:+1:

Issue #919, opened 2018-03-31
Add a minimal Dockerfile


kargakis:
I am suggesting to move the base-build Dockerfile from [grin-docker](https://github.com/mimblewimble/grin-docker/blob/master/base-build/Dockerfile) inside this repo. It will be easier to keep it up to date and track grin's build dependencies. Even better build an image out of it as part of the CI testing of this repo.

sesam:
How would that affect TravisCI build times? We've been having trouble getting the CI to work for, rather than against, us.

Related:
https://github.com/mimblewimble/grin/issues/749
https://github.com/mimblewimble/grin/issues/772
https://github.com/mimblewimble/grin/issues/849
https://github.com/mimblewimble/grin/issues/190

kargakis:
The docker build could be a separate job if it ends up slowing down the main test suite. Unclear whether you can have more than one job with travis though. How is the gitlab work going? Maybe evaluate other options for extra jobs like CircleCI?

ignopeverell:
I agree with the suggestion. We can let people build the image themselves for now but a base Dockerfile could be handy for many.

yeastplume:
I don't think I agree with moving the docker file into the main Grin repo.. it seems a widely-adopted practice to keep docker builds in separate repositories, which makes sense to me as it's very much a layer on top of the core build. Getting people to use it manually is just a documentation issue, we just need to tell people that it's there and tell them how to build it.

The reason it keeps falling out of date is that no automated build is using it and there's no automated testing on it... I'd be in favour of having a nightly docker build that at least checks that it builds and puts the latest code onto docker hub.

kargakis:
> I don't think I agree with moving the docker file into the main Grin repo.. it seems a widely-adopted practice to keep docker builds in separate repositories, which makes sense to me as it's very much a layer on top of the core build.

I am working on production environments that base builds/deployments off of docker images and I can ensure you this is not true. Even though production images are using package managers to do installs, I don't think you are worried about that. There are two different modes optimized for development; one being bind-mounting local repos to simulate production installs and another one being bind-mounting the binaries directly (fastest). In both cases, dockerfiles are part of your source code. Rule of thumb: keeping build/deployment artifacts alongside your code is a best practice (spec files, dockerfiles, config files, what have you).

kargakis:
[...]

PR #1136, opened 2018-06-04
Add a minimal dockerfile

quentinlesceller:
Fix https://github.com/mimblewimble/grin/issues/919 by adding a minimal dockerfile.

Build instructions:
```
$ docker build -t grin .
$ docker run -it --rm --name grin_node grin
```

Still have weird character with the TUI despite having UTF-8 locales.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,13 +25,13 @@ RUN cargo build --release
 # runtime stage
 FROM debian:9.4
 
-RUN set -ex && \
-    apt-get update && \
-    apt-get --no-install-recommends --yes install locales && \
-    apt-get clean && \
-    rm -rf /var/lib/apt && \
-    sed -i '157 s/^##*//' /etc/locale.gen && \
-    locale-gen
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y locales
+
+RUN sed -i -e 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/' /etc/locale.gen && \
+    dpkg-reconfigure --frontend=noninteractive locales && \
+    update-locale LANG=en_US.UTF-8
+
+ENV LANG en_US.UTF-8
 
 COPY --from=builder /usr/src/grin/target/release/grin /usr/local/bin/grin
 COPY --from=builder /usr/src/grin/grin.toml /usr/src/grin/grin.toml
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN cargo build --release
+
+# runtime stage
+FROM debian:9.4
+
+RUN set -ex && \
+    apt-get update && \
+    apt-get --no-install-recommends --yes install locales && \
+    apt-get clean && \
+    rm -rf /var/lib/apt && \
```
ignopeverell: Just curious, why remove all apt data here? Smaller image file?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+
+# Building Grin
+RUN cargo build --release
+
+# runtime stage
+FROM debian:9.4
+
+RUN set -ex && \
+    apt-get update && \
+    apt-get --no-install-recommends --yes install locales && \
```
bladedoyle: Dont we still need ncurses library in the runtime image?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    apt-get clean && \
+    rm -rf /var/lib/apt && \
+    sed -i '157 s/^##*//' /etc/locale.gen && \
+    locale-gen
+
+COPY --from=builder /usr/src/grin/target/release/grin /usr/local/bin/grin
+COPY --from=builder /usr/src/grin/grin.toml /usr/src/grin/grin.toml
```
bladedoyle: Do we want to disable the TUI for running in a container?

Dockerfile
quentinlesceller: Yes, however I'm updating this part and we will not need to remove the apt files.

Dockerfile
quentinlesceller: It looks like we don't need ncurses in the runtime image (tested locally on macOS 10.13.5).

Dockerfile
quentinlesceller: Currently when you do ```docker run -it grin_app```it launch just like ```grin server run```.  This is just a really minimal docker image.

quentinlesceller:
Okay merging that now.

Dockerfile
quentinlesceller: Well actually, I'm leaving the TUI as it's really nice to be able to run in foreground with TUI.

PR #1140, opened 2018-06-05
Add doc and persistent storage for docker

quentinlesceller:
- Add documentation for the Dockerfile.
- Add the possibility to use a persistent storage. (Please note that in that case you'll need to copy paste the grin.toml in it)

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -20,16 +20,13 @@ you can start mining by building and runing grin-miner against your running Grin
         # Build using all available cores
         docker build -t grin .
 
-        # or build using a specific number of cores (reduce RAM requirement)
-        docker build --build-arg NPROC=1 -t grin .
-
-        # either run in foreground
+        # run in foreground
         docker run -it -v grin:/usr/src/grin grin
 
         # or in background
         docker run -it -d -v grin:/usr/src/grin grin
 
-If you decide to use a persistent storage you'll need the grin.toml file in it.
+If you decide to use a persistent storage (e.g. ```-v grin:/usr/src/grin```) you will need grin.toml configuration file in it.
 
 ## Requirements
 
--- a/doc/build.md
+++ b/doc/build.md
@@ -15,6 +15,22 @@ Please note that all mining functions for Grin have moved into a separate, stand
 [grin_miner](https://github.com/mimblewimble/grin-miner). Once your Grin code node is up and running,
 you can start mining by building and runing grin-miner against your running Grin node.
 
+## Docker
+
+        # Build using all available cores
+        docker build -t grin .
+
+        # or build using a specific number of cores (reduce RAM requirement)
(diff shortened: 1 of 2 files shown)


doc/build.md
```diff
@@ -15,6 +15,22 @@ Please note that all mining functions for Grin have moved into a separate, stand
 [grin_miner](https://github.com/mimblewimble/grin-miner). Once your Grin code node is up and running,
 you can start mining by building and runing grin-miner against your running Grin node.
 
+## Docker
+
+        # Build using all available cores
+        docker build -t grin .
+
+        # or build using a specific number of cores (reduce RAM requirement)
+        docker build --build-arg NPROC=1 -t grin .
```
kargakis: Do you think this is going to be used frequently? I would boil these instructions down to a single docker build & run with -d.

doc/build.md
```diff
@@ -15,6 +15,22 @@ Please note that all mining functions for Grin have moved into a separate, stand
...
+
+        # either run in foreground
+        docker run -it -v grin:/usr/src/grin grin
+
+        # or in background
+        docker run -it -d -v grin:/usr/src/grin grin
+
+If you decide to use a persistent storage you'll need the grin.toml file in it.
```
kargakis: You already use a volume for /usr/src/grin. Do you mean something else with "persistent storage"?

doc/build.md
quentinlesceller: Yeah I can probably remove the ```NPROC``` part. However, I think running in foreground and background is nice.

doc/build.md
quentinlesceller: I mean you can run it with or without the ```v``` option. If you decide to go with the ```v``` option you'll need a ```grin.toml```. Let me know if that's not clear and I will change the text.

kargakis: Ah, I just noticed what you do above. Is `VOLUME` in the Dockerfile necessary then since in both examples you bind-mount from the host? I forseee two different scenarios for running this image: one is for development purposes eg. a test suite that stress-tests a grin network, where you don't care about preserving state. Second case is an actual production deployment. In the second case, you either mount from host or you use cloud storage via an orchestrator like kubernetes. I would avoid depending on docker volumes.

kargakis: I think most people will want to bind-mount their host directories as in the examples you provide anyway.

doc/build.md
quentinlesceller: Okay I simplified a bit this part and added a bit of explanation.

doc/build.md
quentinlesceller: Yes, actually after reading https://docs.docker.com/storage/ and https://docs.docker.com/engine/reference/builder/#volume it looks like the ```VOLUME```instructions is not necessary. I'm currently testing and will push if everything works.

Issue #1127, opened 2018-06-02
compactor error and consecutive erros in peer and stratum_server


chri2:
Logfile can be found [here](http://z9.de/log-2018-06-02.bz2).

@jaspervdm wrote on gitter that he'd created a block containing >2000 inputs just before this happened. Could be coincidence, but the first error (compactor) occured just one block before receiving block 94001 with >2000 inputs.

Afterwards peer threw errors each time block 94001 has been received.

After stop and restart of grin server the server synced up to the top of the chain without further problems.

Grin version 0.2.0 (git testnet2-135-gb6c31eb), built for x86_64-unknown-linux-gnu by rustc 1.26.0 (a77568041 2018-05-07). Built with profile "release", features "" on Wed, 30 May 2018 20:09:48 GMT.

ignopeverell:
Can you see if the following reproduces the issue?

```
curl -X POST 127.0.0.1:13413/v1/chain/compact
```

If so, it'd be great to have a stacktrace on a debug build.

chri2:
Didn't reproduce. But also: There is no logging on level trace about anything related to that http-post.

chri2:
Happend again at block height 98000. Log is [here](http://z9.de/log-2018-06-05.bz2).

antiochp:
I reproduced this locally a couple of times - investigating here as well.
It looks to be related to the data file during compaction the 2nd time, the hash file appears to behave correctly.

PR #1141, opened 2018-06-06
[DNM] add failing test to cover case where we compact a previously compacted MMR 

antiochp:
Test case that reliably reproduces the error during compaction.
This happens when we compact the output MMR, spend some subsequent outputs and then compact the (previously compacted) MMR a second time.

Related #1127.

Staring at the code for a while and don't see what's happening yet - but at least we can reproduce this now on a small test MMR.

cc/ @ignopeverell


antiochp:
Closing - fixed in #1143.

PR #1142, opened 2018-06-06
Factor out wallet communications

yeastplume:
Hopefully the last explicit wallet refactoring PR for the time being, the aim of this one being to factor out all inter-wallet and wallet-node communication into a single trait which can be used within libwallet whenever there needs to be communication. A new WalletClient trait will be used within libwallet, while the communication implementation details can be moved out into implementation-specific code.


yeastplume:
All done except for the wallet restore functions (which will be rewritten before T3 anyhow)

PR #1152, opened 2018-06-08
 Missing type annotation keychain

quentinlesceller:
Fix the missing type annotation error when building Grin:
```
error[E0283]: type annotations required: cannot resolve `_: keychain::Keychain`
   --> src/bin/grin.rs:543:16
    |
XXX|         let wallet = FileWallet::new(wallet_config.clone(), passphrase).unwrap_or_else(|e| {
    |                      ^^^^^^^^^^^^^^^
    |
    = note: required by `<wallet::FileWallet<K>>::new`
```


ignopeverell:
Thanks for the fix!

PR #1153, opened 2018-06-10
Fix the way shares are validated

bladedoyle:
It seemed easy to call this existing function, so I did.

But maybe its very inefficient to create a new Cuckoo object for each share and should only build one for each new block?


yeastplume:
Looks good

PR #1154, opened 2018-06-11
Externalize PMMR test

quentinlesceller:
With over 1000 lines, the PMMR tests are quite big.
This PR externalize them in ```tests/pmmr.rs```.


antiochp:
:+1: that file was pretty huge.

hashmap:
I'm late to the party and not sure if it's fully applicable here, but tests in `src` and `tests` folders have different semantics. Unit tests are supposed to be in in `src` as we had in `pmmr.rs`, `tests` folder is for integration tests[1]. It seems to me that we deal here with unit tests which now looks like integration ones.

[1] https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html

antiochp:
Hmm - I'm not sure I fully agree with the recommended approach from https://doc.rust-lang.org/book/second-edition/ch11-03-test-organization.html
I think that makes sense if the files are reasonably small but its harder to deal with when the file is in the 000s of lines.

It also feels kind of arbitrary to have "unit test" in src files and "integration tests" in a separate tests dir.
That way you're looking in two different places for tests. 
And it potentially gets confusing once you've got tests that blur the boundary between unit and integration - as everything outside trivial unit tests always begin to do.
 
I'd much rather organize this around separate files under the tests dir (and face the wrath of the rust community).

quentinlesceller:
I agree with @antiochp, everything is way more readable as two separated files (especially considering that the tests were creating temporary ```struct```).

PR #1158, opened 2018-06-11
Typo fix

quentinlesceller:
~~Was reading through core code and *had* to fix these typos.~~
Thought I would check the entire code for typos.


CODE_OF_CONDUCT.md
```diff
@@ -11,11 +11,11 @@ Please be kind and courteous. There’s no need to be mean or rude. Respect that
 
 Please keep unstructured critique to a minimum. If you have solid ideas you want to experiment with, make a fork and see how it works.
 
-We will exclude you from interaction if you insult, demean or harass anyone. That is not welcome behaviour. We interpret the term “harassment” as including the definition in the [Citizen Code of Conduct](http://citizencodeofconduct.org/); if you have any lack of clarity about what might be included in that concept, please read their definition. In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
+We will exclude you from interaction if you insult, demean or harass anyone. That is not welcome behavior. We interpret the term “harassment” as including the definition in the [Citizen Code of Conduct](http://citizencodeofconduct.org/); if you have any lack of clarity about what might be included in that concept, please read their definition. In particular, we don’t tolerate behavior that excludes people in socially marginalized groups.
```
antiochp: Depends if we are using US spelling 😄

core/src/core/pmmr.rs
```diff
@@ -269,7 +269,7 @@ where
 	/// The last position in the PMMR
 	pub last_pos: u64,
 	backend: &'a mut B,
-	// only needed for parameterizing Backend
+	// only needed to parametrise Backend
```
antiochp: I'm pretty sure "parametrise" is not right?
Old and not in common use any more?

CODE_OF_CONDUCT.md
yeastplume: I'm okay with standardizing on American spelling throughout, it's generally the path of least resistance.

core/src/core/pmmr.rs
yeastplume: Not right, just a missing 'e' typo

quentinlesceller: Both are correct but I will update for ```parameterize```.

CODE_OF_CONDUCT.md
quentinlesceller: No strong opinion on that one. Let me know what you prefer.

ignopeverell: Even if I try hard, I'm pretty sure I'll let some US spelling slip through... for the words I can spell, that is.

Issue #1107, opened 2018-05-31
Store kernel_sum on BlockHeader


antiochp:
These will be very useful for lightweight nodes.
If we trust the BlockHeaders then we know what the sums should be at any given height, allowing us to take a txhashset and verify it for any given height etc.

We currently store the sums on the BlockSum struct stored in the db by block hash.
We can remove this and rely on the sums on the headers themselves.

A block header should commit to these as part of the pow.

antiochp:
As per #1163 we are going to store kernel_sum and the sizes of the kernel and output MMRs on the block header.

antiochp:
Resolved in #1163.

PR #1163, opened 2018-06-12
Kernel sum and MMR sizes in block header

### LINKED ISSUE #1162: fast sync is insecure
In the current fast sync approach, we take the MMR commitments in the horizon header as the indisputable truth (as secured by the cumulative pow difficulty). But an attacker could have branched off from the previous block (or the one before that) and even mined 2000 blocks on top, to make up arbitrary MMRs at horizon,
which are presented to the syncer along with the horizon header. The syncer merely checks that the MMRs
have the correct commitment, but doesn't check their past evolution.

To be certain of the kernel MMR that the horizon cumulative diff is committed to, the syncer needs to go through all intermediate kernel MMR sizes at each height, compute their commitment, and check that this matches the one in the header at that height. Here it might help to have #kernels as a separate header field.

ignopeverell:
Adds the kernel sum total and output and kernel MMR sizes to block headers. This allows validation of all intermediate kernel roots during fast sync. Also allows getting rid of the `BlockSum` index.

Fixes #1162, #1107, and #1165.


core/src/core/block.rs
```diff
@@ -125,6 +127,13 @@ pub struct BlockHeader {
 	/// We can derive the kernel offset sum for *this* block from
 	/// the total kernel offset of the previous block header.
 	pub total_kernel_offset: BlindingFactor,
+	/// Total accumulated sum of kernel commitments since genesis block.
+	/// Should always equal the UTXO commitment sum minus supply.
+	pub total_kernel_sum: Commitment,
```
antiochp: Just clicked that we don't actually need (or want) the `total_output_sum` in here.
We can derive this from the total supply (block height).

core/src/core/block.rs
```diff
@@ -576,24 +596,17 @@ impl Block {
 
 		// now sum the kernel_offsets up to give us
 		// an aggregate offset for the entire block
-		let total_kernel_offset = {
+		kernel_offsets.push(prev.total_kernel_offset);
+		let total_kernel_offset = committed::sum_kernel_offsets(kernel_offsets, vec![])?;
```
antiochp: Yes :+1:.

core/src/core/block.rs
```diff
@@ -675,26 +689,40 @@ impl Block {
...
 
-		let sums = self.verify_kernel_sums(
-			self.header.overage(),
-			self.header.total_kernel_offset(),
-			Some(prev_output_sum),
-			Some(prev_kernel_sum),
-		)?;
+		// take the kernel offset for this block (block offset minus previous) and
+		// verify outputs and kernel sums
+		let block_kernel_offset = if self.header.total_kernel_offset() == prev_kernel_offset.clone()
```
antiochp: Nice - this is a better conditional check for a zero offset (no txs).

antiochp:
Maybe done as a separate PR but I think we can get rid of `BlockMarker` here also as we can derive the "markers" from the MMR sizes at each block?

antiochp:
Also I think the changes specific to fixing #1165 can be backported to master as they don't depend on the header changes?

PR #1172, opened 2018-06-15
validate block against previous total_kernel_offset

### LINKED ISSUE #1165: block validation should validate block in isolation (not total sums)
Right now we validate a block at height n via the following (paraphrasing slightly) - 

```
"total output_sum at height n-1" + "output sum of block at height n" 
= "total kernel_sum at height n-1" + "kernel sum of block at height n" + "total kernel offset at height n"
```

This can (and should) be reworked as the following - 

```
"output sum of block at height n" 
= "kernel sum of block at height n" + "total kernel offset at height n" - "total kernel offset at height n - 1"
```

i.e. instead of summing the block with the total so far, we just check the sums of the block itself (and offset the kernel by the difference in offsets between the block and the previous block).

If everything goes smoothly (as it should) then these two approaches are interchangeable.

From @ignopeverell on gitter - 

> ... but every new fast sync is going to check the total sum and we can have some nodes regularly doing a full validation to do the same, but nothing is going to do a per-block check and I'd rather it being reversed
> if we have a bug in summing and/or tracking those sums, I'd rather both be checked, albeit in different scenarios
> in addition, per-block checks are slightly simpler overall, and simpler also makes it less likely to have an issue

antiochp:
Validate a block against previous `total_kernel_offset`.
This removes the need for `block_sums` and allows us to get rid of these from the db/index.

Resolves #1165.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -23,14 +23,13 @@ use std::time::{Duration, Instant};
 use core::core::hash::{Hash, Hashed};
 use core::core::pmmr::MerkleProof;
 use core::core::target::Difficulty;
-use core::core::Committed;
 use core::core::{Block, BlockHeader, Output, OutputIdentifier, Transaction, TxKernel};
 use core::global;
 use grin_store::Error::NotFoundErr;
 use pipe;
 use store;
 use txhashset;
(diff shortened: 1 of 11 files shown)


antiochp:
Closing - superseded by #1163.

Issue #1171, opened 2018-06-15
txhashset.zip can contain arbitrary files


antiochp:
Currently we just zip the local `.grin/txhashset` dir up and send it across (and implicitly trust we have no local files in there that should not be exposed to other nodes).
And on the receiving side we just unzip it (and implicitly trust there is nothing unwanted or malicious in there).

We should probably explicitly list the contents of the zip file on both ends and filter everything else out.

Related - #1147 (proposing to generate snapshot utxo files and send those over to support rewind).

sesam:
In Steve Gibsons latest, "Security Now 667 Zippity Do... or Don't" has related infotainment.
This is probably a very prioritised issue.
If you worry about security, wipe your boxen - or learn to love living with (intentionally or not permanently root-infested hardware.

quentinlesceller:
I got this.

PR #1174, opened 2018-06-15
Check content before zip/unzip the txhashset

quentinlesceller:
Fix https://github.com/mimblewimble/grin/issues/1171.

- Check the content of the txhashset dir before zip/unzip.
- Add tests
- Rename ```test_txhashset_raw_txs.rs``` → ```test_txhashset.rs```


antiochp:
Hoping to get #1147 in before we tackle this - then we can revisit the list of files that the `txhashset.zip` should actually contain.

quentinlesceller:
Technically, with https://github.com/mimblewimble/grin/pull/1174/files#diff-12bff24bc654a13261536be1084749a8R31, this method is "future-proof".

antiochp:
With #1147 we will need to support a more dynamic list of files.
The new `pmmr_utxo.bin` file will be rewound and a snapshot written to disk with a block hash as a suffix on the filename (we need to rewind it _before_ we send it, as we cannot rewind it after receiving it before we process full blocks).
So we will need to support passing in a block header (or similar) as part of the txhashset creation process, and the filtering of files will need to be aware of this.

quentinlesceller:
I see I will update accordingly.

ignopeverell:
What's the status on this PR? It'd be really nice to have, is it just missing some sort of wildcard pattern matching? Do we know the exact name the snapshot will have (I think it includes the height)?

quentinlesceller:
Yeah, I'll try to finish this PR ASAP.

quentinlesceller:
@ignopeverell @antiochp, Just updated what do you think?

chain/src/txhashset.rs
```diff
@@ -1072,14 +1071,76 @@ pub fn zip_read(root_dir: String) -> Result<File, Error> {
...
+		return Err(ErrorKind::Other(
+			"Unexpected file(s) found in txhashset folder".to_string(),
+		).into());
+	}
+
+	// Checking now every subdirectory
+	let pmmr_files_expected: HashSet<_> = PMMR_FILES
+		.iter()
+		.cloned()
+		.map(|s| if s.contains(PMMR_FILES[2]) {
```
antiochp: This `PMMR_FILES[2]` is easy to miss - better to just be explicit about the filename itself?

quentinlesceller:
Sorry for the delay catching up asap.

chain/src/txhashset.rs
```diff
@@ -1072,14 +1071,76 @@ pub fn zip_read(root_dir: String) -> Result<File, Error> {
...
+	check_files(&txhashset_path, header)?;
 
 	fs::create_dir_all(txhashset_path.clone())?;
 	zip::decompress(txhashset_data, &txhashset_path)
 		.map_err(|ze| ErrorKind::Other(ze.to_string()).into())
 }
 
+/// Check that the txhashset directory does not contains unexpected files
+fn check_files(txhashset_path: &PathBuf, header: &BlockHeader) -> Result<(), Error> {
+	let subdirectories_expected: HashSet<_> = [OUTPUT_SUBDIR, KERNEL_SUBDIR, RANGE_PROOF_SUBDIR]
```
hashmap: do we need to always create the set? perhaps build it once?

chain/src/txhashset.rs
```diff
@@ -1072,14 +1071,76 @@ pub fn zip_read(root_dir: String) -> Result<File, Error> {
...
+		.iter()
+		.cloned()
+		.map(|s| String::from(s))
+		.collect();
+	let subdirectories_found: HashSet<_> = fs::read_dir(txhashset_path)
+		.unwrap()
```
hashmap: `unwrap` looks risky

chain/src/txhashset.rs
```diff
@@ -1072,14 +1071,76 @@ pub fn zip_read(root_dir: String) -> Result<File, Error> {
...
+
+	// Checking now every subdirectory
+	let pmmr_files_expected: HashSet<_> = PMMR_FILES
+		.iter()
+		.cloned()
+		.map(|s| if s.contains(PMMR_FILES[2]) {
+		format!("{}.{}", s, header.hash())}
+		else {String::from(s) })
+		.collect();
+	let subdirectories = fs::read_dir(txhashset_path).unwrap();
```
hashmap: `unwrap` here and below

chain/src/txhashset.rs
```diff
@@ -1072,14 +1071,76 @@ pub fn zip_read(root_dir: String) -> Result<File, Error> {
...
+	for subdirectory in subdirectories {
+		let subdirectory_path = subdirectory.unwrap().path();
+		let pmmr_files = fs::read_dir(subdirectory_path).unwrap();
```
hashmap: this code looks like a copy code above starting from line 1090, can we make euse it somehow?

chain/src/txhashset.rs
```diff
@@ -1072,14 +1071,76 @@ pub fn zip_read(root_dir: String) -> Result<File, Error> {
...
+			})
+		})
+		.collect();
+	let dir_difference: Vec<String> = subdirectories_found
+		.difference(&subdirectories_expected)
+		.cloned()
+		.collect();
+	if !dir_difference.is_empty() {
+		return Err(ErrorKind::Other(
+			"Unexpected file(s) found in txhashset folder".to_string(),
```
hashmap: Should we care about missing files?

quentinlesceller:
@ignopeverell @antiochp @hashmap finally found time to finish this. Now we filter the content of the txhashset folder on both sides.
When we zip, we create a copy of the txhashset dir, we remove any unexpected files from there and zip it.
When we receive a zip, we unzip it and filter out unexpected files.

antiochp:
:+1: Awesome!

PR #1178, opened 2018-06-18
Update Stratum Server to use rust Result for returning errors

bladedoyle:


servers/src/mining/stratumserver.rs
```diff
@@ -344,30 +342,33 @@ impl StratumServer {
...
-					} else {
-						let resp = RpcResponse {
-							id: request.id,
-							jsonrpc: String::from("2.0"),
-							method: request.method,
-							result: Some(response),
-							error: None,
-						};
-						rpc_response = serde_json::to_string(&resp).unwrap();
+					match response {
```
ignopeverell: You could reduce this to one single code path with something like:

```
result: response.ok()
result: response.err()
```

PR #1186, opened 2018-06-21
[DNM] include the (optional) Merkle proof and block hash when generating input hash

### LINKED ISSUE #1185: tx input malleability (block_hash and merkle_proof excluded from hash)
For inputs spending coinbase outputs we include a block hash and a Merkle proof in a tx input to verify "maturity".
Is it possible today to silently swap a Merkle proof (or block hash) out and invalidate a tx while it is in the txpool (and propagate the now invalid version)?

https://github.com/mimblewimble/grin/blob/343904180734dc2f5af9c1a9d5965612a3abbc1d/core/src/core/transaction.rs#L665-L682

We follow a similar pattern with outputs where the _rangeproof_ is excluded in the hash.
But in the case of rangeproofs we commit to the output commitment (so the rangeproof cannot be invalidated in this way).

I don't recall if this was done intentionally for inputs or not...

antiochp:
If the input is malleated (for example by modifying the Merkle proof) I believe we want the hash to change. Otherwise a node can malleate a tx (via an input) without changing the tx hash.

Resolves #1185.

Not actually sure if this is consensus breaking or not.
We do not actually store inputs in the chain state anywhere (only in full blocks).
Block hashes only consider block headers.
Input hashes (I think) are only referred to in the txpool when dealing with hydrating compact blocks.
So I don't think this affects consensus but not 100% sure of this...

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -667,14 +667,12 @@ impl Writeable for Input {
 		writer.write_u8(self.features.bits())?;
 		self.commit.write(writer)?;
 
-		if writer.serialization_mode() != ser::SerializationMode::Hash {
-			if self.features.contains(OutputFeatures::COINBASE_OUTPUT) {
-				let block_hash = &self.block_hash.unwrap_or(ZERO_HASH);
-				let merkle_proof = self.merkle_proof();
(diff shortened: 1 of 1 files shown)


antiochp:
Suspect this is about to be made redundant by #1191.

antiochp:
Superseded by #1203.

PR #1187, opened 2018-06-21
[DNM] proof of concept - conditional locktime on output

antiochp:
One of the outstanding problems in Grin/MW is how to do "relative locktimes" on tx outputs.

We currently have "unconditional" lock_heights on txs (tx is not valid and will not be accepted until lock_height has passed).

Given _unconditional locktimes on txs_ I think we can leverage these to give us _conditional locktimes on outputs_ by "entangling" two outputs on two related txs together.

1) Private keys can be summed together.
Key<sub>3</sub> = Key<sub>1</sub> + Key<sub>2</sub>

2) Commitments can be summed together.
C<sub>3</sub> = C<sub>1</sub> + C<sub>2</sub>

So we can (I believe) construct two txs (Tx<sub>1</sub>, Tx<sub>2</sub>) each with one output Out<sub>1</sub> and Out<sub>2</sub> such that -
  * Out<sub>1</sub> (commitment C<sub>1</sub>) is from Tx<sub>1</sub> and built using Key<sub>1</sub> 
  * Out<sub>2</sub> (commitment C<sub>2</sub>) is from Tx<sub>2</sub> and built using Key<sub>2</sub>
  * Tx<sub>2</sub> has an unconditional lock_height on it
  
If we can do this (and we can manage the keys sufficiently) -
  * Out<sub>1</sub> + Out<sub>2</sub> can be spent as a pair using Key<sub>3</sub> (the original private key)

And (this is critical) -
  * They can _only_ be spent after lock_height from Tx<sub>2</sub>

Tx<sub>1</sub> (containing Out<sub>1</sub>) can be broadcast, accepted and confirmed on-chain immediately.
Tx<sub>2</sub> cannot be broadcast and accepted until lock_height has passed.
So if Alice, say, only knows K<sub>3</sub> and does not know Key<sub>1</sub> or Key<sub>2</sub>, then Out<sub>1</sub> can only be spent after lock_height has passed.
Say Bob, on the other hand, knows Key<sub>2</sub> then they can spend Out<sub>1</sub> immediately.

We have a "conditional" timelock on Out<sub>1</sub> (confirmed, on-chain) 
where it can be spent either with Key<sub>3</sub> (after lock_height), _or_ Key<sub>2</sub> immediately.

So I think the problem can be restated to one of key management across multiple parties -
Can we do the necessary multi-party tx construction in such a way that we do not expose Key<sub>1</sub> or Key<sub>2</sub> to Alice?

\* This does not solve for the problem of "relative" locktimes, but I think there are probably ways to do this also.

### DIFF
--- a/wallet/src/libtx/build.rs
+++ b/wallet/src/libtx/build.rs
@@ -32,6 +32,7 @@ use core::core::pmmr::MerkleProof;
 use core::core::{Input, Output, OutputFeatures, ProofMessageElements, Transaction, TxKernel};
 use keychain::{self, BlindSum, BlindingFactor, Identifier, Keychain};
 use libtx::{aggsig, proof};
+use util::secp::pedersen::Commitment;
 use util::LOGGER;
 
 /// Context information available to transaction combinators.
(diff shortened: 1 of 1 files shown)


yuntai:
you mean?
```
Tx1 (containing Out1) can be broadcast, accepted and confirmed on-chain immediately.
Tx2 cannot be broadcast and accepted until lock_height has passed.
So if Alice, say, only knows K3 and does not know Key1 or Key2, then Out1 can only be spent after lock_height has passed.
Say Bob, on the other hand, knows Key1 then they can spend Out1 immediately.

We have a "conditional" timelock on Out1 (confirmed, on-chain)
where it can be spent either with Key3 (after lock_height), or Key1 immediately.
```

antiochp:
@yuntai yes! Fixed.
Thanks! (took me a minute to see what was different in your copy...)

antiochp:
Closing - documented this for future investigation in #1190.

PR #1193, opened 2018-06-24
Accept multiple Cuckoo graph sizes

ignopeverell:
Adds the cuckoo size shift in the block header and validate taking it into account. Implements @tromp's proposal:

https://lists.launchpad.net/mimblewimble/msg00494.html


ignopeverell:
Couple things still missing but ready for a first review. What's missing is:

* The PoW hash is calculated based on the serialized representation of `Proof`, which right now isn't right. I need to override `hash` to base it solely on the `Vec` of nonces.
* Not very happy with VLQ+delta encoding as it's still slightly larger and more complex. Going to try an _exact_ bit by bit next (so Cuckoo30 is 42*30 bits plus 4 zero bits to byte-align the whole thing).

ignopeverell:
Note: automatic min increase should come as a separate PR that formalizes the activation rule + wait period scheme.

wallet/src/libtx/build.rs
```diff
@@ -118,13 +118,7 @@ where
 			let commit = build.keychain.commit(value, &key_id).unwrap();
 			trace!(LOGGER, "Builder - Pedersen Commit is: {:?}", commit,);
 
-			let rproof = proof::create(
```
yeastplume: Seems to be a lot of un-rustfmtting here?

PR #1203, opened 2018-06-26
verify coinbase maturity via output_mmr_size

### LINKED ISSUE #1185: tx input malleability (block_hash and merkle_proof excluded from hash)

### LINKED ISSUE #1191: timelocked coinbase outputs (relative timelock against coinbase kernel)
If/when we implement #1190 I suspect we can rework our "coinbase maturity" handling follow the same approach.

Currently we "lock" coinbase outputs for 1,000 blocks by requiring a Merkle proof to prove the age of the output being spent.
I _think_ we can rework this to take advantage of the fact we know which _kernel_ is involved in the original coinbase output being created, and require a kernel Merkle proof (rather than an output Merkle proof) on the tx kernel spending it (and not the input spending the output as we have today).

This would also give us the following advantages - 
* if we have a tx spending many coinbase outputs we would only need a single kernel Merkle proof (against the most recent coinbase kernel) to prove they have all matured sufficiently.
Right now we require an output Merkle proof on each one individually (on the individual inputs).
* We sign tx kernels and kernels are not pruned, so this would be potentially be verifiable, even for historical validators.

The coinbase reward (output) for mining a block is associated with a corresponding coinbase kernel.
We can prove maturity against this kernel (and not the output as we do currently).

Related - #1190 and #1187.

antiochp:
We do not need Merkle proofs to spend coinbase outputs.
We only need the `output_mmr_size` from the block header.

* verify coinbase maturity via output pos vs. `output_mmr_size` committed to in block header
* wallet does not need to be concerned with Merkle proofs or block hashes

TODO - 
- [x] any more cleanup?
- [x] test with real running nodes

Resolves #1191.
Resolves #1185.


yeastplume:
Very happy with these simplifications if it all works.. refresh_missing_block_hashes in particular is a bit source of slow-down when trying to use the wallet.

antiochp:
Going to merge this. 
Tests are all passing on travis.
Testing locally with a couple of nodes indicates this is all still working as expected without needing Merkle proofs on the inputs.

PR #1210, opened 2018-07-01
Introduce state to track sync stages

### LINKED ISSUE #1209: Fast sync: show progress in TUI
During fast sync, the chain stays at height 0 for a while because it has to verify all the range proofs. It would be nice if the TUI indicates that it is currently verifying them and if it would show how far along it is

ignopeverell:
A new sync status enum encapsulated in a state struct allows
tracking of where sync is at. Leveraging it in the TUI to provide
more helpful messages.

Still left: see if we can get some percentage completion on state
validation based on the number of kernels and range proofs.

Fixes #1209

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -30,7 +30,7 @@ use grin_store::Error::NotFoundErr;
 use pipe;
 use store;
 use txhashset;
-use types::{BlockMarker, BlockSums, ChainAdapter, ChainStore, Error, Options, Tip};
+use types::{BlockMarker, BlockSums, ChainAdapter, ChainStore, Error, Options, Tip, TxHashsetWriteStatus};
 use util::secp::pedersen::{Commitment, RangeProof};
 use util::LOGGER;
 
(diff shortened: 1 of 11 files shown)


yeastplume:
Looks good... I might suggest throwing the SyncStatus into `servers/src/common/stats.rs` to sit alongside the rest of them, and you're probably already on this but adding percentage complete args to each enum variant to make as best a guess we can for each stage. Since it'll probably be harder to get an overall percentage completion, perhaps we should try to give a numeric indication of what step we're on e.g. (Step 2/5: 43% Complete)

ignopeverell:
I'm on the fence about throwing `SyncStatus` in stats. At first I thought it was a good idea, but then `is_syncing()` is a fairly important test in a few parts of the server that's not really stats dependent. What do you think?

Looking at percentage completions now. It may be more straighforward for some steps than others, I'll see how far I can go.

yeastplume:
All fine, I won't bikeshed about the location of SyncStatus. And I don't think we need to stress too hard about percentage completions for each stage being 100% accurate ATM, they can be refined over time if needed. Showing anything progress related is far better than showing nothing.

PR #1213, opened 2018-07-03
avoid potential 'attempt to add with overflow' panic

yuntai:


ignopeverell:
This looks like it could still be valid, isn't it?

Issue #1207, opened 2018-06-29
txhashset validation failing on testnet3 branch


antiochp:
Fast-sync appears to be failing currently on `testnet3`.

We see a `StoreErr(NotFoundErr, "wrapped")` in the logs so I suspect something got clobbered with all the PR merges going on.

```
Jun 28 21:03:47.811 DEBG Header head before txhashset request: 83 / 4bbcf88a
Jun 28 21:03:47.811 DEBG Asking 127.0.0.1:13414 for txhashset archive at 64 8a3379a6.
Jun 28 21:03:47.880 DEBG handle_payload: txhashset archive for 8a3379a6 at 64 rewind to 152/138
Jun 28 21:03:47.903 DEBG leaf_set: copying rewound file .grin/txhashset/output/pmmr_leaf.bin.8a3379a6 to .grin/txhashset/output/pmmr_leaf.bin
Jun 28 21:03:47.904 DEBG leaf_set: copying rewound file .grin/txhashset/rangeproof/pmmr_leaf.bin.8a3379a6 to .grin/txhashset/rangeproof/pmmr_leaf.bin
Jun 28 21:03:47.910 DEBG txhashset: validated the output|rproof|kernel mmrs, took 0s
Jun 28 21:03:48.050 DEBG txhashset: verified 71 kernel signatures, pmmr size 138, took 0s
Jun 28 21:03:52.312 DEBG txhashset: verified 33 rangeproofs, pmmr size 152, took 4s
Jun 28 21:03:52.313 DEBG Error returned, discarding txhashset extension: StoreErr(NotFoundErr, "wrapped")
Jun 28 21:03:52.314 ERRO Failed to save txhashset archive: StoreErr(NotFoundErr, "wrapped")
Jun 28 21:03:52.314 DEBG handle_payload: txhashset archive for 8a3379a6 at 64, DONE
```

Have not had time to investigate but it appears to be happening somewhere in the txhashset validation step.

yeastplume:
After a round of error handling improvements via https://github.com/mimblewimble/grin/pull/1208, the output is now:

```
Jun 29 15:47:21.648 DEBG txhashset: verified 222 rangeproofs, pmmr size 438, took 17s
Jun 29 15:47:21.872 DEBG Error returned, discarding txhashset extension: Store Error: NotFoundErr("BLOCK HEADER: ffffffff")                        
 Cause: DB Not Found Error: BLOCK HEADER: ffffffff           
[...]

PR #1214, opened 2018-07-03
[WIP] TXHashset Validation problems

yeastplume:
In chasing down the cause for #1207 on testnet3, I've uncovered a couple of issues, one an easy fix and the other needs some re-thought. Simple fix is below, but problem is as follows:

During validation of the kernel history after getting the txhashset during a fast-sync, rewind now relies upon building a bitmap of a particular block (through `get_block_input_bitmap`)... however, in the fast-sync case no block exists, just the header, so it would seem there's currently no way to rewind in order to do the validation.

Is this an oversight or am I missing something?

### DIFF
--- a/chain/src/txhashset.rs
+++ b/chain/src/txhashset.rs
@@ -1066,15 +1066,15 @@ impl<'a> Extension<'a> {
 		// header, rewind and check each root. This fixes a potential weakness in
 		// fast sync where a reorg past the horizon could allow a whole rewrite of
 		// the kernel set.
+		let header_head = header.clone();
 		let mut current = header.clone();
 		loop {
 			current = self.commit_index.get_block_header(&current.previous)?;
 			if current.height == 0 {
 				break;
 			}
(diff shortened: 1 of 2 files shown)


chain/src/txhashset.rs
```diff
@@ -1150,9 +1150,11 @@ fn input_pos_to_rewind(
 		// I/O should be minimized or eliminated here for most
 		// rewind scenarios.
 		let current_header = commit_index.get_block_header(&current)?;
-		let input_bitmap = commit_index.get_block_input_bitmap(&current)?;
-		found = input_bitmap.0;
-		bitmap.or_inplace(&input_bitmap.1);
+		let input_bitmap_res = commit_index.get_block_input_bitmap(&current);
```
yeastplume: Note this is wrong.. in fast sync case `get_block_input_bitmap` is returning an error due to no block or bitmap existing in the DB.

yeastplume:
https://github.com/mimblewimble/grin/pull/1218

Issue #1212, opened 2018-07-02
T3 Initial Difficulty Adjustment 


yeastplume:
Below is the output from running some launch difficulty scenarios, over-estimated, under-estimated etc, as well as the actuals from the launch of Testnet 2. I don't these are moving fast enough, as (particularly in the first case,) you would have expected the network difficulty to keep dropping by half much longer than it does.

@tromp could you take a look through these and perhaps suggest what needs to be done here?

```
running 1 test
adding create: 1530535738, 1000000
*********************************************************
Scenario 1) Grossly over-estimated genesis difficulty 
*********************************************************
Height: 0, Time: 1530535738, Interval: 0, Network difficulty:1000000
Height: 1, Time: 1530539338, Interval: 3600, Network difficulty:1000000
Height: 2, Time: 1530542938, Interval: 3600, Network difficulty:500000
Height: 3, Time: 1530544738, Interval: 1800, Network difficulty:375000
Height: 4, Time: 1530546538, Interval: 1800, Network difficulty:312500
Height: 5, Time: 1530547438, Interval: 900, Network difficulty:273437
Height: 6, Time: 1530548338, Interval: 900, Network difficulty:246093
Height: 7, Time: 1530549238, Interval: 900, Network difficulty:225585
Height: 8, Time: 1530550138, Interval: 900, Network difficulty:210050
Height: 9, Time: 1530551038, Interval: 900, Network difficulty:197336
Height: 10, Time: 1530551938, Interval: 900, Network difficulty:184937
Height: 11, Time: 1530552838, Interval: 900, Network difficulty:176246
Height: 12, Time: 1530553738, Interval: 900, Network difficulty:168954
Height: 13, Time: 1530554638, Interval: 900, Network difficulty:161255
Height: 14, Time: 1530555538, Interval: 900, Network difficulty:153749
*********************************************************
adding create: 1530535738, 1000000
*********************************************************
Scenario 2) Grossly under-estimated genesis difficulty 
*********************************************************
Height: 0, Time: 1530535738, Interval: 0, Network difficulty:1000000
Height: 1, Time: 1530535739, Interval: 1, Network difficulty:1000000
Height: 2, Time: 1530535740, Interval: 1, Network difficulty:1487603
Height: 3, Time: 1530535741, Interval: 1, Network difficulty:1850283
Height: 4, Time: 1530535742, Interval: 1, Network difficulty:2151017
[...]

PR #1222, opened 2018-07-05
Simulnet test fix

yeastplume:
Cleanup following from https://github.com/mimblewimble/grin/issues/1212 LOWER_TIME_BOUND is redundant and can never be attained with the current adjustment algorithm and timestamp resolution.


ignopeverell:
Not sure if it's Travis or not, but I've restarted the tests a couple times and one of them doesn't seem to be terminating?

yeastplume:
It would seem removing the lower time bound is causing some issues with the test miner, so will just leave it in for now. This just fixes the simulnet test (code of which hasn't changed for a good while, travis has just decided to start acting up over it now, for some reason)

Issue #1215, opened 2018-07-03
Investigate reason for high number of **rejected** mined shares


bladedoyle:
A fast miner (1080 ti GPU) sees a high number of both rejected and stale shares.  ex:
 "974 accepted, 44 rejected and 25 stales"

A slow miner (1-core on an i7 cpu) sees far fewer stale shares but still sees many shares rejected due to invalid POW.  ex:
"109 accepted, 16 rejected, 0 stale"

We need to understand why the shares are **rejected**:
* does the miner return invalid pow?
* does the data get corrupted in transit?
* is there some issue with the validation code?
* something else?

---
The grin.log messages associated with rejected shares:
```
Jun 30 00:25:29.786 DEBG pipe: process_block b67f0bcb at 126038 with 0 inputs, 1 outputs, 1 kernels
Jun 30 00:25:29.787 ERRO pipe: validate_header failed for cuckoo shift size 30
Jun 30 00:25:29.787 INFO Rejected block b67f0bcb at 126038: InvalidPow
Jun 30 00:25:29.789 ERRO (Server ID: StratumServer) Failed to validate solution at height 126038: InvalidPow
```

yeastplume:
This appears to be happening just after the `attempt_time_per_block` timeout expires and the server builds a new block with a new timestamp and nonce. If grin-miner returns a solution for the previous block, it's rejected as invalid (presumably because the nonce it returns doesn't match the new block).

yeastplume:
No, scratch that.. it seems to be happening every time a solution is returned very quickly after the job is sent out, which is very odd (why should that pow not verify in that case?)

yeastplume:
This (and potentially some cases of #1216) appear to have been caused by reading the wrong solution from an underlying solution output queue.. so if two solutions are found in rapid succession and one meets difficulty, the next job would read an old solution from the queue. Fixed this within cuckoo-miner and the latest version of grin-miner, so will monitor for a while now to see if this occurs again.

ignopeverell:
Nice! Thanks for looking.

bladedoyle:
I think (at least part of) the problem is in the Stratum Server.

The stratum server just uses whatever header a miner submits and attaches that to the "current block". If the stratum server just built a new block, and the miner sends a solution for the previous one, the block wont validate (even if none of the transactions changed the timestamp will).

1) The stratum server will need to version the templates it sends - probably by adding an "id" field to the job template struct.
AND
[...]

PR #1230, opened 2018-07-09
[WIP] add job_id to stratum job and solution rpc

bladedoyle:
Warning: This changes the RPC protocol.
This is for https://github.com/mimblewimble/grin/issues/1215
Please merge at the same time as grin-miner PR https://github.com/mimblewimble/grin-miner/pull/20

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -28,8 +28,8 @@ use chain;
 use common::adapters::PoolToChainAdapter;
 use common::stats::{StratumStats, WorkerStats};
 use common::types::{StratumServerConfig, SyncState};
-use core::core::{Block, BlockHeader};
-use core::{pow, global};
+use core::core::Block;
+use core::{global, pow};
 use keychain;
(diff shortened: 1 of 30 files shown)


servers/src/mining/stratumserver.rs
```diff
@@ -445,12 +449,18 @@ impl StratumServer {
 				return Err(serde_json::to_value(e).unwrap());
 			}
 		};
-
-		let mut b: Block;
+		let mut b: Block = Block::default();
```
bladedoyle: Im not really happy about creating a throwaway block here.  Is it a performance concern?  I guess it would be better to use Option<Block> and then return invalid share error for None.  I'll make that change unless somebody has a better idea.

bladedoyle:
```Jul 09 20:49:40.043 DEBG Received message: {"id":"0","jsonrpc":"2.0","method":"status","result":{"accepted":171,"difficulty":1,"height":1666,"id":"0","rejected":0,"stale":5},"error":null}```

PR #1231, opened 2018-07-09
Add cuckoo size to block API

### LINKED ISSUE #1229: Add cuckoo size to block api
Now that the cuckoo size is no longer fixed, it would be nice if it gets added to the block api (`/v1/blocks/XX` where `XX` is the block number)

jaspervdm:
Closes #1229. I'm not sure on the naming of the field, so if anyone has a better suggestion I can rename it


jaspervdm:
I just noticed that I used the wrong branch, sorry. Redoing it with the correct branch

Issue #1189, opened 2018-06-23
cannot compile librocksdb-sys v5.11.3 under WSL


mcdallas:
I get the following error when trying to build on WSL (Ubuntu 14.04)

> mike@Mike-ZenBook:~/projects/grin$ cargo build --release
> warning: An explicit [[bin]] section is specified in Cargo.toml which currently
> disables Cargo from automatically inferring other binary targets.
> This inference behavior will change in the Rust 2018 edition and the following
> files will be included as a binary target:
> 
> * /home/mike/projects/grin/src/bin/client.rs
> 
> This is likely to break cargo build or cargo test as these files may not be
> ready to be compiled as a binary target today. You can future-proof yourself
> and disable this warning by adding `autobins = false` to your [package]
> section. You may also move the files to a location where Cargo would not
> automatically infer them to be a target, such as in subfolders.
> 
> For more information on this warning you can consult
> https://github.com/rust-lang/cargo/issues/5330
>    Compiling librocksdb-sys v5.11.3
>    Compiling croaring-sys v0.3.1
> error: failed to run custom build command for `librocksdb-sys v5.11.3`
> process didn't exit successfully: `/home/mike/projects/grin/target/release/build/librocksdb-sys-142584821dfa73fb/build-script-build` (exit code: 101)
> --- stdout
> cargo:rerun-if-changed=build.rs
> cargo:rerun-if-changed=rocksdb/
> cargo:rerun-if-changed=snappy/
> rocksdb/include/rocksdb/c.h:48:9: warning: #pragma once in main file, err: false
> 
> --- stderr
> thread 'main' panicked at 'function not loaded: clang_Type_getNumTemplateArguments', /home/mike/.cargo/registry/src/github.com-1ecc6299db9ec823/clang-sys-0.21.2/src/lib.rs:1456:1
> note: Run with `RUST_BACKTRACE=1` for a backtrace.
> 
> warning: build failed, waiting for other jobs to finish...
> error: failed to run custom build command for `croaring-sys v0.3.1`
> process didn't exit successfully: `/home/mike/projects/grin/target/release/build/croaring-sys-c3ed2cfb8ca9e426/build-script-build` (exit code: 101)
> --- stdout
> TARGET = Some("x86_64-unknown-linux-gnu")
> OPT_LEVEL = Some("3")
> TARGET = Some("x86_64-unknown-linux-gnu")
> HOST = Some("x86_64-unknown-linux-gnu")
[...]

PR #1237, opened 2018-07-09
specify minimum clang version

bladedoyle:
Update build docs to specify minimum clang version needed - see https://github.com/mimblewimble/grin/issues/1189


antiochp:
We're no longer using `rocksdb`. It was replaced with `lmdb`.
And I think the clang version needs to be `3.8.0` now as the croaring bitmap crate needs this...

bladedoyle:
@antiochp testnet3 failed to build croaring with clang 3.8.0.   I tested 5 and 6 and both of those worked.

PR #1241, opened 2018-07-10
Remove Iron dependency and update hyper to version 0.12

### LINKED ISSUE #876: [Enhancement] Removing hyper/iron dependencies on Api library.
Currently Api library depends on hyper and iron with underlying Future and Tokio crates for async processing of http requests between the server and the wallet, and between wallets. The dependencies are heavy and iron is not actively maintained after hyper dumped to 0.11. Also grin uses iron for server and hyper for client, which causes conflict on error handling between the server and the wallet. (refert to https://github.com/mimblewimble/grin/issues/642). 

The requirements of Api library is Http protocol and json support for the server and the client. And the traffic is not significant. The data communication of wallet is limited by the function of locking the wallet data file. So no need of heavy duty web server. And the same for the communication between the server and the wallet.

The goal for this enhancement is easy maintenance with the requirements above fulfilled. The target http stack is a single (or very limited number of) thread(s) based on std.net, std.io, http crate and httparse create.

e-max:
Fixes #876  
Work in progress. We still need to create a test suit for api,  fix usage of this API in wallet and implement shutdown method for ApiServer.


hashmap:
Some random notes explaining why code may looks weird:
* Router is designed to be cpu cache friendly
* Static/unsafe handler of request was done to prevent router to be cloned per request, so now we have just one static instance of it
* Client provides old synchronized interface keeping all async nature of new Hyper under the hood.

sesam:
nice work!

nitpick, needless //commented line:

+ b/api/src/client.rs
@@ -15,12 +15,17 @@
 //! High level JSON/HTTP client API

 use failure::{Fail, ResultExt};
+//use http::uri::InvalidUri;
+use http::uri::{InvalidUri, Uri};

kargakis:
> Router is designed to be cpu cache friendly
> Static/unsafe handler of request was done to prevent router to be cloned per request, so now we have just one static instance of it

Would be nice to see these as comments in the code.

yeastplume:
Thanks for taking this on, and it's looking good. I'll do a more detailed review in a while.

I think we're going to have to do quite a bit of separate testing before merging this. We can talk amongst ourselves about how best to handle this, but this will probably deserve its own branch for a while.

hashmap:
Ready for review.

yeastplume:
Okay, thanks for all of the hard work on this! I've reviewed and tested and most of the core api and wallet functionality appears to work fine. However I've still come across a few issues, which is fair enough given the number of things that were re-implemented.

Given the size of this PR and to lessen any further re-basing headaches (of which you've had many so far, apologies for that,) I'm going to merge this now and then open up separate issues for what I come across in testing. Thanks again!

PR #1242, opened 2018-07-10
Wallet API Test Client

yeastplume:
Trying to implement a test client that acts on an in-memory chain directly, so as to make exercising the wallet API in tests far easier. Also some changes to the underlying API structs to support this.

### DIFF
--- a/servers/tests/framework/mod.rs
+++ b/servers/tests/framework/mod.rs
@@ -335,7 +335,7 @@ impl LocalServerContainer {
 			.unwrap_or_else(|e| panic!("Error creating wallet: {:?} Config: {:?}", e, config));
 		wallet.keychain = Some(keychain);
 		let _ =
-			wallet::controller::owner_single_use(Box::new(wallet), |api| {
+			wallet::controller::owner_single_use(Arc::new(Mutex::new(Box::new(wallet))), |api| {
 				let result = api.issue_send_tx(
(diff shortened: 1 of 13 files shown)


yeastplume:
Attempt 2, this one is working far better than the previous iteration. 

There's now a 'WalletProxy', which acts as a shim between a local chain instance and many test wallets, and a 'LocalWalletClient', which implements the WalletClient trait and sends messages to the WalletProxy, which either simulates a server api action (refresh_outputs, post_tx, etc) or simulates sending a message to another wallet, which in turn invokes a wallet API, performs the given action and responds, simulating an HTTP request/response.

This should allow for much better coverage of the Wallet code, and will allow us to test most functionality that's been difficult to test so far (sending/receiving transactions, building coinbase transactions, querying wallet info and checking heights are correct)... since it's operating against a local chain we can control how many blocks are mined and thus expect precise results to test against (as opposed to having a thread mining unpredictably). Tests can also be run without opening all sorts of ports, making it more CI friendly.

The `transaction_api` test in `wallet/tests/transaction.rs` demonstrates how to set up a wallet test against a local chain.

yeastplume:
Think this is done for now. Some changes to how wallet locks are handled within the API as a result of testing, and tests in place to test the DB wallet implementation... should be able to add more wallet tests as needed now.

PR #1245, opened 2018-07-10
Add blockfound response in stratum

quentinlesceller:
Add a ```BlockFound```message in Stratum to notify the worker that the submitted share is a valid solution for the block.

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -89,6 +89,11 @@ pub struct JobTemplate {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
+pub struct BlockFound {
+	hash: String,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
 pub struct WorkerStatus {
 	id: String,
 	height: u64,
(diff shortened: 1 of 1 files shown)


servers/src/mining/stratumserver.rs
```diff
@@ -88,6 +88,11 @@ pub struct JobTemplate {
 	pre_pow: String,
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+pub struct BlockFound {
```
bladedoyle: I dont think we really need a BlockFound struct if there will be only a single value.

servers/src/mining/stratumserver.rs
```diff
@@ -539,6 +544,12 @@ impl StratumServer {
 				LOGGER,
 				"(Server ID: {}) Solution Found for block {} - Yay!!!", self.id, params.height
 			);
+			worker_stats.num_accepted += 1;
```
bladedoyle: This will be a double count since we already count the share below

servers/src/mining/stratumserver.rs
```diff
@@ -539,6 +544,12 @@ impl StratumServer {
...
+			let block_found = BlockFound {
+				hash: b.hash().to_hex(),
+			};
+			let block_found_json = serde_json::to_string(&block_found).unwrap();
+			return Ok((serde_json::to_value(block_found_json).unwrap(), share_is_block))
```
bladedoyle: We cant return here since that will skip the final step of logging all the details of the accepted share.

servers/src/mining/stratumserver.rs
quentinlesceller: It depends on your other comment, but we can add the same logging as below and keep the return thus the count here. Since we might want to keep the ok response anyway. Or we could have a submit response message with the information that the share was a valid solution.

quentinlesceller: Actually found a simpler way. Update incoming.

Issue #1249, opened 2018-07-11
missing field "job_id"


emmick4:
Trying to run the grin-miner alongside the node, not sure where the bug actually lies. It's spitting out an error as soon as it receives its first job. Happy to test any fixes and provide any info I can. Running on debian amd64.

This is the error from the grin-miner:
```
Jul 11 16:47:53.008 DEBG Received message: {"id":"Stratum","jsonrpc":"2.0","method":"job","params":{"difficulty":1,"height":3803,"pre_pow":"00010000000000000edb05784f3b370b9a5aa5dfa2$

Jul 11 16:47:53.009 DEBG Received request type: job
Jul 11 16:47:53.064 ERRO
thread 'client_controller' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `job_id`", line: 0, column: 0)': libcore/result.rs:945stack backtrace:
   0:     0x5642be930b6c - backtrace::backtrace::trace::hcb6d74f522eafe9d
   1:     0x5642be930192 - backtrace::capture::Backtrace::new::h5075cb690ef81509
   2:     0x5642be81f54f - grin_miner_util::logger::send_panic_to_log::{{closure}}::h8dea98af9827d598
   3:     0x5642be943ac5 - std::panicking::rust_panic_with_hook::h55e258944e31bc6d
                        at libstd/panicking.rs:467
   4:     0x5642be94360c - std::panicking::begin_panic_fmt::ha93d510596c1ce0e
                        at libstd/panicking.rs:350
   5:     0x5642be9435d1 - rust_begin_unwind
                        at libstd/panicking.rs:328
   6:     0x5642be99a302 - core::panicking::panic_fmt::h48a9dc276960b80c
                        at libcore/panicking.rs:71
   7:     0x5642be7bc07d - core::result::unwrap_failed::h94a0a4510ca2aa22
   8:     0x5642be786839 - grin_miner::client::Controller::run::h82ee6a9c68de931c
   9:     0x5642be794cec - std::sys_common::backtrace::__rust_begin_short_backtrace::he9e665643ab55f7b
  10:     0x5642be769217 - std::panicking::try::do_call::h7ad9e58f3a95d8ae
  11:     0x5642be960aa9 - __rust_maybe_catch_panic
                        at libpanic_unwind/lib.rs:105
  12:     0x5642be7a7030 - <F as alloc::boxed::FnBox<A>>::call_box::h41bcb4ab274a4a07
[...]

PR #1252, opened 2018-07-12
Update build.md

emmick4:
Don't specify building from testnet3 since it doesn't include necessary commits, and will cause mining to fail. (#1249)


ignopeverell:
Git clone will put you on master already so I'd just remove the checkout line.

emmick4:
done. any other recommendations?

ignopeverell:
Nope, thanks!

PR #1258, opened 2018-07-13
Update necessary rust version in build doc

conscott:
Building on 1.24 now gives compile errors. It seems current target is now 1.26+

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -30,7 +30,7 @@ If you decide to use a persistent storage (e.g. ```-v grin:/usr/src/grin```) you
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


conscott:
As a follow up (or possibly added to this PR) - it may make sense to mention that you can
```
rustup upgrade
```
If a current version of Rust is already installed. 

On my machine (Ubuntu 16.04) simply following docs with:
```
curl https://sh.rustup.rs -sSf | sh; source $HOME/.cargo/env
```
Does not actually upgrade rust if it's already installed, so I was under the false impression I was at latest version, when it stayed the same. Instead doing the `rustup upgrade` got me to where I needed to be.

ignopeverell:
Adding `rustup update` as a short note in the build docs wouldn't be a bad idea.

conscott:
Updated with `rustup update` note. Can squash commits if necessary.

PR #1259, opened 2018-07-13
LMDB Wallet Backend functional

yeastplume:
This completes the LMDB backend functionality to be on-par with the current file wallet implementation. As of this merge, users can select either DB or File backends for the wallet, and they should behave identically. 

Also removed the `select_coins` function out of the backend trait and into the library where it more properly belongs.

Some notes on use:
* grin.toml now contains a `use_db_wallet` param, which will use a db for the default wallet if selected.
* to create a db wallet instead of a file wallet (default), run `grin init -d` in an empty directory
* when running grin wallet commands, grin will prefer the db wallet implementation if that directory also contains filewallet files.

### DIFF
--- a/grin.toml
+++ b/grin.toml
@@ -65,6 +65,11 @@ run_tui = true
 #Whether to run the wallet listener with the server by default
 run_wallet_listener = true
 
+#whether to use the database backend storage for the default listener wallet
+#true = lmdb storage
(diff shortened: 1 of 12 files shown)


kargakis:
Why would I prefer to use file over lmdb?

Issue #1247, opened 2018-07-11
grin api returns 505 when a block is not found in fast_sync node


yuntai:
`wget -O- http://127.0.0.1:13413/v1/blocks/5
--2018-07-12 01:55:14--  http://127.0.0.1:13413/v1/blocks/5
Connecting to 127.0.0.1:13413... connected.
HTTP request sent, awaiting response... 500 Internal Server Error
2018-07-12 01:55:14 ERROR 500: Internal Server Error.`

which is not very informative

ignopeverell:
We should return a `404` with a more explanatory message, especially if the node is in fast sync mode. Generally, if a node admin wants to be able to retrieve any block locally, she should run it in archival mode.

Issue #1239, opened 2018-07-09
grin wallet info - incorrect block heights


antiochp:
I think `grin wallet info` is showing the wrong block heights.
Not sure if this is a new issue or if we just haven't ever noticed this before...

```
Wallet Outputs - Block Height: 1262
-----------------------------------------------------------------------------------------------------------------------------------
 Key Id                Child Key Index  Block Height  Locked Until  Status       Is Coinbase?  Num. of Confirmations  Value 
===================================================================================================================================
 695bce7618475f1b0243  299              1263          2263          Unconfirmed  true          0                      60.000000000 
-----------------------------------------------------------------------------------------------------------------------------------
 d3a367aadc253684c585  320              1260          0             Unspent      false         3                      16.156000000 
-----------------------------------------------------------------------------------------------------------------------------------
 2dfb9a218e44534d7ada  321              1260          0             Unspent      false         3                      1.000000000 
-----------------------------------------------------------------------------------------------------------------------------------
```

Compare the above - 2 unspent outputs with block height `1260`, to the info about block 1262 (which appears to be the block this tx was actually included in) - 

https://grinscan.net/block/1262

We can see the `lock_height` of the tx kernel is 1260, but the block height is actually 1262.

I suspect we are incorrectly displaying the `lock_height` via `grin wallet info` (and incorrectly calculating the # confirmations from this).

yeastplume:
It's entirely likely it wasn't noticed before... will have a look as part of ongoing wallet work.

yuntai:
It seems related to I am wanting to tackle next - 'wallet restore' which currently doesn't resolve block height for each output. thinking since node does not necessarily have all the blocks scanning pmmr combined with cumulative pmmr size in the block header can be used to restore the correct block for the output. currently, wallet doesn't seem to monitor blockchain so no way to know the correct block height to which a tx is included.

yeastplume:
@yuntai absolutely correct, just haven't got to it yet, please feel free

antiochp:
I think we are/were setting the "next block height" making ther assumption that this is the correct block height for the new txs.
This was _roughly_ correct when txs were likely to be included in the next block.
But Dandelion adds a few mins of delay - which means we're always off by a few blocks now.

@yuntai the wallet does a refresh/update to get block heights here - 
https://github.com/mimblewimble/grin/blob/42bc03f5f32bff88c41353ea1bdbcb242023f06d/wallet/src/libwallet/internal/updater.rs#L134-L136

[...]

PR #1261, opened 2018-07-15
Add API handler for fetching header by range of heights

yuntai:
- Extension of #1248 to handle range queries.
`GET /v1/headers/<hash|height>`
`GET /v1/headers?start_height=30&max=30`

- Added total_kernel_sum,output_mmr_size & kernel_mmr_size to Header json output.
Toward fixing #1239, where the batch of headers along with output_mmr_size needed to extract block information for each output commit. 

- Tried proper handling 404 instead of 500 (#1247)

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -24,6 +24,7 @@ use serde_json;
 use urlencoded::UrlEncodedQuery;
 
 use chain;
+use store;
 use core::core::hash::{Hash, Hashed};
 use core::core::{OutputFeatures, OutputIdentifier, Transaction};
 use core::ser;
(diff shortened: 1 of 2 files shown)


api/src/handlers.rs
```diff
@@ -513,6 +514,7 @@ impl Handler for ChainCompactHandler {
 /// Gets block headers given either a hash or height.
 /// GET /v1/headers/<hash>
 /// GET /v1/headers/<height>
+/// GET /v1/headers/byheight/start_height=200&max=100
```
kargakis: More idiomatic rest api is to use `GET /v1/headers` (without a hash or a height) for listing all headers and control `start_height` and `max` via url parameters.

yuntai: makes sense.
```
/v1/headers/<hash>
/v1/headers/<height>
/v1/headers?start_height=200&max=100
```
I will resubmit the pr.

yuntai:
a single commit moved to #1269

Issue #1257, opened 2018-07-13
Use a working directory under the user home


ignopeverell:
Currently grin uses the current directory to load different databases, write its logs, read configuration, save wallet files, etc. We should instead rely on a `.grin` under the user home like pretty much all other blockchain daemons.

pcdinh:
Should it be configurable? I think that lots of hard core miners and users may want to move the working directory to a dedicated disk (mount point) for good reasons

ignopeverell:
Yes of course, I should have been clearer that under the home folder should be a default, with a custom location configurable.

haarts:
I think that using the [XDG specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) should be used here (that uses `~/.local` and `~/.config`). Most modern pieces of software do and that would be expected behavior.

yeastplume:
Done via #1433

Issue #1256, opened 2018-07-13
Make grin.toml config optional


ignopeverell:
Config should default to all current values without needing the toml file. If options need to be overwritten, we should be able to accept a file with only those few options.

i1skn:
I would like to whelp with one.

i1skn:
I did some expirement with removing `grin.toml` and making possible to run `grin` without it.
It ends up, that default values set in the `grin.toml` sometimes are different from ones in the code.

I did collect ones are different below (this was done by dumping `println!("{:?}",global_config);` with and without config) :
```
// Each field has a format of <value_in_the_code> | <value_in_grin.toml>

GlobalConfig {
    members: Some(ConfigMembers {
        server: ServerConfig {
            chain_type: UserTesting | Testnet3,
            p2p_config: P2PConfig {
                ban_window: Some(10800) | None,
                peer_max_count: Some(25) | None,
                peer_min_preferred_count: Some(8) | None,
            }
            stratum_mining_config: Some(StratumServerConfig {
                enable_stratum_server: None | Some(true),
                stratum_server_addr:  None | Some("127.0.0.1:13416"),
                attempt_time_per_block: 4294967295 | 15,
            })
            skip_sync_wait: None |  Some(false),
            run_tui: None | Some(true),
            run_wallet_owner_api: Some(false) |  Some(true),
            use_db_wallet: Some(false) |  None
        },
        mining_server: Some(StratumServerConfig {
            enable_stratum_server: None |  Some(true),
            stratum_server_addr: None |  Some("127.0.0.1:13416"),
            attempt_time_per_block: 4294967295 | 15,
        }),
        logging: Some(LoggingConfig {
            stdout_log_level: Debug | Warning,
            log_to_file: false | true,
            file_log_level: Trace | Debug,
            log_file_append: false | true,
        }),
        wallet: WalletConfig {
            check_node_api_http_addr: "http://127.0.0.1:13413" | "http://localhost:13413"
        }
    })
}
```

Now, my question is which constants in the code we should change to ones from `grin.toml` file?
Also, do we need both `server.stratum_mining_config` and `mining_server` in `GlobalConfig`?

antiochp:
Thanks for this - this is really useful.
There's one more bit of complexity here, for some of these values we have a "default" deeper in the code if its not set. For example `run_tui` I believe will default to 'true' if its not set (set to None) in `grin.toml`.

[...]

PR #1262, opened 2018-07-15
Force user to either specify an absolute path or set .grin($HOME/.grin) in grin.toml

yuntai:
#1257 
currently commenting out `db_root` in `grin.toml` causes error. should be fixed in a separate pull req for #1256

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -29,7 +29,7 @@ use wallet::WalletConfig;
 /// the config file location
 
 const CONFIG_FILE_NAME: &'static str = "grin.toml";
-const GRIN_HOME: &'static str = ".grin";
+pub const GRIN_HOME: &'static str = ".grin";
 
 /// Returns the defaults, as strewn throughout the code
 
--- a/config/src/lib.rs
+++ b/config/src/lib.rs
@@ -34,3 +34,4 @@ pub mod config;
 pub mod types;
 
 pub use types::{ConfigError, ConfigMembers, GlobalConfig};
+pub use config::GRIN_HOME;
--- a/src/bin/grin.rs
+++ b/src/bin/grin.rs
@@ -44,6 +44,8 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time::Duration;
+use std::path::{PathBuf, Path};
+use std::env;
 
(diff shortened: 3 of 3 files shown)


config/src/config.rs
```diff
@@ -128,8 +129,17 @@ impl GlobalConfig {
 				// Put the struct back together, because the config
 				// file was flattened a bit
 				gc.server.stratum_mining_config = gc.mining_server.clone();
+				if Path::new(&gc.server.db_root).is_relative() {
```
ignopeverell: I think the `config` crate should just worry about finding, reading and parsing the config. Massaging the diverse options seems better under `src/bin`. Maybe add a `config.rs` helper there to encapsulate all config manipulations after reading?

config/src/config.rs
```diff
@@ -128,8 +129,17 @@ impl GlobalConfig {
...
+					if gc.server.db_root != GRIN_HOME {
+						Err(ConfigError::InvalidConfig(format!("specifier either full path or {}", GRIN_HOME)))?
```
ignopeverell: Small typo: specify

yuntai:
How about we follow bitcoind style
- additional top-level command arguments: `--datadir` & `--conf ` 
- `--datadir` defaults to `$HOME/.grin`
- remove `db_root` from grin.toml
- `--conf` specifies either a relative path to `--datadir` or an absulute path (default: `grin.toml`)

#### Usages
- Vanilla (Use $HOME/.grin & $HOME/.grin/grin.tml)
   `$ grin` 
- Use different data directory
   `$ grin --datadir=/bigfatmount/node0
- Mutitple configuratoins using same data directory
   `$ grin --conf=grin2.toml`
- Running multi nodes
   `$ grin --datadir=/mnt/node0`, `$grin --datadir=/mnt/node1`, etc.

haarts:
I think it would be advantageous to use the XDG specification: https://github.com/mimblewimble/grin/issues/1257#issuecomment-405224814
Things like data dir and config dir are there specified in a universal way.

config/src/config.rs
yuntai: logic moved to src/bin

neuronfog:
@yuntai are you still working on this issue: [1257](https://github.com/mimblewimble/grin/issues/1257)?

Looks like you were off to a good start with this PR.

PR #1268, opened 2018-07-16
LMDB Wallet Transaction Log + LMDB Migration for all

yeastplume:
WIP transaction log + FileWallet to DB Wallet Migration
* All operations that create outputs are logged by the transaction log (implemented for DB Wallet backend only), with a new 'TxLogEntry' created. Outputs are linked to TxLogEntries in a many-to-one relationship.
* `grin wallet txs` displays the contents of the transaction log
* Also included here is a FileWallet-LMDBBackend Migration Utility.. when the server is started or a wallet command is invoked, file wallets will be migrated to an LMDBWallet (a simple data copy) and the new LMDBWallet will be instantiated. This is non destructive, with plenty of warnings given to the user about what's happening.
* Outputs within a wallet will not update if the node's chain height is less than whatever height the wallet was last updated at (to prevent everything being marked as spent if someone runs `grin wallet info` against an unsynced node)

### DIFF
--- a/wallet/src/display.rs
+++ b/wallet/src/display.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use core::core::{self, amount_to_hr_string};
-use libwallet::types::{OutputData, WalletInfo};
+use libwallet::types::{OutputData, TxLogEntry, WalletInfo};
 use libwallet::Error;
 use prettytable;
 use std::io::prelude::Write;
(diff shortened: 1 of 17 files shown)


yeastplume:
This this is ready for merge.. think this is review-worthy, particularly for the lmdb migration but also to see what's going on with the new transaction log

wallet/src/display.rs
```diff
@@ -76,6 +82,74 @@ pub fn outputs(cur_height: u64, validated: bool, outputs: Vec<OutputData>) -> Re
 	Ok(())
 }
 
+/// Display transaction log in a pretty way
```
ignopeverell: Very nice!

wallet/src/libwallet/internal/updater.rs
```diff
@@ -60,6 +60,18 @@ where
 	Ok(outputs)
 }
 
+/// Retrieve all of the transaction entries
```
ignopeverell: I'm sure you're already aware of it, but this just might need to accept start/count type arguments at some point for pagination.

PR #1269, opened 2018-07-16
Resolve correct block height for UTXO during `wallet restore`

yuntai:
Fix to make `grin wallet restore` correctly restore the block heights for the identified UTXOs.

related to #1239 and comments there
closing #1261 as this PR containing the same commit.

#### TODO
- [x] batch block header fetching & expose output mmr_size in the REST API
- [x] REST client code in wallet side
- [x] expose mmr_index in txhashet/outputs REST call for UTXO outputs
- [x] resolve block height for each UTXO using mmr_size in the header and mmr_index of the UTXO
- [x] warning message still shows `wallet.dat` instead of `wallet_data` directory.
- [x] Issue with the transaction log ID not in increasing order when restored.
- [x] Restoring creation & confirmation time using the timestamps of block headers. 

~I will rebase when #1268 comes in.~ rebased with #1268

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -543,7 +543,7 @@ pub struct HeaderRangeHandler {
 }
 
 impl HeaderRangeHandler {
-	fn get_header_by_range(&self, start_height: u64, max: u64) -> Result<Vec<BlockHeaderPrintable>, Error> {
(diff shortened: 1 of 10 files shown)


yeastplume:
Why is this closed now? Seems it was nearly there

PR #1277, opened 2018-07-18
Add config option to prevent 'slog-async: logger dropped messages...' for debugging

### LINKED ISSUE #458: ERRO slog-async: logger dropped messages due to channel overflow, count: 7
Soon after starting grin server run, I got ~90 seconds worth of logger errors, like the one in this issue's title.
This could be avoided by having less verbose (bytes) log lines and/or less of them.

It looks like this was triggered after increasing the aggressiveness in connecting to peers. I'll try to make related log calls less verbose. Candidates:

> DEBG connect_and_req: err - Connection(Error { repr: Custom(Custom { kind: TimedOut, error: StringError("the future timed out") }) }), 83.162.208.57:13414, flagging as defunct
> DEBG saving peer to store PeerData { addr: V4(83.162.208.57:13414), capabilities: , user_agent: "", flags: Defunct }
> DEBG with_timeout: ok but nested - () (treating this as timeout)

yuntai:
Fix #458
https://docs.rs/slog-async/2.3.0/slog_async/enum.OverflowStrategy.html
Add an **optional** config in `grin.toml` (`log_no_drop`) to prevent slog dropping log messages but instead blocking on log producer to prevent log messages when debugging.

### DIFF
--- a/util/src/logger.rs
+++ b/util/src/logger.rs
@@ -77,8 +77,14 @@ lazy_static! {
 			let file_decorator = slog_term::PlainDecorator::new(file);
 			let file_drain = slog_term::FullFormat::new(file_decorator).build().fuse();
 			let file_drain = LevelFilter::new(file_drain, slog_level_file).fuse();
-			let file_drain_final = slog_async::Async::new(file_drain).build().fuse();
-
+			let file_drain_final = slog_async::Async::new(file_drain)
+				.overflow_strategy(
+						if slog_level_file == Level::Trace {
+							slog_async::OverflowStrategy::Block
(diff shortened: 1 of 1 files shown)


yuntai:
so only optional for debugging.

antiochp:
I'd be :+1: with this being optional.
There have _definitely_ been cases where I wish we hadn't dropped messages locally when testing something.

ignopeverell:
Can we enable it only in trace mode?

yuntai:
Fixed to use BlockStrategy only when logging to file with Trace level.

PR #1278, opened 2018-07-19
Make grin.toml config optional

i1skn:
This PR mirror exisiting config parameters in `grin.toml` to source code, so binary can run without a config file.

As a part of https://github.com/mimblewimble/grin/issues/1256 was decided to use current `grin.toml` parameters as default ones in source code.

### Test
To keep `grin.toml` and code "defaults" in sync, I've decided to create a test, so when either of them changed - another needs to be updated. While I was creating a test for this, I've encountered a problem, that config diff looks like that:

screen shot 2018-07-19 at 23 16 43

this is not very informative, so I decided to use `pretty_assertions` crate to make it looks like that:

screen shot 2018-07-19 at 23 16 52

I've added it to `[dev-dependency]`, so it should not pollute builds.

### DIFF
--- a/config/tests/config.rs
+++ b/config/tests/config.rs
@@ -7,9 +7,9 @@ use config::GlobalConfig;
 #[test]
 fn file_config_equal_to_defaults() {
 	let x = true;
-	let mut global_config_without_file = GlobalConfig::default();
+	let global_config_without_file = GlobalConfig::default();
 
-	let mut global_config_with_file = GlobalConfig::new(Some("../grin.toml")).unwrap_or_else(|e| {
+	let global_config_with_file = GlobalConfig::new(Some("../grin.toml")).unwrap_or_else(|e| {
 		panic!("Error parsing config file: {}", e);
 	});
 
--- a/config/tests/config.rs
+++ b/config/tests/config.rs
@@ -1,0 +1,20 @@
+#[macro_use]
+extern crate pretty_assertions;
+extern crate grin_config as config;
+
+use config::GlobalConfig;
+
+#[test]
+fn file_config_equal_to_defaults() {
+	let x = true;
+	let mut global_config_without_file = GlobalConfig::default();
+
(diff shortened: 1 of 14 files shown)


config/tests/config.rs
```diff
@@ -0,0 +1,20 @@
+#[macro_use]
+extern crate pretty_assertions;
+extern crate grin_config as config;
+
+use config::GlobalConfig;
+
+#[test]
+fn file_config_equal_to_defaults() {
+	let x = true;
+	let mut global_config_without_file = GlobalConfig::default();
```
hashmap: should it  (and `global_config_with_file`) be `mut`?

src/bin/grin.rs
```diff
@@ -320,33 +320,39 @@ fn main() {
 		panic!("Error parsing config file: {}", e);
 	});
 
-	if global_config.using_config_file {
-		// initialize the logger
-		let mut log_conf = global_config
+	if let Some(file_path) = global_config.config_file_path.clone() {
```
hashmap: for my education - why do we need `.clone()` here?

i1skn:
@ignopeverell but,`Cargo.lock` has changes related to the library `pretty_assertions`, which is added in this commit? Should I just leave `Cargo.toml` than?

src/bin/grin.rs
```diff
@@ -320,33 +320,39 @@ fn main() {
...
-		// initialize the logger
-		let mut log_conf = global_config
+	if let Some(file_path) = &global_config.config_file_path {
+		info!(
+			LOGGER,
+			"Found configuration file at {}",
+			file_path.to_str().unwrap()
+		);
+	} else {
+		info!(LOGGER, "configuration file not found");
```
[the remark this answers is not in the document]
i1skn: @ignopeverell yeah, that's sound good.

ignopeverell:
@i1skn my bad, github was hiding the changeset so I thought (mistakenly) you had many more differences than just that.

Issue #1271, opened 2018-07-17
A lot of "refused by chain: Block is unfit: already known"


garyyu:
Testnet3, new build on master branch, after completely synced and mined some new blocks, the node then becomes always late from the network, for example, on this writing, the latest height is 11424, but my node is on 11419, and keep the words "downloading blocks 99%". 
```

Jul 17 11:52:12.032 DEBG Block 01d92413 at 11413 is unfit at this time: already known
Jul 17 11:52:12.033 DEBG adapter: process_block: block 01d92413 refused by chain: Block is unfit: already known 
... ... (a lot)
Jul 17 11:54:16.337 DEBG Block 14db084d at 11418 is unfit at this time: already known
Jul 17 11:54:16.338 DEBG adapter: process_block: block 14db084d refused by chain: Block is unfit: already known 
... ... (a lot)
... ... (a lot other blocks)
Jul 17 11:57:38.440 DEBG Block 14db084d at 11418 is unfit at this time: already known
Jul 17 11:57:38.440 DEBG adapter: process_block: block 14db084d refused by chain: Block is unfit: already known 
... ... (a lot)
```

garyyu:
[grin.log.gz](https://github.com/mimblewimble/grin/files/2200188/grin.log.gz)
The full logs could help to address the problem.

ignopeverell:
Quick question: is that on a release build?

garyyu:
Just started another fresh installation on another pc with fresh folder, the sync procedure stuck at height 0 after full headers downloaded, also with a lot of "refused by chain" logs. Now I just kill the process.

This log file is much more simple and smaller, please let me know this will help, then I can upload this new log.

ignopeverell:
A pause of 10-15 min after all headers are downloaded is expected. Or are you running an archive node? In any case more logs would be useful.

garyyu:
I test it again from a new clean **release** build on latest code and install, with all config at default (so, this is a fast mode), run it for **one hour**, the TUI status show "`Downloading headers: 100%, step 1/4`".

grep the log and find 435 lines "`refused by chain`". The full log (file name: `grin.0720.1.log.gz`) attached, please take a look at this major bug.
```
$ grep -c "refused"  grin.log 
435
```
[grin.0720.1.log.gz](https://github.com/mimblewimble/grin/files/2212373/grin.0720.1.log.gz)

garyyu:
[...]

PR #1281, opened 2018-07-20
fix bug on block sync (#1271)

garyyu:
To fix block sync issue on https://github.com/mimblewimble/grin/issues/1271 

2 bugs found and fixed:

1. function `check_header_known()` is using `ctx.block_hashes_cache`, but function `check_known()` is using `ctx.block_hashes_cache`,  which should be reversed.
2. `pipe` function `process_block()` will call `check_known()` and ignore the blocks if already received before (i.e. `known`), which is the source of those logs "`unfit at this time: already known`". But for block sync, it's quite normal that blocks received not in ideal sequence, then a lot of blocks will go into cache (orphans, by chain `&self.orphans.add(orphan);` ).  When node receive a block and sequence is ok, block is valid, saved and appended to chain, it will trigger calling chain `check_orphans()` function, which will release some blocks from cache (so-called `orphans`), by calling chain `process_block_no_orphans()`, and eventually call to `pipe` function `process_block()` again. At this time, `check_known()` function will ignore the blocks which is already known in `ctx.block_hashes_cache`,  then, problem happen! sometimes a cache(so-called `orphan`) block is in still in `ctx.block_hashes_cache` but sometimes not, because there's HASHES_CACHE_SIZE (now it's 50) limitation. So only the lucky cached block can be accepted if it's not in `ctx.block_hashes_cache`.

The fix solution is to add a `from_cache` parameter for pipe `process_block()` function.

Please take a code review on this fix before merging it, since it's the core logic on block syncing.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -197,7 +197,7 @@ impl Chain {
 		b: Block,
 		opts: Options,
 	) -> Result<(Option<Tip>, Option<Block>), Error> {
-		let res = self.process_block_no_orphans(b, opts);
+		let res = self.process_block_no_orphans(b, opts, false);
 		match res {
 			Ok((t, b)) => {
 				// We accepted a block, so see if we can accept any orphans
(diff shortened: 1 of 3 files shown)


ignopeverell:
Thanks for tracking down the issue. The fix is fine, but may be better to not add to cache when the block is found to be an orphan instead of adding a boolean switch for those methods.

garyyu:
Thanks, Good idea! will do it soon.

garyyu:
But having one question here:
`check_orphans()` also call `process_block_no_orphans()`, **after orphan block processing, do we need add to cache or not?**   I suggest add,  and then we'd better let `HASHES_CACHE_SIZE` (now is 50) at least same as `MAX_ORPHAN_SIZE` (now is 200). 
Is this suggestion OK for you?

ignopeverell:
Agreed on the caches size.

Adding to cache should be done in `process_block_no_orphans()` when the block is either fully accepted or fully rejected (of which orphans are an exception). As `check_orphans()` calls `process_block_no_orphans()`, addition in the hash cache will be done there once an orphan is accepted.

PR #1283, opened 2018-07-20
Add Stratum documentation

### LINKED ISSUE #1176: Document the StratumServer RPC
Documenting RPC for the StratumServer will make it easier for miners and pools to integrate.  

At least the following should be captured:
* Each Request / Response in JSON
* Each possible error code/message and meaning
* miners SHOULD randomize the job nonce before starting (not the way it is now, the way it should be)
* miners MUST continue mining the same job until the server sends a new one, though a miner MAY request a new job at any time
* miners MUST NOT send an rpc response to a job request from the server 
* miners MAY set the RPC "id" and expect responses to have that same id
* miners MAY send a keepalive message
* miners MAY send a login request (to identify which miner finds shares / solutions in the logs), the login request MUST have all 3 params.
* miners MUST return the supplied job_id with submit messages.
* Link to the "grin-miner" client.rs file as a reference implementation

quentinlesceller:
Add Stratum documentation thus fixes #1176.
Feel free to review it and I'll make the corresponding modifications.

### DIFF
--- a/doc/stratum.md
+++ b/doc/stratum.md
@@ -1,0 +1,556 @@
+# Grin Stratum RPC Protocol
+
+This document describes the current Stratum RPC protocol implemented in Grin.
+
+## Table of Contents
+
+1. [Messages](#messages)
+    1. [getjobtemplate](#getjobtemplate)
+    2. [job](#job)
+    3. [keepalive](#keepalive)
+    4. [login](#login)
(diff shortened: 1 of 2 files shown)


ignopeverell:
Merging now, corrections and improvements can easily be added as additional PRs.

PR #1286, opened 2018-07-23
fix: fast sync mode will fail when 'sync_state' was switched to 'HeaderSync'

garyyu:
Fast sync will offen fail with message `connection lost: BadMessage`, or fail with message `WARN txhashset_write: xxx ignored for sync_state HeaderSync`

To fix it, freeze the header_sync when `SyncStatus` is `TxHashsetDownload`.

Also add some error logs on `TxHashSetArchive` message receive, for easier to find issues in the future.

### DIFF
--- a/p2p/src/protocol.rs
+++ b/p2p/src/protocol.rs
@@ -229,24 +229,45 @@ impl MessageHandler for Protocol {
 				let sm_arch: TxHashSetArchive = msg.body()?;
 				debug!(
 					LOGGER,
-					"handle_payload: txhashset archive for {} at {}",
+					"handle_payload: txhashset archive for {} at {}. size={}",
 					sm_arch.hash,
 					sm_arch.height,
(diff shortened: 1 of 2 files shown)


p2p/src/protocol.rs
```diff
@@ -229,34 +229,56 @@ impl MessageHandler for Protocol {
...
+						"handle_payload: txhashset archive received but SyncStatus not on TxHashsetDownload",
+					);
 					return Err(Error::BadMessage);
 				}
 
 				let mut tmp = env::temp_dir();
 				tmp.push("txhashset.zip");
-				{
-					let mut tmp_zip = File::create(tmp.clone())?;
+				let mut save_txhashset_to_file = |file| -> Result<(), Error>  {
```
ignopeverell: Don't think that needs to be `mut`.

PR #1288, opened 2018-07-23
clean some build warnings

garyyu:
1. remove 2 warnings;
2. suppress 2 duplicate warnings about deprecated api "`aggsig api still subject to review and change`"

### DIFF
--- a/wallet/src/libtx/aggsig.rs
+++ b/wallet/src/libtx/aggsig.rs
@@ -70,7 +70,6 @@ pub fn verify_partial_sig(
 }
 
 /// Just a simple sig, creates its own nonce, etc
-#[allow(deprecated)]
 pub fn sign_from_key_id<K>(
 	secp: &Secp256k1,
 	k: &K,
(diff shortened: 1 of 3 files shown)


wallet/src/libtx/aggsig.rs
```diff
@@ -70,6 +70,7 @@ pub fn verify_partial_sig(
 }
 
 /// Just a simple sig, creates its own nonce, etc
+#[allow(deprecated)]
```
ignopeverell: Those deprecation warnings are actually here on purpose, at least until the underlying secp256k1 API is deemed stable.

garyyu: Yes, I know that purpose. That's why I keep one warning for each function, but suppress the duplicate warning (on same secp256k1 API function).   
So, the remaining 4 warning is enough to remind those secp256k1 API.

yeastplume:
I'm fine with removing most of the warnings, so long as at least a single instance of them pops up during a build (phrased in a way that it's clear it covers the whole API), but I'd probably rather do it on the rust-secp256k1 side rather than adding directives into grin.

Issue #1264, opened 2018-07-16
Get rid of the clang dependency


ignopeverell:
Building croaring requires clang. However the underlying C implementation does not, clang-format is an optional dependency (assumingly, only for code formatting). So there's likely a missing option in the Rust binding. Figure this out.

Nothing personal against clang but it's a large dependency and creates all sorts of wrinkles from one environment to another and we end up with all sorts of user build errors. So if it's not required, better not have it.

jpbriquet:
I have tried to investigate a little bit about croaring and clang.

It seems that croaring/croaring-sys crates relies on the bindgen crate.
The [bindgen crate](https://github.com/rust-lang-nursery/rust-bindgen) has clang as a dependency and use it to generate rust bindings for croaring C code, that can be seen in the croaring-sys crate build system:
https://github.com/saulius/croaring-rs/blob/master/croaring-sys/build.rs

I could be wrong but I think it will not be possible to remove easily this dependency without changing the croaring-sys crate. 
It might be possible to have a specific process to generate these rust bindings outside of the build.rs, however I'm not sure it would be a accepted by the crate maintainer as it changes lots of things, generated bindings have to be stored on git, a CI must launch the generation process etc...

Rust dependency tree (extract):
│   ├── grin_chain v0.3.0
│   │   ├── ...
│   │   ├── croaring v0.3.2
│   │   │   ├── croaring-sys v0.3.5
│   │   │   │   └── libc v0.2.42 (*)
│   │   │   │   [build-dependencies]
│   │   │   │   ├── bindgen v0.37.4
│   │   │   │   │   ├── ...
│   │   │   │   │   ├── clang-sys v0.23.0
│   │   │   │   │   │   ├── glob v0.2.11
│   │   │   │   │   │   ├── libc v0.2.42 (*)
│   │   │   │   │   │   └── libloading v0.5.0
│   │   │   │   │   │       [build-dependencies]
│   │   │   │   │   │       └── cc v1.0.17
│   │   │   │   │   ├── ...
│   │   │   │   └── gcc v0.3.54 (*)
│   │   │   └── libc v0.2.42 (*)

ignopeverell:
Could that be done easily if we forked `croaring-rs`?

jpbriquet:
I will fork it and will try to remove the dep. If this approach is working I'll let you know.

antiochp:
[...]

PR #1292, opened 2018-07-24
Get rid of the clang dependency (#1264)

garyyu:
complete the enhancement requirement of https://github.com/mimblewimble/grin/issues/1264

How:

- Fork the repo of https://github.com/saulius/croaring-rs.git
- Remove the `bindgen` which has dependency on `clang-sys`,and use the static bindgen generated file instead of dynamically generate it.
- By the way, upgrade `croaring-rs` dependency from old `0.3.2` version to the latest `0.3.5` version.
- Modify Grin dependency of `croaring` to:
`croaring = { git = "https://github.com/fastgrin/croaring-rs.git", tag = "0.3.5-grin" }`


knocte:
shouldn't the fork be kept in the mimblewimble github account instead?

garyyu:
@knocte  fork one into mimblewimble/ also OK, but have to be done by limited people:)  in my view, that's not important since anyway they are in github.com :)

yeastplume:
I don't really think it's worthwhile maintaining a special fork of something just to remove a dependency. Is there really no other way of addressing this?

yeastplume:
Sorry just read the other thread to see how we got here... still really less than ideal and wish there were some other alternative.

garyyu:
a fork is the only way :)  unless you don't need crate `croaring-rs`.

yeastplume:
This is a problematic dependency no matter what we do. I've tried this on an OSX machine, and had a whole host of problems trying to get it to build.. even with the correct LLVM/Clang version installed it failed because the wrong version of rustfmt was installed.

I'm not sure this proposed solution will work, as we can't be sure that a static version of the bindgen file will work for all machines/versions of clang that are likely to be out there. Do we know, for instance, if the output is the same regardless of the compiler on the target machine? Given this affects core modules as well I'm uneasy with a separate special fork that we need to maintain for it.

Before going ahead with this perhaps @antiochp can have a look and see if there are any alternatives.

garyyu:
What 'bindgen' generated is [this pure rs file](https://github.com/fastgrin/croaring-rs/blob/master/croaring-sys/src/croaring-sys.rs).  I don't think it's the generated file's problem.

Could you share the error log when building?  I test on my 3 macs all are OK.

garyyu:
To have a quick verification, please generate it on your problem mac:
```
git clone --recursive https://github.com/saulius/croaring-rs/
cd croaring-rs
cargo build --release

```
Then search the generated file:  `croaring-sys.rs`, and compare it with the file I shared above.

yeastplume:
I'm past the issue, it was due to an older version of rustfmt on the system.

The generated .rs file you've posted there appears to have a ton of Apple/darwin specific constants defined which is leading me to believe it will be generated differently between systems. Have you compared that version of the file to one generated on a linux system?

garyyu:
Just tried in Ubuntu 17, unfortunately,  Grin build fail in link stage with error missing croaring***

Then tried to build  https://github.com/fastgrin/croaring-rs  alone, success and quick.

Then tried to build original https://github.com/saulius/croaring-rs.git, need install libclang-dev, and then build still failed with tons of errors, missing a lot of standard header files, link one by one, then at last, failed with tons of these errors, boring:(   seems not a good idea~   

```
--- stderr
/usr/include/ia32intrin.h:112:1: error: definition of builtin function '__rdtsc'
/usr/include/xmmintrin.h:130:10: error: invalid conversion between vector type '__m128' (vector of 4 'float' values) and integer type 'int' of different size
/usr/include/xmmintrin.h:136:10: error: invalid conversion between vector type '__m128' (vector of 4 'float' values) and integer type 'int' of different size
/usr/include/xmmintrin.h:142:10: error: invalid conversion between vector type '__m128' (vector of 4 'float' values) and integer type 'int' of different size
/usr/include/xmmintrin.h:148:10: error: invalid conversion between vector type '__m128' (vector of 4 'float' values) and integer type 'int' of different size
/usr/include/xmmintrin.h:242:10: error: returning 'int' from a function with incompatible result type '__m128' (vector of 4 'float' values)
/usr/include/xmmintrin.h:248:10: error: returning 'int' from a function with incompatible result type '__m128' (vector of 4 'float' values)
/usr/include/xmmintrin.h:254:10: error: returning 'int' from a function with incompatible result type '__m128' (vector of 4 'float' values)
/usr/include/xmmintrin.h:260:10: error: returning 'int' from a function with incompatible result type '__m128' (vector of 4 'float' values)
... ...
```

garyyu:
Since can't find a good way to solve the following problems:

- bindgen result file differentiate in OS
- not clear what's the impact on bindgen result file from OS version upgrading 

I suspend this PR and wait for thinking out a better solution to remove clang dependency.

knocte:
> @knocte fork one into mimblewimble/ also OK, but have to be done by limited people:) in my view, that's not important since anyway they are in github.com :)

The issue here is trust. A person that downloads grin and executes it in their computer are already trusting github and the mimblewimble org. If grin repo depends on some other software forked in an org outside mimblewimble, then all users of grin need to "increase" their trust surface. Placing this fork in the mimblewimble org secludes the trust in just one place.

garyyu:
@knocte  dependencies not the key issue, you can pick any one of most famous projects and search the dependencies, you will wow it:) this is the internet world, sharing, sharing, and sharing.

We want remove Clang just for `croaring-rs` too easy to build fail in all kinds of os env, till now, I still can't make it build on my ubuntu 17, that directly cause Grin build fail.

But unfortunately, I think this task failed just because it could not be a good way to do it.

ignopeverell:
The underlying C/C++ `croaring` impl seems fairly straightforward to implement and has a small API surface. Can't we just have a simple direct binding to it that doesn't rely on `bindgen`?

garyyu:
Good idea!
Try it again and today is lucky:)

1. Instead of using autogenerated ffi (`croaring-sys.rs`), using self ffi (`ffi.rs`) and only put those used api for this lib
2. Build and test OK in mac. But still fail with "undefined reference" in Linux.
3. Finally, I find it's just caused by those `#[link_name = ...]`, remove all those `link_name`, everything is fine then :)

will submit the pull-request again.

antiochp:
Apologies - been out of the loop for a couple of weeks. 
Catching up on the discussion around this today.

garyyu:
This one ready for merge one week ago. Tested OK both in Mac and Linux.

antiochp:
What effect does removing all the `#[link_name = ` lines from the lib have?

garyyu:
@antiochp  removing that is a fix for platforms other than mac. Its due to the different calling conventions between different platforms, On Linux the leading underscore is not needed but it is needed on OSX and Windows, without that we can let compiler to handle it by itself.

garyyu:
polluted by a git rebase. to close it for the moment,  and easy to create new pr if this feature needed in the future.

PR #1300, opened 2018-07-28
Build doc for running a free node on Gcloud & minor fixes   

lehnberg:
* Detailed step by step instructions for setting up a node on Google Cloud from scratch to receive transactions from other wallets. The node can run for free 24/7, costing nothing to the end user. Should be a good way to onboard new users that are curious and want to try Grin out as it makes setup fairly fool-proof. (even I could do it!) I also add some links to the doc in the other documentation.
* Removal of some broken links.
* Adding info in wallet.md about `grin wallet txs` transaction log.

Making a pull request for this as there already are build instructions in `/doc/`. Or would you prefer it to be maintained via the [wiki](https://github.com/mimblewimble/docs/wiki) instead?


ignopeverell:
Seems that would belong in the wiki, no? Maybe just link it from the build docs in a "VPS/Cloud builds" section?

lehnberg:
Yep, makes sense. Closing this down. Moved it to wiki, [link here](https://github.com/mimblewimble/docs/wiki/How-to%3A-Run-a-Grin-node-on-Google-Cloud-for-free) in case anyone wants to reference it in the build docs. ✌️

PR #1302, opened 2018-07-30
[WIP] Wallet documentation

yeastplume:
Update to wallet docs, diagrams, re-org them a bit. Update usage guide to reflect current options


doc/wallet/design/design.md
```diff
@@ -0,0 +1,93 @@
...
+!·tx_id: Option<u32>,          
+) -> Result<Vec<OutputData>, Error>
+where     
+!·T: WalletBackend<C, K>,      
+!·C: WalletClient,
+!·K: Keychain,
+{  
+```
+
+With `T` in this instance being a class that implements  the `WalletBackend` trait, which is further parameterized with implementations of
```
hashmap: It seems that `WalletBackend` which is described above as

> **WalletBackend** - Defines the storage implementation of the wallet

is more than storage layer, but also let say controller which glue all parts together? Is it possible to just drop in some new storage method (like sqlite) and reuse the rest of functionality?

yeastplume: It can be looked at that way, yes. It'll probably be worth factoring out the storage a bit more and creating another trait that's more of a piece of glue that contains all the other components, but not quite there yet. The basic abstraction is good enough for the high-level documentation at the moment, anyhow.

Issue #1304, opened 2018-07-31
Header sync is broken, stuck on 0%


hashmap:
Regression was introduced as part of https://github.com/mimblewimble/grin/pull/1297
As result header chain is considered invalid after block 7:

```
Jul 31 11:41:20.756 DEBG header head 0386b7d5 @ 7
Jul 31 11:41:20.757 DEBG sync head 0386b7d5 @ 7
Jul 31 11:41:20.759 DEBG pipe: sync_block_header: 036a0f74 at 8
Jul 31 11:41:20.761 ERRO validate_header: BANNABLE OFFENCE: header cumulative difficulty 1870 != 1484
Jul 31 11:41:20.761 INFO Invalid block header 036a0f74: Error { inner:

Addition of difficulties on all previous blocks is wrong }.
Jul 31 11:41:20.761 DEBG pipe: sync_block_header: 00e5b31d at 9
Jul 31 11:41:20.762 INFO Invalid block header 00e5b31d: Error { inner:

Orphan }.
.. 
```
(I fixed ` 1870 != 1484` part, the current code shows total difficulty on the right side `94831` which is misleading). 

Network difficulty calculation is broken because it relies on timestamp, there are some differences in `time` and `chrono` timestamp.
https://github.com/mimblewimble/grin/blob/master/chain/src/pipe.rs#L295

I reverted `chrono` commit and was able to sync my node.

garyyu:
The old time crate use 0..11 as `tm_mon`, but chrono has `Utc.ymd(year, month, day)` api with 1..12 as month.
For example, this old code was saying "8th July", not "8th June".
```
            timestamp: time::Tm {
                tm_year: 2018 - 1900,
                tm_mon: 6,
                tm_mday: 8,
                tm_hour: 18,
                ..time::empty_tm()
            },
```

Fixed in https://github.com/mimblewimble/grin/pull/1305.

quentinlesceller:
Fixed in #1305.

PR #1305, opened 2018-07-31
fix: old time crate use 0..11 as the month, make mistake with chrono Utc.ymd (#1304)

garyyu:
The old time crate use 0..11 as `tm_mon`, but chrono has `Utc.ymd(year, month, day)` api with 1..12 as month.
For example, this old code was saying "8th July", not "8th June".
```
            timestamp: time::Tm {
                tm_year: 2018 - 1900,
                tm_mon: 6,
                tm_mday: 8,
                tm_hour: 18,
                ..time::empty_tm()
            },
```

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -297,9 +297,9 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		if target_difficulty != network_difficulty.clone() {
 			error!(
 				LOGGER,
-				"validate_header: BANNABLE OFFENCE: header cumulative difficulty {} != {}",
+				"validate_header: BANNABLE OFFENCE: header target difficulty {} != {}",
 				target_difficulty.to_num(),
-				prev.total_difficulty.to_num() + network_difficulty.to_num()
+				network_difficulty.to_num()
 			);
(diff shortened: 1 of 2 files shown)


chain/src/pipe.rs
```diff
@@ -297,9 +297,9 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		if target_difficulty != network_difficulty.clone() {
 			error!(
 				LOGGER,
-				"validate_header: BANNABLE OFFENCE: header cumulative difficulty {} != {}",
+				"validate_header: BANNABLE OFFENCE: header target difficulty {} != {}",
 				target_difficulty.to_num(),
-				prev.total_difficulty.to_num() + network_difficulty.to_num()
+				network_difficulty.to_num()
```
yeastplume: what's this change for?

garyyu: The misleading log.

garyyu: `ERRO validate_header: BANNABLE OFFENCE: header cumulative difficulty 1870 != 94831`

ignopeverell:
Been confused between months starting at 0 vs 1 more than once (Java scars you forever).

PR #1313, opened 2018-08-03
Improved basic transaction diagram

jaspervdm:
I think there is a small mistake in the diagram (unless things changed very recently in how a transaction is built). The random offset `oS` is subtracted from the excess `xS`, not from the nonce `kS`. The only purpose of `kS` is to generate a signature.


jaspervdm:
Following the suggestion by @garyyu, I added the calculation of `xS`

garyyu:
Thanks @jaspervdm , looks good!

to be a little bit nitpicked, I suggest to keep the step numbers unchanged, and only change the sentence as:
```
9. Calculate total blinding excess xS1 (private scalar), sum for all inputs(-) and outputs(+)
```
and a little change on this step:
```
11: Select a random kernel offset **oS**, and calculate **xS** = **xS1** - **oS** to create **xS**
```

jaspervdm:
The offset is not part of the kernel, so that's why I removed that line.

jaspervdm:
That is already in there
> Create final transaction **tx** from **Slate**

PR #1319, opened 2018-08-06
Fix get outputs call in wallet

hashmap:
It generates an invalid url if there are 1000+ outputs.
* Also switched to async http client for performance reasons (see below)
* Couple unrelated cleanups

I time-d the existing sequential version of the client code vs async (both in debug profile), the later is more scalable, but it brings some complexity to the code. Currently most of the wallets produces 1-5 chunks (chunk requires a separate request) and difference is around 0.1 sec.

Sync version (200 chunks):
```
real    0m4.001s
user    0m2.248s
sys     0m0.249s
```

Async version (200 chunks):
```
real    0m3.636s
user    0m1.572s
sys     0m0.132s
```

Sync (1000 chunks):
```
real     0m5.620s
user     0m6.610s
sys      0m0.899s
```

Async (1000 chunks):
```
real    0m4.104s
user    0m3.470s
sys     0m0.503s
```

### DIFF
--- a/api/src/router.rs
+++ b/api/src/router.rs
@@ -141,7 +141,6 @@ impl Router {
 		for key in keys {
 			node_id = self.find(node_id, key).ok_or(RouterError::RouteNotFound)?;
 			if self.node(node_id).key == *WILDCARD_STOP_HASH {
-				debug!(LOGGER, "ROUTER stop card");
 				break;
 			}
 		}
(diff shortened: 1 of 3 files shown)


ignopeverell:
Makes sense, there's a limit to URL lengths in HTTP. Thanks!

Issue #1293, opened 2018-07-24
[sync] remaining issues


garyyu:
lehnberg @lehnberg 17:12

```
FYI, built the latest from master just now and removed .grin as discussed before, this time around I synced seamlessly. Can’t have taken me more than five minutes. ✌️
git fetch upstream
git merge upstream/master
cargo clean
cargo build --release
rm -rf .grin
RUST_BACKTRACE=1 ./target/release/grin
```

Gary Yu @garyyu 17:34
`less than 5 minutes is also odd :) do you have a powerful PC? even the fastsync mode, the validation procedure is slow on my mac air. will measure how long I need.`

Gary Yu @garyyu 19:08
```
@lehnberg I’m not lucky, fastest time I need 15 minutes to sync. so it’s possible your MBP 2016 is 3 times faster than my macpro 2013.
and sync procedure still buggy :(
I tested 2 times:
1st time on my poor old mac:

- spent 2’30” for header download,
- spent 5 minutes to download the TxHashsets, ( not lucky! could download from a very slow peer! )
- spent 7' to validate chain state,
- spent 24 minutes to download remaining blocks and finish syncing. (buggy here! paused for a * while no any blocks coming and force to quit and start again).

2nd time, on a mac pro 2013

- spent 1’46” for header download,
- spent 6" to download the TxHashsets, (buggy here! paused for Error summing and verifying kernel sums, and have to force quit and start again, 6 seconds is for 2nd time).
- spent 8' to validate chain state,
- spent 5’30" to download remaining blocks and finish syncing.

```

garyyu:
1st question is: what's the root cause of an `Error summing and verifying kernel sums`?

ignopeverell:
It generally means the downloaded kernels failed to validate. It could be a consensus failure but it's a lot more likely to be caused by some download failure or corrupted file.

garyyu:
Is there possibility to restart downloading the TxHashsets in case errors?  in current implementation it will stuck here. I checked before, it seems no existing way to notify sync thread about this error event .

ignopeverell:
We should maybe rely on the sync status to communicate those errors across parts of the system. This would allow fast sync to retry I believe.

garyyu:
Good idea :)
I prefer to add a `sync_error` field into `SyncState` to communicate those errors:

```
/// Current sync state. Encapsulates the current SyncStatus.
pub struct SyncState {
	current: RwLock<SyncStatus>,
[...]

PR #1320, opened 2018-08-06
[WIP] Handle txhashset failure

hashmap:
Allow fast sync to recover if txhashset download or verifiaction fails
Partially fixes #1293

PR #1322, opened 2018-08-07
fix: in case txhashset validation fail, 'TxHashsetDownload' should retry (#1293)

garyyu:
1. To communicate the sync errors across parts of the system, enhance the `SyncState` structure to have a `sync_error` element.
2. In `txhashset_write()`, set this `sync_error` in case of failure. 
3. In `run_sync` thread, read this `sync_error` and give proper handling: for example in this case restart TxHashset download.

And one more related fix:

- When `header_sync_due` is due, must always launch `header_sync()`, otherwise, `if header_head.height == si.highest_height` could always fail and cause `fast_sync_due()` never executed.

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -24,7 +24,7 @@ use std::thread;
 use std::time::Instant;
 
 use chain::{self, ChainAdapter, Options, Tip};
-use common::types::{ChainValidationMode, ServerConfig, SyncState, SyncStatus};
+use common::types::{self, ChainValidationMode, ServerConfig, SyncState, SyncStatus};
 use core::{core, global};
 use core::core::block::BlockHeader;
 use core::core::hash::{Hash, Hashed};
@@ -361,7 +361,9 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 			w(&self.chain).txhashset_write(h, txhashset_data, self.sync_state.as_ref())
 		{
 			error!(LOGGER, "Failed to save txhashset archive: {}", e);
-			!e.is_bad_data()
+			let is_good_data = !e.is_bad_data();
+			self.sync_state.set_sync_error(types::Error::Chain(e));
+			is_good_data
 		} else {
 			info!(LOGGER, "Received valid txhashset data for {}.", h);
 			true
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -15,7 +15,7 @@
 //! Server types
 
 use std::convert::From;
-use std::sync::RwLock;
+use std::sync::{Arc, RwLock};
 
 use api;
 use chain;
@@ -296,7 +296,7 @@ pub enum SyncStatus {
 /// Current sync state. Encapsulates the current SyncStatus.
 pub struct SyncState {
 	current: RwLock<SyncStatus>,
-	sync_error: RwLock<Result<String, String>>,
+	sync_error: Arc<RwLock<Option<Error>>>,
 }
 
 impl SyncState {
@@ -304,7 +304,7 @@ impl SyncState {
 	pub fn new() -> SyncState {
 		SyncState {
 			current: RwLock::new(SyncStatus::Initial),
-			sync_error: RwLock::new(Ok("None".to_owned())),
+			sync_error: Arc::new(RwLock::new(None)),
 		}
 	}
 
@@ -337,19 +337,19 @@ impl SyncState {
 
 	/// Communicate sync error
 	pub fn set_sync_error(&self, error: Error){
-		let mut sync_error = self.sync_error.write().unwrap();
-		*sync_error = Err(format!("{:?}", error));
+		let clone = Arc::clone(&self.sync_error);
+		*clone.write().unwrap() = Some(error);
 	}
 
 	/// Get sync error
-	pub fn sync_error(&self) -> Result<String, String> {
-		self.sync_error.read().unwrap().clone()
+	pub fn sync_error(&self) -> Arc<RwLock<Option<Error>>> {
+		Arc::clone(&self.sync_error)
 	}
 
 	/// Clear sync error
 	pub fn clear_sync_error(&self){
-		let mut sync_error = self.sync_error.write().unwrap();
-		*sync_error = Ok("None".to_owned());
+		let clone = Arc::clone(&self.sync_error);
+		*clone.write().unwrap() = None;
 	}
 
 }
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -337,8 +337,7 @@ impl SyncState {
 
 	/// Communicate sync error
 	pub fn set_sync_error(&self, error: Error){
-		let clone = Arc::clone(&self.sync_error);
-		*clone.write().unwrap() = Some(error);
+		*self.sync_error.write().unwrap() = Some(error);
 	}
 
 	/// Get sync error
@@ -348,8 +347,7 @@ impl SyncState {
 
 	/// Clear sync error
 	pub fn clear_sync_error(&self){
-		let clone = Arc::clone(&self.sync_error);
-		*clone.write().unwrap() = None;
+		*self.sync_error.write().unwrap() = None;
 	}
 
 }
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -296,6 +296,7 @@ pub enum SyncStatus {
 /// Current sync state. Encapsulates the current SyncStatus.
 pub struct SyncState {
 	current: RwLock<SyncStatus>,
+	sync_error: RwLock<Result<String, String>>,
 }
 
 impl SyncState {
@@ -303,6 +304,7 @@ impl SyncState {
 	pub fn new() -> SyncState {
 		SyncState {
 			current: RwLock::new(SyncStatus::Initial),
+			sync_error: RwLock::new(Ok("None".to_owned())),
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


hashmap:
Thanks for the PR! I'm not sure about the design, `sync_error` is like a second dimension for the state machine we have, what advantages of it do you see? Have you considered adding a state  to `SyncStatus` and handling it here for example https://github.com/mimblewimble/grin/blob/master/chain/src/chain.rs#L553  after extending the trait https://github.com/mimblewimble/grin/blob/master/chain/src/types.rs#L125 ?
Also a txhashet request may be ignored by a peer, or just fail, we need to handle this case too.

garyyu:
@hashmap  Thanks for the review~  below are answers to your questions.

Question 1. `sync_error is like a second dimension for the state machine we have,  what advantages of it do you see?`
Answer: 
- The advantage is to communicate `errors` only, not directly touch the `SyncState` change, that's a way to avoid coupling between different modules. 
- Another advantage is you can extend all kinds of error handling with this `sync_error`, not only for this fix, also for any future error handling.

Question 2. `Have you considered adding a state to SyncStatus and handling it ?`
Answer: 
- compare to add a `sync_error`, add one or more `SyncStatus` will need change that status directly, other module perhaps is using this status, that's a deep coupling.
- in future, if other errors need to be handled, you have to add more and more `SyncStatus`, but actually, the errors are already there, we can just reuse the errors.

Question 3. `Also a txhashet request may be ignored by a peer, or just fail, we need to handle this case too.`
Answer: very good question! Do you mean the `MessageHandler for Protocol` in protocol.rs ? What's the indicator of a request ignored or fail?

hashmap:
@garyyu thanks! I'd prefer to rely on state changes only, but it's a matter of taste, I'd consult with @ignopeverell because he introduced tracking of state changes.

Reg Q3 - There are 2 parts - sending request and receiving the response. The first one is easier - replace `unwrap` with proper handling here https://github.com/mimblewimble/grin/blob/master/servers/src/grin/sync.rs#L278  The second part is a bit trickier - my naive attempt was to rely on timer https://github.com/mimblewimble/grin/pull/1320/files#diff-da8815c2e0832a4f8ff91de9eea26c7aR461  because there is no an immediate response , a peer can decide to ignore us or fail to send a file for some reason.

garyyu:
Thanks!
Understood  Q3 now, is this a little bit premature optimization? I never see these errors before on my tests. But the validation fail happened several times on my tests, that's why this fix come.

Anyway, I wait for your consultation result before proceeding for further fix on Q3 :)

hashmap:
it's hard to see the lack of response, basically your node stuck forever on sync without any errors, only restart helps. It happened to me couple times, probably a peer was restarted while handling my  node's request. But I was running fast sync from scratch dozen times per day.

servers/src/common/types.rs
```diff
@@ -296,13 +296,15 @@ pub enum SyncStatus {
 /// Current sync state. Encapsulates the current SyncStatus.
 pub struct SyncState {
 	current: RwLock<SyncStatus>,
+	sync_error: RwLock<Result<String, String>>,
```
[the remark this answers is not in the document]
garyyu: Perhaps better to replace `Result<String, String>` with  `Result<String, Error>`?   I used `Error` at the beginning, forgot what problem lead me to replace `Error` with `String` here.

Rust is new to me:(  still confusing about Result, Error, Option and so on.
Are you sure Option is better here than `Result<String, Error>`?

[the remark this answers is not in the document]
hashmap: I think so. `Result` is needed when you have something which is value or error - usually it's a result of a computation/function (hence Result). Option is needed when you may have value or nothing (null/nil in many languages), actually Option may sound confusing, in other languages it's called `Maybe` which sounds quite natural - `maybe error`. So `Option<Error>` would  give us 2 possible value - `None` (absence of error) or `Some(error)`.

garyyu:
since it's indeed existing issue, ok to add these 2 fix. is testing by error cases simulation for the new code.

servers/src/common/types.rs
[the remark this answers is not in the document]
ignopeverell: I agree with @hashmap that `Option<Error>` would be more idiomatic here.

servers/src/common/adapters.rs
```diff
@@ -361,7 +361,9 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 			w(&self.chain).txhashset_write(h, txhashset_data, self.sync_state.as_ref())
 		{
 			error!(LOGGER, "Failed to save txhashset archive: {}", e);
-			!e.is_bad_data()
+			let is_good_data = !e.is_bad_data();
+			self.sync_state.set_sync_error(types::Error::Chain(e));
```
ignopeverell: As a less verbose alternative, you can also clone the error here. They're very small objects so it's not a big deal.

servers/src/grin/sync.rs
```diff
@@ -451,6 +458,11 @@ impl SyncInfo {
 			false
 		}
 	}
+
```
ignopeverell: Perhaps we could get rid of `fast_sync_due` now. The time seems irrelevant after this fix, so we should be able to only rely on the sync state (including the error), right?

servers/src/common/adapters.rs
garyyu: Tried before, unfortunately `Clone is not implemented for failure::Context<error::ErrorKind>`. Have no idea about how to fix it.

And a long story about Error clone,  many Error can't clone, for some reasons such as io::Error, then all Error which packing io::Error can't clone.

servers/src/grin/sync.rs
garyyu: We keep `fast_sync_due` for the moment, use it for download timeout detection. Perhaps switch to better timeout detection method in the future.

garyyu:
Updated PR according to above review comments: 
- Use Option<Error> for 'sync_error' instead of Result. 

And add 2 more fixes according to @hashmap failure cases:
1. handle sending request failure of `send_txhashset_request`;
2. use a `10-mins timeout` to handle no-response failure, in case a peer ignore our txhashset request, or a peer fail to send a file for some reason.

Note about `Error` type: because of the problem of Error `clone` (such as io::Error, all Error which packing io::Error can't clone directly too).  Instead, I prefer `Arc::clone` since it's cross module and multi-threads access. That's why the final `sync_error` becomes:
```
sync_error: Arc::new(RwLock::new(None)),
```

servers/src/grin/sync.rs
```diff
@@ -120,21 +120,51 @@ pub fn run_sync(
...
 						};
 					}
 
 					if fast_sync_enabled {
-						// run fast sync if applicable, every 5 min
-						if header_head.height == si.highest_height && si.fast_sync_due() {
-							fast_sync(peers.clone(), chain.clone(), &header_head);
-							sync_state.update(SyncStatus::TxHashsetDownload);
+
+						// check sync error
```
hashmap: Nitpicking - a fragment from this line to 151 could be a bit simpler like:
```
						let maybe_sync_error = sync_state.sync_error();
						if let Some(ref sync_error) = *maybe_sync_error.read().unwrap() {
							error!(
								LOGGER,
								"fast_sync: error = {:?}. restart fast sync", sync_error
							);
							si.fast_sync_reset();
							sync_state.clear_sync_error();
						}
```

servers/src/common/types.rs
```diff
@@ -332,6 +334,24 @@ impl SyncState {
 
 		*status = new_status;
 	}
+
+	/// Communicate sync error
+	pub fn set_sync_error(&self, error: Error){
+		let clone = Arc::clone(&self.sync_error);
```
hashmap: No clone is needed:
```
	*(self.sync_error.write().unwrap()) = Some(error);
```

servers/src/common/types.rs
```diff
@@ -332,6 +334,24 @@ impl SyncState {
...
+	}
+
+	/// Get sync error
+	pub fn sync_error(&self) -> Arc<RwLock<Option<Error>>> {
+		Arc::clone(&self.sync_error)
+	}
+
+	/// Clear sync error
+	pub fn clear_sync_error(&self){
+		let clone = Arc::clone(&self.sync_error);
```
hashmap: Also no clone here:
```
*(self.sync_error.write().unwrap()) = None;
```

servers/src/grin/sync.rs
garyyu: Haha~   You're making same error like me:)  Actually, yours (also my previous trying) can't build, because of an odd short live error, plus a dead-lock risk.

hashmap: try it:)

garyyu:
@hashmap Thanks again for your review~   The unnecessary clone in set_sync_error/clear_sync_error has been removed.

servers/src/grin/sync.rs
garyyu: I said that's old mistakes what I made before:) it will cause 2 problems there~

garyyu:
```
error[E0597]: `maybe_sync_error` does not live long enough
   --> servers/src/grin/sync.rs:139:39
    |
139 |                             if let Some(ref sync_error) = *maybe_sync_error.read().unwrap() {
    |                                                            ^^^^^^^^^^^^^^^^ borrowed value does not live long enough
...
160 |                         }
    |                         - `maybe_sync_error` dropped here while still borrowed
    |
    = note: values in a scope are dropped in the opposite order they are created
```

garyyu: @hashmap   I can add one line after your code to pass the build:
```
drop(maybe_sync_error);
```

But then it's same as my original code:)
And another problem in your code is the `read lock` and the `write lock`, it's a risk to dead lock, I don't know why compiler don't complain about that, but better to avoid.

hashmap:
Thanks for taking care of it!

PR #1323, opened 2018-08-07
add block validation rules - single coinbase output and single coinbase kernel

antiochp:
Right now we build new blocks with a single coinbase output and a single coinbase kernel.
But - we do not actually restrict blocks in this way - we could build a valid block with 1,000 coinbase kernels (summing up to the correct overall excess).

This PR proposes adding validation rules to ensure - 
* single coinbase output
* single coinbase kernel

I don't think there is any benefit to letting blocks be overly flexible in terms of multiple coinbase outputs or coinbase kernels.
I think it would be beneficial to prevent miners inflating the size of the MMRs unnecessarily.

Any reasons not to do this?

@ignopeverell @tromp ?


tromp:
> prevent miners inflating the size of the MMRs unnecessarily

Except it doesn't achieve that. Miners are still free to add arbitrarily many regular transactions.
As long as there are not enough fee paying transactions in the mempool to fill up a block, they can do this at zero cost. The only way to dis incentivize miners from block bloat is to effectively reduce the reward for bigger blocks (for example, as in my quadratic burning proposal).
So until we have a better justification, I prefer to keep the flexibility of multiple coinbase outputs/kernels.

antiochp:
Hmm good point...

antiochp:
Closing.

PR #1329, opened 2018-08-09
Update fuzz in core crate

hashmap:
Fix compilation error in fuzz code after the wallet refactoring


hashmap:
@antiochp There are 2 issues with that. Fuzz test runs infinite amount of time (in theory), in practice we can limit it by 1-5 minutes or even less, usually it finds obvious issues pretty quickly. Second, it requires nightly at the moment, I think it may complicate travis build.

PR #1332, opened 2018-08-10
Consolidate and cleanup tx aggregation

ignopeverell:
* Include commitments non-duplicate checks in aggregate
* Remove said check from the pool
* Block building now uses tx aggregation to reduce duplication

Didn't want to get too deep in the rabbit hole but I think we're at a point where there's enough common validation code between `Block` and `Transaction` to introduce a *thing* that both would include. The *thing* could have the `Vec`s of `Input`, `Output` and `Kernel` and all the validation in common.


ignopeverell:
Created #1333 for the block and transaction validation cleanup.

core/src/core/transaction.rs
```diff
@@ -431,14 +431,16 @@ impl Transaction {
 	/// Validates all relevant parts of a fully built transaction. Checks the
 	/// excess value against the signature as well as range proofs for each
 	/// output.
-	pub fn validate(&self) -> Result<(), Error> {
+	pub fn validate(&self, as_block: bool) -> Result<(), Error> {
```
antiochp: This bool flag feels awkward but I think you're tracking this as subsequent cleanup/refactoring in #1333?

core/src/core/transaction.rs
```diff
@@ -431,14 +431,16 @@ impl Transaction {
...
 		if self.inputs.len() > consensus::MAX_BLOCK_INPUTS {
 			return Err(Error::TooManyInputs);
 		}
-		self.verify_features()?;
+		if !as_block {
```
antiochp: Not sure how critical this is but I've been leaning toward ordering these verification rules _roughly_ based on how expensive they might be to run.

core/src/core/transaction.rs
```diff
@@ -525,37 +532,26 @@ pub fn aggregate(transactions: Vec<Transaction>) -> Result<Transaction, Error> {
...
+	}
 
 	let in_set = inputs
 		.iter()
 		.map(|inp| inp.commitment())
 		.collect::<HashSet<_>>();
 
-	let out_set = outputs
-		.iter()
-		.filter(|out| !out.features.contains(OutputFeatures::COINBASE_OUTPUT))
```
antiochp: We need to be careful with coinbase outputs here?
There's a case here where a block spends a coinbase output and its the same coinbase output for the block reward then we cut-through (bypassing the maturity check).
I don't think we ever want to allow cut-through on coinbase outputs.

core/src/core/transaction.rs
ignopeverell: Correct, at least that was the hope. Yes, slightly awkward but at the moment removes a bunch of duplication.

core/src/core/transaction.rs
ignopeverell: I think that's a good idea. I was trying to avoid checking that bool multiple times, maybe something we can fix up as well when we remove it?

core/src/core/transaction.rs
ignopeverell: Yes, I was just thinking about adding an additional transaction-level validation check for this. Going to do it now.

ignopeverell: On 2nd look, `verify_features` already does that for transactions. Regarding the block case, I don't think said block would be valid. The block `verify_coinbase` will fail as coinbase sum won't work, right? And if a miner decided to broadcast the tx spending his coinbase *before* the block it'd get orphaned, and be invalid after (spending an immature coinbase).

antiochp: Actually this block can't be valid at that point.
The output was cut-through maintaining the sums (inputs vs outputs) but simply removing a kernel will break the summing.
I think.

Issue #1327, opened 2018-08-07
Fix aggregation of very large transactions


ignopeverell:
One could make a transaction that's just below our maximum size limits and send it over a Dandelion relay. As soon as it gets aggregated with another transaction, the resulting transaction would be impossible to mine.

antiochp:
So one issue here is we do not actually validate the resulting tx after aggregating txs together.
We verify the kernel sums correctly but that is the only validation check we perform.

https://github.com/mimblewimble/grin/blob/328d832bd6e3ac33cf4c160c4d78085dafb20adf/core/src/core/transaction.rs#L576-L580

It should be straightforward to validate the tx (in isolation) at this point.


The other related issue (as discussed briefly on gitter) is that we have a discrepancy between the tx validation rules and the rules for reading/writing txs over the wire.

Tx validation checks that we have less than `consensus::MAX_BLOCK_INPUTS` inputs - 
https://github.com/mimblewimble/grin/blob/328d832bd6e3ac33cf4c160c4d78085dafb20adf/core/src/core/transaction.rs#L434-L437


But transaction `read()` checks that we have less than `consensus::MAX_TX_INPUTS` inputs (among other checks) - 

https://github.com/mimblewimble/grin/blob/328d832bd6e3ac33cf4c160c4d78085dafb20adf/core/src/core/transaction.rs#L302-L307


And these are significantly different values - 

https://github.com/mimblewimble/grin/blob/328d832bd6e3ac33cf4c160c4d78085dafb20adf/core/src/consensus.rs#L91-L96

antiochp:
I think we should - 

a) add a post-aggregation tx validation step to fully validate the resulting tx.

b) consolidate the input|output|kernel count validation rules in `tx.validate()`

c) increase the input limit to match the block input limit. I think it makes sense to allow a _huge_ tx that would end up in a blocks on its own, effectively filling the block entirely. This way we make aggregation as effective as possible at the tx pool level. @tromp?
This may be trickier for limiting outputs as a block will contain at least one additional output (for the coinbase, and maybe multiple of these). Similar for tx kernel limit.
 
d) move the related count rules out of `tx.read()` and call `tx.validate()` as part of `tx.read()` so we can reuse the existing validation rules without needing to duplicate them explicitly in `tx.read()`.

I can take this on if the proposed approach makes sense.

tromp:
good point about tx limits needing to be a little below block limits to allow for coinbase.
[...]

PR #1334, opened 2018-08-10
validate txs when reading them from the wire

antiochp:
Partial fix for #1327.
Related - #1331.

This consolidates the tx input|output|kernel size validation so we have one consistent and reusable set of validation rules.
This should prevent us producing a tx that is "valid" within a single node but "invalid" during serialization/deserialization as we propagate is across the nework.

* introduce a shared `MAX_INP_OUT_KERN_LEN` consensus constant
  * make it `30_000` which is what our original max block inputs was set to
* use `MAX_INP_OUT_KERN_LEN` to keep the following consistent - 
  * MAX_TX_INPUTS
  * MAX_TX_OUTPUTS
  * MAX_TX_KERNELS
  * MAX_BLOCK_INPUTS
* move input|output|kernel size validation out of `tx.read()` and into `tx.validate()`
* call `tx.validate()` from `tx.read()`


antiochp:
Merging. Consensus related but nothing contentious.
Discussions around these validation rules also ongoing in terms of actual max limits etc.

PR #1336, opened 2018-08-10
Validate block size

antiochp:
~Depends on #1334 (this PR include changes from there also as I branched off from it).~
~This changeset will be smaller once #1334 is merged in.~

We were verifying the max number of block inputs in `verify_weight()` but it was hidden at the end of a larger comparison and it was not that clear what was happening.

This PR introduces a `verify_size()` to be consistent with tx validation.

As discussed in #1327 we allow a tx to have up to `30_000` inputs, outputs and kernels.
This PR allows a single enormous tx to be included in a block iff it is the only tx in the block.
To allow this we ned to allow the following in the block itself -
* max `30_000` inputs
* max `30_000 + 1` outputs (`30_000` outputs from the tx + a single coinbase output for the block)
* max `30_000 + 1` kernels (`30_000` kernels from the tx + a single coinbase kernel for the block)

We should discuss these max counts. This works for now but I'm not sure if these number make sense in the greater scheme of things.
Maybe we want more flexibility baked into the max block sizes to allow for some more flexibility when dealing with enormous (aggregated or otherwise) txs close to the max tx size.

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -90,11 +90,20 @@ pub const MAX_BLOCK_WEIGHT: usize = 80_000;
 
 /// Reused consistently for various max lengths below.
 /// Max transaction is effectively a full block of data.
+/// Soft fork down when too high.
+/// Likely we will need to tweak these all individually, but using a single constant for now.
 const MAX_INP_OUT_KERN_LEN: usize = 300_000;
 
-/// Maximum inputs for a block (issue#261)
-/// Hundreds of inputs + 1 output might be slow to validate (issue#258)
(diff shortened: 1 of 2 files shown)


antiochp:
travis is having issues right now - but lets be careful merging onto master before travis completes (even if it is slow and frustrating).

hashmap:
@antiochp Ok, apologies, was impatient to get your changes to avoid merge conflict.

antiochp:
Yeah no worries - I just don't the blame if master fails and my PR was the last one merged... 😄

antiochp:
That's my last change I think for now in this area (except maybe the new `verify_weight()` on txs).
So we should be good in terms of conflicts now. Thanks!

PR #1346, opened 2018-08-11
Sort pool tx by fee over weight for mining

### LINKED ISSUE #1105: tx pool does not consider fee ordering
The current `prepare_mineable_transactions()` implementation is pretty naive and simply takes the first `n` txs by tx pool insertion time.

Related - #1067.

We check that the per tx is not too low when accepting new txs but we do not take this into consideration when preparing a new block -

https://github.com/antiochp/grin/blob/b18cb8a0ea5254cfb3f9341b0385133470fba635/pool/src/transaction_pool.rs#L155-L160

ignopeverell:
Requires #1345 to add the `block_weight` method to a transaction.

Fixes #1105


antiochp:
I think reordering the txs might be risky to do when reading from the pool.
Right now we add to to the pool in time-order and maintain this ordering.
If we can add txs 1,2,3 to the pool and the pool is valid at each point in time then we know we can read [1], [1, 2] or [1, 2, 3] and be _reasonably_ confident any of these sets of txs will result in a valid block.
If we pull from the pool based on fees we may (extreme example) end up with [3] skipping txs 1 and 2.
But if 3 is dependent (0-conf) on 1 or 2 then we just built an invalid block.

The function `blockchain.validate_raw_txs()` is used when reconciling the pool after seeing a new propagated block.
This adds txs to a "running state" and checks the state is still valid each time, returning a final set of txs that together produce a valid state.

I think we could use this in `prepare_mineable_transactions()` after re-ordering them by fee to produce a set of internally consistent txs that would make up a valid block.

Anything more complex than this feels like it starts to get into "replace by fee" territory and I don't think we necessarily want to tackle this right now.

antiochp:
Yeah I think (1) is probably the best/simplest approach, at least initially.

Note that `validate_raw_txs()` will actually do the incremental adding of txs for us. 
It attempts to add each tx incrementally. If it fails on one tx it does not fail for everything, it will call `rewind_raw_tx()` and continue with the rest of the vec.
This lets us reconcile the full tx pool on seeing a new block, regardless of the original sort order of the pool.

ignopeverell:
The more I think about this, the more complex it gets. It's trivial to build an attack where I'd send a lot of transactions with low fee, spent immediately by another with higher fee. The result will be:

1. Single transaction blocks in the non-iterative case (single pass, we don't try to save rejects to fill blocks).
2. An O(n^2) algorithm in the iterative case.

We also have to think of cut-through, which could allow filling blocks quite a bit more in some cases, but complicates the selection up to block weight.

antiochp:
You mean a long chain of dependent "0-conf" txs, each spending the output of the previous one but with each with an increasingly higher fee?

We can determine relatively easily of a tx being added to the pool is 0-conf or not, based on whether it attempts to spend outputs currently existing in the pool.
So I wonder if we can leverage this somehow?

The original pool implementation tracked the set of "root" txs. 
Defined as rooted in the current chain state. 
A 0-conf spend would be a non-root tx.

And I think the intent was to select exclusively root txs when building the block.
Selecting a root tx would then select all dependent txs in the subtree.
I believe this was not fully implemented as the code did not handle cases where txs could be dependent on multiple roots txs or if the sum total of all dependent txs beneath a root exceeded the block limits etc.


What if we sorted roots txs first?
This would cover the majority of simple cases. Then handled dependent 0-conf differently somehow?
Maybe dependent txs are always left in inserted time order for simplicity (process the big chain of txs in the same order they were originally, successfully added to the pool).

ignopeverell:
This is take two. It mainly adds a first pass that buckets transactions that depend on each other and aggregates. @antiochp what do you think?

pool/src/pool.rs
```diff
@@ -70,11 +71,60 @@ where
...
 	/// Take the first num_to_fetch txs based on insertion order.
 	pub fn prepare_mineable_transactions(&self, num_to_fetch: u32) -> Vec<Transaction> {
-		self.entries
-			.iter()
-			.take(num_to_fetch as usize)
-			.map(|x| x.tx.clone())
-			.collect()
+		// first, group dependent transactions in buckets (vectors), each bucket
+		// is therefore independent from the others
+		// this relies on the entries Vec having parent transactions first
```
antiochp: I assume we can safely rely on this as the pool is sorted internally by tx insertion order.

ignopeverell:
In that case the aggregate won't validate, failing the weight check, and will be skipped. It won't disrupt mining but may just stay in the pool forever until eviction (also TODO).

I'm thinking I should just add a hard limit to those chains (like 20 transactions), which should address the great majority of those corner cases. It won't deal with a huge transaction aggregated with a small one in the same bucket, but that's enough of a edge case that eviction may be the best outcome, at least for now.

PR #1354, opened 2018-08-14
Consolidate validation in Block and Transaction

### LINKED ISSUE #1333: Cleanup: consolidate validation in `Block` and `Transaction`
We're at a point where there's enough common validation code between `Block` and `Transaction` to introduce a *thing* that both would include. The *thing* could have the `Vec`s of `Input`, `Output` and `Kernel` and all the validation in common.

Don't want to put the label `easy` on this, but should be fairly straightforward, except for a likely very large amount of changes for everything that uses the inputs/outputs/kernels directly from blocks and transactions. Also not sure whether introducing a trait would be useful (and doable) or a common struct would be fine.

hashmap:
This PR Introduces TransactionBody which is included into block and tx.

Initially I tied to hide `body` by exposing `inputs/outputs/kernels` as methods on tx and block, but then I found that exposing `body` as a filed makes it conceptually simpler, but perhaps harder to read/use.

Fixes #1333

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -27,8 +27,8 @@ use core::hash::{Hash, HashWriter, Hashed, ZERO_HASH};
 use core::id::ShortIdentifiable;
 use core::target::Difficulty;
 use core::{
-	transaction, Commitment, KernelFeatures, Output, OutputFeatures, Proof, ShortId, Transaction,
-	TransactionBody, TxKernel,
+	transaction, Commitment, Input, KernelFeatures, Output, OutputFeatures, Proof, ShortId,
+	Transaction, TransactionBody, TxKernel,
 };
 use global;
 use keychain::{self, BlindingFactor};
@@ -351,7 +351,7 @@ pub struct Block {
 	/// The header with metadata and commitments to the rest of the data
 	pub header: BlockHeader,
 	/// The body - inputs/outputs/kernels
-	pub body: TransactionBody,
+	body: TransactionBody,
 }
 
 /// Implementation of Writeable for a block, defines how to write the block to a
@@ -452,9 +452,10 @@ impl Block {
 
 		// collect all the inputs, outputs and kernels from the txs
 		for tx in txs {
-			all_inputs.extend(tx.body.inputs);
-			all_outputs.extend(tx.body.outputs);
-			all_kernels.extend(tx.body.kernels);
+			let tb: TransactionBody = tx.into();
+			all_inputs.extend(tb.inputs);
+			all_outputs.extend(tb.outputs);
+			all_kernels.extend(tb.kernels);
 		}
 
 		// include the coinbase output(s) and kernel(s) from the compact_block
@@ -548,7 +549,7 @@ impl Block {
 			let zero_commit = secp_static::commit_to_zero_value();
 			let secp = static_secp_instance();
 			let secp = secp.lock().unwrap();
-			let mut excesses = map_vec!(agg_tx.body.kernels, |x| x.excess());
+			let mut excesses = map_vec!(agg_tx.kernels(), |x| x.excess());
 			excesses.push(prev.total_kernel_sum);
 			excesses.retain(|x| *x != zero_commit);
 			secp.commit_sum(excesses, vec![])?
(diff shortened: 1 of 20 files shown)


ignopeverell:
Yes, I like TransactionBody like this, and it factors in a lot of validation code, which is great. Now for the bad news... I don't really like the `.body.kernels` or `.body.outputs` everywhere. It'd be a lot nicer with a `inputs()`, `outputs()` and `kernels()` accessor on `Transaction` and `Block`. May be a pain to introduce but I think it'd be a lot nicer.

hashmap:
Not a pain, I did one round of refactoring in the opposite direction already, thanks to mighty vim. I’ll do it one more time, unfortunately it will add some amount of `clone()`

ignopeverell:
Perhaps the accessors should return references then?

hashmap:
That’s how I started. Then found that in some places we consume them, eg in `aggregate`. Anyway now I have couple of ideas how to implement it with minimum overhead.

hashmap:
@ignopeverell I removed direct `body` access. In places where we consume `i/o/k` I introduced a type safe zero-cost conversion of `tx` into `tx body`, so we can access `i/o/k` fields without `clone` even if `body` is private.

core/src/genesis.rs
```diff
@@ -25,109 +25,83 @@ use global;
...
-				0x15cc5bc6, 0x16edf697, 0x17ced40c, 0x17d84f9e, 0x18a515c1, 0x19320d9c, 0x19da4f6d,
-				0x1b50bcb1, 0x1b8bc72f, 0x1c7b6964, 0x1d07b3a9, 0x1d189d4d, 0x1d1f9a15, 0x1dafcd41,
-			]),
-			..Default::default()
-		},
-		inputs: vec![],
-		outputs: vec![],
-		kernels: vec![],
-	}
+	core::Block::with_header(core::BlockHeader {
```
antiochp: Nice!

core/src/core/block.rs
```diff
@@ -47,7 +47,7 @@ pub enum Error {
 	CoinbaseSumMismatch,
 	/// Restrict block total weight.
 	TooHeavy,
-	/// Block weight (based on inputs|outputs|kernels) exceeded.
+	/// Block weight (based on.body.inputs.body.outputs.body.kernels) exceeded.
```
antiochp: Comment messed up during search/replace?

core/src/core/block.rs
```diff
@@ -488,9 +452,10 @@ impl Block {
 
 		// collect all the inputs, outputs and kernels from the txs
 		for tx in txs {
-			all_inputs.extend(tx.inputs);
-			all_outputs.extend(tx.outputs);
-			all_kernels.extend(tx.kernels);
+			let tb: TransactionBody = tx.into();
+			all_inputs.extend(tb.inputs);
```
antiochp: Most access appears to be along the pattern of `block.inputs()` or `tx.inputs()` now.
It may feel more consistent if we could do this here also (via an accessor and not a direct field access)?

core/src/core/block.rs
```diff
@@ -533,7 +497,7 @@ impl Block {
 		let mut kern_full = vec![];
 		let mut kern_ids = vec![];
 
-		for k in &self.kernels {
+		for k in &self.body.kernels {
```
antiochp: Can we use the accessor here? `&self.kernels()`

core/src/core/block.rs
hashmap: Accessors borrow references to fields, here we need actual values, not references. Direct field access in Rust allows partial move of values from the struct, which is quite unique for Rust. I tried to make this case explicit by type casting, instead of returning `block` filed from the method (which implicitly destroys the tx here, quite unexpected side effect). Another approach would be clone, but I don't like it here because of loop, overhead may be quite big.

antiochp: That makes sense I think. I still don't have an intuitive grasp of this in Rust... I think I get what you are saying.

hashmap:
@ignopeverell sure, I run my node on it (with fast sync from scratch after each build)

PR #1360, opened 2018-08-15
Fix overage and fee overflow

hashmap:
2 issues were found by fuzzing.  Overage and fee come from the network and may cause overflow if a message was corrupted or crafted to attack a server.


antiochp:
:+1:

Issue #1306, opened 2018-08-01
grin wallet send cause recipient existed money missed


garyyu:
Bug. It happens each time when call this send.

Sender:
```
$ ./grin wallet send -d http://xxx:13415 3.33 -c 6 -s smallest
Aug 01 15:03:59.058 INFO This is Grin version 0.3.0 (git testnet3-54-g84e46c2), built for x86_64-apple-darwin by rustc 1.27.1 (5f2b325f6 2018-07-07).
Aug 01 15:03:59.064 DEBG Built with profile "release", features "" on Tue, 31 Jul 2018 11:02:42 GMT.
Aug 01 15:03:59.064 INFO Found configuration file at /Volumes/ramdisk/grin/grin.toml
Aug 01 15:03:59.064 INFO Using LMDB Backend for wallet
Aug 01 15:03:59.065 INFO Using LMDB Backend for wallet
Aug 01 15:03:59.065 DEBG Using wallet seed file at: ./wallet.seed
Aug 01 15:03:59.091 DEBG Refreshing wallet outputs
Aug 01 15:03:59.184 DEBG Building input (spending coinbase): 60000000000, 062c7cdc42c4789fc67a
Aug 01 15:03:59.195 DEBG Building an output: 56662000000, 423fe7a2e7d3d837bd1e
Aug 01 15:03:59.290 DEBG Posting transaction slate to http://xxx:13415/v1/wallet/foreign/receive_tx
Aug 01 15:03:59.530 DEBG Validating final transaction
Aug 01 15:03:59.617 INFO Tx created: 3.330000000 grin to http://xxx:13415 (strategy 'smallest')
Aug 01 15:03:59.709 INFO Tx sent
```

Recipient: (before this sending)
```
$ ./target/release/grin wallet info

____ Wallet Summary Info as of 32251 ____

 Total                            | 2644.384000000 
 Awaiting Confirmation            | 9.900000000 
 Immature Coinbase                | 900.000000000 
 Currently Spendable              | 1734.484000000 
 ---------                        | --------- 
 (Locked by previous transaction) | 0.000000000 

$ ./target/release/grin wallet outputs

Wallet Outputs - Block Height: 32251
-------------------------------------------------------------------------------------------------------------------------------------------------
 Key Id                Child Key Index  Block Height  Locked Until  Status       Is Coinbase?  Num. of Confirmations  Value          Transaction 
=================================================================================================================================================
...
[...]

PR #1374, opened 2018-08-18
fix: deriv_idx is not updated after wallet restore (#1306)

garyyu:
Fix for https://github.com/mimblewimble/grin/issues/1306.

Root Cause:

After wallet restore, check the wallet database and find `deviv_idx` is still 0, that will cause the overwriting of existing money in wallet in any future receiving.

### DIFF
--- a/wallet/src/libwallet/internal/restore.rs
+++ b/wallet/src/libwallet/internal/restore.rs
@@ -220,7 +220,7 @@ where
 
 			max_child_index = if max_child_index >= output.n_child.unwrap() {
 				max_child_index
-			}else{
+			} else {
 				output.n_child.unwrap()
 			};
 		} else {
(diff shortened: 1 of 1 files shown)


wallet/src/libwallet/internal/restore.rs
```diff
@@ -215,6 +217,12 @@ where
 				is_coinbase: output.is_coinbase,
 				tx_log_entry: None,
 			});
+
+			max_child_index = if max_child_index >= output.n_child.unwrap() {
+				max_child_index
+			}else{
```
ignopeverell: Nitpick: we use `} else {`. Also do you have `rustfmt` configured to run on commit? It would catch all these little details.

wallet/src/libwallet/internal/restore.rs
```diff
@@ -223,6 +231,15 @@ where
 			);
 		}
 	}
+
+	if max_child_index>0 {
```
ignopeverell: Same here with operator spacing.

wallet/src/libwallet/internal/restore.rs
garyyu: thanks. always forgot to link that githook in a new clone, will find auto way to have rustfmt for us

yeastplume:
Looks good.. (wallet restore needs another round of development, will look at it shortly)

PR #1376, opened 2018-08-18
Fix config defaults

antiochp:
master is currently failing on travis due to discrepancy between `grin.toml` and our global defaults.

```
             run_wallet_owner_api: Some(
<                true
>                false
             ),
```


antiochp:
Not sure why cargo.lock is in here. Sorting this out.

yeastplume:
sorry my fault, thanks for fixing

PR #1385, opened 2018-08-19
Fix txpool race condition

antiochp:
Transaction pool validation is based on "current" chain state (by verifying we can potentially add txs from the pool to current chain state).
So adding a new tx to the txpool needs to validate against chain state.
But processing a block will update chain state.

So if we process a block and a new tx we get into various possible race conditions where the chain state is updated and txpool validation can fail (a tx is effectively in both the txpool _and_ the block we just processed).

This PR adds a `header_hash` param to `add_to_pool()` and `validate_raw_txs()` so we now validate the txpool against a known chain state (the latest block at the time the tx was actually received) and not the "current" chain state which may have been updated since.

Operations on the txpool (and stempool) now take block specific chain state into account rather than simply assuming they can safely use latest chain state.

Also tweaks some of the debug logging and error handling around tx api calls to provide more context around tx failures when adding to the pool.

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -116,8 +116,6 @@ impl OutputHandler {
 			}
 		}
 
-		debug!(LOGGER, "outputs_by_ids: {:?}", commitments);
-
 		let mut outputs: Vec<Output> = vec![];
 		for x in commitments {
 			if let Ok(output) = self.get_output(&x) {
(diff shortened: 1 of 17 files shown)


antiochp:
Tested locally and against testnet3 under various conditions.
All tests passing.

Going to merge this as this fixes several weird errors we intermittently see when attempting to `grin wallet send` (if our node happens to process a block being propagated at the same time).

PR #1386, opened 2018-08-20
Not schedule many peers

hashmap:


hashmap:
will add it to https://github.com/mimblewimble/grin/pull/1398

PR #1387, opened 2018-08-20
Don't wait if grin was restarted quickly

hashmap:
Currently it leads to 30 seconds delay, because we have 0 most worked peers (we are up to date), but we wait to get at least 4 before starting sync. To fix it we also check if we have some difficulty and enough connected peers.

Basically logic in sync.rs was changed in line 95 + changes in peers, the rest if rustfmt noise


hashmap:
Rustfmt made this pr too noisy, will try to make it cleaner

PR #1390, opened 2018-08-20
Refactor rewind single header

antiochp:
* pass single header to `extension.rewind()`, no need to pass head_header each time
* safely handle the case where `head_header` is behind the header being rewound to (rewind during fast sync)

While working on this PR discovered we had introduced a bug in the way we were handling `block_input_bitmaps` when we moved to LMDB - 

https://github.com/mimblewimble/grin/blob/82a467ac3c4f9e1caf17540c72f2a8b584817484/chain/src/txhashset.rs#L765-L767

We introduced the concept of a `batch` as part of LMDB (for writing batches of updates to the db).
We should be saving a input_pos_bitmap to the db for _each_ block being rewound.
But this had been rewritten - passing a single `found` flag back and writing a _single_ aggregated `block_input_bitmap` for the head_header block hash only.

I'm having a hard time understanding what this is actually doing but it is not correct.
I don't really know what the impact of this is - I don't see how we haven't seen more serious problems in `testnet3` due to the fact we are not "unwinding" spent inputs correctly like this.

I _suspect_ we have got away with this so far because we have relatively few txs in `testnet3`...

Resolved this by moving the `block_input_bitmap` operations to the batch itself and simply handling read/write/cache of these via the current batch.
The batch gets committed at the end if everything is successful.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -410,13 +410,12 @@ impl Chain {
 		pre_tx: Option<Transaction>,
 		block_hash: &Hash,
 	) -> Result<Vec<Transaction>, Error> {
-		// Get headers so we can rewind chain state correctly.
+		// Get header so we can rewind chain state correctly.
 		let header = self.store.get_block_header(block_hash)?;
-		let head_header = self.store.head_header()?;
 
 		let mut txhashset = self.txhashset.write().unwrap();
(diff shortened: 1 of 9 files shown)


yeastplume:
Good find indeed, but no theory here either as to how we haven't seen more issues related to this on T3.

PR #1393, opened 2018-08-20
Run header sync only when stalled or complete

### LINKED ISSUE #1357: sync head in apparent loop when syncing from scratch
Trying to sync the chain from scatch, and each time I re-start grin the sync head appears to progress for a while, but then gets stuck apparently processing a range of blocks over and over again. Had to restart dozens of times to get the header download to 100%. Haven't had time yet to look into this much further as I'm trying to get some secp256 stuff out of the way, but here's the log in the meantime:

[grin-problem-sync-loop.log](https://github.com/mimblewimble/grin/files/2290099/grin-problem-sync-loop.log)

ignopeverell:
Header sync used to run when complete (right count of blocks
received) or every 10 sec regardless. Sometimes 10 secs is not
enough, especially when a big batch of blocks bodies is arriving
at the same time.

This tweaks the time-based heuristic just a bit to only ask for
headers when stalled, meaning that the header head has not
*progressed* for 10 seconds.

Could maybe fully fix #1357 this time.

### DIFF
--- a/servers/src/grin/server.rs
+++ b/servers/src/grin/server.rs
@@ -264,6 +264,14 @@ impl Server {
 		Ok(())
 	}
 
+	/// Ping all peers, mostly useful for tests to have connected peers share
+	/// their heights
+	pub fn ping_peers(&self)-> Result<(), Error> {
+		let head = self.chain.head()?;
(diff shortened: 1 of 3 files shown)


ignopeverell:
Good point, thanks. Just fixed this.

Issue #1325, opened 2018-08-07
Fluffing a transaction with large number of outputs


jaspervdm:
I have been experimenting with large transactions for a bit, and I noticed that a transaction that has a large amount of outputs doesn't seem to fluff successfully. For reference I first tried a normal 1->2 transaction, which was mined quickly after submitting. Here are what the logs look like:
```
Aug 07 21:23:52.737 INFO Pushing transaction with 1 inputs and 2 outputs to pool.
Aug 07 21:23:52.737 DEBG pool: add_to_pool: cdef9cce, kernels - 1, stem? false
Aug 07 21:23:53.025 DEBG pool [txpool]: add_to_pool: cdef9cce, TxSource { debug_name: "push-api", identifier: "?.?.?.?" }
Aug 07 21:23:53.031 DEBG Send tx cdef9cce to 45.61.156.122:13414
Aug 07 21:23:53.031 DEBG Send tx cdef9cce to 185.53.158.12:13414
Aug 07 21:23:53.031 DEBG Send tx cdef9cce to 195.201.38.140:13414
Aug 07 21:23:53.032 DEBG Send tx cdef9cce to 178.62.100.159:13414
Aug 07 21:23:53.032 DEBG Send tx cdef9cce to 87.117.38.161:13414
Aug 07 21:23:53.032 DEBG Send tx cdef9cce to 128.199.165.106:13414
Aug 07 21:23:53.033 DEBG Send tx cdef9cce to 108.196.200.233:13414
Aug 07 21:23:53.033 DEBG Send tx cdef9cce to 94.130.229.193:13414
Aug 07 21:23:53.229 DEBG Received tx cdef9cce from TxSource { debug_name: "p2p", identifier: "?.?.?.?" }, going to process.
Aug 07 21:23:53.230 DEBG pool: add_to_pool: cdef9cce, kernels - 1, stem? false
Aug 07 21:23:53.336 DEBG Received tx cdef9cce from TxSource { debug_name: "p2p", identifier: "?.?.?.?" }, going to process.
Aug 07 21:23:53.349 DEBG Received tx cdef9cce from TxSource { debug_name: "p2p", identifier: "?.?.?.?" }, going to process.
Aug 07 21:23:53.350 DEBG Received tx cdef9cce from TxSource { debug_name: "p2p", identifier: "?.?.?.?" }, going to process.
[...]

PR #1394, opened 2018-08-21
Refactor compact block body

antiochp:
Mainly a refactor on compact blocks to make the internal implementation more like full blocks.

Also removed `WriteableSorted ` trait and `write_sorted()`.
We should not _ever_ be in a situation where we process a tx locally and _then_ sort it differently during serialization.

* extract compact block code out into `compact_block.rs`
* Introduce `CompactBlockBody` (consistent with TransactionBody)
* add `verify_sorted()` on `CompactBlockBody`
* add `validate()` on `CompactBlock`
* call `validate()` when deserializing a compact block
* get rid of `WriteableSorted ` trait and `write_sorted()`.


core/src/core/compact_block.rs
```diff
@@ -0,0 +1,161 @@
...
+	/// List of full kernels - specifically the coinbase kernel(s)
+	pub kern_full: Vec<TxKernel>,
+	/// List of transaction kernels, excluding those in the full list
+	/// (short_ids)
+	pub kern_ids: Vec<ShortId>,
+}
+
+impl CompactBlockBody {
+	fn validate(&self) -> Result<(), Error> {
+		self.verify_sorted()?;
```
hashmap: May be just `{ self.verify_sorted() }`

core/src/core/compact_block.rs
```diff
@@ -0,0 +1,161 @@
...
+		Ok(())
+	}
+
+	// Verify everything is sorted in lexicographical order.
+	fn verify_sorted(&self) -> Result<(), Error> {
+		self.out_full.verify_sort_order()?;
+		self.kern_full.verify_sort_order()?;
+		self.kern_ids.verify_sort_order()?;
```
hashmap: We may replace last 2 lines with `self.kern_ids.verify_sort_order()`

core/src/core/compact_block.rs
```diff
@@ -0,0 +1,161 @@
...
+/// before block) and can go request missing tx data from peers if necessary to
+/// hydrate a compact block into a full block.
+#[derive(Debug, Clone)]
+pub struct CompactBlock {
+	/// The header with metadata and commitments to the rest of the data
+	pub header: BlockHeader,
+	/// Nonce for connection specific short_ids
+	pub nonce: u64,
+	/// Container for out_full, kern_full and kern_ids in the compact block.
+	pub body: CompactBlockBody,
```
hashmap: do we need it to be `pub`? It's handy, but may be confusing to have methods below and direct access to body

core/src/core/transaction.rs
```diff
@@ -240,7 +239,7 @@ impl PMMRable for TxKernel {
 	}
 }
 
-/// TransactionBody is acommon abstraction for transaction and block
+/// TransactionBody is a common abstraction for transaction and block
```
hashmap: 👍

core/src/core/transaction.rs
```diff
@@ -269,15 +268,9 @@ impl Writeable for TransactionBody {
...
-		// Consensus rule that everything is sorted in lexicographical order on the
-		// wire.
-		let mut inputs = self.inputs.clone();
-		let mut outputs = self.outputs.clone();
-		let mut kernels = self.kernels.clone();
-
-		inputs.write_sorted(writer)?;
-		outputs.write_sorted(writer)?;
-		kernels.write_sorted(writer)?;
+		self.inputs.write(writer)?;
```
hashmap: Nice!

core/src/core/transaction.rs
```diff
@@ -290,9 +283,18 @@ impl Readable for TransactionBody {
 		let (input_len, output_len, kernel_len) =
 			ser_multiread!(reader, read_u64, read_u64, read_u64);
 
-		let inputs = read_and_verify_sorted(reader, input_len)?;
-		let outputs = read_and_verify_sorted(reader, output_len)?;
-		let kernels = read_and_verify_sorted(reader, kernel_len)?;
+		let inputs = read_and_verify_sorted(reader, input_len).map_err(|e| {
```
hashmap: I found a rule - don't log error if you don't handle it - very useful, especially if an error bubbles up and each layer logs it. It could be successfully handled, in this case we will just have lines of noise in the log or logged with the stack trace so we get all data in one place

core/src/core/transaction.rs
```diff
@@ -290,9 +283,18 @@ impl Readable for TransactionBody {
...
+			error!(LOGGER, "tx body: read: inputs: {:?}", e);
+			e
+		})?;
+		let outputs = read_and_verify_sorted(reader, output_len).map_err(|e| {
```
hashmap: I wonder - can we introduce a constructor function for `TransactionBody` and sort there? Granted we need to track all direct instantiations of the struct, but it may simplify code and gives us better guarantees?

core/src/core/compact_block.rs
antiochp: I actually prefer it with an explicit `Ok(())` at the end, to keep the various lines above all consistent.
There's also something in rust (I think `From` is only called automatically from within `try!`?) where errors will not necessarily be converted without the explicit `...?` syntax.
We don't need this everywhere but I think its better to just be consistent with this.

core/src/core/compact_block.rs
antiochp: Same comment as below - I prefer it being explicit and consistent with other examples where we have multiple lines of `...?;` before the `Ok(())`.

core/src/core/transaction.rs
antiochp: This was actually intentional (and temporary) as we try and track down the underlying issue for #1325.
Right now we're just getting a very generic (and context free) `CorruptedData` error propagating up.

core/src/core/transaction.rs
antiochp: I think I was close to trying something similar to this. I considered a `sort()` fn on `TransactionBody`.
But I like the idea of a constructor and doing it there - I'll take another pass at this.

PR #1395, opened 2018-08-21
Ensure wallet output heights are always correct

yeastplume:
Node API will now return the output's height while doing an unspent check, which will then be updated int the wallet when refreshing a wallet node. In theory, an unspent and validated output in the wallet should always have the correct block height after this change.

Much simpler to just return the height every time rather than force clients to make an explicit call to get the height. The block height search for an output against the chain is O(log n) at worst so hopefully shouldn't create too much of an overhead.

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -69,7 +69,8 @@ fn get_output(chain: &Weak<chain::Chain>, id: &str) -> Result<(Output, OutputIde
 
 	for x in outputs.iter() {
 		if let Ok(_) = w(chain).is_unspent(&x) {
-			return Ok((Output::new(&commit), x.clone()));
+			let block_height = w(chain).get_header_for_output(&x).unwrap().height;
+			return Ok((Output::new(&commit, block_height), x.clone()));
 		}
 	}
 	Err(ErrorKind::NotFound)?
@@ -320,6 +321,7 @@ impl TxHashSetHandler {
 			spent: false,
 			proof: None,
 			proof_hash: "".to_string(),
+			block_height: None,
 			merkle_proof: Some(merkle_proof),
 		})
 	}
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -157,14 +157,17 @@ pub enum OutputType {
 pub struct Output {
 	/// The output commitment representing the amount
 	pub commit: PrintableCommitment,
+	/// Height of the block which contains the output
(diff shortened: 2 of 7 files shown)


antiochp:
I'd be interested to see how much of a perf hit we do take with this.
If its O(log n) for _every_ output in the wallet that we need to do an unspent check for this might add up in a wallet with lots of outputs.

yeastplume:
Indeed, will keep an eye on it. However, each operation in the lookup is just pulling a block header from the header cache, which should be optimised and cached and available in memory anyhow, so in theory the hit should be negligible. 

The alternative is to have to keep another state var in a wallet output keeping track of whether a particular output has had its height updated and calling the API again to get the correct height if not, which leads to more thinks to keep track of. Would much rather try the simple and straightforward way first and do it the other way if it turns out to be too much of a hit.

yeastplume:
Or just make the explicit call when an output is being changed to confirmed.. but this isn't enough to handle a re-org case.

yeastplume:
Just some quick test output on a release build:

```
Aug 21 11:23:09.213 DEBG GET_HEADER_FOR_OUTPUT_TIME_START: 213764
Aug 21 11:23:09.214 DEBG GET_HEADER_FOR_OUTPUT_TIME_END: 214068
Aug 21 11:23:09.214 DEBG GET_HEADER_FOR_OUTPUT_TIME_ELAPSED: 0.304 ms
Aug 21 11:23:09.214 DEBG GET_HEADER_FOR_OUTPUT_TIME_START: 214417
Aug 21 11:23:09.214 DEBG GET_HEADER_FOR_OUTPUT_TIME_END: 214693
Aug 21 11:23:09.214 DEBG GET_HEADER_FOR_OUTPUT_TIME_ELAPSED: 0.276 ms
Aug 21 11:23:09.215 DEBG GET_HEADER_FOR_OUTPUT_TIME_START: 215105
Aug 21 11:23:09.215 DEBG GET_HEADER_FOR_OUTPUT_TIME_END: 215369
Aug 21 11:23:09.215 DEBG GET_HEADER_FOR_OUTPUT_TIME_ELAPSED: 0.264 ms
```

So each call on the current chain size takes say about 0.3 ms at worst. 1000 outputs in a wallet would add .3 seconds to the query time, 10000 would add a full 3 seconds (but this wouldn't be the only thing slowing it down)

PR #1398, opened 2018-08-21
Get peers as quick as possible

hashmap:
Currently we get 2 peers and then wait for 20 seconds before trying to get more. As result `Waiting for peers` stage of sync takes 31 seconds on my machine, with this fix it is 3-7 seconds.

### DIFF
--- a/api/src/client.rs
+++ b/api/src/client.rs
@@ -90,8 +90,7 @@ where
 	IN: Serialize,
 {
 	let req = create_post_request(url, input)?;
-	let resp = send_request(req)?;
-	println!("Resp: {}", resp);
+	send_request(req)?;
 	Ok(())
 }
 
--- a/api/src/client.rs
+++ b/api/src/client.rs
@@ -90,7 +90,8 @@ where
 	IN: Serialize,
 {
 	let req = create_post_request(url, input)?;
-	send_request(req)?;
+	let resp = send_request(req)?;
+	println!("Resp: {}", resp);
 	Ok(())
 }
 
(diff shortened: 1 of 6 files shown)


ignopeverell:
Nice! Seems to be failing a `servers` test, hopefully an easy fix.

hashmap:
@ignopeverell yeah, bad timing, grin now is fast enough to add itself as a potential peer but not fast enough to remove itself from the peer list, so test gets 2 peers instead of 1. Which is actually nice catch, it seems that we treat any peer in a the peer list as connected one, I'll add some logic to filter it out. 
```
Peers: [PeerData 
{ addr: V4(127.0.0.1:40002), capabilities: UNKNOWN, user_agent: "", flags: Healthy, last_banned: 0, ban_reason: None }, 
PeerData { addr: V4(127.0.0.1:40004), capabilities: FULL_HIST | TXHASHSET_HIST | PEER_LIST | FULL_NODE, user_agent: "MW/Grin 0.3.0", flags: Healthy, last_banned: 0, ban_reason: None 
}]
```

hashmap:
Some scope creep happened. Now this fix includes:

* More rustfmt, please ignore fast sync formatting changes
* `connected_peers` return connected peers, not all
* p2p test was changed to reflect change in `connected_peers` and more aggressive peer discovery (we have self in peer list now, in Defunct state)
* Schedule less peers to connect (config.peer_max_count - because we call this function more often) and make sure that peers are unknown (otherwise a peer may be "rediscovered" and change state for example from `Banned` to `Healthy`)
* Fix logging in tests in `servers/tests/api.rs`

hashmap:
Don't merge yet, found that TUI reports different number of peers in `Basic Status` and `Peers and Sync`, will investigate and fix.

hashmap:
Ready for review

Issue #1372, opened 2018-08-18
HeaderSync Speedup


garyyu:
Currently I need **11 minutes** to complete a full fast-sync on my mac air(Early 2015).

And reported by @antiochp :
> ~7mins for full fast sync

Reported by @hashmap :
> 10 mins for fast sync from scratch for me

I hope this fast sync on current ~50k height should be **less than 5 minutes** on a slow machine like mine, and less than 3 minutes on a fast new machine.
 
Giving a profile for my **11 mintues**:

No. | Time | Percentage | Tasks
------------ | ------------ | ------------- | -------------
1 | 6 m | 54.5% | HeaderSync
2 | 12 s | 1.8% | txhashset download (size example: 46M bytes)
3 | 1 m 12 s | 10.9% | copy and save file: txhashset.zip
4 | 5 s | 0.7% | unzip and check txhashset.zip
5 | 50 s | 7.6% | validate the output|rproof|kernel  mmrs
6 | 14 s | 2.1% | validate_roots + verify_kernel_sums + verify_kernel_signatures
7 | 38 s | 5.8% | Range proof validation (batch)
8 | 6s | 0.9% | txhashset::extending()
9 | 1 m 30 s | 13.6% | BodySync
10 | 14s | 2.1% | others

Looks like the spent time in No.3 "copy & save file: txhashset.zip" is unreasonable, will check later.

And let's come to the biggest part: header sync!

Current implementation in header sync is using **sequential** design, each time the node only request one package (512 headers) from one peer:
> sync: request_headers: asking x.x.x.x:13414 for headers

This request normally need average 3 seconds on my machine to receive the requested header package:
> Aug 17 23:46:03.749 DEBG sync: request_headers: asking 51.x:13414 for headers, 
> Aug 17 23:46:07.557 INFO Received block headers  from 51.x:13414
> Aug 17 23:46:08.813 DEBG sync: request_headers: asking 195.x:13414 for headers, 
> Aug 17 23:46:11.804 INFO Received block headers  from 195.x:13414

So, I propose to change it from sequential to parallel (at least requesting from 3 peers in the same time), this suppose to suppress the 6 minutes HeaderSync to 2 minutes.

Before trying this, do you think it's a good direction or not?  or is there any constrain on that parallel ?

ignopeverell:
[...]

Issue #1358, opened 2018-08-15
Aggsig signature hash being computed incorrectly


yeastplume:
See: https://github.com/mimblewimble/secp256k1-zkp/pull/21/files

Only the first 8 bytes of the public key in the sighash are being hashed into the signature hash, whereas the full 32 bytes should be written. Will need to wait for a hard fork and provide a switch to that function in order to fix this.

ignopeverell:
Oops. Yes. Nicely spotted @garyyu

ignopeverell:
Is this merged? Can we close this now?

garyyu:
Yes. Closed by #1501 and the PR in secp256k1-zkp repo: https://github.com/mimblewimble/secp256k1-zkp/pull/22

PR #1400, opened 2018-08-21
HeaderSync optimization (#1372)

garyyu:
With this optimization for https://github.com/mimblewimble/grin/issues/1372, the HeaderSync procedure can be 200% faster than before on my mac air(early 2015), i.e, **3 minutes** compared to **6 minutes** before:
```
Aug 21 19:02:20.615 DEBG sync_state: sync_status: Initial -> HeaderSync { current_height: 0, highest_height: 61131 }
Aug 21 19:05:19.367 DEBG sync_state: sync_status: HeaderSync { current_height: 60698, highest_height: 61134 } -> TxHashsetDownload
```


ignopeverell:
This is a great performance improvement! However I'd like to propose a couple changes:

1. It seems most of the new code is just about buffering reads instead of consuming straight from the TCP connection. Couldn't we get the same gain by using `BufStream` as described in #1402 ?
2. If this is indeed the gain, we should do it for every message, not only headers.

garyyu:
Please look my comment in https://github.com/mimblewimble/grin/issues/1402 , `BufStream` won't help anything.

Regarding:
> It seems most of the new code is just about buffering reads instead of consuming straight from the TCP connection

Why you have this feeling? This new code's not related to any buffering reads.
Actually, this new code gave 2 changes:

1. Instead of receiving all 511 Headers and then go to process all Headers in one call, I split it into pieces, each piece containing 8 Headers and go to process immediately, then go to download next 8 Headers. I call this new procedure as `streaming`. This can avoid a long (about 1 second on my pc) processing time on 511 Headers at one call.
2. 2nd change is in `get_locator()`, which you gave the comment and I need rework it.

garyyu:
I just realized it's better to split `get_locator()` optimization from this PR, for the good rule of `one PR only solve one problem`.

Another reason is that `get_locator()` is security related, so better to give dedicated code review on it.

Will remove it from this PR and submit a new PR for it.

ignopeverell:
Sounds good, thank you for removing the `get_locate` changes! To summarize a gitter discussion:

1. Add a simpler method on `Message` that just returns the &`'mut TcpStream` so it can be read from `protocol.rs`
2. Move your code to a simple utility function in `protocol.rs`
3. Potentially simplify given that all headers have equal size
4. Remove the `rustfmt` changes from `sync.rs` as you're not touching it.

And thank you for the great work!

garyyu:
Thanks for a good review! 
Refactoring is finished according to above comments.

ignopeverell:
Have you seen my previous comment? Looks like header size calculation will fail in presence of headers with different Cuckoo sizes.

garyyu:
> Just meant to amend my comments first as I realized I made a mistake. Point 3 is actually incorrect as we allow Cuckoo sizes greater than 30 now. So the proof of work part of the header could be 30*42 bits, 31*42 bits, 32*42 bits, etc. Sorry for that.

Sorry, I don't know why I thought that doesn't impact :)  (it could because you firstly said `simplify `, and then you said `made a mistake`, so I took it as `-1 * -1 = +1`) . But indeed it does!

The Headers Vector with different size Header is totally out of control, both in this new code and in the old code, **I can't imagine how possible to deserialize a Vector with each different size of element**. I will read more code to confirm.  If it's an existing issue, that will be a major bug and better to be handled in an independent new PR.

garyyu:
The **good thing**:  After reading `Headers` deserialization code, I confirm there's no problem for the old code, since `Proof` has its own deserialization code which can handle different Cuckoo size.

Now the **bad** thing: a very upset finding, we don't have a `size` field in `BlockHeader` structure! then, without a deserialization, we don't know the exact size of this BlockHeader.  But, to split the `Headers`, I need to know the size of this BlockHeader before deserialization!

This fact could make this optimization not feasible anymore!

Any suggestion to this?  perhaps it's too late to add `size` field to `BlockHeader` structure, which will be consensus breaking?

ignopeverell:
It's too late and in theory not needed, as each header has enough information to get deserialized (the cuckoo size is in them).

One idea would be to read enough bytes for say Cuckoo34 headers. You'll overshoot, but then you can reuse the unused bytes in the next iteration, placing them first in the buffer.

garyyu:
> One idea would be to read enough bytes for say Cuckoo34 headers. You'll overshoot, but then you can reuse the unused bytes in the next iteration, placing them first in the buffer.

sounds a feasible solution. but :)    a little bit ugly, and need rewrite a deserialize function for that.

Even I agree it could be too late to add `size` for BlockHeader, I still strongly propose a hard fork on Testnet3.  Because it's not only for this minor optimization, also for another maybe more important use case:  Block receiving.

When a Grin node receive a Block, currently we complete this whole block receiving from network, then start block validation, if find this received block already received before (or Header validation fail), we throw this block.  That could be a hole for attacking. A better solution is to receive its Header firstly, than  start a validation for this Header, if not needed, we don't need receive the bigger body part of this block.

Does this make sense to worth a hard fork?

And remember we have a security bug in https://github.com/mimblewimble/grin/issues/1358, which also need a hard fork.

Anyway, up to your decision:)

ignopeverell:
You're arguing for header-first propagation. We already have compact blocks, which are pretty much the same, maybe better. By the way none of these are strictly hard forks (nothing forks), they're just breaking protocol changes that can lead to network partitioning if not properly handled. Also we only receive a single block/header in that cases, the total size is the same as the message size.

I may be mistaking but it doesn't seem the lack of header size requries requires rewriting a deserialization. You can just use the same deserialization, it'll do the right thing from a `Reader` as long as you feed it enough bytes.

garyyu:
Complete the enhancement to support variable BlockHeader size in one Headers message, from Cuckoo30 to Cuckoo36.

core/src/core/block.rs
```diff
@@ -265,6 +299,20 @@ impl BlockHeader {
 	pub fn total_kernel_offset(&self) -> BlindingFactor {
 		self.total_kernel_offset
 	}
+
+	/// Ser size of this header
+	pub fn size_of_ser(&self) -> usize {
```
ignopeverell: How about just `serialized_size`? Similar for the utility functions above.

garyyu:
Thanks for approving!

And function name has been changed.

The travis-ci panic at `grin_chain::txhashset::extending`  could be a real report. But I can't reproduce on Mac, could only happen in Linux platform. I have to wait to tomorrow for Linux test since now only laptop with me, in a trip.

ignopeverell:
I think you're running into an issue with the fast sync test that I fixed a few days ago. Can you merge master back to double-check?

ignopeverell:
Starting to lose track with other PRs looking at Travis failures. Are there still test regressions caused by this PR? I've restarted tests many times but they always fail (while master mostly passes).

hashmap:
@ignopeverell on my machine I see a pr issue:
```
Aug 27 14:16:28.767 DEBG Client 127.0.0.1:12001 connection lost: Connection(Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })
Aug 27 14:16:28.771 DEBG Client 127.0.0.1:12000 connection lost: Connection(Custom { kind: InvalidData, error: StringError("headers_streaming_body") })
Aug 27 14:16:28.771 WARN sync: no peers available, disabling sync
Aug 27 14:16:28.771 DEBG sync_state: sync_status: HeaderSync { current_height: 0, highest_height: 30 } -> NoSync
```

garyyu:
@ignopeverell  please let me finish https://github.com/mimblewimble/grin/pull/1434 firstly, to fix the boring false alarm of `travis-ci`. then will come back to check this PR again.

ignopeverell:
Indeed finally! Sorry it was so difficult, but thanks for another great PR!

PR #1401, opened 2018-08-21
Add a peers_preferred option in Grin

quentinlesceller:
Add a peers_preferred option in Grin.toml.
Grin will attempt to connect to this peer no matter what.

### DIFF
--- a/grin.toml
+++ b/grin.toml
@@ -25,14 +25,6 @@ api_http_addr = "127.0.0.1:13413"
 
 db_root = ".grin"
 
-#How to seed this server, can be None, List, WebStatic or DNSSeed
-#
-#seeding_type = "None"
-
-#If seeding_type = List, the list of peers to connect to.
-#
-#seeds = ["192.168.0.1:13414","192.168.0.2:13414"]
-
 #The chain type, which defines the genesis block and the set of cuckoo
 #parameters used for mining. Can be:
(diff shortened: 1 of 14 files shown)


antiochp:
We already have the following - 

```
#hardcoded peer lists for allow/deny
#will *only* connect to peers in allow list
#peers_allow = ["192.168.0.1:13414", "192.168.0.2:13414"]
#will *never* connect to peers in deny list
#peers_deny = ["192.168.0.3:13414", "192.168.0.4:13414"]
```

Is this intended to be more flexible (and in addition to) the existing allow/deny lists?

quentinlesceller:
Yes all peers are allowed. Plus we will peer to a selected list of peers. E.g. have a cluster of grin nodes that connect to each others and the rest of the network.

grin.toml
```diff
@@ -33,6 +33,9 @@ db_root = ".grin"
 #
 #seeds = ["192.168.0.1:13414","192.168.0.2:13414"]
 
+#A list of preferred peers to connect to
```
antiochp: Can we nest this under p2p config so it sits alongside `peers_allow` and `peers_deny`?

quentinlesceller:
Done @antiochp. I also picked up the TODO along the way :).

antiochp:
:+1:

PR #1407, opened 2018-08-22
remove break from msg read_exact

antiochp:
What is this `break` for?

https://github.com/mimblewimble/grin/blob/105bfacaa6ff2f5dbc24c00e0b3164999d025c8b/p2p/src/msg.rs#L138-L143

Right now if we hit an `Interrupted` or a `WouldBlock` earlier and we have 0 bytes in the buffer (not yet successfully read anything) then we just `break` and do not keep looping.

This feels wrong but likely I'm misunderstanding something?
Wrong as in there is a _huge_ edge case here where we just stop reading msgs before we have even started.

@ignopeverell

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -138,8 +138,6 @@ pub fn read_exact(
 		if !buf.is_empty() {
 			thread::sleep(sleep_time);
 			count += 1;
-		} else {
-			break;
 		}
 		if count > timeout {
 			return Err(io::Error::new(


hashmap:
Copy from gitter - We read to a buffer and assign local var `buf` to the tail of the buffer which has to be filled. When it’s empty it actually means that it’s filled (or full:) so we can exit. 

Reminds me list processing using recursion

ignopeverell:
Just to complement, `read_exact` is still meant to be async and collaborative. So it returns immediately if there's nothing to read, so it can be looped on cooperatively with other tasks (like writes).

antiochp:
> buf is actually uninitialized part of actual buffer
Thanks @hashmap that makes a lot of sense now. That's pretty neat now that I get what its doing...

PR #1419, opened 2018-08-25
cache rangeproof and kernel signature verification results

antiochp:
Still a work in progress.
Some code cleanup still to do, but the basics are there and it works both locally and against testnet3.

* Introduced a new `VerifierCache` trait to handle caching the results of rangeproof verification and kernel signature verification.
* The `LruVerifierCache` uses a `LruCache` to cache verification results based on rangeproof hash and kernel hash so we only verify rangeproofs and kernel signatures once.
* Introduced a shared `LruVerifierCache` through from the top level server down to both the transaction pool and the block processing pipeline.

TODO - 
- [x] cleanup build warnings
- [x] docs/comments
- [x] how big should these caches be? (enough to cover at least one full block and included txs)
- [x] is an lru_cache the best thing to use here? We only need a bool per key (hash).

For another time - 
- [ ] ~investigate caching the following in addition (sums are non-trivial to compute) -~
  - [ ] ~block.verify_coinbase()~
  - [ ] ~verify_kernel_sums()~
  - [ ] ~block header kernel sum verification~

---------------
### Initial Performance Numbers

Scenario -
* Wallet pushes a tx with 20 inputs and 501 outputs to Node 1.
* Node 1 broadcasts the tx to Node 2.
* Node 1 mines a block containing the tx.
* Node 1 broadcasts a compact blocks to Node 2.

Node 1 (`300ms` from pushing tx to pool to sending to peers) - 
```
Aug 25 19:09:32.311 INFO Pushing transaction 312cba9b to pool (inputs: 20, outputs: 501, kernels: 1)
Aug 25 19:09:32.349 DEBG get_coinbase: BlockFees { fees: 0, height: 195, key_id: Some(Identifier(e341b9c9afbb7640bfda)) }
Aug 25 19:09:32.616 DEBG pool [txpool]: add_to_pool: 312cba9b, TxSource { debug_name: "push-api", identifier: "?.?.?.?" }, inputs: 20, outputs: 501, kernels: 1 (at block 34b94f62)
Aug 25 19:09:32.620 DEBG Send tx 312cba9b to 127.0.0.1:14414
```

Node 1 (`18ms` from processing new block to broadcasting to peers) -
```
Aug 25 19:09:32.939 DEBG get_coinbase: BlockFees { fees: 1985000000, height: 196, key_id: Some(Identifier(c70877169b2e58bdb6fd)) }
Aug 25 19:09:32.956 DEBG Built new block with 20 inputs and 502 outputs, network difficulty: 9, cumulative difficulty 695
...
Aug 25 19:09:32.982 DEBG pipe: process_block ff24f2f6 at 196 with 20 inputs, 502 outputs, 2 kernels
Aug 25 19:09:32.999 DEBG pipe: chain head ff24f2f6 @ 196
Aug 25 19:09:33.000 DEBG adapter: block_accepted: ff24f2f6
Aug 25 19:09:33.000 DEBG broadcast_compact_block: ff24f2f6, 695 at 196, to 1 peers, done.
```

Node 2 (`450ms` from handling tx payload to adding tx to pool) - 
```
Aug 25 19:09:32.626 DEBG handle_payload: received tx: msg_len: 360067
Aug 25 19:09:32.630 DEBG Received tx 312cba9b, inputs: 20, outputs: 501, kernels: 1, going to process.
Aug 25 19:09:33.091 DEBG pool [txpool]: add_to_pool: 312cba9b, TxSource { debug_name: "p2p", identifier: "?.?.?.?" }, inputs: 20, outputs: 501, kernels: 1 (at block 34b94f62)
```

Node 2 (`45ms` from handling compact block payload to accepting full block) -
```
Aug 25 19:09:33.121 DEBG handle_payload: received compact block: msg_len: 1192
Aug 25 19:09:33.121 DEBG Received compact_block ff24f2f6 at 196 from 127.0.0.1:13414, outputs: 1, kernels: 1, kern_ids: 1, going to process.
Aug 25 19:09:33.121 DEBG pipe: process_block_header at 196 [ff24f2f6]
Aug 25 19:09:33.121 DEBG adapter: txs from tx pool - 1
Aug 25 19:09:33.143 DEBG adapter: successfully hydrated block from tx pool!
Aug 25 19:09:33.143 DEBG pipe: process_block ff24f2f6 at 196 with 20 inputs, 502 outputs, 2 kernels
Aug 25 19:09:33.167 DEBG pipe: chain head ff24f2f6 @ 196
Aug 25 19:09:33.168 DEBG adapter: block_accepted: ff24f2f6
```


core/src/core/ok_verifier.rs
```diff
@@ -0,0 +1,178 @@
...
+	Secp(secp::Error),
+}
+
+impl From<secp::Error> for Error {
+	fn from(e: secp::Error) -> Error {
+		Error::Secp(e)
+	}
+}
+
+pub trait OKVerifier {
```
ignopeverell: I'm not sure I like having this in core and I'm not sure I like it facading validation. Wouldn't a bucketed (for each type, range proof or kernel, etc) hash LRU in `util` be simpler? We'd have a few more ifs but it's very localized and it seems making it explicit would also be better.

antiochp: I'm not sure I follow what you're saying.
I get that this probably shouldn't live in core - that makes sense.
But what do you mean by facading validation?

We need to be using the same instance of the verifier everywhere (to take advantage of the shared cache) so we either passit around or use a global singleton and I was trying to avoid introducing a singleton here.
Is that what you are proposing here? Have a singleton in util and just use it where necessary?

antiochp: Oh I think I understand (after taking a step back) - you are suggesting we keep the validation code in core (where it is currently) and pass cache(s) around directly instead of passing some kind of verifier thing around.
And then use the cache in the validation code.

ignopeverell: Correct, sorry for not being clearer.

antiochp:
Reworking this to pass a "cache" around and not a "verifier".

I'm not sure we can pull this out from core cleanly though. 
Even if its just a simple cache we need to know the types of the cache keys. 

If we're using `core:core::hash::Hash` as cache keys then its going to be hard to move this into `util`.

antiochp:
@ignopeverell ok rewritten to pass a cache around and not a verifier.
But - we call `tx.validate()` which lives in core and needs to be "cache aware" (as currently implemented). 
The cache itself needs to know about `core::core::hash::Hash` for cache keys at a minimum. 
So I'm not sure how to get the cache stuff out of core (without the validation logic itself living outside of core).

core/src/core/transaction.rs
```diff
@@ -558,12 +533,68 @@ impl TransactionBody {
...
-		self.verify_kernel_signatures()?;
+
+		// TODO
+		// move verify_rangeproofs() and verify_kernel_signatures()
+		// back in here
+		// but we need something in util that will let us call with
+		// a vec of outputs and a cache
+		// given a vec of outputs filter into smaller vec based on cached values
+		// read cache for multiple outputs
+		// write cache for multiple outputs etc.
```
ignopeverell: Maybe have the verifier itself do that filtering, from a `Vec<output>` and `Vec<TxKernel>`?

ignopeverell:
Sorry I wasn't clear enough, didn't mean we could remove all core dependencies completely. More that it's a core utility passed around all the way down, instead of half passed around and half facade.

Looks good to me!

hashmap:
I wonder if we should consider adding `Verify` trait to separate validation logic from the data structs. I see we already do some parts of validation on different level like `Output::batch_verify_proofs(&commits, &proofs)`. 

It would allow to decouple caching logic (or any other optimizations) from verification, e.g. `CashingVerifier` is a wrapper of `BasicVerifier` (middleware pattern), so when `verify_block` calls `verify_kernel` (of the same verifier) it first checks it in the cache and returns immediately if found or delegate to the inner verifier and then caches the result.

hashmap:
I suppose it may not play well with batch verification but I don't know enough details to judge.

antiochp:
Yeah may be worth exploring (separate PR though).

PR #1421, opened 2018-08-26
Improve file-based tx send to not use private file

ignopeverell:
Instead of generating a private file on send that needs to be
provided again in finalize, the private context information is
saved in the wallet db.

@yeastplume starting to wonder whether `Context` should be upgraded from `internal` now, how about `libwallet::types`?


yeastplume:
yes, makes sense, seems like a better place to expose it now

PR #1429, opened 2018-08-27
Cleanup generics

antiochp:
Cleanup and simplify a lot of this - 

```
impl<T> Handler for PoolPushHandler<T>
where
	T: pool::BlockChain + Send + Sync + 'static
{...}
```

We don't have many different `BlockChain` impls that we need to be generic over.

Same for `TxHashsetWriteStatus`.


hashmap:
👍  I think it makes code much easier to read.

PR #1430, opened 2018-08-27
Fix LocalServerContainer and remove backtraces

hashmap:
* LocalServerContainer used a different folder than regular server, so test could not find txhashset files
* Error stacktrace was printed in cases when error was expected, which created some noise and confusion


ignopeverell:
PR looks good but seems that you lost the rustfmt race. Can you merge?

hashmap:
@ignopeverell thanks, rebased, waiting for travis

PR #1432, opened 2018-08-28
Fix flakiness of simple_server_wallet test

hashmap:
It produces 1000+ otputs on a fast machine (?) and generates invalid URL as result (too many paramaters). We don't care about sending all outputs in this test, we just need something.


hashmap:
Hm, it's a result of https://github.com/mimblewimble/grin/pull/1430 because before all mined blocks were rejected (because of the wrong state folder).

hashmap:
discussed in gitter

PR #1433, opened 2018-08-28
Config + Default directories

yeastplume:
Mostly for https://github.com/mimblewimble/grin/issues/1257

Starting off by giving the executable the ability to spit out a default (hopefully versioned) `grin.toml` file (complete with comments, which has been most of the work so far), then will move on to rejigging the default paths and options.


yeastplume:
Generally:
* The stratum server and wallet listener are disabled by default
* Configuration files are now generated by the executable
* Configuration is split into two files, grin-server.toml and grin-wallet.toml
* Running `grin server` or `grin_wallet` commands from anywhere creates or uses files in the default `~/.grin` directory
* However if a `grin-server.toml` (for server commands) or a `grin-wallet.toml` file exists in the current directory, the options in those files are used instead. When these files are generated, the paths within are set up to use the current directory.

In more detail, the process of using grin would be like this:

Normal Use:

* Running `grin` from anywhere on a virgin system will create a `~/.grin` directory. This directory will contain a `chain_data` directory which contains the chain, a new default `grin-server.toml` file with paths filled out to point at the default directory location, and the `grin-server.log` file

* To have a wallet, the user needs to run `grin wallet -p [password] init` from anywhere. This will also create the `~/.grin' directory if it doesn't exist, generate a `grin-wallet.toml` file in there, and create a `wallet_data` directory which contains `wallet.seed` and lmdb files. All further calls to `grin wallet` will use those data files

* You don't have to have both a server and a wallet, but if you do the structure will look something like:
```
~./grin
   -chain_data/
      -lmdb/
      -txhashset/
   -wallet_data/
      -db/
         -/lmdb
      wallet.seed   
   grin-server.toml
   grin-wallet.toml
   grin-server.log
   grin-wallet.log
```

To run the server in a particular directory, the process is as follows (from the target directory).

* `grin server config` will create a `grin-server.toml` file in the current directory, with the paths configured to create the same structure for server files as above. (It then exits, creating no more files)

* `grin server [whatever args] run` will use the `grin-server.toml` file in the current directory if one exists. The first `grin server` command performed in this directory will create the database and log files.

To run the wallet in a particular directory:

* `grin wallet -p [password] init -h` will create a `grin-wallet.toml` file with paths configured to use the current directory and generate the seed and database there, same structure as above.

* Subsequent calls to `grin wallet` will use settings the `grin-wallet.toml` file if one exists in the current directory, or the default in '~/.grin` otherwise

That is more or less it. The main differences in approach are:
* Anyone wanting to use a wallet will have to initiate it explicitly before running any commands (so one isn't created by default with an empty password)
* Anyone wanting to mine will need to have a bit more awareness than a regular user, but it should be within their capabilities to run `grin wallet init` and `grin wallet listen` before enabling the stratum server.

Issue #1439, opened 2018-08-29
unexpected fail of peer connect


garyyu:
Since https://github.com/mimblewimble/grin/issues/1424 had been closed by github automatic, I open this new issue to track this remaining issue, which was found when fixing https://github.com/mimblewimble/grin/issues/1424.

The following error log is a better way to describe this problem than words description:

```
Aug 29 08:44:34.920 WARN Error accepting peer 127.0.0.1:63069: Connection(Custom { kind: WouldBlock, error: StringError("read_exact") })
Aug 29 08:44:34.921 DEBG connect_and_req: from 0.0.0.0:10020 to 127.0.0.1:10000, is Defunct. Connection(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" })

Aug 29 08:44:35.923 DEBG connect_and_req: from 0.0.0.0:10020 to 127.0.0.1:10000, retrying 1
Aug 29 08:44:35.927 DEBG connect_and_req: from 0.0.0.0:10020 to 127.0.0.1:10000, ok. attempting send_peer_request
Aug 29 08:44:36.929 DEBG monitor_peers: on 0.0.0.0:10020, 1 connected (1 most_work). all 5 = 5 healthy + 0 banned + 0 defunct

```

will give further research later.

garyyu:
Test again with the network capture with `wireshark`:
```
Aug 29 13:16:17.644 DEBG monitor_peers: on 0.0.0.0:10030, 0 connected (0 most_work). all 0 = 0 healthy + 0 banned + 0 defunct
Aug 29 13:16:17.644 DEBG monitor_peers: no preferred peers
Aug 29 13:16:17.644 DEBG monitor_peers: no dandelion relay updating
Aug 29 13:16:17.645 DEBG Could not update dandelion relay
Aug 29 13:16:17.645 WARN Error accepting peer 127.0.0.1:50332: Connection(Custom { kind: WouldBlock, error: StringError("read_exact") })
Aug 29 13:16:17.645 DEBG connect_and_req: on 0.0.0.0:10030. connect to 127.0.0.1:10000 is Defunct. Serialization(UnexpectedData { expected: [30], received: [0] })
...
Aug 29 13:16:17.646 DEBG Connected! Cumulative 1 offered from V4(127.0.0.1:10000) "MW/Grin 0.3.0" FULL_HIST | TXHASHSET_HIST | PEER_LIST | FULL_NODE
[...]

PR #1440, opened 2018-08-29
fix for peer connect fail on 'WouldBlock' Error

garyyu:
This is the fix for https://github.com/mimblewimble/grin/issues/1439.

Root caused is analyzed in #1439.

To limit this fix only for solving peer connection problem, I choose to modify `accept()` function in `handshake.rs`, giving 10ms retrying to call `read_message`.

### DIFF
--- a/p2p/src/handshake.rs
+++ b/p2p/src/handshake.rs
@@ -13,8 +13,10 @@
 // limitations under the License.
 
 use std::collections::VecDeque;
+use std::io;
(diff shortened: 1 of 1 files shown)


ignopeverell:
Nice catch! Maybe you could reuse the blocking variation of `read_exact` instead of re-implementing the loop here? Just `read_exact` the `Hand` message?

garyyu:
Thanks, now it looks much more simple.

PR #1441, opened 2018-08-29
complete the 'simulate_seeding' test case for automatic test

garyyu:
Test case `simulate_seeding` looks like an unfinished job: 
```
#[test]
fn simulate_seeding() {
	...
	let servers = pool.run_all_servers();
	stop_all_servers(servers);	<<<< run servers then stop server immediately!
	...
```
 Just add a small piece of code to complete this test case.

**Note**:
1. This test case need the fix in https://github.com/mimblewimble/grin/pull/1434, otherwise it will have 50% probability to fail for:
```
thread 'simulate_seeding' panicked at 'assertion failed: `(left == right)`
  left: `3`,
 right: `4`', servers/tests/simulnet.rs:129:2
```

2. Please feel free to ignore this PR. But at least add a `#[ignore]` before it, if it's not useful anymore.

### DIFF
--- a/servers/tests/simulnet.rs
+++ b/servers/tests/simulnet.rs
@@ -105,6 +105,9 @@ fn simulate_seeding() {
 
 	pool.create_server(&mut server_config);
 
+	// wait the seed server fully start up before start remaining servers
+	thread::sleep(time::Duration::from_millis(1_000));
+
 	// point next servers at first seed
 	server_config.is_seeding = false;
 	server_config.seed_addr = String::from(format!(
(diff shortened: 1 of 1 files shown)


servers/tests/simulnet.rs
```diff
@@ -116,10 +119,27 @@ fn simulate_seeding() {
...
 	}
 
-	// pool.connect_all_peers();
-
 	let servers = pool.run_all_servers();
+	thread::sleep(time::Duration::from_secs(5));
+
+	// Check they all end up connected.
+	let url = format!("http://{}:{}/v1/peers/all", &server_config.base_addr, 30020);
+	let peers_all = api::client::get::<Vec<p2p::PeerData>>(url.as_str()).map_err(|e| Error::API(e));
```
hashmap: Looks good! Nitpick - why do we need a custom error and map here to it?

hashmap: Remove `.map_err()` and `Error` struct at the end of the file

Issue #1404, opened 2018-08-22
thread::yield_now() vs thread::sleep()


garyyu:
I open this issue to record this little research.

To recap the discussion in Gitter:

> Ignotus Peverell @ignopeverell 11:53
> 
> makes me think there should be a better alternative to sleep(1) in read_exact as well, that seems suboptimal
>
> Gary Yu @garyyu 12:01
> yield_now() will give more chance to read since read is on a async TcpStream, Yes? I can try it to check whether the count number increase. We want less sleep() call, right?
>
> Ignotus Peverell @ignopeverell 12:03
> the sleep exists just to let some time for the buffer to fill and have the next read get something, but 1ms is a long time, we actually a short pause to avoid a busy loop, so yield_now() might be it
alternatively, sleep for some nanosecs instead of 1 ms
>
> Gary Yu @garyyu 12:04
> OK, will try it
> 
> Ignotus Peverell @ignopeverell 12:05
> thanks! I was rewriting the whole p2p crate when I wrote this so didn't really have time to finick or check perfs around it
> so thanks for looking into it!
> 

1. About `std::thread::yield_now()`, I test it and it's very strange that it doesn't work like our expectation, look like once call it, the `scheduler` will never come back to wake up this thread anymore. Then suddenly I realized that `yield_now()` is not what we want, because in our `read_exact()` context, we indeed want to stick in this thread for the TCP stream data integrity.

2. And about changing `thread::sleep(1 ms)` to `thread::sleep(1 us)` (or `10 us`), need more test to check the impact and will update later.

garyyu:
I change to `10us`, and add a line of log for test here:
```
pub fn read_exact(
    ...
		match conn.read(buf) {
			Ok(0) => break,
			Ok(n) => {
				let tmp = buf;
				buf = &mut tmp[n..];
				read += n;
				debug!(                        <<<< add a line of log here!
					LOGGER,
					"read_exact - read {} bytes, progress: {}/{}, yield count: {}", n, read, exact, count,
				);
			}

```

And then test with the `txhashset` downloading.

Sometimes it looks good, only sleep for tens or at most hundreds of macro seconds:
```
...
Aug 22 16:11:23.152 DEBG read_exact - read 2896 bytes, progress: 3376/8000, yield count: 1
Aug 22 16:11:23.153 DEBG read_exact - read 2896 bytes, progress: 6272/8000, yield count: 3
[...]

PR #1453, opened 2018-08-31
improve: use 10us sleep instead of 1ms in read_exact

garyyu:
Upon the discussion which is recapped in https://github.com/mimblewimble/grin/issues/1404, it's better to sleep for some nanosecs instead of 1 ms in `read_exact()`.

This PR include 2 little changes:
1. use 10us sleep instead of 1ms in `read_exact`;
2. use `time::Duration` for the `timeout` parameter, instead of a counter of `10000`, `20000` and so on.


garyyu:
Oh, do you mean that `continue outer` i just added is also not needed ?
Then it will put another orphan (if multiple in same height) again into the orphans queue. Do you need it?

Issue #1428, opened 2018-08-27
Wallet BIP-32 Compatibility


yeastplume:
At the moment, the wallet key derivation is a simple master/child keys setup, but it should be modified to conform to BIP-32, allowing for master node/child wallet key derivations (which should make the life of wallet authors and service providers much easier).

https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki

mcdallas:
I would like to give this a go, however BIP32 uses HMAC-SHA512 instead of blake2 as well as SHA256 and RIPEMD160 (also pbkdf2 if BIP39 support is to come later). This needs one or more crypto libraries as dependancies. What should be the approach to this?

antiochp:
If we can I think we should stick with blake2 which does not require any special HMAC construction and keep as close as this allows to BIP32.

hashmap:
However we may want to consider hw wallets support there.  Trezor SDK supports BIP-32 directly, Ledger Nano S SDK provides HMAC-SHA512, SHA256  and RIPEMD160 (perhaps BIP-32 on higher level) and doesn't support blake2. Of course we can bring additional libraries as part of the app, but size limits are pretty tough.

yeastplume:
I was looking into getting dirty with crypto functions again, but it seems everyone's favourite cryptographer @apoelstra has already gone and done the bulk of the work for us here: https://github.com/rust-bitcoin/rust-bitcoin/blob/master/src/util/bip32.rs

So in my head this work should entail:

* Cut and paste what's needed from the above code
* Modifying the above code to use use blake2 (or perhaps allow for many hash algos given @hashmap's very valid point above? Probably best to slot the hashing into a trait for now and perform tests with a few different hashing algos, ensuring the BIP32 test vectors match at all times). 
* Interfacing into grin's BIP32 extended key via an implementation of the Keychain trait
* Modifying Keychain trait in accordance with whatever findings come from trying to integrate the BIP32 key derivation
* Using trait in wallet by default (for new wallets) but also allowing existing wallets to use the existing key derivation (need to flag this somewhere). To keep annoyance to a minimum, wallets should use one derivation or the other, not both
* Wallet restore needs to support both types of derivation

apoelstra:
[...]

PR #1454, opened 2018-08-31
BIP32 Lib Integration

yeastplume:
First part of #1428, with this PR I just intend to

* Integrate @apoelstra's BIP-32 implementation into our code in the least impactful way possible
* Figure out what hashing we'll support and how. Most likely thinking for now I want to abstract the hashing into a trait and provide a Blake2 implementation of it, then import and have an implementation for the actual BIP-32 SHA/ripemd hashing just for test cases
* Get tests in place

The way this is getting integrated really making me think we need a better separation between wallet and server code, (for instance, I'd like to be able to import whatever hash libs into the wallet code without worrying about bloating grin) but we can handle that another day.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -666,6 +666,7 @@ dependencies = [
  "byteorder 1.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "grin_util 0.3.0",
  "rand 0.3.22 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rust-crypto 0.2.36 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde 1.0.70 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_derive 1.0.70 (registry+https://github.com/rust-lang/crates.io-index)",
  "serde_json 1.0.24 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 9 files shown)


mcdallas:
@yeastplume I already have done some work on this but it's nowhere near working state, I can push my branch if you want before you start working at this

yeastplume:
@mcdallas Sorry didn't mean to invalidate what you're doing. Andrew's implementation is already there and been tested at least somewhere (and mostly uses the same libraries that we do), so it seems to make sense just to integrate that. All that really needs to be done is to swap out the hash functions, so this is just about there already.

Perhaps I can just finish integrating this bit so we have our working extended key in place and you can pick it up from there?

mcdallas:
@yeastplume No problem at all. I was just trying to save you some work because you seem to be doing the same thing. I have already created two structs ExtendedPrivKey and ExtendedPubKey mostly based on andrew's code and swapped the hmac-sha512 with blake2. Code [here](https://github.com/mcdallas/grin/commit/27853582b3948f3d3adce5a5dfc8f0cc7efbd72a).

> Perhaps I can just finish integrating this bit so we have our working extended key in place and you can pick it up from there?

Whatever suits you is fine with me.

yeastplume:
Thanks! No worries, should be in fairly shortly. Actually, I see what's gone on now, I think @ignopeverell or some early contributor took that code originally and cut it down for a simplified key derivation. Not sure why that might have been done (was before my time).

yeastplume:
@antiochp @ignopeverell So we have the BIP32 module integrated there now, passing unit tests but otherwise not being used. I've factored out the actual hashing calls and some seeds into a trait, in case we wanted to replace any call with Blake2. (We'll certainly want to change a couple of seed values, I think).

This ends up re-introducing rust-crypto, but only for Sha/Ripemd hashing. I think we may be okay to stick with the BIP32 hash algos as much as possible; performance shouldn't be an issue since we should never need to grind through them on restore if we wind the key indices into an output's bullet proof as we've been talking about. Thoughts?

apoelstra:
https://github.com/rust-bitcoin/bitcoin_hashes will eventually be a better choice than rust-crypto (no deps, supports specifically the Bitcoin hashes). But it has open PRs for SHA512 and HMAC so right now it can't be used for BIP32.

ignopeverell:
My take: at this point, anything half decent would be a real improvement.

yeastplume:
bitcoin_hashes looks like a great idea. I'll merge what's there for now (using rust-crypto) so we can move on, but looks like it should just be a drop in replacement once it's ready.

PR #1455, opened 2018-08-31
[DNM] kernel height index

antiochp:
* add a new kernel height index in lmdb (block height per kernel excess)
  * given a kernel we can now quickly find the corresponding block height

If we do go down the path of "relative kernels" as a way to get relative lock heights we need this index to be able to quickly find the block height for a given kernel.

* new index
* update the index when we apply_kernel on an extension
* update the index during fast sync
* backfill the index on `chain::init()` if not already populated (backward compatibility on existing nodes)

Note: An alternative implementation would be to index block hashes by kernel and do it via an additional level of indirection, but this seemed unnecessary.

Edit: The simple height index is insufficient. We also need an index to the kernel by pos (as we do for outputs) as we need to locate the kernel in the MMR to show it exists.
This also implies uniqueness of kernels in the MMR (as we do for outputs).

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -164,7 +164,11 @@ impl Chain {
 		// open the txhashset, creating a new one if necessary
 		let mut txhashset = txhashset::TxHashSet::open(db_root.clone(), store.clone(), None)?;
 
+		// Setup the chain head.
 		setup_head(genesis, store.clone(), &mut txhashset)?;
+
+		// Setup the kernel height index (introduced later, may not exist on all nodes).
+		setup_kernel_height_index(store.clone(), &mut txhashset)?;
 
 		// Now reload the chain head (either existing head or genesis from above)
 		let head = store.head()?;
@@ -599,6 +603,7 @@ impl Chain {
 		txhashset::extending(&mut txhashset, &mut batch, |extension| {
 			extension.rewind(&header)?;
 			extension.rebuild_index()?;
+			extension.rebuild_kernel_height_index(header.clone())?;
 			Ok(())
 		})?;
 
@@ -620,7 +625,9 @@ impl Chain {
 			batch.save_body_head(&head)?;
 			batch.save_header_height(&header)?;
 			batch.build_by_height_index(&header, true)?;
+
 		}
+
 		batch.commit()?;
 
 		debug!(
@@ -879,6 +886,30 @@ impl Chain {
 			.block_exists(&h)
 			.map_err(|e| ErrorKind::StoreErr(e, "chain block exists".to_owned()).into())
 	}
+}
+
+fn setup_kernel_height_index(
+	store: Arc<store::ChainStore>,
+	txhashset: &mut txhashset::TxHashSet,
+) -> Result<(), Error> {
+	let header = store.head_header()?;
+	let kernel = txhashset::extending_readonly(txhashset, |extension| {
+		extension.kernel_for_block_height(1)
+	});
+
+	if let Ok(kernel) = kernel {
+		if let Err(_) = store.get_kernel_height(&kernel) {
+			warn!(LOGGER, "***** backfilling kernel height index from chain init...");
+			let mut batch = store.batch()?;
+			txhashset::extending(txhashset, &mut batch, |extension| {
+				extension.rebuild_kernel_height_index(header)?;
+				Ok(())
+			})?;
+			batch.commit()?;
+			warn!(LOGGER, "***** ...DONE backfilling");
+		}
+	}
+	Ok(())
 }
 
 fn setup_head(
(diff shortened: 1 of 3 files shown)


ignopeverell:
Main question: are we sure we need a full index instead of trying to recover one when it gets referenced? Could kernels that will be referenced later be marked by a feature so only those are indexed?

antiochp:
Actually realizing this "height" index is also insufficient anyway.
We need the following (for at least the kernels in question) - 

  * kernel pos index (we need to be able to find it in the MMR as only the MMR is authoritative)
  * kernel uniqueness in the MME

And yes it would be nice if we could avoid building a full index when some (hopefully) tiny percentage of kernels are actually involved.

Thinking on this some more.

antiochp:
Taking a step back I think the options here are - 

1) every node maintains a full `kernel_pos` index and potentially a related `kernel_height` index
2) we maintain no indices and simply lookup kernels from the MMR as necessary
3) we provide a Merkle proof for each relative kernel

(1) and (2) both make the assumption that all nodes will maintain a full kernel MMR forever
(3) does not make this assumption

*If* we were to go down the Merkle proof path - where would the Merkle proofs live? 
Would they part of a tx? 
Would they be part of a block?
Can they be discarded after accepting a tx or a block? Do we need to keep them?

If we consider the flexibility of a transaction being built with a lock height relative to a kernel that is _not yet_ confirmed and in the kernel MMR then we cannot build the Merkle proof at transaction creation time.

* build tx 1 with kernel A
* build tx 2 with kernel B (and specifiy a relative lock height relative to kernel A of 100 blocks)
* broadcast tx 1
* several blocks are produced
* Tx 1 is included in block at height 105
* Tx 2 can now be accepted at block height 205, but no earlier (we do not know this until block height 105 when kernel A appears in the chain)

So the Merkle proof would appear to be separate to the tx itself.
But we would need to provide the Merkle proof (for kernel A) for tx 2 to be accepted.

Is this even possible?

Would the block (say at height 205 in example above) that include kernel B also need to include the Merkle proof for kernel A?
So would these Merkle proofs be committed to in block 205? Or are they somehow separate?

Would a historical validator validating block 205 need the Merkle proof to validate this block?
(Not necessarily as they would be able to determine the height of kernel A if they had the full kernel MMR at validation time, say during sync.)

So can we do this with Merkle proofs?
And if so where and how do we provide them?

@ignopeverell @tromp

antiochp:
And I guess for completeness and greatest flexibility the "kernel proof" here would actually consist of - 

* A Merkle proof showing the kernel at a given pos in the kernel MMR
* A block header (and previous header?) showing the kernel_mmr_size >= the pos in question
* A proof showing the header is in the most work chain (flychain style)

Which in total would allow us to prove the kernel was included in a block at a particular height on the most work chain.

tromp:
I think a full node needs to keep a full kernel MMR, since these are needed to serve in a fast-sync.
Then they may as well maintain an index on them, taking space on the same order.
Having to sequentially scan through all MMR leaves seems unacceptably inefficient and a DoS vector.
So that would be option (1).

If nodes don't maintain MMR leaves, then syncing can only be done from archival nodes.
I think we're then forced to include Merkle Proofs for any kernel-relative timelocked tx in the block itself, so that all nodes can verify timelock compliance. They could be pruned by any non-archival node.

ignopeverell:
I think 3 brings in too much complexity at this point. 2 would be best and I could see requiring the MMR position to avoid any index or scan. But as @antiochp noted, we can't anchor anything in the MMR when the dependent transaction isn't confirmed yet. So it appears we have no choice but to go with 1...

ignopeverell:
I'm wondering if we could be a little clever about the kernel signature construction. Somehow I remembered this post:

https://lists.launchpad.net/mimblewimble/msg00426.html

There may a good idea here based on repeated hashing (or an offset on the kernel pubkey). I think I need to sleep on this.

tromp:
Igno, how would the hash chain apply in our case where we no longer have switch commitments?

ignopeverell:
Deriving a public key `k*G` from a private can be seen as a hash function. And `n*k*G` could be seen as repeated application of the hash function. We definitely have a spot for those in kernels.

tromp:
n*k*G is not like repeated application of a hash function, since you can just divide n out again...

antiochp:
Even if we do manage to hide the "lock height" in the kernel somehow it still needs to be relative to something (we have absolute lock heights today, height is just in the clear in the kernel). 
So I'm assuming we still need/want to reference prior kernels.

antiochp:
Closing this - replacing with the simpler #1487.
We need to maintain `kernel_pos` index similar to `output_pos` index and we cannot safely jump straight to a `kernel_height` index.

PR #1457, opened 2018-08-31
one day coinbase maturity

tromp:


antiochp:
:+1: on making this less arbitrary.

"24 hours" _is_ still an arbitrary value though - is it worth having a discussion around what this period should be (and the intentions and reasons why)?

If this is purely to minimize the risk of re-orgs (and dependent coins becoming unspendable) then 1 day is presumably still completely arbitrary.
But "1 day" is a nice simple, rememberable (and presumably un-controversial) rule.

I think there was a discussion in bitcoin around making the 100 blocks period significantly longer? To better align long term interests of miners?

I don't want to open a huge can of worms but what if we were to set this to a larger value? 
Would it deter miners? Would it filter out miners who are interested solely in making a quick profit (would this even be desirable?)

Strawman - What if miners could only spend their rewards after a month? 

Feels like this overlaps (a little) with the ASIC and PoW discussion - in terms of aligning interests and motives for getting involved in mining early on.

(Probably not the best place the start a discussion like this but either way - :+1: on changing this from 1,000 blocks to 1,440 blocks for the sake of simplicity).

tromp:
> Strawman - What if miners could only spend their rewards after a month?

Miners will prefer to be able to sell at their discretion, e.g. when they feel the price will likely drop over the next month. So there will be calls to reduce the maturity.

That means a one month maturity needs a good justification. I cannot think of one right now. It would seem that 1440 blocks is long enough that re-orgs of such length are all but impossible. So a one month maturity cannot be justified on the need for better coping with re-orgs.

In economic terms, a longer maturity reduces liquidity, which seems counterproductive for a currency.
Some people might argue that short maturity creates a downward pressure on price, but even if I bought that argument, I don't see that as a bad thing.

antiochp:
Related (but arguing the extreme opposite of what I suggested) - https://github.com/zcash/zcash/issues/953
Zooko arguing for a 6 block coinbase maturity in ZCash to keep it aligned with the "standard confirmation depth" of 6 blocks.

Regarding "standard confirmation depth" our wallet implementation in testnet3 is simply waiting for 1 block. There's a task somewhere (I think) to define what this should actually be and update the default wallet impl to match it.

Apologies for polluting your PR with this, but putting it here for reference.

:+1: on 1,440 blocks (regardless of what I'm posting here).

tromp:
A maturity of 6 seems a little low, but one can wonder why it needs to be on the order of a day.
Perhaps it could be reduced to one hour, i.e. 60 blocks, same as the difficulty retargeting window.
60 deep reorgs should be nearly as impossible as 1440 deep ones.
I have no strong feelings either way. To me both 1440 and 60 are decent non-arbitrary choices.

ignopeverell:
While a 60 block reorg would seem enough normally, there's a security argument against it: miners are in the best position to figure out how to do long forks. So I would be a bit uncomfortable with 60 or anything lower. I'm happy with 1440.

antiochp:
I'm going to merge this.
Barring any other changes this is scheduled to activate at height `100,000`.
We are at approx `80,000` currently.

PR #1458, opened 2018-08-31
Fix amount parsing (from string)

### LINKED ISSUE #1384: Amount sent is 1 nanogrin short
Edit - please check further down, fee was right, but amount sent was 1 nanogrin short of 66.6.

Okay here's one for the grin crew. When I sent 66.6, with a 0.005 fee, it actually didn't use a 0.005 fee, it used a 0.004999(etc) fee, like this:
Sent Tx 4d23408a-b8a8-4d63-82ec-b71a56c55e5c 2018-08-19 12:11:23 true 2018-08-19 16:11:05 4 1 118.395000001 185.0 0.005 -66.604999999

hashmap:
Fixes #1384

I decided to split string into 2 parts (grins and nanogrins) and then parse it separately. Unfortunately for nanogrins I had to implement conversion from the scratch, I hope we don't need to parse billions of strings.

### DIFF
--- a/core/src/core/mod.rs
+++ b/core/src/core/mod.rs
@@ -28,7 +28,6 @@ pub mod verifier_cache;
 use consensus::GRIN_BASE;
 #[allow(dead_code)]
 use rand::{thread_rng, Rng};
-use std::num::ParseFloatError;
 use std::{fmt, iter};
 
 use util::secp::pedersen::Commitment;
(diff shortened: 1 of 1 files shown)


core/src/core/mod.rs
```diff
@@ -198,12 +197,56 @@ impl BitVec {
...
+	if amount == "" {
+		Ok(0)
+	} else {
+		amount
+			.parse::<u64>()
+			.map_err(|_| Error::InvalidAmountString)
+	}
+}
+
+fn parse_ngrins(amount: &str) -> Result<u64, Error> {
```
ignopeverell: Why not just right-pad with zeroes and then parse as u64?

hashmap: Mostly to not allocate a tmp string. I can try to implement both (actually 3 approaches) and write benchmarks

PR #1466, opened 2018-09-03
Fix typos and unused imports

quentinlesceller:
Small PR to remove some unused imports and fix trivial typos.

### DIFF
--- a/chain/src/txhashset.rs
+++ b/chain/src/txhashset.rs
@@ -476,7 +476,8 @@ impl<'a> Extension<'a> {
 		let kernel_pos = self.kernel_pmmr.unpruned_size();
 
 		// Build bitmap of output pos spent (as inputs) by this tx for rewind.
-		let rewind_rm_pos = tx.inputs()
+		let rewind_rm_pos = tx
+			.inputs()
 			.iter()
 			.filter_map(|x| self.get_output_pos(&x.commitment()).ok())
 			.map(|x| x as u32)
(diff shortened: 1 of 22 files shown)


chain/src/txhashset.rs
```diff
@@ -476,8 +476,7 @@ impl<'a> Extension<'a> {
 		let kernel_pos = self.kernel_pmmr.unpruned_size();
 
 		// Build bitmap of output pos spent (as inputs) by this tx for rewind.
-		let rewind_rm_pos = tx
-			.inputs()
+		let rewind_rm_pos = tx.inputs()
```
antiochp: I see PRs flip-flopping on this code style _all the time_ (first chained fn call on same line or next line).
Is this a difference between local rustfmt configs?

I'm on - 
```
rustfmt --version
rustfmt 0.8.2-stable (08da30d7 2018-06-06)
```

And I _think_ my rustfmt puts these on the next line which contradicts this PR.

core/tests/consensus.rs
```diff
@@ -137,7 +139,8 @@ fn get_diff_stats(chain_sim: &Vec<Result<(u64, Difficulty), TargetError>>) -> Di
 
 	// Obtain the median window for the latest time period
 	// i.e. the last  MEDIAN_TIME_WINDOW elements
-	let mut window_latest: Vec<u64> = last_blocks.clone()
```
antiochp: But then here it appears to be the opposite... (was on same line, now on next line).

sesam: :+1: for a whole-project rustfmt commit soon-ish. Then if rustfmt issues continue to pop up, maybe check tab size configs and maybe use .editorconfig in project root

chain/src/txhashset.rs
quentinlesceller: Agree, I think this is something related to https://github.com/rust-lang-nursery/rls-vscode/issues/210. Investigating.

antiochp: :+1:

PR #1473, opened 2018-09-04
Save transaction as part of TxLogEntry after transaction completion

yeastplume:
Saves the entry on transaction completion, will also add a couple of commands to dump and/or resend a stored transaction.

Might optionally delete it once the transaction becomes verified


yeastplume:
Ready now, details on how to repost a transaction are in the `repost` section of doc/wallet/usage.md

PR #1474, opened 2018-09-04
Remove vulnerable smallvec 0.2 from dependencies

hashmap:
Reqwest 0.8.7 removed internal dependency on tokio-proto, which removed unsafe small-vec
dependency. It was a build dependency, but still nice to have `cargo audit` output.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -434,6 +434,14 @@ version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "encoding_rs"
+version = "0.8.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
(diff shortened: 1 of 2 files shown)


hashmap:
```
cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 11 security advisories (from /home/arch/.cargo/advisory-db)
    Scanning Cargo.lock for vulnerabilities (269 crate dependencies)
error: Vulnerable crates found!

ID:	 RUSTSEC-2018-0003
Crate:	 smallvec
Version: 0.2.1
Date:	 2018-07-19
URL:	 https://github.com/servo/rust-smallvec/issues/96
Title:	 Possible double free during unwinding in SmallVec::insert_many
Solution: upgrade to: >= 0.6.3 OR ^0.3.4 OR ^0.4.5 OR ^0.5.1
```

PR #1475, opened 2018-09-04
improve "check known" steps during process_block()

antiochp:
Currently we do a fast in-memory check `check_known()` during `process_block`.

This checks block hashes against the 200 most recent blocks in `block_hash_cache`.

We can do an additional check against the full block index for the case where we see an "old" block that is beyond the scope of `block_hash_cache`.
This will be more expensive than the in-memory check but still considerably faster than full block validation on the duplicate block.
This PR introduces an additional `check_known_store()` to do this.

For even older blocks (on the main chain) there is an additional way we can know if we are processing a "known" block - we can look in the txhashset MMRs.
Rather than rewinding past the block and reapplying it we can rewind _to_ the block and check the MMR roots and sizes.
This PR adds a `check_known_mmr()` that accomplishes this.

We now have 3 steps during `process_block` where we check if the block being processed is already known (with increasing cost of doing the check) -

    * check_known() # fast in-memory check for recent blocks (200 recent blocks)
    * check_known_store() # slower (but still fast) lookup against full block index (2,000 recent blocks on fast sync node)
    * check_known_mmr() # leverage our MMRs for old blocks on most work chain (limited to most work chain, no forks supported)

If a block is not "known" via any of these then we can proceed to validating the block and applying it to the txhashset.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -416,16 +416,9 @@ fn apply_block_to_txhashset(b: &Block, ext: &mut txhashset::Extension) -> Result
 
 	// Check the output and rangeproof MMR sizes here against the header.
 	let (output_mmr_size, _, kernel_mmr_size) = ext.sizes();
-<<<<<<< HEAD
 	if b.header.output_mmr_size != output_mmr_size || b.header.kernel_mmr_size != kernel_mmr_size {
 		return Err(ErrorKind::InvalidMMRSize.into());
 	}
-=======
-	if b.header.output_mmr_size != output_mmr_size ||
-		b.header.kernel_mmr_size != kernel_mmr_size {
-			return Err(ErrorKind::InvalidMMRSize.into());
-		}
->>>>>>> document what we do during pipe::apply_block()
 
 	Ok(())
 }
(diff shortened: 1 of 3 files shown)


antiochp:
Edit: just realized we actually reject blocks older than 50 blocks prior in `validate_block()`...
So maybe these changes are unnecessary.

I'm tempted to pull the "older than 50 blocks ago" check out into a separate, earlier check in process_block() to make it more explicit.

Thoughts @ignopeverell?

ignopeverell:
Making it more explicit and cleaning it up a bit sounds good.

chain/src/pipe.rs
```diff
@@ -236,17 +244,115 @@ fn check_header_known(bh: Hash, ctx: &mut BlockContext) -> Result<(), Error> {
...
+fn check_known_orphans(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), Error> {
+	if ctx.orphans.contains(&header.hash()) {
+		Err(ErrorKind::Unfit("already known in orphans".to_string()).into())
+	} else {
+		Ok(())
+	}
+}
+
+// Check if this block is in the store already.
+fn check_known_store(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), Error> {
```
garyyu: > check_known_store() # slower (but still fast) lookup against full block index (2,000 recent blocks on fast sync node)

Where is this `2000` ?

antiochp: Its the threshold for triggering compaction in `check_compact()` -
https://github.com/mimblewimble/grin/blob/86aea875de1c4c6595fa1e1403048cea8134c71f/servers/src/common/adapters.rs#L545-L547

antiochp:
Merging.

PR #1478, opened 2018-09-05
Hard fork tryout: better PoW encapsulation in block header

ignopeverell:
Introducing a struct to encapsulate all PoW fields. This has the effect to reorder block header fields.


tromp:
Will it also add the scaling factor for secondary PoW?

ignopeverell:
Just add it and keep it unused for now? Wasn't planning on adding seconday PoW validation, factor adjustment, etc

tromp:
Yep; just add "for future use". Makes hard-fork slightly more interesting due to change in header size.

ignopeverell:
Alright then, I'll add that up.

ignopeverell:
Ready for review from whoever wants to.

PR #1482, opened 2018-09-06
Update intro.md

jaredbrubaker:
Correcting what I think may be a typo.


quentinlesceller:
Indeed well spotted. Thanks for the PR.

PR #1484, opened 2018-09-06
Add file based transaction in owner API

quentinlesceller:
Add file transaction endpoint into owner API.
- ```issue_send_tx?method=file``` created. Add a  ```method``` parameters to the send options in the cli and in the ```SendTxArgs```. 
- ```finalize_tx```
- Fix misplaced ```file_receive_tx``` which was in OwnerAPI instead of ForeignAPI.


src/bin/grin.rs
```diff
@@ -213,6 +213,13 @@ fn main() {
 				.long("change_outputs")
 				.default_value("1")
 				.takes_value(true))
+			.arg(Arg::with_name("method")
```
ignopeverell: Not against an explicit argument but don't forget to update the docs.

src/bin/cmd/wallet.rs
```diff
@@ -192,7 +195,7 @@ pub fn wallet_command(wallet_args: &ArgMatches, config: GlobalWalletConfig) {
 					.expect("Failed to parse number of change outputs.");
 				let fluff = send_args.is_present("fluff");
 				let max_outputs = 500;
-				if dest.starts_with("http") {
+				if method == "http" {
```
ignopeverell: Perhaps output an explicit error if the destination provided doesn't look like a URL?

quentinlesceller:
@ignopeverell there is actually already a receive_tx method. https://github.com/mimblewimble/grin/blob/master/wallet/src/libwallet/controller.rs#L420

ignopeverell:
As noted by @gavinmcdermott, `cancel_tx` is also not exposed. It seems it'd be useful, both to file-based and regular transactions

gavinmcdermott:
Hey @quentinlesceller: As @ignopeverell mentioned above, this is definitely a useful call from the perspective of a user-friendly wallet/client. I don't think we can expect regular folks to hop over to the cli in order to free up outputs ;) this endpoint would be quite useful.

quentinlesceller:
@ignopeverell @gavinmcdermott I just added cancel_tx endpoint and also dump_stored_tx to the wallet owner API.

gavinmcdermott:
🎉Thank you @quentinlesceller—sincerely appreciated!

PR #1495, opened 2018-09-10
cache get_shift() and get_leaf_shift() in prune_list

### LINKED ISSUE #1448: validating mmrs in fast sync (weird debug logs)
Saw this when doing a fast sync from scratch in `testnet3`.
This should take a non-zero amount of time and something seems wrong here - 

```
Aug 30 21:05:01.705 DEBG txhashset: validated the output|rproof|kernel mmrs, took 0s
```

antiochp:
Calls to `get_shift()` and `get_leaf_shift()` get progressively slower on a pruned/compacted node as the PMMRs grow in size. 

We need to look back over _all_ previous pruned nodes to calculate the shift for the file offset.

Also caching `is_pruned()` on the prune_list to avoid needing to look back over full paths in the PMMR each time.

We populate these caches on startup when we read the prune_list from disk.
We also rebuild these caches when we call `check_compact()` as part of flushing the prune_list back to disk.

Resolves #1448.

As part of this change we have had to normalize the `prune_list` behaviour.
We were only storing "roots" in there if they were height>=1 (i.e. not leaves).
We are now storing _all_ roots (including those at height 0 (leaves are also roots).
This affected some of the tests in terms of checks if certain pos were included/excluded in the prune_list.

We now see this when validating a txhashset, even on a pruned/compacted node - 

```
Sep 10 14:30:41.545 DEBG txhashset: validated the output 205150, rproof 205150, kernel 188652 mmrs, took 0s
```

This should save us approx 45s on a fast sync.

And we should see some performance improvement when processing both txs and blocks. Basically whenever we call `get_data()` or `get_hash()` against the compact PMMRs.


antiochp:
This is a full chain compaction from a freshly sync'd archival node.
We get the benefit of a fast txhashset validation here as well -  

```
Sep 11 19:57:30.602 DEBG Starting blockchain compaction.
Sep 11 19:57:31.821 DEBG txhashset: validated the output 208648, rproof 208648, kernel 192135 mmrs, took 0s
Sep 11 19:57:49.711 DEBG txhashset: verified 96072 kernel signatures, pmmr size 192135, took 16s
Sep 11 19:57:49.711 DEBG Compaction remove blocks older than 88246.
Sep 11 19:58:15.487 DEBG Compaction removed 88245 blocks, done.
```

antiochp:
Merging this. Tested in various scenarios. 
Updating an existing node and restarting works as expected.
Chain compaction and restart works as expected.

Issue #1492, opened 2018-09-10
0 connected peers for new node


garyyu:
With the latest code, a new build and run with clean `.grin` folder, can't start sync and 0 connected peers.

```



Sep 10 12:43:27.197 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 109.74.202.16:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:43:27.198 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 198.245.50.26:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:43:38.200 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 198.245.50.26:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:43:38.201 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 109.74.202.16:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:43:49.205 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 109.74.202.16:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:43:49.206 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 198.245.50.26:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }



Sep 10 12:45:17.398 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 109.74.202.16:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:45:17.400 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 198.245.50.26:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:45:28.404 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 198.245.50.26:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:45:28.404 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 109.74.202.16:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
Sep 10 12:45:39.405 DEBG connect_peer: on 0.0.0.0:13414. Could not connect to 109.74.202.16:13414: Custom { kind: TimedOut, error: StringError("connection timed out") }
[...]

PR #1496, opened 2018-09-10
add 2 more dns seeds

garyyu:
Related to https://github.com/mimblewimble/grin/issues/1492
To give at least 3 dns seeds:

```
$ nslookup  t3.seed.grin-tech.org

Non-authoritative answer:
Name:	t3.seed.grin-tech.org
Address: 198.245.50.26
Name:	t3.seed.grin-tech.org
Address: 204.48.26.36
Name:	t3.seed.grin-tech.org
Address: 109.74.202.16

$ nslookup seed.grin.lesceller.com

Non-authoritative answer:
Name:    seed.grin.lesceller.com
Address: 198.245.50.26
Name:    seed.grin.lesceller.com
Address: 109.74.202.16
Name:    seed.grin.lesceller.com
Address: 46.4.91.48
Name:    seed.grin.lesceller.com
Address: 192.241.160.172

$ nslookup t3.grin-seed.prokapi.com

Name:    t3.grin-seed.prokapi.com
Address: 95.216.163.175 
Name:    t3.grin-seed.prokapi.com
Address: 195.201.38.140


```

After removing the duplication, actually, the seed nodes are:
1. `198.245.50.26`
2. `109.74.202.16`
3. `204.48.26.36`
4. `192.241.160.172`
5. `46.4.91.48`
6. `95.216.163.175`
7. `195.201.38.140`

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -35,7 +35,9 @@ use util::LOGGER;
 const SEEDS_URL: &'static str = "http://grin-tech.org/seeds.txt";
 // DNS Seeds with contact email associated
 const DNS_SEEDS: &'static [&'static str] = &[
-	"t3.seed.grin-tech.org", // igno.peverell@protonmail.com
+	"t3.seed.grin-tech.org",    // igno.peverell@protonmail.com
+	"seed.grin.lesceller.com",  // q.lesceller@gmail.com
+	"t3.grin-seed.prokapi.com", // info@prokapi.com
 ];
 
 pub fn connect_and_monitor(
@@ -359,6 +361,7 @@ pub fn dns_seeds() -> Box<Fn() -> Vec<SocketAddr> + Send> {
 				),
 			}
 		}
+		debug!(LOGGER, "Retrieved seed addresses: {:?}", addresses);
 		addresses
 	})
 }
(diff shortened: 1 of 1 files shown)


antiochp:
My understanding was we had added a "layer of indirection" and the single DNS seed was actually fronting three (at least three?) actual ip addresses (as you show above).

Do we also want/need to have multiple DNS entries on top of this?

Just want to make sure we're all on the same page with this.

garyyu:
@antiochp  There're two layers of redundancy, one is multiple ip addresses for **one** dns, another one is multiple dns. Both are needed for normal design of a stable high availability service.

garyyu:
possible reasons but not limited for a DNS entry fail to resolve:
1. DNS server is down or by DoS attacking
2. DNS entry is blocked by the firewall
3. Domain expired

Another important reason for multiple DNS entry is:  global areas design for speed, but since Grin are server solution only and not for client for the moment, this one is not so important for now, but could need to be designed in the future if huge clients connections.

ignopeverell:
Right. I guess for a testnet that's going to get killed in a month or less that's not major but for mainnet we definitely want those 2 layers. Some jackass could also try to trigger a domain name transfer.

kargakis:
Why not make this config-driven instead of hardcoding?

kargakis:
You wouldn't need to rerelease in the unlike case of a domain seizure.

kargakis:
Presumable this needs to be configurable anyway to separate between testnet and mainet seeds?

PR #1499, opened 2018-09-10
default to 10 confirms before spendable

antiochp:
TODO - 

- [x] `grin wallet send` now defaults to 10 confirmations during coin selection
- [ ] Does `grin wallet info` also need to be aware of this?
  - [ ] In a flexible way of just hard-coded to the default 10 confirmations?


antiochp:
Closing this for now - focused elsewhere.
If anybody else wants to pick this up go for it!

Issue #1470, opened 2018-09-03
alternative full archival node sync?


antiochp:
I wonder if we can bring a full archival node online using an alternative sync process.

Currently we sync a full archival node by syncing the full set of headers then requesting blocks from height 1 to height n.

Could we - 
* sync full set of headers
* request a (full, un-compacted) txhashset from an existing full archive node
* complete the fast sync, then
* "fill in the gaps" by requesting the older blocks
  * possibly with additional validation to compare against txhashset MMRs?

This would allow a new full archival node to be "usable" sooner.
And we could "fill in the gaps" at our leisure (within reason) without this blocking use of the node.

(I think something similar _may_ have been proposed before, but can't find the issue if there is one).

ignopeverell:
Some considerations:

* Traditionally, a full node means the full chain has been verified block by block. I understand your proposal doesn't change that but it introduces a time where the node is operational without it being the case.
* However, fast sync does not deteriorate the traditional full node security model. It really is just as secure, even though each block hasn't been verified individually. Hence the name archive node instead of full node. Most people do not really get this.

I like your proposal because it normalizes sync around our own "full node" security model. But we'll likely have to explain the above very clearly somewhere to avoid misunderstandings.

garyyu:
I like this.

Parity (for Ethereum) has a similar feature, but if I remember well, it use bittorrent to download archive.

>  request a (full, un-compacted) txhashset from an existing full archive node

That would be too heavy for **an** node in the future, better to request from multiple peers in the same time, each for a split range.

antiochp:
Yeah long term we probably want to consider how to request the txhashset in multiple pieces.
But right now its all or nothing and a coin toss in terms of whether we get a compacted one or a full one back when we request it.

ignopeverell:
This may be controversial, but I've been thinking about removing the archival mode as it is now. Getting all those past blocks is going to get more and more chancy in a network the rightfully defaults on fast sync. It's one more thing we have to maintain, and doing so at the detriment of privacy,

So my vote would be to just have a switch to disable pruning. That's it.

kargakis:
> We want to avoid the situation of there not being any archival nodes in existence (I think)?

naive question: what's the utility of an archival node?

antiochp:
Good question.
I think the main thing an archival node gives us is the confidence that we can recover* from a fork at an arbitrary depth (* for some definition of recover).
[...]

PR #1500, opened 2018-09-10
Prevent duplicate output commitments

### LINKED ISSUE #1497: duplicate commitments or duplicate outputs?
We currently check for "duplicate outputs" via - 

https://github.com/mimblewimble/grin/blob/da87f5fef4a3830c7c0f3dc8c7ec9d3414e63f44/chain/src/txhashset.rs#L661-L675

* look in the `output_pos` index for the commitment
* if we find an entry, lookup the hash at that pos
* compare the hash with the hash of the output

But I think this is not an exhaustive check and I _think_ maybe there is an issue here.

Note that we use the __commitment__ to lookup the pos in the MMR.
But we compare the hash of the __output__.

It is possible to have two different _outputs_ (with different hashes), both with the same _commitment_, specifically one could be a coinbase output and one a regular output (differing only in features).

So if we generated a (60 grin) coinbase output and reused the commitment for a (again 60 grin) regular output then we would have two different outputs, but they would share a non-unique commitment, even though they would hash to different values.

@ignopeverell - my understanding was that we wanted to disallow duplicate commitments in the output MMR. 
Is this correct? 
Or is it that we want to disallow duplicate outputs (which we do currently)?

antiochp:
Resolves #1497.

We were checking for duplicate _outputs_ (based on the hash in the MMR).
But we actually want to be checking for duplicate output _commitments_ directly.

TODO - 

- [x] Can we use a full archive node to check for any violations of this? (unlikely but will fork the chain if a duplicate is present).

### DIFF
--- a/chain/src/txhashset.rs
+++ b/chain/src/txhashset.rs
@@ -573,7 +573,9 @@ impl<'a> Extension<'a> {
 
 			// Find the "cutoff" pos in the output MMR based on the
 			// header from 1,000 blocks ago.
-			let cutoff_height = height.checked_sub(global::coinbase_maturity(height)).unwrap_or(0);
+			let cutoff_height = height
+				.checked_sub(global::coinbase_maturity(height))
+				.unwrap_or(0);
 			let cutoff_header = self.commit_index.get_header_by_height(cutoff_height)?;
 			let cutoff_pos = cutoff_header.output_mmr_size;
 
--- a/chain/src/txhashset.rs
+++ b/chain/src/txhashset.rs
@@ -659,16 +659,8 @@ impl<'a> Extension<'a> {
 		let commit = out.commitment();
 
 		if let Ok(pos) = self.batch.get_output_pos(&commit) {
-			// we need to check whether the commitment is in the current MMR view
-			// as well as the index doesn't support rewind and is non-authoritative
-			// (non-historical node will have a much smaller one)
-			// note that this doesn't show the commitment *never* existed, just
-			// that this is not an existing unspent commitment right now
-			if let Some(hash) = self.output_pmmr.get_hash(pos) {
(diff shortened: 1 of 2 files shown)


chain/src/txhashset.rs
```diff
@@ -659,16 +661,8 @@ impl<'a> Extension<'a> {
...
-			// as well as the index doesn't support rewind and is non-authoritative
-			// (non-historical node will have a much smaller one)
-			// note that this doesn't show the commitment *never* existed, just
-			// that this is not an existing unspent commitment right now
-			if let Some(hash) = self.output_pmmr.get_hash(pos) {
-				// Check the hash matches what we expect.
-				// We may be on a fork which may result in the entry at that pos being
-				// different to the one we expect.
-				if hash == OutputIdentifier::from_output(out).hash_with_index(pos - 1) {
+			if let Some(out_mmr) = self.output_pmmr.get_data(pos) {
```
ignopeverell: As you mention this will hardfork archive nodes because `get_data` bypasses the spent check. We should only check UTXOs.

antiochp: We are only checking UTXOs here.

The variants are `get_data()` and `get_data_from_file()`. 
`get_data()` _is_ "spent aware", the other goes straight to the file.

We should think about some better naming for these two...

antiochp:
Kicked off a "full archival sync" to see if we successfully sync with this.
We should see a duplicate unspent commitment as a failure to sync fully.

Related - #1470 (noticing the early blocks are _slow_ to sync as not many nodes have them on the network).

chain/src/txhashset.rs
ignopeverell: Oh right, sorry for the confusion!

PR #1501, opened 2018-09-10
[WIP, DNM] BIP32 Wallet Compliance

yeastplume:
(Technically not consensus-breaking but will give everyone a bad day if this is merged. Intended for T4)

Modify wallet key derivation to be BIP-32 Compliant, involves: 

* Key derivation paths will have to become the identifier internally
* For the moment, just support a single derivation off of the master, m/0/1, m/0/2, m/0/3 etc, so the user functionality stays exactly as it is now
* Wallet can later be extended to support multiple accounts (at m/1, m/2 etc) 
* Derivation paths will be stored in the bullet proof (via https://github.com/mimblewimble/secp256k1-zkp/commit/e6f1191da78b031c5c7bd00527756b74adecf0e3), supporting up to 4 levels deep for now (m/0/0/0/0)
* Wallet restore can use these paths to avoid having to do any more key grinding (added bonus)

Please feel free to review and comment even as this is in progress, will be messing around with this a bit to get the path structure and KeyChain trait interface feeling right.

PR #1502, opened 2018-09-10
BIP39 Support

mcdallas:
Not sure if we need it now, but here it is.


yeastplume:
Looks very good, thanks for putting this together. Wallet merges are a bit messy at the moment due to upcoming T4, so going to wait a little while to merge this (once I'm ready to merge it on top of https://github.com/mimblewimble/grin/pull/1501)

Hopefully once T4 is up we can incorporate this into the wallet workflow user-side.

yeastplume:
Would you mind rebasing on the current (I think it should just be a matter of the Cargo lock files). I'm kind of of two minds as to whether this should go into the keychain lib or whether it belongs a bit higher up in the wallet (happy to take opinions here).

mcdallas:
Rebased and replaced unmaintained rust-crypto with individual crypto crates (as per #1573)

mcdallas:
@yeastplume maybe it makes more sense to be in the wallet crate but I put it in keychain to contain all the crypto dependencies in there.

ignopeverell:
No strong opinion either way but it does seem a little more keychainy.

yeastplume:
Yeah, of two minds as well. Keeping the deps contained is a good enough argument to keep it keychainy for now, so let's go with that. @mcdallas thanks for putting this together

mably:
Is usage documented somewhere?

yeastplume:
This is just library support, it's not actually integrated into the wallet yet.

Issue #1464, opened 2018-09-02
Healthy peers ratio very low


garyyu:
In the beginning, I have 106/172 healthy peers:
```
Sep 02 11:55:55.068 DEBG monitor_peers: 0 connected (0 most_work). all 0 = 0 healthy + 0 banned + 0 defunct
Sep 02 11:55:56.069 DEBG monitor_peers: 2 connected (1 most_work). all 115 = 115 healthy + 0 banned + 0 defunct
Sep 02 11:56:29.974 DEBG monitor_peers: 0 connected (0 most_work). all 115 = 115 healthy + 0 banned + 0 defunct
Sep 02 11:56:30.980 DEBG monitor_peers: 2 connected (1 most_work). all 115 = 115 healthy + 0 banned + 0 defunct
Sep 02 11:56:31.988 DEBG monitor_peers: 3 connected (2 most_work). all 115 = 110 healthy + 0 banned + 5 defunct
Sep 02 11:56:32.993 DEBG monitor_peers: 6 connected (3 most_work). all 172 = 163 healthy + 0 banned + 9 defunct
Sep 02 11:56:53.038 DEBG monitor_peers: 9 connected (6 most_work). all 172 = 121 healthy + 0 banned + 51 defunct
Sep 02 11:57:13.095 DEBG monitor_peers: 10 connected (7 most_work). all 172 = 106 healthy + 0 banned + 66 defunct

```

For somehow reason, this ratio was decreasing from time to time:
```
Sep 02 22:10:03.031 DEBG monitor_peers: 0 connected (0 most_work). all 214 = 109 healthy + 0 banned + 105 defunct
Sep 02 22:10:23.101 DEBG monitor_peers: 17 connected (10 most_work). all 214 = 109 healthy + 0 banned + 105 defunct
Sep 02 22:10:43.164 DEBG monitor_peers: 17 connected (10 most_work). all 214 = 88 healthy + 0 banned + 126 defunct
...
Sep 02 22:11:54.044 DEBG monitor_peers: 0 connected (0 most_work). all 214 = 88 healthy + 0 banned + 126 defunct
Sep 02 22:12:14.101 DEBG monitor_peers: 17 connected (10 most_work). all 214 = 88 healthy + 0 banned + 126 defunct
Sep 02 22:12:34.154 DEBG monitor_peers: 17 connected (10 most_work). all 214 = 67 healthy + 0 banned + 147 defunct
...
[...]

PR #1503, opened 2018-09-11
return an Err when read return 0

garyyu:
Fix:  Return an Err when read return 0, instead of return an all zeroed buff without any error.

This PR is partially related to https://github.com/mimblewimble/grin/issues/1464 , but in #1464, there's  also another existing issue to be tracked there. 

It's clear now why we often see such kind of Error:
```
Serialization(UnexpectedData { expected: [30], received: [0] })
```
(BTW, `30` means the 1st byte of the magic word `0x1e, 0xc5` which we're using for the Grin message.)

It's because of one case of `read_exact()`:
```
		match conn.read(buf) {
			Ok(0) => break,       <<<< problem is here!
			Ok(n) => {
				...
			}
```

If we `break` at the `Ok(0)`, that means we will successfully return the `read_exact()` function call with the read buffer but all zeroed.

But actually, it's indeed an Error case, could be caused by some reasons, such as `connection reset`, `connection close`, etc.   Unfortunately, in these cases, we don't get the following case matched:
```
			Err(e) => return Err(e),
```
then we don't get the official `io::ErrorKind::ConnectionAborted` or `io::ErrorKind::NotConnected`.

I'm not sure which one is the real matched `io::ErrorKind`.  But it doesn't matter here if I use `ConnectionAborted` for `Ok(0)`.

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -118,10 +118,19 @@ pub fn read_exact(
 	let sleep_time = time::Duration::from_micros(10);
 	let mut count = time::Duration::new(0, 0);
 
+	let exact = buf.len();
 	let mut read = 0;
 	loop {
 		match conn.read(buf) {
-			Ok(0) => break,
+			Ok(0) => {
+				if read < exact {
+					return Err(io::Error::new(
+						io::ErrorKind::ConnectionAborted,
+						"read_exact",
+					));
+				}
+				break;
+			}
 			Ok(n) => {
 				let tmp = buf;
 				buf = &mut tmp[n..];
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -118,18 +118,14 @@ pub fn read_exact(
 	let sleep_time = time::Duration::from_micros(10);
 	let mut count = time::Duration::new(0, 0);
 
-	let exact = buf.len();
 	let mut read = 0;
 	loop {
 		match conn.read(buf) {
 			Ok(0) => {
-				if read < exact {
-					return Err(io::Error::new(
-						io::ErrorKind::ConnectionAborted,
-						"read_exact",
-					));
-				}
-				break;
+				return Err(io::Error::new(
+					io::ErrorKind::ConnectionAborted,
+					"read_exact",
+				));
 			}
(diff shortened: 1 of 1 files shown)


antiochp:
Nice catch!

Looks good to me - but I want to wait for @ignopeverell or @hashmap to chime in here as they know more about the networking stack and expected behavior around this.

p2p/src/msg.rs
```diff
@@ -118,10 +118,19 @@ pub fn read_exact(
 	let sleep_time = time::Duration::from_micros(10);
 	let mut count = time::Duration::new(0, 0);
 
+	let exact = buf.len();
 	let mut read = 0;
 	loop {
 		match conn.read(buf) {
-			Ok(0) => break,
+			Ok(0) => {
+				if read < exact {
```
hashmap: Nice catch! What do you think about always return the error here? I don't see how we can get here with full (well, empty) buffer. We read and check the buffer below, if we are done we break https://github.com/mimblewimble/grin/blob/5a1d69744e19dafe023b58d1f8a15d3f37e1929b/p2p/src/msg.rs#L151. So we read again only if we expect to read at least 1 byte.

garyyu: I agree it looks like `if read < exact` will be always `true` and `break` will never be reached.

But that `break` make us 100% safe from an infinite loop, so, if it doesn't hurt, want to keep it there:)

btw, that comes from a habit to avoid static check tool complain, because normally I use static analysis tool for big projects. 

we could also need use static analysis tool for Grin.

hashmap: No, I mean we should unconditionally return error, so we exit loop here anyway, just now we have 2 branches which is harder to read and to reason about. 
reg static analysis - we use Rust already:) on a serious note I use `clippy` from time to time, but not sure it should be a part of CI.

garyyu: Sorry, don't understand. Do you mean just put a `return Err(io::Error::new(...))` there and remove `if read < exact {` and `break`?

hashmap: Right, I think we both think that if we read `0` here  it's an error for us

PR #1505, opened 2018-09-11
fix: clean_peers is using connected_peers, which make it do nothing

garyyu:
This is the partial fix for https://github.com/mimblewimble/grin/issues/1464.

1. First fix 

We will call `peers.clean_peers()` in `monitor_peers()` function, periodically in 20s. Currently this `clean_peers()` function is using the `connected_peers()` to check the banned peer and the disconnected peer, and remove them from the peers list.

This definitely is wrong, of course the connected peers list will not contain any banned peer or any disconnected peer. Then, function `peers.clean_peers()` actually do thing.

I replace it with the peers list, to fix this bug.

2. Second fix  (I forgot to split it into another PR,  but since it's highly related to clean_peers(), should be OK to put into one PR).

Currently, the banned state of peer.rs is not used, we don't actually ban it and just set a flag there. It's different from the `ban_peer()` in the peers.rs, which will save the banned state to the database and stop the peer.

BTW, we have 2 `State` enum definition, one is in `peer.rs`, another one is in `p2p/src/store.rs`. Both `State` have a `Banned` state, and it's not easy to reconcile them into a single enum, at least for me:)  perhaps a better developer can do this refactor in the future.

To handle the `Banned` state from the `peer.rs` which come from `Error::Serialization` error locally, I have to add a new state: `ToBeBanned`, and let peers module have chance to handle the real ban in the `peers.clean_peers()`  function.


antiochp:
I'm not sure that's entirely true (but my understanding of this may be incorrect, I've just always assumed this is what it was intended to do...)

I _think_ we mark connected peers as banned - and then on the periodic `clean_peers()` call this is where we actually remove them and close the connections.

So connected_peers _can_ (and will) contain banned peers prior to them actually getting removed (at least that's how I've assumed it works).

@ignopeverell may know more about the intended behavior here.

garyyu:
My description of 2nd fix is indeed not very clear, better to read the code change than reading my description:)  The point is in the new added `ToBeBanned` state.

antiochp:
Looking at it from a different perspective - if there are no banned peers in the connected peers list then there is nothing for `clean_peers()` to do. 
We would not need to iterate over all the known peers and try and clean the banned ones up from the connected peers list.

garyyu:
> if there are no banned peers in the connected peers list then there is nothing for clean_peers() to do.

Look like you didn't realize the concept of `connected peers list` = `peers with state Connected`, here we should use all peers in the `peers` vector (but of course not the all known peers from the database), and it's a mistake using `peers with state Connected` for `clean_peers()`.

garyyu:
To give an overview on the old code regarding to the `clean_peers()`:
```
	pub fn clean_peers(&self, max_count: usize) {
		...
		// build a list of peers to be cleaned up
		for peer in self.connected_peers() {     <<<<  problem here!  
			let peer_inner = peer.read().unwrap();
			if peer_inner.is_banned() {          <<<< never true
				rm.push(peer.clone());
			} else if !peer_inner.is_connected() {       <<<< never false
				rm.push(peer.clone());
			}
		}
		...
```

and the following is what the meaning of `connected_peers()`:
```
	pub fn connected_peers(&self) -> Vec<Arc<RwLock<Peer>>> {
		let mut res = self
			.peers
			.read()
			.unwrap()
			.values()
			.filter(|p| p.read().unwrap().is_connected())    <<<< look the detail of is_connected()
			.cloned()
			.collect::<Vec<_>>();
		thread_rng().shuffle(&mut res);     <<<< btw, I don't understand why we need shuffle here!
		res
	}
```

The detail of `is_connected()`:
```
	pub fn is_connected(&self) -> bool {
		if !self.check_connection() {
			return false;
		}
		let state = self.state.read().unwrap();
		*state == State::Connected                    <<<<  here it is!
	}
```

Now should be clear regarding the reason of my 1st fix.

garyyu:
Any more comments?

The core question for 2nd fix is:  

Do we need ban the peer if having serialization error for messages Hand/Shake, Ping/Pong,  GetPeerAddrs, PeerAddrs, and so on?  just like the banning for errors in messages: Block, Header, Headers, CompactBlock, txhashset.

antiochp:
Let me take another look over this - I see what you are talking about with 1st fix.

Regarding your question on the 2nd fix - no, we only want to ban peers for "bad" behavior.
Sending an invalid block would be an example of bad behavior.
But a serialization error on a handshake message would not be enough to cause a ban.

garyyu:
@antiochp Thanks for look again.

About the 2nd fix, the old code is taking it as `Banned` (but no processing in store level, and without connection closing):

```
	fn check_connection(&self) -> bool {
		match self.connection.as_ref().unwrap().error_channel.try_recv() {
			Ok(Error::Serialization(e)) => {
				let mut state = self.state.write().unwrap();
				*state = State::Banned;    <<<<  I'm Here 
				info!(
					LOGGER,
					"Client {} corrupted, ban ({:?}).", self.info.addr, e
				);
				false
			}
			Ok(e) => {
				let mut state = self.state.write().unwrap();
				*state = State::Disconnected;
				debug!(LOGGER, "Client {} connection lost: {:?}", self.info.addr, e);
				false
			}
			Err(_) => true,
		}
	}
```

According to your comment:
> a serialization error on a handshake message would not be enough to cause a ban.

Then, what should we do in this case?  return `true`?

garyyu:
@antiochp or @ignopeverell , I need your help regarding the above question:
> Then, what should we do in this case? return true?

garyyu:
Updated according to the discussion.

Disconnect the peer connection when `Error::Serialization`, instead of banning.

p2p/src/peers.rs
```diff
@@ -446,13 +446,19 @@ impl Peers {
 		let mut rm = vec![];
 
 		// build a list of peers to be cleaned up
-		for peer in self.connected_peers() {
+		for peer in self.peers.read().unwrap().values() {
```
antiochp: Yeah I see it now - `connected_peers()` is not the set of peers we are connected to, but the set of peers we are connected to that are in state `Connected`...

antiochp: :+1:

PR #1509, opened 2018-09-11
Simplify mining header serialization

ignopeverell:
Avoids adding the nonce just to strip it right after. Removes the specialized writer.


yeastplume:
:+1:

PR #1513, opened 2018-09-12
a bit of refactoring on seed

garyyu:
- a bit of refactoring on seed to have simpler code.
- a fix on `connect_and_monitor` condition:

related old code for the fix:
```
pub fn connect_and_monitor(...) {
		...
		if (peers.peer_count() < p2p_server.config.peer_min_preferred_count()
			&& current_time - prev > Duration::seconds(start_wait_base.pow(start_attempt)))
			|| current_time - prev > Duration::seconds(20)
		{
```

related new code for the fix:
```
		if Utc::now() - prev > Duration::seconds(cmp::min(20, 1 << start_attempt)) {
```

I removed the condition of `peers.peer_count()` checking, because we should keep this 20s periodic calling to `monitor_peers()`, even if the connected peers number reach `peer_min_preferred_count` (default is 8).

And we call `peers.clean_peers()` in function `monitor_peers()`, otherwise, we will never clean those `Disconnected` peers in this case.

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -253,10 +253,8 @@ fn listen_for_addrs(
 	rx: &mpsc::Receiver<SocketAddr>,
 ) {
 	if peers.peer_count() >= p2p.config.peer_max_count() {
-		let mut iter = rx.try_iter();
-		let _ = (0..50)
-			.filter(|&_| !iter.next().is_none())
-			.collect::<Vec<i32>>();
+		// clean the rx messages to avoid accumulating
+		for _ in rx.try_iter() {}
 		return;
 	}
 
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -17,16 +17,11 @@
 //!
 
 use chrono::prelude::Utc;
-use chrono::Duration;
-use std::io;
+use chrono::{Duration, MIN_DATE};
 use std::net::{SocketAddr, ToSocketAddrs};
-use std::str;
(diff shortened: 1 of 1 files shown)


garyyu:
Could somebody take the code review on this and merge it?  

This refactoring include an improvement (or say fix) which can ensure `peers.clean_peers()` can be called periodically, without skip.

servers/src/grin/seed.rs
```diff
@@ -268,73 +252,49 @@ fn listen_for_addrs(
 	capab: p2p::Capabilities,
 	rx: &mpsc::Receiver<SocketAddr>,
 ) {
-	let pc = peers.peer_count();
+	if peers.peer_count() >= p2p.config.peer_max_count() {
```
ignopeverell: Not sure I understand what this is doing. It filters 50 things from the iterator but discards the result? Should this just be removed?

garyyu: Thanks for review~

In this early return, we need make sure **not** making the `rx: &mpsc::Receiver` accumulate too much messages (`SocketAddr` to connect).

But unfortunately `mpsc` don't have a `clear()` function, we have to call `rx.try_iter()`.

And again, unfortunately `mpsc` don't have a `len()` function, we don't know the exact number of messages, that's why I just put a `0..50` iterator there.

Perhaps replace this with a empty `for` loop?
```
for addr in rx.try_iter() {}
```

Any suggestion solution?

ignopeverell: Yes, that makes more sense and the comments helps as well. Thanks!

PR #1516, opened 2018-09-12
fix the comments of test bitmap

garyyu:
avoid to be confused by the comments.


antiochp:
These comments were referring to what we _used_ to do - we stored them as `u64`, we never stored these as `u32` values.
But I guess the comparisons are more valid in terms of reduced storage needs if we use the assumption that we could do this with vecs of u32 values.

PR #1521, opened 2018-09-13
fix: infinite loop in peak_map_height for max

garyyu:
1. fix: When input `pos` is `u64::MAX`, function `peak_map_height` will enter infinite loop.
2. a bit of optimization, saving 35% time for this function. Here is the bench test result:
```
$ cd core
$ cargo test --release bench_peak_map  -- --nocapture

running 1 test
  1000000 peak_map_height () in    28.182ms
  1000000 peak_map_height'() in    44.700ms
 10000000 peak_map_height () in   341.778ms
 10000000 peak_map_height'() in   521.723ms
100000000 peak_map_height () in  3787.511ms
100000000 peak_map_height'() in  6007.633ms
```
**Note**: `peak_map_height'` means the old function before this optimization.

### DIFF
--- a/core/src/core/pmmr.rs
+++ b/core/src/core/pmmr.rs
@@ -37,6 +37,7 @@
 //! either be a simple Vec or a database.
 
 use std::marker;
+use std::u64;
 
 use croaring::Bitmap;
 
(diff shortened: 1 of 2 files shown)


ignopeverell:
When pos reaches `u64::MAX` I think we'll be running into larger issues but your change makes the code faster and cleaner, :+1:

Issue #1508, opened 2018-09-11
transaction propagation (can we broadcast "kernel first")?


antiochp:
Saw this earlier today and it got me thinking a bit about how inefficient our tx propagation is right now.
https://bitcointalk.org/index.php?topic=1377345.0 
"Blocksonly mode BW savings, the limits of efficient block xfer, and better relay"

Grin transaction propagation currently consists of -
  * broadcast the full transaction to 8 (I think) of our connected peers,
  * unless we believe that peer has seen it before (we track recent tx ids per peer connection).

This contrasts with block propagation where we -
  * broadcast "header first"
  * then request "compact blocks"
  * then fallback to "full block" if necessary

----

What if we broadcast a list of tx kernel excesses? 
This would be roughly analogous to the `INV` msg in Bitcoin.
Instead of saying "here's a new tx" we say "here's the id of a new tx".
But instead of tx ids we use our tx kernel excesses to identify txs.

Say we had a tx -
`[In1], [Out1, Out2], [Kern1]`

We send `[Kern1]` out to all our peers.
Any peer not having seen a Tx with matching kernel set will then request the full Tx matching `[Kern1]`.

For a multi-kernel transaction we would do something similar - 

Say we have a multi-kernel tx - 
`[In1, In2], [Out1, Out2, Out3], [Kern2, Kern3]`

We send `[Kern2, Kern3]` to all our peers.
Each peer then looks in their txpool for a Tx matching `[Kern2, Kern3]`.
If not yet seen then request the full Tx matching `[Kern2, Kern3]`.

But what if they have seen `Kern2` or `Kern3` before, just not the full kernel set `[Kern2, Kern3]`?

Maybe they have seen `[Kern2]` before, in which case they can ask for `[Kern3]` knowing this is a valid tx. If they successfully receive it then they can then process it and broadcast out `[Kern3]` to let _their_ peers know about it.

But maybe they have seen `[Kern3, Kern4]` before, which conflicts with `[Kern2, Kern3]` as the aggregation is different.
In this case they can request `[Kern2]` which is the (de-aggregated) part they have not yet seen.
If they receive a tx matching `[Kern2]` then they can process it and broadcast out `[Kern2]`.
If not then they just ignore `[Kern2, Kern3]` as there is nothing they can do with this.

i.e. We end up doing something similar to our existing de-aggregation step in the txpool to reduce 
the impact of malicious transaction aggregators (we only allow aggregation during Dandelion stem).

----
[...]

PR #1522, opened 2018-09-13
[WIP] broadcast CompactTransactions with short_ids (similar to CompactBlocks)

antiochp:
Still very much a work in progress.

Takes our `CompactBlock` impl and does something similar (but simpler) for "kernel first" transaction propagation.

* Introduce a new `CompactTransaction` (to hold a vec of `short_ids` representing tx_kernels)
* broadcast out a `CompactTransaction` msg when we process a new transaction
* if we receive a `CompactTransaction` look in our txpool to see if we have it all (TODO)
  * if not already seen, ask the peer for the full tx via a `GetTransaction` msg (passing those same `short_ids`)
* if we receive a `GetTransaction` go look for the tx in our txpool
  * if we have it send it back as a full `Transaction` msg

TODO - 
  - [ ] Need to identify on receiving side if we've already seen the tx represented by the compact_tx (we have the tx hash, the tracking adapter, and we can look in our txpool)
  - [ ] Still need to think about what we do with aggregation/deaggregation here.

Note: I think this opens up some interesting possibilities for handling aggregation safely.
When we broadcast a `compact_tx` neither party (sender or receiver) cares if this is a single kernel tx or a (potentially aggregated) multi-kernel tx.

We may have seen none, all or some subset of it before.
So there is some flexibility around exactly we go ask for if we don't have all of it.

And when asking for a tx via `GetTransaction` the same holds - we don't care if we are asking for a single tx or an aggregated tx. And there is flexibility there again.
Need to think some more about this.

----
A `CompactTransaction` is hash (32) + nonce (8) + vec of short_ids (6 bytes each).
So a `CompactTransaction` for a single kernel tx is 46 bytes.
This is larger than if we simply sent over the tx hash (32 bytes) or the tx kernel excess (33 bytes).

But a `CompactTransaction` with multiple kernels adds only 6 bytes to this for each additional kernel.
2 kernels -> 52 byte `CompactTransaction`.

It might end up making more sense to just pass kernel excesses around but thought it was worth exploring using `short_ids` for this.
Having identifiers in there for all the underlying kernels makes this potentially pretty flexible in terms of aggregation.


antiochp:
Related - #1508.

antiochp:
This PR is growing too big. going to step back and try something smaller initially.

PR #1523, opened 2018-09-13
Example code of conduct

cr-walker:
Example simpler code of conduct.


ignopeverell:
Thanks a lot for taking the time to do this, it's really appreciated. I'll point just point out what, for me, is the value of a more exhaustive CoC like the Rust one (which the currrent CoC is mostly derived from):

1. A clearer code, so community participants have a better idea of what's expected, what's grey area, and what would be bad. With a much shorter version, it's all in the ends of moderators basically, and while now we all have the best intentions, it's good to future-proof it a bit more.
2. More protection against trolls. The number 1 attack you'll get when trying to block someone who's being abusive is around your legitimacy in doing so. It's a lot more effective to say "you obviously violated rule such and such in the code". This helps moderators greatly in doing their job. It also limits greatly how far down the shit hole a troll attack can go.
3. The above point is easy to ignore. But constant troll interactions causes great fatigue (it's a war of attrition). Moderators typically have 5 or 10 other jobs. A community is a fragile thing. They need as much help as possible.
4. The goal of a CoC of conduct is helping the community in limiting the sad but inevitable abuse. In my kinda long career with open source, the only people who question this are almost without exception, the most egotistic, tone deaf, empathy lacking, abusive ones. So why help them with a short CoC?

quentinlesceller:
Closing this PR since no further comments were made.

PR #1526, opened 2018-09-14
Small cleanup and perf improvement for `peak_sizes_height` and `peaks`

garyyu:
Find 2 more similar cases as the small cleanup and perf improvement for `peak_map_height()`.


antiochp:
These are some nice tricks with `u64::MAX`.
:+1:

garyyu:
😃   a const definition could be enough. let me see if it's feasible in this PR.

garyyu:
const `ALL_ONES` defined for `u64::MAX`.

Issue #1477, opened 2018-09-05
Turtle speed API - /v1/blocks/{hash}


garyyu:
A terrible slow API of:
```
http://127.0.0.1:13413/v1/blocks/{hash}
```

I spent **9 hours** to import the whole block data (from height 80,399) into a local database.  There must be a performance evil there to make this terrible slow.  will check it later.

```
Sep 04 22:29:02.284 DEBG txhashset: merkle_proof: output: Commitment(08fb8902d00265c6fc7911e600b5ac03e4b8ece9dd38b2eb1ecc773234d2a80f01), block: 1d2902f6
Sep 04 22:29:02.285 DEBG merkle_proof  186771, last_pos 186771
Sep 04 22:29:02.404 DEBG txhashset: merkle_proof: output: Commitment(0842ad33e2b85fea96e340a4fcee45c7e39547a754e4c8171213d767f0f525cb18), block: 07ce5af6
Sep 04 22:29:02.405 DEBG merkle_proof  186769, last_pos 186770

...
Sep 05 07:47:25.235 DEBG txhashset: merkle_proof: output: Commitment(099c1a79f74f33499dbb3e12e55803b3c76dea00cf9c0047d4877aa922b4fb3f28), block: 036974b8
Sep 05 07:47:26.583 DEBG merkle_proof  17, last_pos 18
Sep 05 07:47:26.635 DEBG txhashset: merkle_proof: output: Commitment(098bae744202a19d270ac99dfe497fd808b084efdf299c6d055c739147e002261a), block: 0386b7d5
Sep 05 07:47:27.994 DEBG merkle_proof  11, last_pos 11

```

looks like it's very fast at the beginning, and more and more slower when height decreasing.

sesam:
Possibly we have throttling / rate limiting on the API endpoints. Sorry I don't
remember where / how / if it was implemented. Maybe see beginning of
handlers.rs or there around.

quentinlesceller:
Agree also observed than when I updated my explorer. I don't remember of any kind of rate limiting in the API though.

garyyu:
The 'evil' function could be `rewind()` function in txhashset.rs, will take this chance to dig into txhashset and pmmr, for better understanding of this code.

garyyu:
#1528 should have great improvement for this issue. The original `9 hours` could be `1 hour` now already, but not tested.

[...]

PR #1528, opened 2018-09-14
optimization for txhashset rewind

garyyu:
Optimization for `rewind()` function in `chain/src/txhashset.rs`. It's about as 700% faster as the old version, for this function call, in case the rewind block is very far from the Tip.

Compare between old version and new version:

- Test with block height 40,000:
```
# curl -0 -XGET  http://127.0.0.1:13413/v1/blocks/40000

Old Version:
Sep 14 20:09:05.106 DEBG txhashset: merkle_proof: output: Commitment(09d84a9c6b5939fbd12739421e6c99d26fc313f269046353d8eecf7879cbef1d56), block: 007453cb
Sep 14 20:09:06.070 DEBG merkle_proof  99312, last_pos 99312
                    <<<< 964ms spent

New Version:
Sep 14 20:31:54.413 DEBG txhashset: merkle_proof: output: Commitment(09d84a9c6b5939fbd12739421e6c99d26fc313f269046353d8eecf7879cbef1d56), block: 007453cb
Sep 14 20:31:54.548 DEBG merkle_proof  99312, last_pos 99312
                    <<<< 135ms spent
```
- Test with block height 50,000:
```
# curl -0 -XGET  http://127.0.0.1:13413/v1/blocks/50000

Old Version:
Sep 14 20:05:18.304 DEBG txhashset: merkle_proof: output: Commitment(09994dd2a6dd7cd7fbf9104e10c304cd8eaf16f74ae5d405e2b26c7e0e3b85c816), block: 02e79e8f
Sep 14 20:05:19.084 DEBG merkle_proof  119420, last_pos 119421
                    <<<<  780ms spent

New Version:
Sep 14 20:30:14.842 DEBG txhashset: merkle_proof: output: Commitment(09994dd2a6dd7cd7fbf9104e10c304cd8eaf16f74ae5d405e2b26c7e0e3b85c816), block: 02e79e8f
Sep 14 20:30:14.948 DEBG merkle_proof  119420, last_pos 119421
                    <<<< 106ms spent
```
- Test with block height 80,000:
```
# curl -0 -XGET  http://127.0.0.1:13413/v1/blocks/80000

Old Version:
Sep 14 20:04:21.345 DEBG txhashset: merkle_proof: output: Commitment(093941106eb2c3392e0f839dbd2d19926f71ae81d2368492f8b6b4b5854bb9bd79), block: 1316d80e
Sep 14 20:04:21.612 DEBG merkle_proof  185971, last_pos 185977
                    <<<< 267ms spent

New Version:
Sep 14 20:28:56.290 DEBG txhashset: merkle_proof: output: Commitment(093941106eb2c3392e0f839dbd2d19926f71ae81d2368492f8b6b4b5854bb9bd79), block: 1316d80e
Sep 14 20:28:56.325 DEBG merkle_proof  185971, last_pos 185977
                    <<<< 35ms spent
```

**another** optimization is for `Bitmap`, use `Bitmap::fast_or()` to replace `or_inplace()`. This was a `TODO` item in the old code. I also add a bench test for it:
```
$ cd store
$ cargo test --release bench_fast_or -- --nocapture

running 1 test
  or_inplace():     5.098ms. bitmap cardinality: 225785
     fast_or():     0.778ms. bitmap cardinality: 226018
fast_or_heap():     9.203ms. bitmap cardinality: 225628
```

A main optimization is using `get_hash_by_height()` instead of `get_block_header()`.

BTW, this PR came from the analysis result of issue: https://github.com/mimblewimble/grin/issues/1477

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -140,6 +140,13 @@ impl ChainStore {
 		}
 	}
 
+	pub fn get_hash_by_height(&self, height: u64) -> Result<Hash, Error> {
+		option_to_not_found(
+			self.db.get_ser(&u64_to_key(HEADER_HEIGHT_PREFIX, height)),
+			&format!("Hash at height: {}", height),
+		)
+	}
+
 	pub fn get_header_by_height(&self, height: u64) -> Result<BlockHeader, Error> {
 		option_to_not_found(
 			self.db.get_ser(&u64_to_key(HEADER_HEIGHT_PREFIX, height)),
--- a/chain/src/txhashset.rs
+++ b/chain/src/txhashset.rs
@@ -1214,8 +1214,8 @@ fn input_pos_to_rewind(
 	head_header: &BlockHeader,
 	batch: &Batch,
 ) -> Result<Bitmap, Error> {
-	let mut bitmap = Bitmap::create();
 	let mut current = head_header.hash();
+	let mut height = head_header.height;
 
 	if head_header.height < block_header.height {
 		debug!(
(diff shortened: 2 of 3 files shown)


chain/src/store.rs
```diff
@@ -140,6 +140,13 @@ impl ChainStore {
 		}
 	}
 
+	pub fn get_hash_by_height(&self, height: u64) -> Result<Hash, Error> {
```
antiochp: Nice. Saves us a redundant read from the db each time (and reading the full block header just to get the hash).

antiochp: I think this has almost been added several times in the past...

chain/src/txhashset.rs
```diff
@@ -1224,27 +1224,45 @@ fn input_pos_to_rewind(
...
-	//
-	// TODO - rework this loop to use Bitmap::fast_or() on a vec of bitmaps.
-	//
-	loop {
-		if current == block_header.hash() {
-			break;
+	let bitmap_fast_or = |b_res, block_input_bitmaps: &mut Vec<Bitmap>| -> Option<Bitmap> {
+		if let Some(b) = b_res {
+			block_input_bitmaps.push(b);
+			if block_input_bitmaps.len() < 256 {
```
antiochp: Do we really need this additional complexity for short rewinds?
Might be easier to maintain (and understand) if we just did the `fast_or` for every rewind.

garyyu: thanks for review.
Do you mean push all the bitmaps into the vec[]? then do `fast_or` only once?
For short rewinds, it's Ok. But if the long rewinds (for example the API: http://127.0.0.1:13413/v1/blocks/40000), the memory consumption could be huge, that's why I limit the vec[] size to 256.

Any idea to improve it?

chain/src/txhashset.rs
```diff
@@ -1224,27 +1224,45 @@ fn input_pos_to_rewind(
...
+				return None;
+			}
 		}
+		let bitmap = Bitmap::fast_or(
+			&block_input_bitmaps
+				.iter()
+				.map(|x| x)
```
antiochp: I _think_ you can get rid of the `map(|x| x)` here and just call `.iter().collect()`.

chain/src/txhashset.rs
antiochp: I think I was misreading the code here. Thought we were only doing `fast_or()` if we had > 256 bitmaps.
But I _think_ what you're doing is batching them up and running `fast_or()` on every batch of 256 bitmaps so we don't need to maintain a huge vec of bitmaps.

:+1: if that's the case.
But maybe add a comment explaining that?

garyyu: Yes, that's the purpose.
Will add the code comment for that.

chain/src/txhashset.rs
garyyu: Ok, thanks, I learn a new trick, Yay!

garyyu:
updated for the review comments:
- add code comments for function `bitmap_fast_or`
- remove `map(|x| x)`

PR #1529, opened 2018-09-14
EC terminology edit proposal.

opacey:
The original states "v1*H' is a value times an elliptic curve 'H'. Surely H should be a POINT on a elliptic curve shouldn't it? I've made an edit to line 95 to illustrate. If my edit is correct then a few more edits would be required further down the file.

### DIFF
--- a/doc/intro.md
+++ b/doc/intro.md
@@ -92,7 +92,7 @@ fundamental properties are achieved.
 Building upon the properties of ECC we described above, one can obscure the values
 in a transaction.
 
-If _v_ is the value of a transaction input or output and _H_ an elliptic curve, we can simply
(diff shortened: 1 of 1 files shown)


ignopeverell:
Right, as @garyyu points out (heh), it's not just any point, it's a generator. But saying "elliptic curve point generator" each time is a bit heavy, especially for something that tries to be more of a layman explanation. Given that the whole discrete field is derived from the generator, I think it's not too bad of an approximation.

opacey:
okay! I could follow the doc, (ps, it's excellent), but that part cost me some time re-reading to make sure I understood.

ignopeverell:
I can understand. It is hard to strike a good balance between technical exactness and more generally easy to understand.

PR #1534, opened 2018-09-15
Dual proof of work

ignopeverell:
First PR toward dual proof of work. All this does is:

* Some cleanup on PoW related types (`Difficulty`, `Proof`, `ProofOfWork) and move them to the `pow` crate.
* Introduce a primary PoW subject to size adjustment, a scaling factor of 1, and Cuckoo size >=30.
* Introduce a secondary PoW subject to the header scaling factor and Cuckoo size of 29.

My preference would be to stop here for this PR and merge it on master to minimize the size of the final T4 PR. Merging this, the forking risk is very minimal on T3 given that we have no tools to produce a Cuckoo29 solution (and T3 won't live for too much longer).

Then we can have another PR to add the constraints on the scaling factor, Cuckatoo and adjust Cuckatoo sizes.

@tromp @yeastplume opinions?


tromp:
Oh, ignore those last 2 remarks; this is not supposed to change behaviour just yet...

tromp:
We don't need this:

		// Scaling between 2 proof of work algos
		let difficulty = if scaling & HIGH_BIT == 1 {
			difficulty / (scaling ^ HIGH_BIT)
		} else {
			difficulty * scaling
		};

we just set scaling to the natural scaling factor for cuckoo30, which is 2^(30-30)*29 = 29.
In T4, the variable scaling factor for cuckatoo29 can be set to 2^(29-16+1)*29 = 475136.
No need to mix multiply & divide...

tromp:
I think the dual pow distinction is better not done in function PoW::to_difficulty, but in Difficulty::from_proof_adjusted, since it has a lot in common with Difficulty::from_proof_scaled and has full knowledge of cuckoo_sizeshift.

ignopeverell:
> For T4, I propose lowering REFERENCE_SIZESHIFT to 16
> In T4, Cuckatoo is indexed with log#edges rather than log#nodes, so this will need to be rewritten

Both sound good, I'll refer to this once I start the PR meant for T4.

> We don't need this, we just set scaling to the natural scaling factor for cuckoo30

You lost me here. This leverages the scaling factor between the AF and AR so the ratio of rewards is split under the slow ramp-up schedule from 10% to 100%. How could this be a fixed number as we can't predict the amount of hashpower on each side?

tromp:
In T4, it's not a fixed number, but a dynamically adjusted variable as you note.
I'm proposing we just multiply it with the raw difficulty, as we do with the AF natural scaling factors,
so none of that HIGH_BIT testing and division stuff.
For the current PR we must preserve old behaviour, so an initial value of 2^(29-16+1)*29 and
no dynamic adjustment.

ignopeverell:
Ah ok, that makes more sense. If we just multiply, it means we don't allow it to get lower than the raw difficulty. What if the AF PoW pushes the raw difficulty so high it becomes unattainable on the AR side?

tromp:
If AF PoW pushes difficulty high, then AR side can just push scaling factor up all the way to u64::MAX to compete.
Your worry should be high AR hashrate reducing the AR scaling factor to 1 making AF unable to get 10% of blocks. But that's prevented by the minimum AF scaling factor of 4194304, meaning the ASICs only need to find 1/9 * 1/4194304 as many solutions as the GPUs...

yeastplume:
I am very far behind right now on my knowledge of Cuckatoo, but as far as the merge strategy, sounds like a good approach. (We used to have a pow crate, but there ended up being so little in it I moved it back to core) Will start getting into it properly once https://github.com/mimblewimble/grin/pull/1501 is ready, which is going to end up a massive PR covering everything non-PoW for T4, so anything that can minimise merge headaches is fine with me.

PR #1536, opened 2018-09-16
Update version of some crates

garyyu:
1. The `enumset 0.3.11` building problem reproduced, fix it by switching it to `0.3.10` version.
2. Update `slog` from `version = "~2.2"` to `~2`, which will match the latest 2.x.x version. For the moment, it will update from `slog v2.2.3` to `slog = "2.3.3"`, which can fix a minor issue I saw: The last line of log always lose.
3. Just removing the whole `Cargo.lock` and regenerated, then some packages are automatic updated:

- "aho-corasick" : "0.6.6" -> "0.6.8"
- "backtrace-sys": "0.1.23" -> "0.1.24"
- "base64": "0.9.2" -> "0.9.3"
- "bitflags": "1.0.3" -> "1.0.4"
- "byteorder": "1.2.4" -> "1.2.6"
- "bytes": "0.4.9" -> "0.4.10"
- "cc": "1.0.18" -> "1.0.25"
- "cfg-if": "0.1.4" -> "0.1.5"
- "chrono": "0.4.5" -> "0.4.6"
- "crossbeam-channel": "0.2.4" -> "0.2.5"
- "crossbeam-deque": "0.3.1" -> "0.6.1"
- "crossbeam-epoch": "0.4.3" -> "0.5.2"
- "crossbeam-epoch": "0.5.2" -> "0.6.0"
- "encode_unicode": "0.3.1" -> "0.3.2"
- "enumset": "0.3.11" -> "0.3.10"       <<<< still strange here! we're using 0.3.11 before but no building error.
- "env_logger": "0.5.11" -> "0.5.13"
- "failure": "0.1.1" -> "0.1.2"
- "failure_derive": "0.1.1" -> "0.1.2"
- "futures": "0.1.23" -> "0.1.24"
- "git2": "0.7.4" -> "0.7.5"
- "h2": "0.1.11" -> "0.1.12"
- "http": "0.1.8" -> "0.1.13"
- "hyper": "0.12.7" -> "0.12.10"
- "isatty": "0.1.8" -> "0.1.9"
- "itoa": "0.4.2" -> "0.4.3"
- "lazy_static": "1.0.2" -> "1.1.0"
- "lazycell": "0.6.0" -> "1.1.0"
- "libc": "0.2.42" -> "0.2.43"
- ...

Packages upgrading is indeed a pain, too many dependencies.

But anyway keeping packages updated is good.

The rough test on mac and debian linux are ok for these updating.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,10 +5,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "aho-corasick"
-version = "0.6.6"
+version = "0.6.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -51,7 +51,7 @@ name = "atty"
 version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.42 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.43 (registry+https://github.com/rust-lang/crates.io-index)",
  "termion 1.5.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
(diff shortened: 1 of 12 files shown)


ignopeverell:
All good except I wouldn't do the `slog` fuzzy version. We don't know what they'll come up with on 2.4 or 2.5 and it may not be backward compatible.

garyyu:
OK, `slog` version changed to "~2.3" instead of "~2".

knocte:
> But anyway keeping packages updated is good.

I disagree with this. Updating dependencies should be done in a case by case basis and with enough reason to do so, otherwise it's very easy to introduce unnoticed regressions.

If the updates involve only security updates, then yes, updating is good.

garyyu:
@knocte  thanks for your comment~
Agree with "If the updates involve only security updates, then yes, updating is good." 👍 

Regarding:
> Updating dependencies should be done in a case by case basis and with enough reason to do so

Absolutely yes for released branch. And IMO, it should be Okay to update for a not released branch (mainly the master).

BTW, I think we indeed need start the release versions and give dedicated version branch to maintain versions, that's another topic:)

knocte:
> And IMO, it should be Okay to update for a not released branch (mainly the master).

Even if it's for a not released branch, updating for the sake of updating is not a good strategy.

sesam:
@knocte we're probably already off-topic. I'm hearing A.B.U. - Always be Updating. For development branches, and elsewhere too. You never know ahead if there's a critical vuln being fixed quitely, or even unknowingly just as a side effect.

:) If your worry is about getting bad updates that compromises your security, you already had "none" :-/. Routers, phones, HP printers, and X86 CPUs. But upgrading helps. It sure helps.

Issue #569, opened 2018-01-02
Make block reward 60 grin (~1/sec)


antiochp:
Right now the block reward is 50 grin per block.
A block reward of 60 grin per block was discussed on gitter - which would give us the nice property of approx 1 grin per second (based on 60s blocks).

PR #1537, opened 2018-09-16
Fix documented emission rate

ivy:
The emission rate was changed to 60 grin per minute (or approximately 1 grin/second) in #569. Unfortunately, I haven't been able to find an official statement on this outside of a [recent interview](https://medium.com/grin-mimblewimble/grin-mimblewimble-meetup-in-london-with-yeastplume-f11476b3a590) from @yeastplume:

> Q: What about the emission rate?
>
> A: Right now it’s planned to be linear. 60 grins every minute for the rest of eternity. […]

This PR only addresses the outdated Grin4bitcoiners document and I wasn't able to find anything else under `/doc`. Perhaps this is something that should be clarified in a separate document?

### DIFF
--- a/doc/grin4bitcoiners.md
+++ b/doc/grin4bitcoiners.md
@@ -31,7 +31,7 @@ Maybe you've heard that MimbleWimble doesn't support scripts. And in some way, t
 
 ## Emission Rate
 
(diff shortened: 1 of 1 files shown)


ignopeverell:
`conensus.rs` serves as a poor-man consensus parameters document but we sure should be able to do better than that. In the meantime, thank you for the PR!

Issue #1524, opened 2018-09-13
Node stuck in sync loop


quentinlesceller:
Attaching the `grin-server.log`
[grin-server.log](https://github.com/mimblewimble/grin/files/2380730/grin-server.log)

ignopeverell:
Looks like an issue in the generation of the header locator, possible a regression after @garyyu optimization? Just taking a few lines out of the logs (didn't know you could do 22MB attachments on github btw):

```
Sep 13 14:32:40.936 DEBG sync_state: sync_status: BodySync { current_height: 93760, highest_height: 93766 } -> HeaderSync { current_height: 93760, highest_height: 93766 }
```

So the node is at 93760 and needs to get 6 blocks to sync up to 93766. However the generated locator looks like this:

```
Sep 13 14:32:40.934 DEBG sync: locator heights': [93130, 93128, 93124, 93116, 93100, 93068, 93004, 92876, 92620, 92108, 91084, 89036, 84940, 76748, 60364, 27596, 0]
Sep 13 14:32:40.934 DEBG sync: history locators: len=17, shrunk=0
Sep 13 14:32:40.935 DEBG sync: locator: [014a6cc3, 082b95aa, 07588dd0, 04002d2b, 006f15f9, 010464cb, 0337b8f3, 0216ed75, 029d2984, 01ab762a, 0af59373, 0476c5b7, 08806be1, 1527d2ff, 12a566dd, 0e418688, 006642e0]
```

Instead of starting at 93760, the locator starts at 93130, which is too far back. So other nodes send us headers we already have. There could be another issue in addition as we logs already show a lot of:

```
Sep 13 14:32:51.000 INFO All known, ignoring
```

Normally, even if we ask at 93130, we should still receive the next 36 headers, 30 of them would get discarded but the last 6 would help. Unfortunately grinscan.net looks down at the moment which complicates lookup by short hashes from the logs.

garyyu:
This looks like something is wrong in the sync head:
```
Sep 13 14:32:39.885 DEBG sync: locator heights : [93130, 93128, 93124, 93116, 93100, 93068, 93004, 92876, 92620, 92108, 91084, 89036, 84940, 76748, 60364, 27596, 0]
Sep 13 14:32:40.934 DEBG sync: locator heights': [93130, 93128, 93124, 93116, 93100, 93068, 93004, 92876, 92620, 92108, 91084, 89036, 84940, 76748, 60364, 27596, 0]
```
[...]

PR #1539, opened 2018-09-17
[DNM] Reset sync head

antiochp:
This should fix the issue with the `sync_head` being out of date.
This is all we should need to do.

Related - #1531.
Related - #1524.

@garyyu @ignopeverell 

TODO - 

- [ ] as @garyyu pointed out, some timing/ordering issues around the state transitions

### DIFF
--- a/servers/src/grin/sync.rs
+++ b/servers/src/grin/sync.rs
@@ -147,6 +147,10 @@ pub fn run_sync(
 							match status {
 								SyncStatus::TxHashsetDownload => (),
 								_ => {
+									// Reset sync_head to header_head
+									// on transition to HeaderSync.
+									chain.init_sync_head(&header_head).unwrap();
+
 									sync_state.update(SyncStatus::HeaderSync {
 										current_height: header_head.height,
 										highest_height: si.highest_height,
--- a/servers/src/grin/sync.rs
+++ b/servers/src/grin/sync.rs
@@ -93,8 +93,9 @@ pub fn run_sync(
 						// * timeout
 						if wp.len() > MIN_PEERS
 							|| (wp.len() == 0
-								&& peers.enough_peers()
-								&& head.total_difficulty > Difficulty::zero()) || n > wait_secs
+								&& peers.enough_peers() && head.total_difficulty
+								> Difficulty::zero())
+							|| n > wait_secs
 						{
 							break;
 						}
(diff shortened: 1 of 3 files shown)


servers/src/grin/sync.rs
```diff
@@ -147,6 +148,10 @@ pub fn run_sync(
 							match status {
 								SyncStatus::TxHashsetDownload => (),
 								_ => {
+									// Reset sync_head to header_head
+									// on transition to HeaderSync.
+									chain.init_sync_head(&header_head).unwrap();
```
antiochp: Ignore all the rustfmt changes, this is the important change.

garyyu: 👍 
And some review comments:
1. Once entering syncing and `header_sync_due`, it will be called. that means this `init_sync_head()` could be called in every 10 seconds. Do we really need update so frequently?
2. `init_sync_head()` should be called on front of `header_sync()`.
3. If it doesn't matter whether we reset before receiving new Headers or after receiving new Headers, I still prefer after, just let `get_locator()` using `header_head` instead of `sync_head` in case of they're not matched. It make sense to update `sync_head` ONLY after we really received that Headers, doesn't it?
4. If we reset `sync_head` in this way, we will lose the ability to convince ourselves in the future about whether this `sync_head` comes from Sync OR come from this Reset.

@ignopeverell  what's your opinion on this?

garyyu: In another words: 

IMO, the problem of this issue #1524 is in the `get_locator()`, not about the `sync_head` updating / or reset.

ignopeverell: 1. I agree we need this change.
2. I agree we also need to guard against unwanted state changes that could mess this up.
3. Tackle the other issues separately in additional PRs.

garyyu: Please feel free to use @antiochp 's fix solution here 😄   if nobody agree `get_locator()` is a better solution, leaving `sync_head` untouched if no new Headers received.

Anyway, I continue to cleanup the code of #1531, to split the unrelated modification into new PR.

antiochp:
I'm closing this as it was just to demonstrate how we could potentially update `sync_head`.
It's not in a working state and on closer examination the sync status needs some rework to only do this on the initial transition to HeaderSync (I thought we had a hook in the code already for this but its not doing quite what I thought it was).

Issue #1379, opened 2018-08-18
Provide 64 bits Mac and Linux binaries


ignopeverell:
It shouldn't be too complicated at this point, but it implies a release process and schedule. I'm proposing to start with loose monthly releases. It's a good balance of frequency and by not setting a fixed date we give ourselves some latitude to move it by a week or two if it happens to be a bad time or unstable. This should include `grin-miner`.

It may be a little more controversial but we may want to include a kill switch after say 2 months. At least until mainnet.

Toughts?

antiochp:
Kill switch as in the local binary itself stops running after a particular date?
Or is this possible (feasible or even desirable?) to do with a hardfork on the chain itself?

yeastplume:
Makes sense to start getting into the habit. Are you thinking about automating this? Providing .deb or .pkg packages? I've also had issues in the past trying to use binaries from one distro on another.

ignopeverell:
Haven't thought about automation yet, I was thinking that step 1 would be to try a build and see what are the portability issues.

@antiochp I was thinking about the binary stopping after say 3 months (for a 1-month release schedule).

quentinlesceller:
Do we want to do deterministic builds such as gitian build (https://github.com/bitcoin-core/docs/blob/master/gitian-building.md) ?

antiochp:
Yeah - I think having the binaries time limited makes a lot of sense. 
Its just a communication issue to make sure users are clear that this is what they are signing up for initially.

antiochp:
@quentinlesceller deterministic builds would be a nice thing to aim for.
I can't find the link now but I thought I'd read something about this being hard to do with Rust/Cargo though. Could be wrong though.

quentinlesceller:
Ah yes I remember that too. This is what you are looking for @antiochp https://github.com/rust-lang/rust/issues/34902.

ignopeverell:
Let's make that a goal eventually but not start with it. Right now a good first step would simply be to have 2 portable binaries.

kargakis:
@stratakis can help with RPMs in Fedora once you start releasing binaries.

mcdallas:
For distributing the binaries may I suggest keybase.io? 

Each keybase user has a public folder using the keybase filesystem ([kbfs](https://keybase.io/docs/kbfs)). Everything you put in your public folder is automatically signed and there are public proofs that the key belongs to the owner (if you link your github account). 

[...]

PR #1540, opened 2018-09-17
chore: automatic binaries release

garyyu:
Working In Progress for https://github.com/mimblewimble/grin/issues/1379, providing 64 bits Mac and Linux binaries.

TODO:
- [x] build for both mac and linux platform.
- [x] compress grin as a tar.gz file, and name the binary with the version number
- [x] md5 checksum for generated binary
- [x] script to automatically create change log.
- [x] `release instruction` document.
- [x] modify `.travis.yml` to deploy auto release.


quentinlesceller:
Not sure it is https://github.com/rust-lang/rust/issues/34902

ignopeverell:
Deterministic binaries would definitely be nice, but it's not easy and Rome wasn't built in a day.

@garyyu a difficulty here is to do a fully statically bound build, that's the only way we can have decent portability. See here for what I mean:

https://doc.rust-lang.org/1.9.0/book/advanced-linking.html#static-linking

This document might be outdated, so it'll likely requiring some testing and trying things out.

jaspervdm:
@ignopeverell yes sorry I was only wondering about the technical possibility, not necessarily asking for it now :) But it seems that it is hard with Rust

ignopeverell:
@jaspervdm no worries, still good to keep on the back of your mind!

garyyu:
Finally done!  It's ready to take a review now 😸 

Please take a serious review on this PR, both on the code and the doc, because it's regarding the version naming rule, and release process, etc.

kargakis:
Another thing to keep in mind while working on this is [secure verification](https://www.grin-forum.org/t/grin-binaries-w-secure-verification/782)

Maybe not necessary now but the sooner it's in place, the better for tooling that wants to make use of it. :)

.auto-release.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+repo_slug="garyyu/grin"
```
jaspervdm: Shouldn't this be `mimblewimble/grin`?

garyyu: Ah, Yes!   this one is forgot 😄   Thanks @jaspervdm

garyyu: fixed now.

jaspervdm:
Maybe its a bit too early to define release versions? At some point we need it, but at that same time we should formalize the release cycle, commits, reviews etc. (GIPs anyone?) Might be worth to talk about in the next meeting :)

antiochp:
Agree with @jaspervdm - maybe worth breaking a smaller PR out of this that just focuses on building the binaries?

We can build binaries based on the git SHA (at least initially) without trying to lockdown a full release structure and set of naming conventions etc.

CONTRIBUTING.md
```diff
@@ -6,17 +6,34 @@ The [list of issues](https://github.com/mimblewimble/grin/issues) is a good plac
...
+# Find Us
 
+If any help is needed during your effort to contribute on this project, please don't hesitate to contact us:
 * Chat: [Gitter](https://gitter.im/grin_community/Lobby).
+* [Forum](https://www.grin-forum.org/)
+* [Website](https://grin-tech.org)
 * Mailing list: join the [~MimbleWimble team](https://launchpad.net/~mimblewimble) and subscribe on Launchpad.
+* News: [Twitter](https://twitter.com/grinmw). Twitter bot that scrapes headlines, mailing list, and reddit posts related to MimbleWinble/Grin.
+
+# Pull-Request Name Prefix
```
quentinlesceller: Was it discussed on Gitter? Not against it but I don't see why it belongs in this PR.

CONTRIBUTING.md
```diff
@@ -6,17 +6,34 @@ The [list of issues](https://github.com/mimblewimble/grin/issues) is a good plac
...
+- **test**:     Adding missing tests
+- **chore**:    Maintain. Changes to the build process or auxiliary tools/libraries/documentation
+
+For example: `fix: a panick on xxx when grin exiting`. Please don't worry if you can't find a suitable prefix, this's just optional, not mandatory.
 
 # Grin Style Guide
 
-Grin uses `rustfmt` to maintain consistent formatting.
-Please use the git commit hook as explained below.
+Grin uses `rustfmt` to maintain consistent formatting, and we're using the git commit hook as explained below.
```
quentinlesceller: Same here not sure this changes is needed or belong in this PR.

CONTRIBUTING.md
```diff
@@ -49,20 +61,26 @@ The output will be:
 core.hookspath=./.hooks
 ```
 
-## Running rustfmt
+**Note**: The pre-commit hook will not prevent commits if style issues are present, instead it will indicate any files that need formatting, and it will automatically run `rustfmt` on your changed files, each time when you try to do `git commit`.
```
quentinlesceller: 👍 Indeed good for newcomers .

CONTRIBUTING.md
```diff
@@ -49,20 +61,26 @@ The output will be:
...
+For example:
+```
+rustfmt client.rs
+```
+
+**Notes**:
+1. *Please add the rustfmt corrections as a separate commit at the end of your Pull Request to make the reviewers happy.*
 
-Any errors or rustfmt failures? Fix those manually.
+2. *most important: `rustfmt` is just a tool to make your code having pretty formatting, your changed code itself importance is definitely above the code format, since anyway the formatting can be done by running `rustfmt` later. So the point is, if not sure what should do on this, or the `rustfmt` is changing a lot of untouched code, please feel free to ignore it. We like to make you also happy to contribute on this open source project!*
```
quentinlesceller: I understand the intention but the wording is confusing.

.travis.yml
```diff
@@ -37,12 +53,25 @@ env:
...
+  - if [[ "$TEST_DIR" == "none" ]]; then cargo build --release;                fi
+  - if [[ "$TEST_DIR" != "none" ]]; then cd $TEST_DIR && cargo test --release; fi
+
+before_deploy:
+  - if [[ "$TEST_DIR" == "none" ]]; then ./.auto-release.sh;                   fi
+
 deploy:
   provider: releases
   api_key:
     secure: PBTFcoUmiQITkDdtFzrBlNR/5OgYHTCw+xVWGYu205xwTlj/ARBgw7DNt8dIdptLx+jOM2V5SbJqSFxs/CJ2ZcOHQZ6ubwpAJlRfuk3xDAi5JmuHYfcY+4SQ9l/0MgHnGfuml093xP7vTIYm2Vwwgdq8fd3jdWmvwgk9zgaGXB4UIXQA0yIs3EzxZpqiLg629Ouv7edMfyffwlG+rgQ1koe6sqeMCxIs0N3p97GCx19kNe0TV4dC7XAN74HreMdHmwxPKAK4xG/jtA1Snm0pMQ50Z0Kizt+0yrGOPMLnWwO9sS38iosBn3Vh1R8HKle2xBGflTtT/LG9lHdQZ5NF572q6681x6t7str4OjJ5bboy1PtNLFxG7RJCVIpp9gbouzdxIaJWRTxIdlk8UNQMrD8ieiNE6V1vZtbHGtJHRSJN1vO/XxsLlQDCyakLhG/nmSKXgiT9wIsu+zj/3oDe+LBt5QetEGYGBrCwUewjaQ7EP1rsT7alQrHTMad5DPjYftJuvfR+yBtz1qbzQwZVJpQC1KY1c476mXPQsaywuUrj56hH92p7P3vl6aMN2OPJZP+zENOVSURHc56KeTsDS55+KKzcRjCMA2L0LR1hP33+V5kavMHgCRrWIkxAkZ4eRqnermalzp8vlzL6EEoGm0VFLzv4mJmzrY1mC1LyCHo=
```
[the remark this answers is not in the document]
garyyu: what do you mean?  what can be pull out of travis.yml ?

doc/release_instruction.md
```diff
@@ -0,0 +1,132 @@
+# Grin - Instruction of Release 
+
+## Version Number Rule
+
+In Grin, we're using [Semantic Versioning 2.0.0](https://semver.org). For a short description of the rule:
```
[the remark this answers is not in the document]
garyyu: Anyway we have a 1st time to release binary, why not now?

CONTRIBUTING.md
garyyu: Ah, just copy & paste.  Do you mean `News` is not official?

I'm touching this doc, see the different `Find us`, so, ...   don't want to wait 😺

CONTRIBUTING.md
garyyu: 😸

CONTRIBUTING.md
quentinlesceller: No I mean the prefixes. Sorry if that wasn't clear.

CONTRIBUTING.md
garyyu: ha ha~   I'm not a good english writer, so perhaps somebody help to rewrite this paragraph?  just want to ease the reader about the minor format issue.

garyyu:
@jaspervdm  @antiochp  

>  Maybe its a bit too early to define release versions?

How can we deliver binaries but without version?  each binary must have a version attached and identified.

And regarding to `too early`, is it talking about 1 month?  3 months? or 1 year?  IMO, if it's in 3 months, then now it's good time to start it.  Even this initial doc about version rule has a lot of items can be discussed, but it's already a good start, no?

CONTRIBUTING.md
garyyu: Okay 😄   Yes, it belongs. Because in this PR, the `change log` will be generated from these `issues` and `pull-request`. We don't want to generate a mess change log 😸 

And that's why I said I request `serious` review and send invite message in Gitter to give review on this PR.

jaspervdm:
@garyyu like @antiochp suggested, i think for now we can couple each binary to the hash of the commit.

The main reason it is too early in my opinion is that because we have to define more things than just version number (see my previous comment).

.travis.yml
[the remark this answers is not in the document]
hashmap: @antiochp I assume the value is encrypted and the key is stored outside of `travis.yml`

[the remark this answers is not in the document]
antiochp: Ahh ignore that then. 
Yes - this is an encrypted version of the key and according to travis docs is the "official" way of doing this.

Still makes me nervous though to have credentials like this checked in...

ignopeverell:
I'd also be happy with "snapshot" as release number for now (or 0.0.0) to make it obvious. Release management and policy is a much longer discussion. But it's great to have a binary build to start!

garyyu:
> I'd also be happy with "snapshot" as release number for now (or 0.0.0)

Agree with `0.0.0`, and it's still feasible with this PR, just making a tag with `0.0.0-pre.0.3.1` or something like that, that's all.
We can set all these versions as `pre-release` in Github.

> Release management and policy is a much longer discussion. 

hmm, look like it's absolutely Yes. 
So, let me change all the `process` and `policy` parts with an obvious indicator such as `[draft! to be reviewed and discussed]`, then go ahead with this PR. And create several forum posts on these topics to follow up the discussion. Does this make sense?

> But it's great to have a binary build to start!

Fully agree! That's why I take this task as my 1st priority 😺  

@ all other reviewers:

Please continue review these `release management and policy` parts in forum (will create today), and discuss it in Gitter anytime when needed. And if this this PR don't have functional problem for auto binary releasing, please approve it to make the binaries happen in github 😄  (and as I said above, I need  add `[draft! to be reviewed and discussed]` on those doc parts, will do it now.)

And I took a look at some other chains **release** page which are same at **Testnet**, some of them already released many versions. Here just lists 3 of them which I search quickly:
- aeternity:   https://github.com/aeternity/epoch/releases
- cardano:    https://github.com/input-output-hk/cardano-sl/releases Launched by Input Output Hong Kong(IOHK), lead by Charles Hoskinson, former CEO of Ethereum. Market cap rank no.9: https://www.coingecko.com/en/coins/cardano
- Even the Beam-MW which was just opened in Github at this month:  https://github.com/beam-mw/beam/releases

aeternity and cardano at above even name the testnet binary versions as `0.22.0` or `1.3.0`. And if we release binaries with `0.0.0-pre.0.3.1` something like that, we should have nothing to worry.

garyyu:
Two notes added:
1. pull-request title prefix:
```
# Pull-Request Name Prefix

**Note**: *[draft part! to be reviewed and discussed]*
```

2. Release Instruction Doc
```
# Grin - Instruction of Release 

**Note**: *[totally draft doc! to be reviewed and discussed]*
```

garyyu:
Forum post created:
- https://www.grin-forum.org/t/github-pr-pull-request-title-prefix-rule-optional/842
- https://www.grin-forum.org/t/release-management-and-policy/843

CONTRIBUTING.md
garyyu: I update and add the explicit [draft, to be reviewed and discussed] words there.  
and a forum post for this discussion is submitted on:  https://www.grin-forum.org/t/github-pr-pull-request-title-prefix-rule-optional/842  , so let's continue discussion there.

antiochp:
Can we break out the "binary release" part of this PR into a separate PR?

garyyu:
@antiochp  OK for me.  I guess you mean split 2 docs into another PR. Could you confirm it?
- CONTRIBUTING.md
- doc/release_instruction.md

garyyu:
Done for the docs splitting:
- #1570 
- #1571

garyyu:
After splitting the docs, is there any further review comments on this PR? 😃

.auto-release.sh
```diff
@@ -0,0 +1,68 @@
...
+    /bin/ls -ls *-md5sum.txt  | awk '{print $6,$7,$8,$9,$10}'
+    cd - > /dev/null;
+    echo "tarball generated\n"
+fi
+
+version="$tagname"
+branch="`git symbolic-ref -q --short HEAD`"
+
+# automatic changelog generator
+gem install github_changelog_generator
```
kargakis: I guess it makes sense to use the CI to release the binaries but has it been thought through whether the release process should depend on travis as opposed to a script that can run both in travis and locally?

.travis.yml
```diff
@@ -37,12 +53,25 @@ env:
   - TEST_DIR=core
   - TEST_DIR=util
   - TEST_DIR=config
-script: cd $TEST_DIR && cargo test --release
+  - TEST_DIR=none
+
+script:
+  - if [[ "$TEST_DIR" == "none" ]]; then cargo build --release;                fi
```
kargakis: I have hit issues with c shared libraries when trying to build a docker image based on debian that bind-mounts a grin binary I built on fedora so I am afraid if this is not static linked, it may cause issues in some linux systems.

garyyu: Do you know which shared libraries are missing? We can update the `How to use` doc (and all related doc) for the dependent shared libraries, and update the docker file.

I'm not a fans of static linking because it make the binary much bigger, and the most important:
- the bugs (especially the security related bugs) in the shared libraries need to be managed by OS security upgrading process, not managed by Grin binary.

garyyu:
I'm merging this one if no more comment or against.

So as to enable this function, and start doing more tests on change log generation on the real project environment.

.auto-release.sh
kargakis: You wouldn't use travis at all when running locally, you would push releases directly in github.

.travis.yml
kargakis: There wasn't a library missing, I think the issue was that the location of shared libraries differs between different linux systems. I want to try and reproduce by fiddling with the docker build but haven't had time to do so far. I don't know that the size of a binary is an issue but I can buy the argument of not having to care about OS security updates.

.auto-release.sh
garyyu: Thanks @kargakis for the review~
I'm afraid no, because we can't push directly in github 😄   It's a rule here any change in repo need use PR instead of directly pushing, for better quality control and review.

.travis.yml
garyyu: @kargakis  OK.  If need help, please submit an issue to describe more detail about the problem, then I could look into it with a docker.

Issue #1290, opened 2018-07-23
Secure access to the grin API by default


kargakis:
Access to the node API is unsecure by default, which means I can easily write a script to determine all nodes that expose their API to the public, and fiddle with them (eg. ban other peers). grin should provide an option to selectively limit access to its APIs via some sort of auth layer. I can see use cases where someone would want to expose some parts of the API like chain height but not others like peers or access to mutating calls. If exposing the web server publicly is not recommended and we don't care about auth, we should at least document it.

bladedoyle:
By default, access to the API is limited by the sockets listening only on 127.0.0.1

https://github.com/mimblewimble/grin/blob/54d3fc0af2c6e012f234358b5dd2722172c61238/grin.toml#L21

PR #1544, opened 2018-09-17
[WIP] Add a private rest API

quentinlesceller:
Fixes https://github.com/mimblewimble/grin/issues/1290.

Introduce another REST API with auth (`?secret=SECRET) in url listening by default on localhost:13412.
Private api calls are for now:
- Everything peers related
- Node status
- Pool info
- Chain compact/validate

Thoughts?


ignopeverell:
Wouldn't HTTP Basic Auth be better? It's standard, your browser would do the right thing, it doesn't involve the server seeing your secret. It can be a little bit more complicated for clients, but a lot of libraries support it.

kargakis:
Avoid using secrets in url parameters. Usually you do http auth via headers.

quentinlesceller:
Closing that one in favor of a single protected API. I will link the new PR here.

kargakis:
Replaced by https://github.com/mimblewimble/grin/pull/1566

PR #1546, opened 2018-09-18
fix: early detection of peer connection lost when downloading txhashset

garyyu:
For the moment, the txhashset archive has increased to 89M. For fast sync, at txhashset download stage, I have seen more times than before that the peer connection is lost during the downloading procedure.

Currently, we need wait 10 minutes to restart the fast sync:
```
			Some(prev) => {
				if now - prev > Duration::minutes(10) {    <<<< here!
					download_timeout = true;
				}
				(false, download_timeout)
			}
```

In this PR, I add the peer connection status checking, so we can quickly know the connection lost and restart the fast sync.

### DIFF
--- a/servers/src/grin/sync.rs
+++ b/servers/src/grin/sync.rs
@@ -15,7 +15,7 @@
 use chrono::prelude::{DateTime, Utc};
 use chrono::Duration;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 use std::time;
 use std::{cmp, thread};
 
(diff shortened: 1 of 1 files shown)


servers/src/grin/sync.rs
```diff
@@ -167,12 +167,30 @@ pub fn run_sync(
...
-										si.fast_sync_reset();
-										sync_error_need_clear = true;
+										sync_need_restart = true;
 									}
 									drop(clone);
 								}
-								if sync_error_need_clear {
+
+								// check peer connection status of this sync
+								if let Some(ref peer) = si.fast_sync_peer {
```
ignopeverell: Perhaps extract those 3 cascading `if`s as a utility function? `sync.rs` is starting to balloon a bit and this main loop even more so.

garyyu: Agree 😺    Actually I plan to do a refactor on the whole `run_sync()`, which is quite hard to do a code review because of a lot of line wrapped.

but prefer to do it in next PR to avoid mixing up with this fix.

PR #1552, opened 2018-09-19
fix: proper tx confirmation calculation

### LINKED ISSUE #1507: wallet showing 0 confirmations (should be 1 confirmation)
I think our # confirms is incorrect in the wallet currently.

If we are at height `90865` and an output is in the most recent block (i.e. included in block `90865`) then that output should have 1 confirmation (and in the the current wallet impl, where we wait for 1 confirmation, be spendable).

```
Wallet Outputs - Block Height: 90865
---------------------------------------------------------------------------------------------------------------------------------------------------
 Output Commitment                                                   Block Height  Locked Until  Status   Coinbase?  # Confirms  Value          Tx 
===================================================================================================================================================
 09d472af6eae3924e56cf2bd95996fb2f305ffa298dfd0199616165e6fb4fa8b65  90865         0             Unspent  false      0           159.990000000  18 
---------------------------------------------------------------------------------------------------------------------------------------------------
 09730c03c63dea83d923bf79020aa50f5ed6e80835f5a6e185755562504c5522fe  90865         0             Unspent  false      0           50.000000000   17 
---------------------------------------------------------------------------------------------------------------------------------------------------
```

    0 conf - "output not yet included in a block, unconfirmed"
    1 conf- "output included in most recent block"
    2 conf - "output included in block at depth 2"  etc.

The impact here is we cannot spend an output immediately after seeing it confirm right now.

Related - #1467 (proposal to change "spendable" to be 10 confirmations, not 1).

aupiff:
fixes #1507 

The method `num_confirmation` on the struct `OutputData` had a small error. The check

```rust
if self.height >= current_height {
	return 0;
}
```

returned # of confirmations = 0 when `self.height == current_height` though @antiochp explained that confirmations should = 1 in this state. The fix required replacing `>=` with `>`.

### DIFF
--- a/wallet/src/libwallet/types.rs
+++ b/wallet/src/libwallet/types.rs
@@ -248,7 +248,7 @@ impl OutputData {
 	/// so we do not actually know how many confirmations this output had (and
 	/// never will).
 	pub fn num_confirmations(&self, current_height: u64) -> u64 {
-		if self.height >= current_height {
+		if self.height > current_height {
 			return 0;
 		}
 		if self.status == OutputStatus::Unconfirmed {


antiochp:
Nice! Thanks for tracking that down.

PR #1556, opened 2018-09-19
Cleanup new output commits

antiochp:
We used to track "new outputs" being applied to the txhashset extension.
We no longer need this as the lmdb batch semantics gives us this for free.

We just need to read from the batch while it is open and remember to commit it at the end.


antiochp:
Those batches are actually really nice once you get used to the transactional semantics.

PR #1557, opened 2018-09-19
Refactor hyper router

hashmap:
* Make it simpler to implement middleware
* Switch from the current thread runtime to the default one. It enables us to inject TLS support later one and potentially more scalable, unfortunately it involves some additonal cloning of the router, because we can't rely on thread local vars anymore
* Introduce `call` entrypoint for Handler, so it's possible to handle any HTTP method in one place, handy for middleware
* Implement example of middleware

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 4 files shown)


quentinlesceller:
👍 This is really neat and will make implementation of middlewares (such as the one needed in #1290) really easy.

PR #1558, opened 2018-09-19
Rust Cuckatoo for verifier and test-miner

yeastplume:
I think this is ready and okay to merge now. What I ended up doing is:

* Implementing simple cuckatoo test solver and verifier
* Refactoring the pow interface into a `PoWContext` trait, that can be instantiated with either a Cuckoo or Cuckatoo context (or whatever comes along), and making sure this trait is used throughout.
* Ensuring that the instantiation is now done in 'global.rs', so there's currently only one line that needs to change in order to swap to a cuckatoo PoW (more will be needed for Dual-PoW, but this is a good start and can be merged now into T3.

Added some Cuckatoo 29 test vectors (should add more as we get them) and all appear to be consistent with the latest @tromp output.

@ignopeverell have a look over it when you can (or just hit the 'merge' button if you're okay with it)


core/src/pow/cuckatoo.rs
```diff
@@ -0,0 +1,487 @@
...
+		];
+		self.solutions = vec![Proof::zero(self.proof_size); 1];
+		self.visited = Bitmap::create();
+		Ok(())
+	}
+
+	pub fn byte_count(&self) -> Result<u64, Error> {
+		Ok(2
+			* self.max_edges.to_u64().ok_or(ErrorKind::IntegerCast)?
+			* mem::size_of::<Link<T>>() as u64 + mem::size_of::<T>() as u64 * 2 * self.max_nodes)
```
ignopeverell: rustfmt is trolling us

PR #1563, opened 2018-09-20
Remove a dead link from docs

mcdallas:
The referenced file no longer exists, it was removed with #1444


doc/intro.md
```diff
@@ -249,7 +249,7 @@ the number. We won't elaborate on the range proof, but you just need to know
 that for any `r.G + v.H` we can build a proof that will show that _v_ is greater than
 zero and does not overflow.
 
-It's also important to note that in order to create a valid range proof from the example above, both of the values 113 and 28 used in creating and signing for the excess value must be known. The reason for this, as well as a more detailed description of range proofs are further detailed in the [range proof primer](rangeproofs.md).
+It's also important to note that in order to create a valid range proof from the example above, both of the values 113 and 28 used in creating and signing for the excess value must be known.
```
garyyu: Thanks @mcdallas to spot this missing 👍 

Perhaps instead of removing it, we could change the link to origin bullet range proof paper?   https://eprint.iacr.org/2017/1066.pdf

Just in case the reader want to know more detail about it.

mcdallas:
@garyyu Fixed to point to the original paper

PR #1565, opened 2018-09-20
Support TLS in ApiServer

hashmap:
This is a ground work to support TLS in Grin APIs (like wallet or node). Particular API implementation needs to decide if TLS is used or not and pass certificate data etc.

* P12 format support
* New method to start TLS server
* Transparent TLS support in API client (depends on URL scheme http/https)
* Refactoring
* Initial support for graceful shutdown (commented out int this PR, unstable for now)
* API server tests (TLS server test is disabled by default, hyper client rejects self-signed certificates, so extra step is needed to install local CA (I used mkcert)


kargakis:
Is there any documentation on how to use this? Do I need to do anything config-wise or will a new node use it by default?

quentinlesceller:
@kargakis Next PR from HashMap should enable this functionality.

hashmap:
@kargakis There will be for sure. This PR doesn't introduce TLS support for node nor wallet, just enables it on the low level. Next pr(s) will add options to configure it and enable in user facing APIs. TLS will be optional because it requires extra steps which are not feasible for some users, like getting a certificate and domain name (or getting a cert  for IP, which is also problematic)

PR #1566, opened 2018-09-20
Implement Basic Auth for API and Owner API

quentinlesceller:
Fixes https://github.com/mimblewimble/grin/issues/1290.
@hashmap many thanks for the middleware implementation.
Grin wallet generates an `api_secret` and `owner_api_secret` that can be used to query both API with Basic Auth.
This basic authentication can be disabled/enabled in the ```grin-server.toml/grin-wallet.toml``` but is enabled by default.
Feedbacks are welcome. 

I'll add a bit of documentation later today.


yeastplume:
:+1: Looks like a good start at wallet auth (and as a bonus doesn't look like it conflicts with T4 wallet changes)

quentinlesceller:
Great, I'm waiting for @hashmap per router middleware instead of per route middleware improvement to reduce the number of line of codes.

api/src/handlers.rs
```diff
@@ -878,7 +885,7 @@ pub fn build_router(
 	};
 
 	let mut router = Router::new();
-	// example how we can use midlleware
+
```
hashmap: 👍

src/bin/cmd/client.rs
```diff
@@ -117,7 +127,16 @@ pub fn unban_peer(config: &ServerConfig, peer_addr: &SocketAddr) {
 pub fn list_connected_peers(config: &ServerConfig) {
 	let mut e = term::stdout().unwrap();
 	let url = format!("http://{}/v1/peers/connected", config.api_http_addr);
-	match api::client::get::<Vec<p2p::PeerInfo>>(url.as_str()).map_err(|e| Error::API(e)) {
+	let peers_info: Result<Vec<p2p::PeerInfo>, api::Error>;
+	if config.api_basic_auth.is_some() && config.api_basic_auth.unwrap() {
```
hashmap: What if we hide this logic in the client API?
* add param `api_secret: Option<String>` to `api::client::get/post
* generate `api_basic_auth` in the client's request builder
* Set `AUTHORIZATION` header if `api_secret` param is not `None`

We may be able to remove all auth-specific methods (`build/get/post*_basic_auth`) in the client and don't need any specific handling here except blindly passing `config.api_basic_auth`

wallet/src/types.rs
```diff
@@ -52,6 +57,8 @@ impl Default for WalletConfig {
 			chain_type: Some(ChainTypes::Testnet3),
 			api_listen_interface: "127.0.0.1".to_string(),
 			api_listen_port: 13415,
+			owner_api_basic_auth: Some(true),
+			owner_api_secret: thread_rng().sample_iter(&Alphanumeric).take(20).collect(),
```
hashmap: Do I understand correctly we regenerate secret every time but it will be used only if it was explicitly persisted?

quentinlesceller: No this is generated for every new config file. On restart it will persist.

src/bin/cmd/client.rs
quentinlesceller: I did what you suggested. Feel free to take another look :).

servers/src/common/types.rs
```diff
@@ -113,6 +114,12 @@ pub struct ServerConfig {
 	/// Network address for the Rest API HTTP server.
 	pub api_http_addr: String,
 
+	/// Whether to enable basic access authentication on API
+	pub api_basic_auth: Option<bool>,
+
+	/// Secret for basic auth on Rest API HTTP server.
+	pub api_secret: String,
```
kargakis: Is this something that is going to get written down to the grin server config? I would urge this to be a path to the actual secret vs mixing credentials with config.

api/src/client.rs
```diff
@@ -27,27 +27,32 @@ use futures::future::{err, ok, Either};
...
 use rest::{Error, ErrorKind};
+use util::to_base64;
 
 pub type ClientResponseFuture<T> = Box<Future<Item = T, Error = Error> + Send>;
 
 /// Helper function to easily issue a HTTP GET request against a given URL that
 /// returns a JSON object. Handles request building, JSON deserialization and
 /// response code checking.
-pub fn get<'a, T>(url: &'a str) -> Result<T, Error>
+pub fn get<'a, T>(url: &'a str, api_basic_auth: bool, api_secret: &'a str) -> Result<T, Error>
```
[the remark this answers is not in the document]
quentinlesceller: Done.

servers/src/common/types.rs
quentinlesceller: You mean using environment variable for example or something else?

kargakis: Not an env variable but a path to a different file (similar to `db_root` or `log_file_path`) where you store the secret. The reason for doing this is to avoid treating the main config file as a secret blob (you need to if you start adding sensitive information to it).

quentinlesceller: Okay I see, let me see if I can add this here.

hashmap:
👍

quentinlesceller:
@hashmap  Just adding @kargakis idea on using another file should land in a few minutes here :)

quentinlesceller:
@hashmap @kargakis can you take a look. Now `grin` will generate a `.api_secret` file under grin directory. You can comment the line to disable basic auth in server and wallet configs. Basic Auth is enabled by default. Both APIs use the same secret.

kargakis:
Can't do a code review atm but the changes based on your comment sgtm
(sounds good to me)

PR #1573, opened 2018-09-21
 Replace rust-crypto with RustCrypto

quentinlesceller:
Replace the unmaintained rust-crypto crates with https://github.com/RustCrypto.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,6 +30,11 @@ version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "arrayref"
+version = "0.3.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "arrayvec"
 version = "0.3.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -125,6 +130,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "block-buffer"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "arrayref 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "byte-tools 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "bufstream"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


garyyu:
Indeed, it could be a problem with `rust-crypto` since its last update was 2 years ago and the last merged PR was 7th Sep. 2016.

But looks like there's no better choice so far.

The `RustCrypt` crate seems quite new, I'm afraid it could be still buggy, just my feeling.

I see `rust-bitcoin` project is also still using `rust-crypto`:

https://github.com/rust-bitcoin/rust-bitcoin/blob/master/Cargo.toml#L26

I'm not against this replacement, just want to inform these info 😃

garyyu:
I have a bad idea:  **we clone `rust-crypto` into `mimblewimble` and start maintain it by ourselves**, since this crate is quite important for Grin.

**Anybody second me?**

sesam:
@garyyu there should another branch in the new crypto lib repo where all the new stuff is, and that's likely what we should use. IIRC

garyyu:
@sesam 
> there should another branch in the new crypto lib repo where all the new stuff is

What do you mean `the new crypto lib repo`?

sesam:
Sorry, I just added confusion and realize I don't know what this is all
about. I do think there was a discussion about crypto libs and that we'll
need to swap something for something else, and that apoelstra recommended a
lib that is at least "nearly there" with some needed features, but I guess
this PR is not really related to that discussion.

Please, if someone can add 1-2 lines explaining the rationale for this
change. Good practice also so future grin-devs can go back through PR
history and understand. cc @quentinlesceller etc :)

quentinlesceller:
It's just that rust-crypto relies on old dependencies and is unmaintained since two years.
Whereas https://crates.io/crates/sha2 is two years old (so not that new) and relies on modern dependencies.

mcdallas:
> 
> 
> I have a bad idea: **we clone `rust-crypto` into `mimblewimble` and start maintain it by ourselves**, since this crate is quite important for Grin.
> 
> **Anybody second me?**

I think the idea is to migrate to https://github.com/rust-bitcoin/bitcoin_hashes once it's ready (according to [this](https://github.com/mimblewimble/grin/pull/1454#issuecomment-418294933))

antiochp:
@mcdallas good catch - I knew we'd discussed _something_ as a replacement but couldn't remember where/what it was...

garyyu:
So, we'd better open an issue and leave it there to remind us this https://github.com/rust-bitcoin/bitcoin_hashes .

But I feel this `bitcoin_hashes` is not at progress state, and the `once it's ready` could become a dream and wake for nothing?

Again, nobody second me to clone `rust-crypto` to mimblewimble?  😄

apoelstra:
I just need @TheBlueMatt (nudge, nudge) to merge the open PRs on bitcoin_hashes, then I think it will be done and in maintenance mode.

knocte:
> It's just that rust-crypto relies on old dependencies and is unmaintained since two years.
> Whereas https://crates.io/crates/sha2 is two years old (so not that new) and relies on modern dependencies.

This reasoning should be written in the commit message, not only in the PR history.

ignopeverell:
@garyyu that's a lot of risky code to maintain when we only use a fraction of it.

This PR is a net improvement, possibly with more improvements to come. So let's merge.

PR #1574, opened 2018-09-21
Fix wallet APIs launch command

hashmap:
We used to launch a thread for API server inside the wallet crate, now we do it inside api crate, so the cmd tool launches API and exit. This fix makes sure that command will wait for API thread.


hashmap:
Merging because it's a hot fix, feel free to review.

PR #1575, opened 2018-09-21
Replace println with warn

hashmap:


servers/src/common/types.rs
```diff
@@ -163,18 +163,21 @@ impl ServerConfig {
...
+					.p2p_config
+					.capabilities
+					.contains(p2p::Capabilities::FULL_HIST)
 			{
 				// if conflict, 'archive_mode' win
 				self.p2p_config
 					.capabilities
 					.toggle(p2p::Capabilities::FULL_HIST);
-				println!(
+				warn!(
```
garyyu: Hey 😺   Did you see the comment at line 178?  
```
"note: slog not available before config loaded, only print here."
```
I prefer to remove this warning message since it make user confusing.

hashmap:
@garyyu I tested and it worked for me, but then I found a bug in my test script, I was building `debug` profile but deploying `release` profile which was built from a different branch. So I removed the `println` because it makes users nervous.

servers/src/common/types.rs
```diff
@@ -163,19 +163,16 @@ impl ServerConfig {
 		// check [server.p2p_config.capabilities] with 'archive_mode' in [server]
 		if let Some(archive) = self.archive_mode {
 			// note: slog not available before config loaded, only print here.
-			if archive != self
-				.p2p_config
-				.capabilities
-				.contains(p2p::Capabilities::FULL_HIST)
+			if archive
+				!= self
```
garyyu: What's your `rustfmt` version?  the format is a little bit strange. 
Or what's your rust version?

PR #1577, opened 2018-09-22
test: add a script to run local tests for all sub-modules

garyyu:
This is the test script for running all tests on local machine.

Travis-CI tests are similar to this test script, except for dual platforms.

We can use this to reproduce Travis-CI failure locally, in case there's any true failure test cases.

**Usage**:
```
./run_tests.sh
```

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+#------------------------------------------------------------------
+# This is the test script for runing all tests on local machine.
+# Travis-CI tests are similar to this, except for dual platform.
+#
+# We can use this to reproduce Travis-CI failure locally, in case
+# there's any true failure test cases.
+#
+# Usage:
+#   ./run_tests.sh
+#------------------------------------------------------------------
+
+abort()
+{
+    printf >&2 "*** ABORTED ***\n"
+    printf "An error occurred. Exiting...\n" >&2
+
+    if [[ $ulimit < 512 ]]; then
+        sudo ulimit -n $ulimit
+        printf "Info: Your ulimit has been restored to $ulimit \n"
+    fi
+    exit 1
+}
+
+trap 'abort' 0
+
+set -e
+
+# Add your script below....
+# If an error occurs, the abort() function will be called.
+#----------------------------------------------------------
+# ===> Your script starts here
+
+ulimit=`ulimit -n 2>/dev/null`
+if [[ $ulimit < 512 ]]; then
+    printf "Info: Your ulimit $ulimit is not enough for this test. You need 'sudo ulimit -n 512' to continue this test... \n"
+    sudo ulimit -n 512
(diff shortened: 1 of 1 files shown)


ignopeverell:
Isn't this achieving the same as this command?

```
cargo test --all --release
```

garyyu:
It's almost same, but:
1. `rm -rf target/tmp` before each test on sub-module, some tests without this clean will fail.
2. `RUST_TEST_THREADS=1` for `server` test.
3. `ulimit -n 512` if it's too small, some tests could fail without this.

sesam:
I remember cases where `cargo test --all` would pass, but `cd`-ing into
each folder separately and running `cargo test` from inside a folder would
fail some tests. That could've been due to different toml config, but
output also at least used to be different between the two. I volunteer to
verify my claim and add some comments about what I find inside this .sh
script, if it gets merged :)

hashmap:
So basically `cargo test --all` is even more demanding, why not to run it locally always? It may be fine to do extra step to help travis cope with limited resources or other issues, but I'm not sure we need it on dev machines

garyyu:
@hashmap   Look like a mis-understand on this test script 😄 
I add it purely for the convenience of **local** test, not used in Travis-CI.

Because I find [these issues](https://github.com/mimblewimble/grin/pull/1577#issuecomment-424155362), they are easy to forget for us or unknown trap for new comer, and better to write down into a script file. Anyone need test locally will find this script useful.

hashmap:
Perhaps I wasn't clear, sorry.  I just run`cargo test --all` without any additional settings and it works fine. Those settings are needed to compensate some Travis issues/limitations.

garyyu:
Indeed, `cargo test --all --release` do it well now!

Not sure whether somebody fixed something for this?  or just because of cargo upgrading?
I remember before I can't pass the test, if I don't manually clean the `target/tmp` folder.

Now, only 2 cases left:  If the ulimit is small (default on some low end pc, like my mac air), for example 128 default, the test will report:
```
TxHashSetErr: Too many open files (os error 24) })', libcore/result.rs:945:5
test simulate_seeding ... FAILED
```
and
```
test simulate_block_propagation ... FAILED
Sep 25 22:32:11.777 INFO Starting server, genesis block: 569ed9e4
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Chain(Error { inner: 

TxHashSetErr: Too many open files (os error 24) })', libcore/result.rs:945:5
```
and then fail.

another case is `RUST_TEST_THREADS=1 for server test.`  , actually I'm not clear why Travis-CI has this config, I just copy it.

ignopeverell:
I think `RUST_TEST_THREADS=1` is also unnecessary now. We had it due to bad server shutdown cleanup and port reuse IIRC.

garyyu:
And I will give an optimization on socket connection numbers, then the last one `ulimit -n 512` is also not necessary.

So, close this PR.

PR #1595, opened 2018-09-26
Expose post_tx in OwnerAPI

### LINKED ISSUE #1592: Accept Slate JSON in Pool Push
Would be extremely useful to broadcast tx like that instead of using the `tx_hex` wrapper.

quentinlesceller:
Fixes #1592.
Just adding another useful endpoint in Owner API.
Add `wallet/owner/post_tx` where a user can post a slate.
Transaction can be fluffed by adding the `?fluff` param at the end of the query.

### DIFF
--- a/wallet/src/libwallet/controller.rs
+++ b/wallet/src/libwallet/controller.rs
@@ -340,6 +340,32 @@ where
 		}
 	}
 
+	fn post_tx(
+		&self,
+		req: Request<Body>,
+		api: APIOwner<T, C, K>,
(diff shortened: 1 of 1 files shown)


hashmap:
@antiochp it's `POST` git diff doesn't show it but the call is added in `handle_post_request`. According HTTP spec It's legal to pass a query string in `POST` request, but it's not very common.

quentinlesceller:
I'm just going to put this one on hold until #1597 is merged.

PR #1607, opened 2018-09-29
Conform auth check to rfc2616

kargakis:
According to rfc2616[1], the response from a server to a request with
bad credentials should be a 401 instead of a 403. Grin does not have
the concept of identities so it does not actually recognize a user
request with bad credentials.

[1] https://tools.ietf.org/html/rfc2616#section-10.4.2

@quentinlesceller @hashmap

### DIFF
--- a/api/src/auth.rs
+++ b/api/src/auth.rs
@@ -38,13 +38,10 @@ impl Handler for BasicAuthMiddleware {
 		req: Request<Body>,
 		mut handlers: Box<Iterator<Item = HandlerObj>>,
 	) -> ResponseFuture {
-		if req.headers().contains_key(AUTHORIZATION) {
-			if req.headers()[AUTHORIZATION] == self.api_basic_auth {
-				handlers.next().unwrap().call(req, handlers)
-			} else {
-				// Forbidden 403
-				forbidden_response()
-			}
+		if req.headers().contains_key(AUTHORIZATION)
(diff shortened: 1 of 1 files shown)


api/src/auth.rs
```diff
@@ -38,13 +38,10 @@ impl Handler for BasicAuthMiddleware {
...
 	) -> ResponseFuture {
-		if req.headers().contains_key(AUTHORIZATION) {
-			if req.headers()[AUTHORIZATION] == self.api_basic_auth {
-				handlers.next().unwrap().call(req, handlers)
-			} else {
-				// Forbidden 403
-				forbidden_response()
-			}
+		if req.headers().contains_key(AUTHORIZATION)
+			&& req.headers()[AUTHORIZATION] == self.api_basic_auth
```
kargakis: I think I just changed the behavior here and always require the auth header to be set. Will open a follow-up to fix that

kargakis: Nevermind, this is fine.

quentinlesceller:
Not sure removing this behaviour was a good idea. Currently we use the user `grin` but we could have changed the rules later on.

kargakis:
Even if you have a user, bad credentials is a 401 not a 403. Please read
the linked doc.

quentinlesceller:
Yeah sorry I removed my comment after (but it was still emailed).

PR #1617, opened 2018-10-01
fix: mutex dead lock between difficulty_iter and txhashset_write

### LINKED ISSUE #1611: regression: fast sync from scratch freeze after txhashset validated
sync from scratch on the latest master, TUI freezing after validating the txhashset, even `q` doesn't work.

garyyu:
This PR fix #1611


antiochp:
:+1:

Issue #1616, opened 2018-10-01
long processing on store will cause TUI freeze


garyyu:
During the analysis of https://github.com/mimblewimble/grin/issues/1611, we find that the TUI thread periodically call `get_server_stats` which call `difficulty_iter`, and there's a `store.batch` lock in `difficulty_iter`, that means if there's a long processing on the `store.batch`, for example the txhashset validation in fast sync procedure, the TUI will wait a long time before successfully get the lock.

We need an improvement on this to avoid TUI freeze too long.

PR #1621, opened 2018-10-01
fix: remove unused DiffStats from get_server_stats

garyyu:
Improvement for https://github.com/mimblewimble/grin/issues/1616.

TUI thread periodically call `get_server_stats` which call `difficulty_iter`, and there's a `store.batch` lock in `difficulty_iter`, which cause the TUI may wait a long time before successfully get the lock, looks like a frozen TUI.

### DIFF
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -102,7 +102,7 @@ pub struct StratumStats {
 }
 
 /// Stats on the last WINDOW blocks and the difficulty calculation
-#[derive(Clone)]
+#[derive(Clone, Default)]
 pub struct DiffStats {
 	/// latest height
 	pub height: u64,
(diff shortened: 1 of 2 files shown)


antiochp:
Do we not actually use this anywhere in the TUI?

garyyu:
No, TUI don't use this info.
But I just find the TUI still freeze when txhashset validating, after removing this DiffStats. I'm investigating what else block it.

garyyu:
Oops~  I'm testing on another working branch:(   
I switch back to this fixing branch and confirm this fix works well. It's ready for merge if no other review comments.

yeastplume:
The TUI does indeed use this info, under mining/difficulty to show the stats for the previous difficulty adjustment window

yeastplume:
I'm sorry, but please don't merge this. That's actually a fairly useful info screen, and we should be aiming for a proper fix for the root issue rather than ripping things out.

PR #1622, opened 2018-10-01
fix: no need switching to HeaderSync in FastSync state

garyyu:
A bit of improvement on sync state transition. 

In case of fast sync, during the following 4 states:
- TxHashsetDownload
- TxHashsetSetup
- TxHashsetValidation
- TxHashsetSave

we don't need the switching back to HeaderSync.

The only state which need switching back to HeaderSync (and again and again) is **BodySync**.

### DIFF
--- a/servers/src/grin/sync.rs
+++ b/servers/src/grin/sync.rs
@@ -199,7 +199,7 @@ fn do_header_sync(
 ) {
 	let status = sync_state.status();
 
-	let doit = match status {
+	let enable_header_sync = match status {
 		SyncStatus::BodySync { .. } | SyncStatus::HeaderSync { .. } => true,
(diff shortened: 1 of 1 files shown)


servers/src/grin/sync.rs
```diff
@@ -199,8 +199,8 @@ fn do_header_sync(
 ) {
 	let status = sync_state.status();
 
-	let update_sync_state = match status {
-		SyncStatus::TxHashsetDownload => false,
+	let doit = match status {
```
ignopeverell: A nitpick but something like `enable_header_sync` would be a lot more explicit.

garyyu: forgive my english~  updated.

PR #1630, opened 2018-10-01
Demo lean miner, minor PoW improvements

ignopeverell:
The main part of this PR introduces the simplest, most idiomatic implementation of a Cuckatoo lean miner I could come up with. It's mostly for education purpose (for myself and everyone interested) and fits in a single page. It does provide some speedup to the dev miner but I'm not sure it's worth using. I think it's still worth having to illustrate generally how Cuck(at)oo optimizations can work.

This lead to some minor improvements and refactoring in the Cuckatoo implementation.


antiochp:
This is great - I need to set some time aside and go over this in more detail (so as to better understand what mining in Grin actually involves...)

> Cuck(at)oo optimizations

Cuck(@)oo maybe?

core/src/global.rs
```diff
@@ -132,7 +131,7 @@ where
 	// Perform whatever tests, configuration etc are needed to determine desired context + edge size
 	// + params
 	// Hardcode to regular cuckoo for now
-	CuckooContext::<T>::new(edge_bits, proof_size, easiness_pct, max_sols)
+	CuckooContext::<T>::new(edge_bits, proof_size, EASINESS, max_sols)
```
yeastplume: Sure, I gues Easiness is the only thing that's not likely to change between pows

core/src/pow/common.rs
```diff
@@ -151,3 +128,80 @@ macro_rules! to_edge {
...
+		let easiness = if cuckatoo {
+			to_u64!(easiness_pct) * num_nodes / 100
+		} else {
+			to_u64!(easiness_pct) * num_edges / 100
+		};
+		let edge_mask = if cuckatoo {
+			to_edge!(num_edges - 1)
+		} else {
+			to_edge!(num_edges / 2 - 1)
+		};
```
yeastplume: We're probably better off consolidating 'num_edges' between cuckoo and cuckatoo (I think cuckatoo29 = cuckoo 30)?

yeastplume:
Yep, Params struct makes sense, good to have a lean option in there as well.

yeastplume:
Need to build on top of this now, so just going to merge then fix tests with another PR.

ignopeverell:
@yeastplume sorry for the lateness in fixing those, was just about to. And thanks for the test fixes!

PR #1640, opened 2018-10-02
[DNM] do not maintain head in ctx (go back to db via batch for this)

antiochp:
[Do not merge without discussion]

I think this eliminates a source of potentially stale data during block and header processing.
Need to do more testing to be confident this actually works as expected.

But like this we go look in the db (accessible via the batch) and we do not keep a potentially stale `head` or `header_head` in the ctx passed into `process_block()`.
So we should be less likely to process a block in one thread and in a separate thread go sync block headers because we got out of date and revert the `header_head` to some previous (and now incorrect) value.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -59,8 +59,8 @@ pub struct BlockContext {
 
 // Check if this block is the next block *immediately*
 // after our current chain head.
-fn is_next_block(header: &BlockHeader, ctx: &mut BlockContext) -> bool {
-	header.previous == ctx.head.last_block_h
+fn is_next_block(header: &BlockHeader, head: &Tip) -> bool {
+	header.previous == head.last_block_h
 }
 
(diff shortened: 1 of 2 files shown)


chain/src/pipe.rs
```diff
@@ -148,7 +149,13 @@ pub fn process_block(
 		// First we rewind the txhashset extension if necessary
 		// to put it into a consistent state for validating the block.
 		// We can skip this step if the previous header is the latest header we saw.
-		if is_next_block(&b.header, ctx) {
+
+		//
+		// TODO - Is this safe to do? Can we just go look in the db via the batch?
+		//
+		let head = extension.batch.head()?;
```
ignopeverell: Yes, I think it's fine now that the lock is taken earlier. What matter the most is that another can't update the head without us knowing because we're in different batches.

antiochp:
Closing this for #1650.

Issue #1632, opened 2018-10-02
Node Stuck in Sync Loop


quentinlesceller:
Attaching the grin log, no specific error. Running on 4d2cbe6596462e609a528182e57a965ca4d5aa8d.
[grin.log](https://github.com/mimblewimble/grin/files/2436090/grin.log)

ignopeverell:
Needs more digging but this looks like we enter body sync properly but fail to request any block.

ignopeverell:
Actually, it's more that there's nothing to request because we're not requesting headers either. It looks like your node has very few peers (6 total, only 3 at the head) so you'd likely benefit from cleaning up your peer db regardless.

Would you have logs around when this particular node went out of sync?

quentinlesceller:
Going to clean my peers db. Here is the previous 20 minutes
https://pastebin.com/JKMUhCdj

garyyu:
Hey, we don't have `peers db` anymore! don't know when and who merged into chain db. So, you have no way to `clean my peers db` 😃 
But I think it's indeed needed a way to clean the peers, either by an independent db, or by a API.

garyyu:
>  node has very few peers (6 total, only 3 at the head) 

Perhaps we need adjust the default configuration from `peer_min_preferred_count = 8` to `peer_min_preferred_count = 16`?  btw, I always change it to 16.

ignopeverell:
From his logs, he can't even get 8 so increasing it wouldn't help.

```
Oct 01 22:34:34.702 INFO sync: total_difficulty 49733148, peer_difficulty 49735111, threshold 1934 (last 5 blocks), enabling sync
Oct 01 22:34:34.703 DEBG sync: initial transition to HeaderSync. sync_head: e733c350 at 119872, reset to: c22f3857 at 119876
Oct 01 22:34:34.706 DEBG sync_state: sync_status: NoSync -> HeaderSync { current_height: 119876, highest_height: 119881 }
Oct 01 22:34:34.706 DEBG sync: locator heights : [119876, 119874, 119870, 119862, 119846, 119814, 119750, 119622, 119366, 118854, 117830, 115782, 111686, 103494, 87110, 54342, 0]
```

Following which the node quickly catch up to 119881. Then again:

```
Oct 01 22:39:24.710 INFO sync: total_difficulty 49735111, peer_difficulty 49737151, threshold 1963 (last 5 blocks), enabling sync
[...]

PR #1643, opened 2018-10-03
Don't push mined blocks to the chain during sync

ignopeverell:
Related to #1632

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -502,7 +502,7 @@ impl StratumServer {
 			return Err(serde_json::to_value(e).unwrap());
 		}
 		// If the difficulty is high enough, submit it (which also validates it)
-		if share_difficulty >= self.current_difficulty {
+		if share_difficulty >= self.current_difficulty && !self.sync_state.is_syncing() {
 			// This is a full solution, submit it to the network
 			let res = self.chain.process_block(b.clone(), chain::Options::MINE);
 			if let Err(e) = res {


servers/src/mining/stratumserver.rs
```diff
@@ -502,7 +502,7 @@ impl StratumServer {
 			return Err(serde_json::to_value(e).unwrap());
 		}
 		// If the difficulty is high enough, submit it (which also validates it)
-		if share_difficulty >= self.current_difficulty {
+		if share_difficulty >= self.current_difficulty && !self.sync_state.is_syncing() {
```
garyyu: Could you please hold this PR for a while? 😄   to give us some time to address that 'impossible' happening found in #1632, and confirm that problem is well fixed.
I'm afraid if this fix is merged, then not easy to reproduce that root cause.

BTW, shall we log a warning message here? to remind the reason of this ignored 'submit' block.

ignopeverell: Good point, we can definitely wait a few days for this. But I think it's already pretty obvious from the logs when you mine a block while syncing is enabled, no?

garyyu: > But I think it's already pretty obvious from the logs when you mine a block while syncing is enabled

Yes, fully agree. 

I mean the fix in this PR is good and necessary, but the root cause for that strange **impossible** happening (https://github.com/mimblewimble/grin/issues/1632#issuecomment-426188635 - `update_header_head` return `None` but `header_head` has been updated!) could be a major risk and deserve a deep checking.

garyyu: I find it's quite difficult to reproduce that special scenario: a bunch of Headers coming and in the same time mined a new block. So giving up to try reproducing problem, and please go on to merge this fix :)

ignopeverell:
@antiochp not sure what it'd be, mining on top of an old block is unlikely to be productive and in itself could be seen as detrimental to the chain.

antiochp:
Just to clarify. My question was around a peer lying about a fork.

* Miner is fully sync'd
* Miner happily mining on latest block
* peer advertises new fork with greater cumulative difficulty
* Miner kicks into sync mode and stops mining
* turns out peer was lying about the fork

ignopeverell:
@antiochp very good point, I filed #1654 to think about that some more.

antiochp:
The whole header MMR can of worms has got me thinking about sync a lot recently...

ignopeverell:
Some more major code removal in the works?

PR #1644, opened 2018-10-03
Don't stall sync when header chain regresses

ignopeverell:
This should normally never happen but this is meant to catch
unforeseen edge cases or block acceptance bugs.

Related to #1632

### DIFF
--- a/servers/src/grin/sync.rs
+++ b/servers/src/grin/sync.rs
@@ -743,7 +743,7 @@ impl SyncInfo {
 		let all_headers_received =
 			header_head.height >= prev_height + (p2p::MAX_BLOCK_HEADERS as u64) - 4;
 		// no headers processed and we're past timeout, need to ask for more
-		let stalling = header_head.height == latest_height && now > timeout;
+		let stalling = header_head.height <= latest_height && now > timeout;
 
 		// always enable header sync on initial state transition from NoSync / Initial
 		let force_sync = match sync_state.status() {


servers/src/grin/sync.rs
```diff
@@ -743,7 +743,7 @@ impl SyncInfo {
 		let all_headers_received =
 			header_head.height >= prev_height + (p2p::MAX_BLOCK_HEADERS as u64) - 4;
 		// no headers processed and we're past timeout, need to ask for more
-		let stalling = header_head.height == latest_height && now > timeout;
+		let stalling = header_head.height <= latest_height && now > timeout;
```
garyyu: Thought it before, and I see you already have a logic to handle the case of header chain change: both `latest_height` and `timeout` will be updated if `header_head.height` is changed.  So, perhaps it's better to keep this `==` and modify this?

```
			// resetting the timeout as long as we progress
			if header_head.height > latest_height {
						<<<< change this `>` as `!=` 
				self.prev_header_sync =
					(now + Duration::seconds(2), header_head.height, prev_height);
			}
			false
```

ignopeverell: I may be biased but this is harder to read for me. If the header has gotten older than last time, we're even more stalled so that makes sense. But going through an additional update of `prev_header_sync` so we can have updated values the next time, that will then trigger `stalling` is less immediate. No? That being said, doing both may not be a bad thing.

garyyu: I'm thinking `less immediate` could be better. 

For example, in case we have sent the Headers request, and a bunch of Headers are coming **on the way**, and at this time our `header_head` is updated (for example in this case by a new mined block, or by a new received block, or any other reason), if you can just wait 1 second for example, the requested Headers arrived and will correct our `header_head`. So, you see, sometimes a less immediate trigger could do things better:)

ignopeverell: That's what the timeout condition of `stalling` is meant to address, and by updating that time for any update and not just forward progression, we'd be somewhat undermining that timeout.

Issue #1586, opened 2018-09-25
0 healthy peers


garyyu:
> Blade Doyle @bladedoyle Sep 24 23:04
How can I reset my peers list? I used to be able to delete the peers file but I dont see that anymore
Sep 24 15:03:35.417 DEBG monitor_peers: on 0.0.0.0:13414, 0 connected (0 most_work). all 117 = 0 healthy + 0 banned + 117 defunct
I have no peers that will talk to me ;(

> Gary Yu @garyyu 06:46
@bladedoyle I’m interested in the reason which cause this 0 healthy peers. Could you share me the logs?
btw, there’s no way to reset peers, you’d better submit an issue to request this feature 

> Blade Doyle @bladedoyle 13:28
@garyyu I can share the log but its 18M compressed. I think you should be able to get it here:
https://drive.google.com/file/d/1GLVmga9jz1EJRVJV-MXdRkjVSpNSMwp5/view?usp=sharing

```
Sep 23 17:43:42.667 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
Sep 23 17:44:02.672 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
Sep 23 17:44:22.681 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
Sep 23 17:44:42.685 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
Sep 23 17:45:02.691 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
Sep 23 17:45:22.696 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
Sep 23 17:45:42.701 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
Sep 23 17:46:02.710 DEBG monitor_peers: on 0.0.0.0:13414, 12 connected (10 most_work). all 117 = 67 healthy + 0 banned + 50 defunct
[...]

PR #1646, opened 2018-10-03
Resurrect a random peer on every monitor

ignopeverell:
Picks a random defunct peer and marks it healthy again anytime
we're querying peers to find more. Over enough time, any peer will
see another as defunct, leading to peer list atrophy. This
allows us random retries.

In addition, we need to cleanup peers after a while, we quickly
accumulate hundreds of dead peers over time. This should be
tracked by a different issue however.

Related: #1632 and #1586

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -18,6 +18,7 @@
 
 use chrono::prelude::Utc;
 use chrono::{Duration, MIN_DATE};
+use rand::{thread_rng, Rng};
 use std::net::{SocketAddr, ToSocketAddrs};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{mpsc, Arc};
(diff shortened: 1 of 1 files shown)


servers/src/grin/seed.rs
```diff
@@ -176,6 +177,13 @@ fn monitor_peers(
 		None => debug!(LOGGER, "monitor_peers: no preferred peers"),
 	}
 
+	// take a random defunct peer and mark it healthy: over a long period any
```
garyyu: How about just retry connecting, instead of indirectly change the peer state.

i.e.  move this paragraph to the end of this function, and 
```
	// try connecting to a random defunct peer: over a long period any peer will see another as defunct eventually,
	// gives us a chance to retry
	if defuncts.len() > 0 {
		thread_rng().shuffle(&mut defuncts);
		tx.send(defuncts[0].addr).unwrap();
	}
```

ignopeverell: I found that to be perhaps less elegant because you'd also need to check whether the maximum peer counts is  exceeded. And if you're at max peer count would never try any resurrection (because it'd be ditched from the channel), which could still be damaging longer term. Just a simple update seemed to mesh better with existing logic.

PR #1650, opened 2018-10-03
maintain head and header_head on the chain, pass them wrapped in arcs to pipe

antiochp:
Refactor `chain` and `pipe` to make pipe aware of "current" chain head.
This allows one thread (say a peer processing a block) to be aware of another thread (say header sync) updating `header_head`.

We now only ever update `head` if we are sure the total work has increased.

And we only ever update `header_head` if total work has increased on the header chain.

This should prevent any situations during sync where we update to a older (less work) `head` or `header_head`.

* Maintain both `head` and `header_head` on chain (wrapped in arcs)
* Pass `chain.head` and `chain.header_head` into pipe in the ctx (wrapped in arcs)
* `sync_block_headers` now returns the updated `header_head` tip if we see more total work
* use a `RwLock` and not a `Mutex` around `head` so code is more explicit around when exactly we mutate the `head` (vs. just reading the current `head`)

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -18,7 +18,7 @@
 use std::collections::HashMap;
 use std::fs::File;
 use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
 
 use lmdb;
@@ -34,7 +34,6 @@ use error::{Error, ErrorKind};
 use grin_store::Error::NotFoundErr;
 use pipe;
 use store;
-use store::Batch;
 use txhashset;
 use types::{ChainAdapter, NoStatus, Options, Tip, TxHashsetWriteStatus};
 use util::secp::pedersen::{Commitment, RangeProof};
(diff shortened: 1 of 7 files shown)


antiochp:
Also added "header chain height" and "header cumulative difficulty" to TUI so we can more easily see what what's going on during sync -

screen shot 2018-10-03 at 4 29 25 pm

ignopeverell:
Thinking more about this: why? It makes things more complicated. Why not abandon the head reference altogether, rely only on the txhashset lock and the batch, and just not update the header head in db on less work?

antiochp:
> Why not abandon the head reference altogether, rely only on the txhashset lock and the batch, and just not update the header head in db on less work?

Right now this is not trivial to do because we don't use the txhashset during header processing.
Header sync in particular is running blind and has no way of knowing if other threads are updating the db at the same time.

* A starts a batch and reads the value of `head` from the db (header sync, no txhashset to lock)
* B locks the txhashset, starts a batch, updates `head` after processing a block
* A still sees the old value of `head` (and things potentially regress backwards)

So I think there needs to be _something_ shared between the various threads tracking some kind of state (for example a struct representing all the current heads).

Let me think a bit more about combining these all into a single "chain lock". Maybe we can pass a ref to the `chain` instance itself into `pipe`?

I don't think the overhead will be too great during normal block propagation. We moved the txhashet lock earlier and earlier to avoid multiple threads processing the same block from different peers in parallel anyway (its all wasted work, even if we _can_ do some of it in parallel).

antiochp:
Ok - closing this PR, will replace with another shortly.
I think I've got it working in a simpler way, similar to #1640 but with the header_head changes to prevent it from regressing (only update on more work).

PR #1655, opened 2018-10-03
API Documentation: Node API

quentinlesceller:
Add a skeleton for the APIs documentation and introduce a documentation for the Node API.
Easier to read here: https://github.com/quentinlesceller/grin/tree/realdocapi/doc/api.


ignopeverell:
Very nice!

PR #1665, opened 2018-10-05
Support pem TLS certificates

hashmap:
Mostly to support let's encrypt. It requires to switch from native-tls and friends to rustls and friends, which perhaps is a good thing per se, rustls looks more modern and for sure more Rusty.
Alternative would be manually convert pkcs12 certificates to pem, which requires openssl tools to be installed and make transparent integration with let's encrypt much harder (this is out of the scope for now, perhaps in near future).

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -17,11 +17,12 @@ serde_json = "1"
 slog = { version = "~2.3", features = ["max_level_trace", "release_max_level_trace"] }
 tokio = "0.1.7"
 tokio-core = "0.1.17"
-tokio-tls = "0.2"
-native-tls = "0.2"
+tokio-tcp = "0.1" 
+tokio-rustls = "0.7"
 http = "0.1.5"
-hyper-tls = "0.3"
+hyper-rustls = "0.14"
 futures = "0.1.21"
+rustls = "0.13"
 url = "1.7.0"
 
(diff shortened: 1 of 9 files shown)


hashmap:
I found we did not use `ring`, `rustls` built on top of it, any objections to have it as a dependency?

api/Cargo.toml
```diff
@@ -17,11 +17,12 @@ serde_json = "1"
 slog = { version = "~2.3", features = ["max_level_trace", "release_max_level_trace"] }
 tokio = "0.1.7"
 tokio-core = "0.1.17"
-tokio-tls = "0.2"
-native-tls = "0.2"
```
kargakis: Does this mean you drop the openssl dependency? If so, can you update the Dockerfile?

hashmap: Thanks for spoting it, I checked Cargo.tomsl, it seems we don't use `openssl` in other places, will send a pr for dockerfile

hashmap:
No, when I removed `openssl` from Dockerfile it  failed with:
 ````
Step 12/14 : RUN grin server config &&     sed -i -e 's/run_tui = true/run_tui = false/' grin-server.toml
 ---> Running in 3715c5a8d161
grin: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```

PR #1666, opened 2018-10-05
Reduce the scope of the chain new block lock

ignopeverell:
When processing a new block, attemps to make the scope of the
corresponding `txhashet` lock as small as possible. The idea is
to avoid bad interactions with orphan locks or block post
processing (i.e. adapter events, etc)


antiochp:
Yeah...
`process_block_single` calls `adapter.block_accepted`, which in turn takes a write lock on the `txpool` _and_ involves locks on peers as it broadcasts out to them.

PR #1669, opened 2018-10-06
optimization of peer state write lock

garyyu:
- Peer `check_connection()` function is being heavily called on peer status checking everywhere.
- Peer state is a slow changing element
- Currently we have the expensive RwLock `write` lock for each `check_connection()` call, it's better to optimize it as `read` lock and only call `write` lock when needed.

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -326,14 +326,19 @@ impl Peer {
 	}
 
 	fn check_connection(&self) -> bool {
-		let state = self.state.read().unwrap();
 		match self.connection.as_ref().unwrap().error_channel.try_recv() {
 			Ok(Error::Serialization(e)) => {
-				if State::Banned != *state {
-					drop(state);
+				let need_stop = {
 					let mut state = self.state.write().unwrap();
-					*state = State::Disconnected;
-					info!(
+					if State::Banned != *state {
+						*state = State::Disconnected;
+						true
+					} else {
+						false
+					}
+				};
+				if need_stop {
+					debug!(
 						LOGGER,
 						"Client {} corrupted, will disconnect ({:?}).", self.info.addr, e
 					);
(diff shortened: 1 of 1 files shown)


antiochp:
@garyyu think I made almost exactly the same changes in #1671.
I think you need to take the read lock and the write lock separately. Don't think you can nest them like that.

garyyu:
Just saw the mistake of `read` then `write`, it's surprised it can pass CI test 😄 
will correct it right now.

p2p/src/peer.rs
```diff
@@ -340,9 +342,13 @@ impl Peer {
...
 			Ok(e) => {
-				*state = State::Disconnected;
-				debug!(LOGGER, "Client {} connection lost: {:?}", self.info.addr, e);
-				self.stop();
+				if State::Disconnected != *state {
+					drop(state);
+					let mut state = self.state.write().unwrap();
+					*state = State::Disconnected;
+					debug!(LOGGER, "Client {} connection lost: {:?}", self.info.addr, e);
+					self.stop();
```
antiochp: Might be worth limiting the scope of the write lock to simply updating the state.
And keeping `stop()` outside the write lock.

p2p/src/peer.rs
```diff
@@ -326,10 +326,12 @@ impl Peer {
 	}
 
 	fn check_connection(&self) -> bool {
-		let mut state = self.state.write().unwrap();
+		let state = self.state.read().unwrap();
 		match self.connection.as_ref().unwrap().error_channel.try_recv() {
 			Ok(Error::Serialization(e)) => {
 				if State::Banned != *state {
+					drop(state);
```
hashmap: Strictly speaking with this change `if` on the line 332 doesn't make sense, we unlock `self.state` here, it may be changed in a different thread, so we update it in any case, even if state is `Banned` already.

p2p/src/peer.rs
```diff
@@ -340,9 +342,13 @@ impl Peer {
 				false
 			}
 			Ok(e) => {
-				*state = State::Disconnected;
-				debug!(LOGGER, "Client {} connection lost: {:?}", self.info.addr, e);
-				self.stop();
+				if State::Disconnected != *state {
+					drop(state);
+					let mut state = self.state.write().unwrap();
```
hashmap: same here

PR #1670, opened 2018-10-06
refactor: remove thread p2p-moniter, move to the connect_and_monitor

garyyu:
The thread `p2p-monitor` in `p2p/src/serv.rs` only do the periodic ping on the connected peers, for every 10 seconds.

We can simply it and move this peer monitor into one place:  seed.rs  `connect_and_monitor()`.

### DIFF
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -90,21 +90,6 @@ impl Server {
 	/// Starts a new TCP server and listen to incoming connections. This is a
 	/// blocking call until the TCP server stops.
 	pub fn listen(&self) -> Result<(), Error> {
-		// start peer monitoring thread
-		let peers_inner = self.peers.clone();
-		let stop = self.stop.clone();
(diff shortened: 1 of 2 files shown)


antiochp:
I'm not sure this makes sense. Lets keep `seed.rs` for the initial seeding. 
I don't think we should put more generic peer monitoring in there.

hashmap:
The question is do we need to support 10 seconds interval or not, because moving it into one thread gives as much weaker guarantee, like "ping not sooner than in 10 seconds", because we don't know how long we will be doing other work.

garyyu:
thanks for the review~

@antiochp I guess your eventual meaning is OK, right?

@hashmap I think when peers `connect_and_monitor` is busy and even can't catch the 10s interval loop, the new `Ping` messages sending (in another thread) will make things even worse.

And the Ping detection in 10s interval is not needed to be accurate, for example even we send a Ping after 15s, that should not be a problem.

PR #1671, opened 2018-10-06
[DNM] smaller scope on write lock around peer state

antiochp:
* smaller scope on write lock around peer state (disconnect outside the lock now)
* read lock on non mut peer (where write lock not necessary)


antiochp:
Ha! Just saw #1669 and #1670...

antiochp:
Closing. Dupe of #1669.

PR #1672, opened 2018-10-06
[DNM] replace std lib sync primitives with parking_lot

garyyu:
parking_lot: https://crates.io/crates/parking_lot
- provides implementations of Mutex, RwLock, Condvar and Once
- **smaller, faster and more flexible** than those in the Rust standard library
- **a experimental deadlock detector** that works for Mutex, RwLock and ReentrantMutex

[copy from https://crates.io/crates/parking_lot]:
When tested on x86_64 Linux, `parking_lot::Mutex` was found to be **1.5x faster** than std::sync::Mutex when uncontended, and **up to 5x faster** when contended from multiple threads. The numbers for `RwLock` vary depending on the number of reader and writer threads, but are almost **always faster** than the standard library `RwLock`, and even **up to 50x faster in some cases**.

- [x] replace all `RwLock` with `parking_lot::RwLock`
- [x] replace all `Mutex` with `parking_lot::Mutex`
- [x] add deadlock detector

To be discussed, to be reviewed, and to be tested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -542,6 +542,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "fixedbitset"
+version = "0.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "flate2"
 version = "1.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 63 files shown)


garyyu:
In case anybody want to have a code review, the main feature added is here:  https://github.com/mimblewimble/grin/blob/d188eeb941b64b3119486d9f0d54cbe99e65daae/servers/src/grin/server.rs#L284-L305

Other changes are just global replacements for RwLock, by bash script.

antiochp:
This looks interesting but :-1: on doing this a few days before T4...

garyyu:
@antiochp  I thought we agreed to complete the fix of syncing problem (which is mainly blocked by dead lock issue) before going to T4. no?

tromp:
btw, my recent grin server run being hung at height 125078 was related to deadlock:

http://tromp.github.io/grin-server.log

Would this code change have any effect on that behaviour?

antiochp:
The "deadlock logger" looks useful if we can use it to reproduce the problem locally (on a branch). 


I think `parking_lot` is definitely worth taking a closer look at. Read a little about it and I know we have it as a dependency on some of our other libs today.

But I'm not 100% sure we should be considering swapping out `std::sync` on master pre T4.

garyyu:
@antiochp  why 👎 me here? I'm trying to help to solve dead lock problem in another way.

garyyu:
@tromp  your attached log has an explicit deadlock:
```
55114 thread 'peer' panicked at 'rwlock read lock would result in deadlock': libstd/sys/unix/rwlock.rs:59stack backtrace:
```
And several `PoisonError`.

but that deadlock could already been fixed by several consequent refactor PRs, since your version is https://github.com/mimblewimble/grin/commit/bcf41438dcd40580c1d865691967710f31d60702 two days ago, and now it's not easy to address which PR fixed which issue.

garyyu:
I close this one for now and will give another clean PR for T4 after.

Issue #1675, opened 2018-10-07
Travis-CI Cache Problem?


garyyu:
Recently I draft a PR with a build error or with an obvious dead lock problem, but Travis-CI test passed!

I suspect there must be a problem caused by Travis-CI cargo cache, perhaps imported by Travis-CI Speed Optimization PR:  https://github.com/mimblewimble/grin/pull/1647

I have to revert the PR #1647  and try to address the root cause of this Travis-CI problem later.

garyyu:
https://travis-ci.org/mimblewimble/grin/jobs/438144451
```
For more information about this error, try `rustc --explain E0425`.
error: Could not compile `grin_p2p`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```
Looks not like a cache issue, could be the bash script error code acquiring issue.
Anyway, I will give more tests later before re-import the speed optimization.

garyyu:
closed by #1687

PR #1676, opened 2018-10-07
revert travis-ci speed optimization for cache problem

garyyu:
As found in https://github.com/mimblewimble/grin/issues/1675, the current Travis-CI seems have a major problem on the cache, even the building error PR can pass the Travis-CI test.

I have to revert the speed optimization of #1647 for the moment, and will investigate the cache problem later.

Sorry for this!  It's also an unexpected surprise for me:(

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ git:
 
 os:
   - linux
-#  - osx
+  - osx
 
 dist: trusty
 sudo: required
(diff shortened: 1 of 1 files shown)


garyyu:
Note: I know there's a revert button on merged PR.  I give manual revert just because that button can't work and warn something like "revert doesn't work because of some conflict".

PR #1678, opened 2018-10-07
Skip sigs txhashset validation

antiochp:
The *only* time we need to do a "full" (expensive) chain validation is after receiving a txhashset.zip from a peer during fast sync.

When we run compaction we do not need to re-verify kernel signatures or rangeproofs (as we are not adding any new data).

We were already skipping rangeproof verification during compaction.

Skipping kernel signature verification shrinks chain compaction down to approx 3s - 

```
time curl -X POST 127.0.0.1:13413/v1/chain/compact
curl -X POST 127.0.0.1:13413/v1/chain/compact  0.01s user 0.01s system 0% cpu 3.192 total
```

And "fast validation" on its own takes 2.5s - 

```
time curl 127.0.0.1:13413/v1/chain/validate 
curl 127.0.0.1:13413/v1/chain/validate  0.01s user 0.01s system 0% cpu 2.528 total
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -462,7 +462,7 @@ impl Chain {
 	}
 
 	/// Validate the current chain state.
-	pub fn validate(&self, skip_rproofs: bool) -> Result<(), Error> {
+	pub fn validate(&self, fast_validation: bool) -> Result<(), Error> {
 		let header = self.store.head_header()?;
 
 		// Lets just treat an "empty" node that just got started up as valid.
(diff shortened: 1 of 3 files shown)


chain/src/chain.rs
```diff
@@ -246,54 +246,52 @@ impl Chain {
 
 				Ok(head)
 			}
-			Err(e) => {
```
antiochp: Ignore changes in this file. This is all just rustfmt deciding to do it now, not sure why...

PR #1689, opened 2018-10-08
small improvement on the capabilities config comments

garyyu:
> John Tromp @tromp 16:23
it says Bit flags for FULL_NODE and then bits = 6
#7 = Bit flags for FULL_NODE, this structure needs to be changed
#internally to make it more configurable
bits = 6
---

> Gary Yu @garyyu 16:24
it says 7= … FULL_NODE. Confused?
---

> John Tromp @tromp 16:25
6 is for FAST_SYNC_NODE
oh i see why im confused
the 7 is not separate from #
---

> John Tromp @tromp 16:27
i read it as a garbled # bit flags for FULL_NODE
---

> Gary Yu @garyyu 16:28
Ok, will give a minor change on this comment.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -210,8 +210,9 @@ fn comments() -> HashMap<String, String> {
 
 	retval.insert(
 		"[server.p2p_config.capabilities]".to_string(),
-		"#7 = Bit flags for FULL_NODE, this structure needs to be changed
-#internally to make it more configurable
+		"# 7 = Bit flags for FULL_NODE
+# 6 = Bit flags for FAST_SYNC_NODE
+#This structure needs to be changed internally, to make it more configurable
 ".to_string(),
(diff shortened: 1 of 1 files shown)


antiochp:
Do we still have an `archival_mode` flag in the config or is this fully replaced with this value `6`?

jaspervdm:
According to [comments.rs:88](https://github.com/mimblewimble/grin/blob/fc0c4a47298ed1e96b603339b9796aba112b3e84/config/src/comments.rs#L88) it still exists

jaspervdm:
So if `archival_mode` takes priority over `bits`, why is `bits` configurable?

garyyu:
good point!  I guess that's why there's a comment there 😄  a remaining todo item.
```
#This structure needs to be changed internally, to make it more configurable
```

Issue #1551, opened 2018-09-19
BorrowError on `header_received`


ignopeverell:
Multiple `BorrowError`s occurred when trying to process a new block header, resulting in my node not processing that block. It then fell out far enough to start sync, which triggered #1524.

```
thread 'peer' panicked at 'already mutably borrowed: BorrowError': libcore/result.rs:945stack backtrace:
   0:     0x563c3b559fbc - backtrace::backtrace::trace::h802abb1eb78e0eed
   1:     0x563c3b559d72 - <backtrace::capture::Backtrace as core::default::Default>::default::hf307ae8df96e50cf
   2:     0x563c3b559de8 - backtrace::capture::Backtrace::new::h76e7aaf5ebd7a453
   3:     0x563c3b4f4dcf - grin_util::logger::send_panic_to_log::{{closure}}::h97e1cd71910fd2f9
   4:     0x563c3b662c73 - std::panicking::rust_panic_with_hook::hba05450791beb933
                        at libstd/panicking.rs:479
   5:     0x563c3b662819 - std::panicking::continue_panic_fmt::hd2a4220d2e2291e2
                        at libstd/panicking.rs:390
   6:     0x563c3b662715 - rust_begin_unwind
                        at libstd/panicking.rs:325
   7:     0x563c3b6bdb8b - core::panicking::panic_fmt::h2c383e888dce42c0
                        at libcore/panicking.rs:77
   8:     0x563c3b10d795 - core::result::unwrap_failed::hbe0edcd4423696ba
   9:     0x563c3b17918a - <grin_servers::common::adapters::NetToChainAdapter as grin_p2p::types::ChainAdapter>::header_received::he3e2896540a63ae4
  10:     0x563c3b3ad47b - <grin_p2p::peers::Peers as grin_p2p::types::ChainAdapter>::header_received::he87decbe59a644ff
  11:     0x563c3b3b8fee - <grin_p2p::peer::TrackingAdapter as grin_p2p::types::ChainAdapter>::header_received::h5ae1b5991a84c62e
  12:     0x563c3b3e2d13 - <grin_p2p::protocol::Protocol as grin_p2p::conn::MessageHandler>::consume::hfc5cb2b0b30cdcb7
  13:     0x563c3b3c2546 - std::sys_common::backtrace::__rust_begin_short_backtrace::h42ccb37c1e69be98
  14:     0x563c3b3c3cc8 - std::panicking::try::do_call::hba67bcf4d45cb2b9
[...]

PR #1694, opened 2018-10-08
Replace RefCell with RwLock in OneTime (and cleanup Weak usage)

antiochp:
This builds and runs... why do we need the complexity of weak refs?

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -791,9 +791,9 @@ impl Handler for PoolPushHandler {
 /// except during tests).
 pub fn start_rest_apis(
 	addr: String,
-	chain: Weak<chain::Chain>,
-	tx_pool: Weak<RwLock<pool::TransactionPool>>,
-	peers: Weak<p2p::Peers>,
+	chain: Arc<chain::Chain>,
+	tx_pool: Arc<RwLock<pool::TransactionPool>>,
+	peers: Arc<p2p::Peers>,
 	api_secret: Option<String>,
 	tls_config: Option<TLSConfig>,
 ) -> bool {
@@ -813,9 +813,9 @@ pub fn start_rest_apis(
 }
 
 pub fn build_router(
-	chain: Weak<chain::Chain>,
-	tx_pool: Weak<RwLock<pool::TransactionPool>>,
-	peers: Weak<p2p::Peers>,
+	chain: Arc<chain::Chain>,
+	tx_pool: Arc<RwLock<pool::TransactionPool>>,
+	peers: Arc<p2p::Peers>,
 ) -> Result<Router, RouterError> {
 	let route_list = vec![
 		"get blocks".to_string(),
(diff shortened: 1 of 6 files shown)


hashmap:
Technicaly we create a memory leak, but if we have more or less static objects (fixed number, static lifetime) - I think adapters and peers/chain are such objects - we can live with that.

antiochp:
That's kind of what I was thinking. We have one `peers` instance and one `chain` instance. They are basically singletons. We are never going to (at least in the current impl) swap them out at runtime for other instances etc.

antiochp:
https://doc.rust-lang.org/std/sync/struct.Arc.html#thread-safety

> Arc<T> will implement Send and Sync as long as the T implements Send and Sync. Why can't you put a non-thread-safe type T in an Arc<T> to make it thread-safe? This may be a bit counter-intuitive at first: after all, isn't the point of Arc<T> thread safety? The key is this: Arc<T> makes it thread safe to have multiple ownership of the same data, but it doesn't add thread safety to its data. Consider Arc<RefCell<T>>. RefCell<T> isn't Sync, and if Arc<T> was always Send, Arc<RefCell<T>> would be as well. But then we'd have a problem: RefCell<T> is not thread safe; it keeps track of the borrowing count using non-atomic operations.

Big alarm going off from that last sentence - 

> RefCell<T> isn't Sync, and if Arc<T> was always Send, Arc<RefCell<T>> would be as well. But then we'd have a problem: RefCell<T> is not thread safe; it keeps track of the borrowing count using non-atomic operations.

@hashmap

ignopeverell:
With 2 `Arc`s referencing each other you can never cleanly stop the server. The issue is described here:

https://doc.rust-lang.org/std/sync/struct.Arc.html#breaking-cycles-with-weak

Regarding `RefCell`, you may have found the cause of those `BorrowError`s. Indeed the borrowing counters aren't atomic. That being said we only borrow mutably once at startup, so that's still a tad weird. But perhaps still worth replacing with a `RwLock`.

garyyu:
Thanks for digging into this, both @hashmap and me dug days with headache on this `Weak` in https://github.com/mimblewimble/grin/issues/1551.  Perhaps it's better to find another way out like you.

As Igno said, we have 2 Arc referencing each other (the root reason to use this `Weak`), Peers contains Adapter, and Adapter contains Peers:
```
pub struct Peers {
	pub adapter: Arc<ChainAdapter>,
	...
}
```

```
pub struct NetToChainAdapter {
	...
	peers: OneTime<Weak<p2p::Peers>>,
	...
}

pub struct ChainToPoolAndNetAdapter {
	...
	peers: OneTime<Weak<p2p::Peers>>,
}

pub struct PoolToNetAdapter {
	peers: OneTime<Weak<p2p::Peers>>,
}
```

As you (@antiochp ) said, we just use Peers as singleton (only ONE instance), why not split `peers` from Adapters and just use a util function to get it, instead of put `peers` as a field of Adapters structure.

hashmap:
@antiochp really nice catch! It explains BorrowError, which happens when the counter is -1, which could happen with non-atomic updates. We could wrap RefCell into RwLock, wdyt? We still need interior mutability here.

antiochp:
I think a RwLock gives us the interior mutability that we need no?

We take a write lock to initialize the OneTime.
Then read locks subsequently.

We should definitely not be using a RefCell across threads as it is explicitly marked as _not_ thread-safe.

antiochp:
Hmm. Everything seems to be fine when running a node (it does appear to cleanly shutdown).

But we see this when running tests (`grin_servers`) on MacOS - 

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Chain(Error { inner: 

TxHashSetErr: Too many open files (os error 24) })', libcore/result.rs:945:5
test simulate_seeding ... FAILED
```

Which I guess is related to the "leak" - we're not cleaning up the various grin servers in the tests while the tests are running.

Going to take another look at reintroducing the `Weak` refs now that we are using `RwLock` in `OneTime`.

garyyu:
@antiochp  
> Too many open files ...
> Which I guess is related to the "leak"

This error on the test is because of OS file limitation, you need change default value by:
```
sudo ulimit -n 512
```
And check it before change it:
```
ulimit -n
```

antiochp:
But why would it start happening all of a sudden?
My theory was this is related to "servers" in the tests not getting cleaned up due to refs hanging around? And these servers are keeping file handles open.

garyyu:
Not only the **file** handler, I think most of them are sockets. On the server test, we open too many sockets. Perhaps need to check if they are closed quickly.

And I remember there's some tool to check the opened file handlers.

And did you upgrade your computer OS version recently?  it could change your default `ulimit -n`.

Issue #1698, opened 2018-10-09
dead lock when calling API v1/blocks/height during body sync


garyyu:
Thanks @jaspervdm for an important hint:
> jaspervdm @jaspervdm 08:42
ah it was working until i did a GET request
https://pastebin.com/uUkXnUVu @garyyu 

I reproduced in my local node, with latest master branch. I think it's very easy to reproduce this dead lock problem.
1. clean the `chain_data`
2. start the grin for a fast sync
3. wait until the body sync stage
4. issue a API call `curl -0 -XGET  http://127.0.0.1:13413/v1/blocks/127900`

Then TUI freeze, and sync stuck.

```
Oct 09 13:01:07.001 DEBG sync_state: sync_status: BodySync { current_height: 128130, highest_height: 130214 } -> BodySync { current_height: 128226, highest_height: 130214 }
Oct 09 13:01:07.069 DEBG txhashset: merkle_proof: output: Commitment(08b54fbf893d8871a83b16a37582b4c3ab0730bfdf93dd62eb45e8704298a2d0a5)
Oct 09 13:01:07.069 DEBG merkle_proof  282948, last_pos 282949
...
Oct 09 13:01:08.040 DEBG pipe: process_block 12ec14a5 at 128275 with 0 inputs, 1 outputs, 1 kernels
Oct 09 13:01:08.064 DEBG pipe: head updated to 12ec14a5 at 128275
Oct 09 13:01:08.067 DEBG check_orphans: get block 0f8ff131 at 128276
...
Oct 09 13:03:03.920 DEBG Received compact_block 01d6b5d7 at 130216 from 185.80.129.70:13414, outputs: 1, kernels: 1, kern_ids: 0, going to process.
Oct 09 13:03:04.280 DEBG Received compact_block 01d6b5d7 at 130216 from 95.216.70.54:13414, outputs: 1, kernels: 1, kern_ids: 0, going to process.
Oct 09 13:04:20.237 DEBG Received block header 04b62bc9 at 130219 from 206.189.236.9:13414, going to process.
Oct 09 13:07:08.449 DEBG Received block header 0f4de306 at 130221 from 52.23.180.2:13414, going to process.
```

garyyu:
Analysis on the stuck backtraces:
```
  thread #19, name = 'peer'
    frame #0: 0x00007fff7dfdbaee libsystem_kernel.dylib`__psynch_rw_wrlock + 10
[...]

PR #1699, opened 2018-10-09
fix: to avoid dead lock, always put txhashset.write before store.batch

garyyu:
Fix for https://github.com/mimblewimble/grin/issues/1698

And we have to always remember in the future that `txhashset.write()` must always before `store.batch()`.

Perhaps somebody has better fix solution for this?  wrap a dedication function? or define a macro?

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -219,8 +219,8 @@ impl Chain {
 	fn process_block_single(&self, b: Block, opts: Options) -> Result<Option<Tip>, Error> {
 		let maybe_new_head: Result<Option<Tip>, Error>;
 		{
-			let batch = self.store.batch()?;
 			let mut txhashset = self.txhashset.write().unwrap();
+			let batch = self.store.batch()?;
 			let mut ctx = self.new_ctx(opts, batch, &mut txhashset)?;
 
 			maybe_new_head = pipe::process_block(&b, &mut ctx);
@@ -299,8 +299,8 @@ impl Chain {
 
 	/// Process a block header received during "header first" propagation.
 	pub fn process_block_header(&self, bh: &BlockHeader, opts: Options) -> Result<(), Error> {
-		let batch = self.store.batch()?;
 		let mut txhashset = self.txhashset.write().unwrap();
+		let batch = self.store.batch()?;
 		let mut ctx = self.new_ctx(opts, batch, &mut txhashset)?;
 		pipe::process_block_header(bh, &mut ctx)?;
 		ctx.batch.commit()?;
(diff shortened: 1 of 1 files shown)


garyyu:
Note: just find this PR can fix this problem, but it will have dead lock for `TxHashsetSave`, I'm investigating on it.

garyyu:
Find it:
```
	pub fn txhashset_write(
		...
		let mut batch = self.store.batch()?;
			...
			let mut txhashset_ref = self.txhashset.write().unwrap();
```

@antiochp  May I move this `txhashset.write()` after  `batch.commit()?;`?    will update the code and request review.

garyyu:
Tested by 2 times of the whole fast sync procedure, and with API call (by following script) on every 0.3s, no more dead lock issue.

```
#!/bin/bash

interval=0.3

if [ "$1" != "" ]; then
    interval=$1
fi

SECONDS=0
while true
do
    variableA=$(curl -0 -XGET  http://127.0.0.1:13413/v1/blocks/128000 2>null | grep block_height | tr -d ,)
    printf "`date '+%Y-%m-%d %H:%M:%S'`  $variableA\n"
    sleep $interval
done
```

NgoziDim:
smh

chain/src/chain.rs
```diff
@@ -693,6 +682,17 @@ impl Chain {
 			"chain: txhashset_write: finished committing the batch (head etc.)"
 		);
 
+		// Replace the chain txhashset with the newly built one.
+		{
```
antiochp: The concern here is that we have a (small) period of time between committing the batch and swapping the txhashset out for the new one.
The db/index is no longer consistent with the old txhashset at this point.
So if any other threads reads from the db we are at risk of reading inconsistent data.

I'm not sure how much of a problem this is in practice - need to think about it a bit more.

garyyu: Or maybe move to exact front of `batch.commit()` ?  more safe?

https://github.com/mimblewimble/grin/blob/7b0c8d36fc9063aec82aef220aa73f1dc460a23f/chain/src/chain.rs#L678

antiochp: There's still a non-atomic operation happening here, still a period when batch is committed but txhashet is not yet updated.

garyyu: You want a mutex to link both txhashset write and batch commit ?

antiochp:
We have a race condition either way - 
* update txhashset first, another thread could attempt to read it based on old indices
* commit batch first, another thread uses new indices to read the old txhashset

So I think we're fine making the change you propose (and keeping batch vs. txhashset ordering consistent everywhere).

:+1:

PR #1709, opened 2018-10-10
[T4] Secondary proof of work difficulty adjustments

ignopeverell:
Set the secondary proof of work difficulty adjustment factor within a ratio of the primary. This means a few things:

1. There's only one network difficulty.
2. Before comparing a potential PoW solution difficulty with the network one, the "block difficulty" is adjusted by a factor, different for the primary and secondary PoW.
3. The primary PoW difficulty is adjusted using a fixed Cuckatoo size factor of `2^(N-M)*(N-1)`, giving an advantage to higher Cuckatoo sizes.
4. The secondary PoW difficulty is adjusted using a fixed multiplier committed to in every header (so only makes it harder).
5. For each difficulty check, we also check the blocks within the window to see what's the actual ratio between primary and secondary PoW blocks. The header multiplier is adjusted based on deviation from ideal.
6. The ideal starts at 90% secondary and loses a percent every 10,000 blocks (roughly a week).

This PR focuses on 5 and 6, the previous points were introduced during T3. There's also a bit of a cleanup on `is_primary` and `is_secondary` which now encapsulate some additional validation.


ignopeverell:
Could use a few more tests but should be ready for review.

yeastplume:
Just running it here to follow the progression on a sample chain. Would you mind giving a quick overview?

core/src/consensus.rs
```diff
@@ -244,7 +245,60 @@ where
...
+		.collect::<Vec<_>>();
+	if scalings.len() == 0 {
+		return 1;
+	}
+	scalings.sort();
+	let scaling_median = scalings[scalings.len() / 2];
+
+	// what's the ideal ratio at the current height
+	let ratio = secondary_pow_ratio(height);
+	println!(
```
yeastplume: Shame. *ding*

ignopeverell:
Sorry I forgot I had originally created the PR without a description. I've updated it now, hopefully it's helpful.

ignopeverell:
Merging before another PR conflicts with this again or more refactoring creeps in. Could use some tuning on the secondary adjustment factor calculation (mostly dampening and bounding) and a test for the secondary PoW (by finding a couple Cuckatoo29 solutions for example).

Issue #1702, opened 2018-10-09
feature: tui: add downloading progress percent for txhashset download


sesam:
From this user experience report https://gitter.im/grin_community/dev?at=5bbbed431c100a4f295167d3

PR #1710, opened 2018-10-10
chain difficulty_iter don't need a WriteTransaction batch

garyyu:
- A little bit optimization for chain store batch usage. the chain::difficulty_iter() is a typical `ReadTransaction` and batch for `WriteTransaction` is not necessary.  And `read` is not blocking, so it's much better for performance improvement.

- This PR can fix https://github.com/mimblewimble/grin/issues/1616, to avoid cause TUI freeze.
- And it's the base of https://github.com/mimblewimble/grin/issues/1702.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -943,8 +943,8 @@ impl Chain {
 	/// difficulty calculation (timestamp and previous difficulties).
 	pub fn difficulty_iter(&self) -> store::DifficultyIter {
 		let head = self.head().unwrap();
-		let batch = self.store.batch().unwrap();
-		store::DifficultyIter::from(head.last_block_h, batch)
+		let store = self.store.clone();
+		store::DifficultyIter::from(head.last_block_h, store)
 	}
 
 	/// Check whether we have a block without reading it
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -434,7 +434,7 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		// the _network_ difficulty of the previous block
 		// (during testnet1 we use _block_ difficulty here)
 		let child_batch = ctx.batch.child()?;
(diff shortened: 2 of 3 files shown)


garyyu:
@antiochp  Please take review on this. I'm not sure whether the read is safe for us.

In case a `WriteTransaction` is ongoing, this `read` can continue execute, so in race condition we will get a `DifficultyIter` which is being updated.

Is this OK? since the next call will get the updated one.

antiochp:
OK. Now I feel like I have a better grasp of the "writes on multiple threads" behavior - 

http://www.lmdb.tech/doc/index.html

> The library is fully thread-aware and supports concurrent read/write access from multiple processes and threads. Data pages use a copy-on- write strategy so no active data pages are ever overwritten, which also provides resistance to corruption and eliminates the need of any special recovery procedures after a system crash. Writes are fully serialized; only one write transaction may be active at a time, which guarantees that writers can never deadlock. The database structure is multi-versioned so readers run with no locks; writers cannot block readers, and readers don't block writers.

Specifically - 
> Writes are fully serialized; only one write transaction may be active at a time, which guarantees that writers can never deadlock.

So there are two things we need to be aware of - 

1. We _cannot_ have multiple write transactions in a single thread (this will __deadlock__).
2. We _can_ safely have multiple threads each with a write transaction (but this will __block__).

And those two are different, even though they can temporarily look similar to an outside observer.

garyyu:
> Ugly but it works...

Fully agree!  Hope you will give a READ ONLY batch for us to call, then it will be much more clean 😄

garyyu:
> So there are two things we need to be aware of -
> 
> 1. We cannot have multiple write transactions in a single thread (this will deadlock).
> 2. We can safely have multiple threads each with a write transaction (but this will block).

I want to add the 3rd thing which need to be remembered:
3. Readers run with no locks. **Writers cannot block Readers, and Readers don't block Writers.**

PR #1716, opened 2018-10-10
The Header MMR (One MMR To Rule Them All)

antiochp:
* Add support for a db backed MMR (hash file, data maintained in db).
* Add additional MMRs to `txhashset`
  * header MMR (headers based on `header_head`)
  * sync MMR (headers based on `sync_head`)
* `txhashset::extending` now extends header MMR (in addition to output, rangeproof, kernel)
* `txhashset::readonly_extending` extends header MMR (in addition to output, rangeproof, kernel)
* `txhashset::header_extending` extends header MMR _in isolation_ (for header validation, where we do not yet have full blocks)
* `txhashset::sync_extending` extends  sync MMR only (for handling batches of headers during sync)
* after receiving output, rangeproof, kernel MMRs from peer via `txhashset.zip` receiving node now re-builds header MMR based on headers in db prior to validating the reconstructed txhashset
* on entering sync state we rebuild the sync MMR based on current `header_head` - this allows us to safely rewind and reapply headers as necessary based on current most work chain

None of this is consensus breaking, we just add the MMRs in this PR.
Consensus breaking changes will come in a separate PR.

----

TODO (this PR) - 
- [x] Cleanup code.
- [x] Cleanup comment/doc warnings.
- [x] More testing.
- [x] ~Optimize `rebuild_header_mmr()` and `rebuild_sync_mmr()` or are they acceptable for now?~
- [x] ~skip `rewind()` on "next header" for `header_extending` and `sync_extending`.~
- [x] Header MMR tracks `header_head` - confirm this is ok if/when `head` and `header_head` differ. 
  - [x] "Catch up" on same chain if nodes falls behind.
  - [x] Short forks where two blocks get mined simultaneously.
- [x] Support migrating an existing node
  - [x] Create/backfill header MMRs as necessary on `chain::init()`

----

TODO (potentially separate PR) - 
* This adds a header MMR but we __do not__ commit to the MMR root yet in block headers -
  * Replace `header.previous` with `header.prev_root`.
  * DB indices to allow us to find previous header easily?
  * Validate `prev_root` prior when validating blocks _and_ headers.
  * What would be involved in hard-forking this in later?

----

Optimizations/Improvements to think about - 
* Skip `rewind()` on "next header" during header extension (we can live without this initially).
* "Rebuild" sync MMR from header MMR by simply copying the files themselves (no need to actually rebuild the MMR from the hashes) 
* Can we extend the header MMR during header validation then simply reuse the updated header MMR when applying full blocks? (Right now we rewind and reapply a couple of times, first the validate header, then to validate and apply the block).

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -97,9 +97,9 @@ impl TxHashSet {
 	pub fn from_head(head: Arc<chain::Chain>) -> TxHashSet {
 		let roots = head.get_txhashset_roots();
 		TxHashSet {
-			output_root_hash: roots.0.to_hex(),
-			range_proof_root_hash: roots.1.to_hex(),
-			kernel_root_hash: roots.2.to_hex(),
+			output_root_hash: roots.output_root.to_hex(),
+			range_proof_root_hash: roots.rproof_root.to_hex(),
+			kernel_root_hash: roots.kernel_root.to_hex(),
 		}
 	}
 }
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -35,7 +35,7 @@ use grin_store::Error::NotFoundErr;
 use pipe;
 use store;
 use txhashset;
-use types::{ChainAdapter, NoStatus, Options, Tip, TxHashsetWriteStatus};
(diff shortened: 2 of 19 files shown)


elenka2430:
Привет,все хорошо

antiochp:
Merging.

PR #1629, opened 2018-10-01
perf: update sync_head once per batch of headers

antiochp:
In block processing we - 
  * update header_head (if most work)

In header sync -
  * update header_head (if most work)
  * update sync_head (always)

But - we only need to update the sync_head and header_head _once_ at the end per batch of headers.
We can save a bunch of writes to the db by avoiding doing this for _every_ header.

PR #1719, opened 2018-10-11
small speed optimization for header sync

garyyu:
near 35% speed up for header sync.

Before this PR: 
- 463s for 132145 headers,  or 285.4 headers/s

After this PR: 
- 340s for 132866 headers, or 390.7 headers/s

Root Cause:
- After a nice performance optimization in https://github.com/mimblewimble/grin/pull/1629 by @antiochp , we only call one `batch.commit` for one `headers_received()` function call. So, the function `headers_received()` is much faster than before.
- Currently we split 511 headers into 8-headers batch for one `headers_received()` function call. Since this function is much faster now, we can change this `8` to a bigger number.
- I tested 16 and 32, and 64.  The best number for now is `32`.

### DIFF
--- a/p2p/src/protocol.rs
+++ b/p2p/src/protocol.rs
@@ -186,7 +186,7 @@ impl MessageHandler for Protocol {
 					let headers: Headers = headers_streaming_body(
 						conn,
 						msg.header.msg_len,
-						8,
+						32,
 						&mut total_read,
 						&mut reserved,
 						header_size,


antiochp:
Is this magic number going to produce consistent results across different platforms and environments?

garyyu:
> Is this magic number going to produce consistent results across different platforms and environments?

No, I'm afraid not. But if we don't want (I really don't, too complex) to expose this to be a configurable item, we have to give a supposed best fixed one on all platforms.

So far, I only estimate this value on one mac pc, with 200Mbps network bandwidth. We need more accurate test results to get better estimation, in the future.

But for now, I'm OK to quickly go with this value 😄   since it's supposed working well on most of situations, and our code is still in fast changing.

You can test it to see the difference if having time.

Issue #1715, opened 2018-10-10
trigger compaction every 2000 blocks


garyyu:
the const value `2000` can be moved to `consensus.rs`, and perhaps can be change to `1440` for consistency with all our other seemingly arbitrary numbers.

https://github.com/mimblewimble/grin/blob/master/servers/src/common/adapters.rs#L530-L532

**recap the discussion in Gitter**:
https://gitter.im/grin_community/dev?at=5bbdf266ef4afc4f284c8c2f

Gary Yu @garyyu 20:39
hardcoded as 2000.

Antioch Peverell @antiochp 20:40
yeah it needs to be a fixed value across all nodes
should probably be in our constants though
and could probably be 1440 for consistency with all our other seemingly arbitrary numbers...

John Tromp @tromp 20:41
could be temporarily set much lower to ensure issue is resolved
like for remainder of t3

Antioch Peverell @antiochp 20:42
not easily because it impacts all other nodes on the network (syncing blocks etc.)
if you compact all your blocks away then other nodes still assume they can safely sync from you (up to 2000 blocks back)

John Tromp @tromp 20:43
ic

Antioch Peverell @antiochp 20:43
it can't safely be modified per node
(so I guess it probably belongs in consensus.rs)

Gary Yu @garyyu 20:43
Okay, so you will change it now? @antiochp
Ah, is this a consensus const?
OK.

Antioch Peverell @antiochp 20:44
I think we should discuss and agree before we change this - and it would be a T4 only change
feel free to propose the change via a PR though

jaspervdm @jaspervdm 20:44
by the way, what happens if there's a fork that's longer than 2000 blocks? will fast sync nodes be able to switch chains without problems, even in the absence of archival nodes?

Gary Yu @garyyu 20:44
better to submit an issue to avoid forget it 

Antioch Peverell @antiochp 20:45
@jaspervdm nope - if we see a fork beyond 2000 blocks (maybe 1440) then fast sync nodes will have to sync from scratch via a full fast sync
i.e. download a full txhashset

jaspervdm @jaspervdm 20:46
ah
but only the node claiming to have the fork will be able to supply it

Antioch Peverell @antiochp 20:46
yes
it works in theory - its something we need to think through a bit more in practice...
you wouldn't want to advertise that fork suddenly from a single mining node, put it that way

jaspervdm @jaspervdm 20:47
yeah

Antioch Peverell @antiochp 20:48
but if you're the only node on the network with that data then everyone else needs to get it from you somehow
[...]

PR #1721, opened 2018-10-11
[T4] change compaction check trigger to 1 day and cut_through_horizon to 1 week

garyyu:
This PR will close https://github.com/mimblewimble/grin/issues/1715
- compaction blocks change from 2000 to 10080, i.e. one week.
- use a `constants.rs` for constants, for the moment, only one is there. will move more to there when see them.


garyyu:
Remind myself about how the `compact()` is using this `cut_through_horizon`:
```
	pub fn compact(&self) -> Result<(), Error> {
		...
		let horizon = global::cut_through_horizon() as u64;
		...
		let mut current = batch.get_header_by_height(head.height - horizon - 1)?;
		loop {
			match self.store.get_block(&current.hash()) {
					...
					batch.delete_block(&b.hash())?;
			...
			match batch.get_block_header(&current.previous) {
		...
		}
```

antiochp:
I think that's right yes - that way we get regular compaction steps happening, but we have a reasonably large history to work with to handle forks.

We should check if anyone has concerns with these numbers though.

garyyu:
Updated. And request @ignopeverell  to take a look.

antiochp:
Something to think about - 

Do we introduce problems network wide if _all_ nodes trigger compaction at exactly the same height?
Basically every 1440 blocks (i.e. at height 1440, 2880, 4320 etc.) all nodes run compaction.
We basically have a big (relatively speaking) "garbage collection" pause at these heights.

To be good citizens all nodes should maintain at least 10080 historical blocks, but _when_ compaction runs can be more flexible and potentially randomized?

garyyu:
Good point 👍 
So, let's change to:
```
if tip.height % global::COMPACTION_CHECK == get_ip_addr() % global::COMPACTION_CHECK
```

antiochp:
I was actually thinking simpler than that - 

Why do we not just "roll a die" every block and set it up so on average we run compaction once per day?
Which is roughly 0.07% chance per block (better check my math though).
Then we can replace the trigger value with a percentage value (and this _can_ be configurable per node). 

This would be similar to what we do for Dandelion stem vs fluff (where we fluff a tx 10% of the time).

garyyu:
Roll the dice:
```
let mut rng = thread_rng();
if 0 == rng.gen_range(0, global::COMPACTION_CHECK) {
    ...
}
```

Is happy now? 😄

garyyu:
Sorry, I just submit the modification 😄  now the discussed solution code is ready in PR.

PR #1724, opened 2018-10-11
Chain init reset sync head

### LINKED ISSUE #1700: fast sync always start from 0 if quit in the middle
I use latest master branch and if I quit during a sync (for example the txhashset downloading stage), next time when run it again, it will always start from the Header sync, exactly from 0 height.

Looks like a regression bug.

antiochp:
Resolves #1700.

We only need to reset the sync_head here, no need to reset header_head as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ script:
     fi;
 
 before_deploy:
-  - if [[ "TEST_SUITE" == "pool-p2p" ]]; then
+  - if [[ "$TEST_SUITE" == "pool-p2p" ]]; then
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -67,7 +67,7 @@ script:
     fi;
 
 before_deploy:
-  - if [[ "TEST_SUITE" == "pool-p2p" ]]; then
+  - if [[ "$TEST_SUITE" == "pool-p2p" ]]; then
```
garyyu: 😄

PR #1731, opened 2018-10-13
Removed kernel sum from header

### LINKED ISSUE #1568: Remove total kernel sum from header
It seems we're now in agreement this was a bad direction (@antiochp). This issue tracks its removal for T4.

ignopeverell:
Fixes #1568


ignopeverell:
Totally missed #1723. Closing.

PR #1738, opened 2018-10-14
Use secp crate directly without extra use statement

rrybarczyk:
I noticed this extra use statement, but it doesn't appear to be necessary.


garyyu:
thanks @rrybarczyk for this finding, it works 😄

PR #1741, opened 2018-10-14
small correction for H

fixone:
change H from 'elliptic curve' to 'generator of an elliptic curve'

### DIFF
--- a/doc/intro.md
+++ b/doc/intro.md
@@ -91,7 +91,7 @@ fundamental properties are achieved.
 Building upon the properties of ECC we described above, one can obscure the values
 in a transaction.
 
-If _v_ is the value of a transaction input or output and _H_ an elliptic curve, we can simply
+If _v_ is the value of a transaction input or output and _H_ the generator of an elliptic curve, we can simply
 embed `v*H` instead of _v_ in a transaction. This works because using the ECC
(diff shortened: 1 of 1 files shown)


ignopeverell:
I've been hesitant about those changes (you're not the first to try to make the terminology more accurate). The guide is aimed at people with little ECC knowledge and I think it does a pretty good at that. If you know what a generator point is, you'll get what all of this is anyway.

sesam:
:+1: on keeping it easy to grasp for newcomers.

How about instead adding a link like "[More details](https://github.com/mimblewimble/docs/wiki/Jargon-file-and-glossary#elliptic-algebra)
and there we can define (and also further teach newcomers) on the meaning of the different symbols commonly used.

fixone:
I'm for simplicity, but also for consistency. Note how a bit further down it says "(practically G is just another generator point on the same curve group as H)". So it already makes it clear that H is not a curve but a point on it

sesam:
good catch.
So the options are: 
curve (group) / generator (point)
The simplest is maybe to change to curve throughout the text, then explain the details and G vs H relation either at the end of the document, or the wiki.

cc @ignopeverell et al

ignopeverell:
How about moving and adapting the clarification between parentheses to that first occurrence of the "elliptic curve" simplification.

PR #1743, opened 2018-10-14
[T4] Change network code for extended keys

jaspervdm:
The base58 encoded prefixes for the private and public extended keys are `xprv` and `xpub` for Bitcoin, I propose to change it to `gprv` and `gpub` for Grin.


yeastplume:
Sure, can try this for now

Issue #1684, opened 2018-10-08
Full node sync stuck on infinite requesting a block


garyyu:
```
Oct 07 18:59:50.854 DEBG Received block 09508fc2 at 124823 from 185.80.129.70:13414, inputs: 0, outputs: 1, kernels: 1, going to process.
Oct 07 18:59:50.859 DEBG pipe: process_block 09508fc2 at 124823 with 0 inputs, 1 outputs, 1 kernels
Oct 07 18:59:50.901 DEBG Error returned, discarding txhashset extension: Store Error: NotFoundErr("Block sums for block: 17006ba7") 
 Cause: DB Not Found Error: Block sums for block: 17006ba7 
 Backtrace: 
Oct 07 18:59:50.902 INFO Rejected block 09508fc2 at 124823: Error { inner: NotFoundErr("Block sums for block: 17006ba7")

Store Error: NotFoundErr("Block sums for block: 17006ba7") }
Oct 07 18:59:50.902 DEBG pipe: process_block 13d7f6a3 at 124896 with 0 inputs, 1 outputs, 1 kernels
Oct 07 18:59:50.902 DEBG adapter: process_block: block 09508fc2 refused by chain: Other Error: Error { inner: NotFoundErr("Block sums for block: 17006ba7")

Store Error: NotFoundErr("Block sums for block: 17006ba7") }
...
Oct 07 18:59:51.724 DEBG Received block 09508fc2 at 124823 from 185.80.129.70:13414, inputs: 0, outputs: 1, kernels: 1, going to process.
Oct 07 18:59:51.724 DEBG pipe: process_block 09508fc2 at 124823 with 0 inputs, 1 outputs, 1 kernels
Oct 07 18:59:51.724 DEBG Block 09508fc2 at 124823 is unfit at this time: already known in cache
...
Oct 08 11:50:55.875 DEBG body_sync: body_head - 065e2a03, 124823, header_head - 0ef763f9, 128760, sync_head - 0ef763f9, 128760
Oct 08 11:50:55.972 DEBG block_sync: 124823/128760 requesting blocks [09508fc2] from 17 peers
Oct 08 11:50:55.972 DEBG Requesting block 09508fc2 from peer 139.162.168.18:13414.
Oct 08 11:50:56.126 DEBG handle_payload: received block: msg_len: 1260
[...]

PR #1745, opened 2018-10-15
[DNM] avoid infinite sync loop to consume bandwidth for nothing

garyyu:
A small improvement on https://github.com/mimblewimble/grin/issues/1684#issuecomment-429494942

In case a node is stuck for some reason, we will enter an infinite loop to download same Headers or same block from peers, this infinite loop which consume the Grin network bandwidth for nothing.

This PR will detect such kind of sync stuck and quit sync thread.

### DIFF
--- a/servers/src/grin/sync/body_sync.rs
+++ b/servers/src/grin/sync/body_sync.rs
@@ -35,6 +35,7 @@ pub struct BodySync {
 	prev_body_received: Option<DateTime<Utc>>,
 	prev_tip: chain::Tip,
 	prev_orphans_len: usize,
+	stuck_detector: (Hash, u64),
 }
 
 impl BodySync {
@@ -53,6 +54,7 @@ impl BodySync {
 			prev_body_received: None,
 			prev_tip: chain::Tip::new(ZERO_HASH),
 			prev_orphans_len: 0,
+			stuck_detector: (ZERO_HASH, 0),
 		}
 	}
 
@@ -142,6 +144,17 @@ impl BodySync {
 			.collect::<Vec<_>>();
 
 		if hashes_to_get.len() > 0 {
+			// Stuck detection:
+			// In case stuck detected, exit sync thread to avoid useless infinite loop.
+			if hashes_to_get.len() == 1 && hashes_to_get[0].clone() == self.stuck_detector.0 {
+				self.stuck_detector.1 += 1;
+				if self.stuck_detector.1 >= 100 {
+					panic!("sync: body sync stuck detected! sync thread exited!");
+				}
+			} else {
+				self.stuck_detector = (hashes_to_get[0].clone(), 0);
+			}
+
 			debug!(
 				LOGGER,
 				"block_sync: {}/{} requesting blocks {:?} from {} peers",
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -18,7 +18,7 @@ use std::sync::Arc;
 
 use chain;
 use common::types::{Error, SyncState, SyncStatus};
-use core::core::hash::{Hash, Hashed};
+use core::core::hash::{Hash, Hashed, ZERO_HASH};
 use p2p::{self, Peer};
 use util::LOGGER;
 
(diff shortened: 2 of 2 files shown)


servers/src/grin/sync/body_sync.rs
```diff
@@ -142,6 +144,17 @@ impl BodySync {
 			.collect::<Vec<_>>();
 
 		if hashes_to_get.len() > 0 {
+			// Stuck detection:
+			// In case stuck detected, exit sync thread to avoid useless infinite loop.
+			if hashes_to_get.len() == 1 && hashes_to_get[0].clone() == self.stuck_detector.0 {
+				self.stuck_detector.1 += 1;
+				if self.stuck_detector.1 >= 100 {
+					panic!("sync: body sync stuck detected! sync thread exited!");
```
sesam: sorry I'm out of the loop, what does panic! do? Kill the current thread? Will it get resurrected, and if yes, where would that thread usually be killed if no stuck detected?

garyyu: `panic!` here is used here to exit the sync thread.
no, this thread will not get resurrected.

garyyu:
> partial code duplication with magic numbers (100 and +1).
No, it's not `duplication`. One is for HeaderSync, and another one is for BodySync.

ignopeverell:
I empathize with the intent but I think this will mask 2 issues that need fixing (and so shouldn't be masked). Namely:

1. A node can get stuck.
2. Other peers don't do a great job at protecting themselves against abusive or malfunctioning peers.

antiochp:
I think I agree with @ignopeverell on (2) here - this PR means a node will decide to kill its own sync process based on data from a connected (and potentially misbehaving, intentional or otherwise) peer.

* Say I'm a peer and I advertise a fork/chain with more work. 
* But I never actually give you those new headers and instead just keep giving you the same set of headers repeatedly.
* You notice this and eventually kill your own sync process.
* All the other nodes on the network do the same thing.

We introduce a vector for a network wide DDoS attack with this.

It's really hard to know if we are actually "stuck" ourselves or if we appear to be "stuck" because of data from another peer.
And I think we need to assume the worst (or at least be prepared to handle the worst) in each situation.

ignopeverell:
Agree with @antiochp too, it seems gameable. And what's good with having a lagging node? We might as well shutdown the node at this point.

garyyu:
@antiochp  thanks for your review~

> this PR means a node will decide to kill its own sync process based on data from a ... peer.

Pls note, not based on **a** peer, I use **100** that should cover all connected peers.
But I do agree I need check whether the connected peers number is just **ONE** or less than a threshold. Will give improvement on this.

garyyu:
@ignopeverell  thanks for your review~

> Agree with @antiochp too, it seems gameable. And what's good with having a lagging node? We might as well shutdown the node at this point.

Sorry, I'm confused by this comment. 
- When you said "`might shutdown the node at this point`", that means you second this PR. 
- When you said "`Agree with @antiochp `", but @antiochp is obvious against this PR because it looks like this PR introduce a network wide DDoS attack risk (which I can improve to avoid) 😄

antiochp:
> Pls note, not based on a peer, I use 100 that should cover all connected peers.

But this is not how the network works. We sync headers from the "most work peer" which may be a single malicious peer or a group of several malicious peers, either way the result is the same.

We need to verify that a peer is actually being honest when it advertises a fork with more work. 
The _only_ way we can do that is by downloading all the headers on that fork to validate the PoW.

We have not control over what other nodes tell us.
Peers can be lying to us, either unintentionally or maliciously.
All we can do is decide to do is ignore them and/or ban them.

My point was that we cannot make decisions locally around shutting our node down (or part of the node, like the sync thread in this case) based on what other (untrusted) nodes on the network tell us.

I don't think this is the right approach to solving this problem.

garyyu:
@antiochp 
> My point was that we cannot make decisions locally around shutting our node down (or part of the node, like the sync thread in this case) based on what other (untrusted) nodes on the network tell us.

Agree. Let me think it again.

Issue #1692, opened 2018-10-08
dropped peer connections not getting closed or cleaned up?


antiochp:
Added `last_seen` to peers and exposed this on the TUI in https://github.com/mimblewimble/grin/pull/1688.

This is highlighting some interesting behavior.
See below for a "screenshot", note the two "oldest" peers (right hand side of the total difficulty column) that we have not heard from in over 1,190s.
The `last_seen` should be updated on every ping/pong message and these happen every 10s (in both directions).

So we are not seeing `ping` or `pong` messages from these peers? 
Does this imply we are not seeing _any_ messages from these peers?

```
┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│Grin Version 0.3.0                                                                                                                                                         │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
┌──────────────────┐┌───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│Basic Status      ││                                                                                                                                                       │
│Peers and Sync    ││ Total Peers: 8                                                                                                                                        │
│Mining            ││ Longest Chain: 52473368 D @ 129466 H vs Us: 52473368 D @ 129466 H                                                                                     │
│Version Info      ││                                                                                                                                                       │
│                  ││ ┌────────────────────────────────────────────────────────────────┤ Connected Peers ├────────────────────────────────────────────────────────────────┐ │
│                  ││ │ Address                   [ ] ╷ State                     [ ] ╷ Direction                 [ ] ╷ Total Difficulty          [v] ╷ Version       [ ] │ │
│                  ││ │ ──────────────────────────────┴───────────────────────────────┴───────────────────────────────┴───────────────────────────────┴────────────────── │ │
│                  ││ │ 206.189.236.9:13414           ┆ Connected                     ┆ Outbound                      ┆ 52473368 D @ 129466 H (1s)    ┆ 1                 │ │
│                  ││ │ 139.162.168.18:13414          ┆ Connected                     ┆ Outbound                      ┆ 52473368 D @ 129466 H (1s)    ┆ 1                 │ │
│                  ││ │ 54.90.220.136:13414           ┆ Connected                     ┆ Outbound                      ┆ 52473368 D @ 129466 H (1s)    ┆ 1                 │ │
[...]

PR #1746, opened 2018-10-15
kick stuck peer out of connected peers

garyyu:
https://github.com/mimblewimble/grin/issues/1692#issuecomment-429590729

A small improvement:  kick out those stuck peers in half an hour, and mark it as `Defunc` state.

2peers

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -424,6 +424,11 @@ impl Peers {
 			} else if !peer.is_connected() {
 				debug!(LOGGER, "clean_peers {:?}, not connected", peer.info.addr);
 				rm.push(peer.clone());
+			} else if peer.is_stuck() {
+				debug!(LOGGER, "clean_peers {:?}, stuck peer", peer.info.addr);
+				peer.stop();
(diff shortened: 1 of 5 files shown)


jaspervdm:
Shouldn't there be another check to prevent the case in where there is no new blocks for 30min in the network? In that case the whole network goes down because all peers kick each other. This would be very very rare but it might be better to be sure.

p2p/src/peers.rs
```diff
@@ -424,6 +424,14 @@ impl Peers {
 			} else if !peer.is_connected() {
 				debug!(LOGGER, "clean_peers {:?}, not connected", peer.info.addr);
 				rm.push(peer.clone());
+			} else {
+				let (stuck, diff) = peer.is_stuck();
+				if stuck && diff < self.adapter.total_difficulty() {
```
garyyu: Taking into account @jaspervdm 's comment: now, only kick out stuck peer when peer's difficulty is lower than own difficulty.

garyyu:
Updated for taking into account both comments.

PR #1753, opened 2018-10-15
Change position of seeding_type in server config

quentinlesceller:
Currently `seeding_type` is after `seeds` which might be confusing.
This PR moves this parameter before.

PR #1754, opened 2018-10-15
[T4] Change position of seeding_type

quentinlesceller:
T4 version of https://github.com/mimblewimble/grin/pull/1753


garyyu:
IIRC, it will cause some config items not effect.

PR #1762, opened 2018-10-16
Remove median time window in difficulty calculation

ignopeverell:
Since we enforce strictly monotonous time progression, it's not necessary.


ignopeverell:
Superseded by #1765

PR #1770, opened 2018-10-17
 feat: add peers used bandwidth calculation and display in TUI

eupn:
# Description

This pull-request makes Grin able to know and to show how many bytes has been sent and received to/from peer in a TUI.

## Changes made

1. Implemented counting of number of sent and received bytes in `Tracker` (this was TODO at [conn.rs:144](https://github.com/mimblewimble/grin/blob/7754adb834fda6ecc14bdd747562593e67b3925e/p2p/src/conn.rs#L144))
2. Added new `Used bandwidth` column to the TUI table at `Peers and Sync` tab (see screenshot)

TUI screenshot

# How Has This Been Tested?

Launched a TUI version of Grin by `cargo run --release` and observed that `Used bandwidth` column is appeared under `Peers and Sync` tab and is updating accordingly.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -22,6 +22,7 @@
 
 use std::fs::File;
 use std::io::{self, Read, Write};
+use std::mem::size_of;
 use std::net::TcpStream;
 use std::sync::{mpsc, Arc, Mutex};
 use std::{cmp, thread, time};
@@ -141,7 +142,6 @@ impl<'a> Response<'a> {
 	}
 }
 
-// TODO count sent and received
 pub struct Tracker {
 	/// Bytes we've sent.
 	pub sent_bytes: Arc<Mutex<u64>>,
@@ -161,7 +161,14 @@ impl Tracker {
 		T: ser::Writeable,
 	{
 		let buf = write_to_buf(body, msg_type);
+		let buf_len = buf.len();
 		self.send_channel.try_send(buf)?;
+
+		// Increase sent bytes counter
+		if let Ok(mut sent_bytes) = self.sent_bytes.lock() {
+			*sent_bytes += buf_len as u64;
+		}
+
 		Ok(())
 	}
 }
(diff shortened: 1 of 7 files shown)


p2p/src/conn.rs
```diff
@@ -177,14 +184,24 @@ where
...
+		send_rx,
+		error_tx,
+		close_rx,
+		received_bytes.clone(),
+	);
 
 	Tracker {
 		sent_bytes: Arc::new(Mutex::new(0)),
-		received_bytes: Arc::new(Mutex::new(0)),
+		received_bytes: received_bytes.clone(),
```
hashmap: What do you think about using `RWLock` for both counters?

src/bin/tui/peers.rs
```diff
@@ -71,9 +92,22 @@ impl TableViewItem<PeerColumn> for PeerStats {
...
 		Self: Sized,
 	{
+		// Compares used bandwidth of two peers
+		fn cmp_used_bandwidth(curr: &PeerStats, other: &PeerStats) -> Ordering {
+			let curr_recv_bytes = curr.received_bytes.unwrap_or(0);
+			let curr_sent_bytes = curr.sent_bytes.unwrap_or(0);
+			let other_recv_bytes = other.received_bytes.unwrap_or(0);
+			let other_sent_bytes = other.sent_bytes.unwrap_or(0);
+
+			curr_recv_bytes
```
hashmap: Do I understand correctly that it sorts by `recv_bytes`? Why not by sum of send + recv

eupn: Sorts by `recv_bytes` and in case `recv_bytes` are equal sorts by `sent_bytes` as well. But yes, I also think that comparing total number of bytes went in both direction seems more correct.

p2p/src/conn.rs
eupn: It seems more logical to use single-producer multiple-consumer enabled lock, yes.

ignopeverell:
@eupn nice work on your first PR!

Unfortunately the cherry pick robot failed to automatically apply on milestone/testnet4, could you submit another PR for the merge on that branch? Thanks!

eupn:
@ignopeverell thank you! Yes, I will rebase.

PR #1777, opened 2018-10-17
[T4 Update minimum share difficulty

quentinlesceller:
Update `minimal_share_difficulty` for stratum server. Use `consensus::MINIMUM_DIFFICULTY`.

### DIFF
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -20,7 +20,7 @@ use api;
 use chain;
 use chrono::prelude::{DateTime, Utc};
 use core::global::ChainTypes;
-use core::{core, pow};
+use core::{consensus, core, pow};
 use p2p;
 use pool;
 use store;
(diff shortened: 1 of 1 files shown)


yeastplume:
It's a bit more complicated here in that grin-miner has no notion of secondary scaling, it's just returning the simple difficulty as calculated. So if this change is made, the adjusted difficulty needs to be communicated to grin-miner somehow, otherwise it will won't return any solutions under the (very high according to it) share difficulty.

yeastplume:
Actually, looking through this again it might be okay. Can you just ensure grin-miner will run properly against the chain after this change before merging?

quentinlesceller:
Sure. I'm doing some local testing now.

ignopeverell:
I'm not sure we need this. There's no difficulty lower than 1 and at this point, the "any Cuckatoo solution is a share" behavior is prolly still what makes the most sense.

quentinlesceller:
Okay in that case closing.

PR #1784, opened 2018-10-18
[T4] feat: add peers used bandwidth calculation and display in TUI 

eupn:
* Cherry-picking from `master` for #1770


garyyu:
Thanks @eupn for this nice PR. and Yes, it indeed can be closed since we already switch back to master branch.

PR #1785, opened 2018-10-18
feat: table cleanup to kill the zombie army

### LINKED ISSUE #1645: Add peer last connect time and table cleanup to kill the zombie army
Over time, a node will see hundreds and eventually thousands of peers that never reappear. In addition to a needless accumulation, this slows down queries and more importantly prevents us from retrying effectively because of all the noise from the dead ones.

We should add a field to track when was the last time we connected to a peer, and regularly delete the ones we've never connected to for a couple weeks for example.

eupn:
# Description

This pull-request makes Grin able to kick peers that wasn't seen for too long. Now the timeout for individual peer before it will be kicked is `two` days (discussable ofc) of silence.

Closes #1645.

## Changes made

1. Added function `is_dead(&self) -> bool` to `Peer` structure. This function checks `Peer`'s `PeerLiveInfo` instance's `last_seen` field to not exceed `now + global::DEAD_PEER_KICK_TIME`
2. Added check for `is_dead()` being `true` for individual peers during `Peers`' `clean_peers` to kick those peers that are seemed to be dead

# How Has This Been Tested?

* A ridiculously small timeout of 5 seconds is set up
* Launched a TUI version of Grin by `cargo run --release` and observed that peers in `Peers and Sync` tab are disappearing after 5 seconds timeout


garyyu:
@eupn thanks to pick this help wanted issue 👍 

I'm afraid that's a different meaning from #1645.

> Over time, a node will see hundreds and eventually thousands of peers that never reappear. In addition to a needless accumulation, this slows down queries and more importantly prevents us from retrying effectively because of all the noise from the dead ones.
> 
> We should add a field to track when was the last time we connected to a peer, and regularly delete the ones we've never connected to for a couple weeks for example.

The meaning of #1645 is removing zombie peers from database, but this PR is just removing from connected peers. Pls note the function `clean_peers()` is only cleaning the `self.peers`.

And 2 weeks could be better than 2 days for this purpose. 

Could you please revise this PR? Thanks~

eupn:
> @eupn thanks to pick this help wanted issue 👍
> 
> I'm afraid that's a different meaning from #1645.
> 
> > Over time, a node will see hundreds and eventually thousands of peers that never reappear. In addition to a needless accumulation, this slows down queries and more importantly prevents us from retrying effectively because of all the noise from the dead ones.
> > We should add a field to track when was the last time we connected to a peer, and regularly delete the ones we've never connected to for a couple weeks for example.
> 
> The meaning of #1645 is removing zombie peers from database, but this PR is just removing from connected peers. Pls note the function `clean_peers()` is only cleaning the `self.peers`.
> 
> And 2 weeks could be better than 2 days for this purpose.
> 
> Could you please revise this PR? Thanks~

Thanks for the review! Yes, I will. Can you, please, point out where the peers DB is located (in codebase)?

eupn:
@garyyu revised in #1794

PR #1789, opened 2018-10-18
Replace logging backend to log4rs and add log rotation

### LINKED ISSUE #1656: Log rotation
Currently we always append to the log which results in uncontrolled growth which is problematic if a user wants to share it for investigation and could easily consume all disk space on a small server.

We may implement a size cap and/or splitting by days.

eupn:
# Description

This pull request replaces `slog` logging backend with [log4rs](https://github.com/sfackler/log4rs) which enables log rotation and uses [log](https://github.com/rust-lang-nursery/log) as a frontend.

Intended to close #1656

## Changes made

* Replaced `slog` with `log4rs`
* Modified `logger.rs` to setup `log4rs` as logging backend
* Make Grin use industry-standard `log` crate as logging frontend
* Got rid of `LOGGER` instance and of accessing it from all over the codebase
* Modified logging options structure to add optional limit of log file size to enable log rotation

# How Has This Been Tested?

Following checks has been performed:

- [x] Logging occur when enabled
    - [x] To `stdout`
    - [x] To file
- [x] Logging to `stdout` doesn't occur when disabled
- [x] Filtering by logging level is performed correctly
- [x] Correct behaviour when levels for `stdout` and log file are different
- [x] Interoperability with TUI is preserved (i.e. filtering of `cursive`-related messages and other garbage)
- [x] Logging to file doesn't occur when disabled
- [x] Log file rotation occur when enabled
- [x] Compressed `.gz` logs are created under log file directory after log rotation occurred
- [x] Log file rotation doesn't occur when disabled
- [x] Log file appending occur when enabled
- [x] Log file rewriting occur when appending is disabled

Feedback is much appreciated!


garyyu:
Thanks @eupn for this PR!
I see the `grin_logs` folder is created on the current folder, instead of the `~/.grin` folder default. Could you revise this?

There is some configurable items for log in `grin-server.toml`,  such as `log_file_path` and etc. Please give some tests on this configuration items to avoid regression.

garyyu:
Some problems:
- It will generate huge log (1.5G for tens minutes), a lot of `DEBUG [cursive::views...`. 
- No timestamp.

eupn:
@garyyu thank you for giving it a shot and feedback, I'll address those issues in consequent commits

garyyu:
OK, Thanks.  
And look like `log4rs` is much more popular than `flexi_logger`. Is there any benefit on `flexi_logger` than `log4rs`? if both support the log rotation.

eupn:
@garyyu 
>Is there any benefit on flexi_logger than log4rs

Can't say on the first glance, `flexi_logger` was just a first drop-in replacement for `slog` but since this PR uses `log` frontend anyway, I will try to replace `flexi_logger` backend with `log4rs` and give it a shot, since it is more popular and supports log rotation.

eupn:
@garyyu please review for any issues or problems. Problems mentioned above were addressed and additional checks for possible regressions has been done.

garyyu:
@eupn  thanks for update. Is it working on your local node? I run it and look like the rotation doesn't work.

eupn:
@garyyu yes it works. For backwards compatibility purposes the `log_max_size` parameter is optional and disabled by default. Can you set it up in the config or regenerate config file? Then rotation should be working.

garyyu:
Yay! it works after I removed the `grin-server.toml` file. Thanks @eupn 

All good for me and just 3 minor change requests remaining:
- You can add comments for this new config item `log_max_size`, for pretty format config file. Here:
https://github.com/mimblewimble/grin/blob/7a007242ca6ec7e91c529c19f8a5c21df9cbe6c4/config/src/comments.rs#L395-L400
- About the default value, I see you use `2 megabytes default`. Because the text file gzip compressing is very efficient, I propose to enlarge this default value properly, such as 16M default.
- Regarding `For backwards compatibility purposes the log_max_size parameter is optional and disabled by default.`, Can we enable it by default? limit the log file size is a good to be a default behavior.

PR #1791, opened 2018-10-18
improve minimum difficulty handling

tromp:


garyyu:
> Gary Yu @garyyu 20:34
The test case next_target_adjustment will fail randomly. Any hint about how to fix it?
```
test next_target_adjustment ... FAILED
 
 failures:
 
 ---- next_target_adjustment stdout ----
thread 'next_target_adjustment' panicked at 'assertion failed: `(left == right)`
  left: `HeaderInfo { timestamp: 1, difficulty: Difficulty { num: 1 }, secondary_scaling: 2784, is_secondary: false }`,
 right: `HeaderInfo { timestamp: 1, difficulty: Difficulty { num: 1 }, secondary_scaling: 1, is_secondary: false }`', 
core/tests/consensus.rs:386:2
```

> John Tromp @tromp 20:54
oh, wait. my changes are not in yet. can you check if my mimblewimble/grin#1791 fixes this?

> John Tromp @tromp 20:59
@garyyu my PR indeed fixes that

John Tromp @tromp 21:02
**they are consensus breaking if difficulty gets below 3**
**but i suspect that will never happen on t4**

jaspervdm:
We can prevent a split if everyone upgrades

PR #1792, opened 2018-10-19
Make check_api_secret_file private

rrybarczyk:
This does not appear to be used outside of the crate.


garyyu:
Thanks @rrybarczyk again 👍

PR #1794, opened 2018-10-19
feat: remove expired peers from the storage

### LINKED ISSUE #1645: Add peer last connect time and table cleanup to kill the zombie army

eupn:
# Description

This pull-request makes Grin able to remove expired peers by the time they were last connected to.

Intended to close #1645.

## Changes made

1. Added `last_connected` field to the `PeerData` structure. This field indicates last moment when the successful connection with the peer occurred
2. Added function `delete_peers` to `PeerStore`. This function creates a batch for deletion of peers which are satisfying some predicate
3. Added function `remove_expired` to `Peers`. This function removes those peers which are seemed to be expired from the storage and peers list
4. Added call to the `remove_expired` from `seed.rs`'s `monitor_peers()`

## Open questions

There're some questions for the Grin people (maintainers probably) who are more familiar with codebase than I and have more insight into some subtle consequences of choices made during implementation of this feature:

- [x] Where/when to check for expired peers and remove them (for now it's done in `monitor_peers()`)?
- [x] Where/when to update peer's `last_connected` value to the current one?
- [x] It is necessary to also remove expired peers from list of peers list besides removing them from the storage?
- [x] If for some reason there're huge bunch of peers to remove, wouldn't removing them in single `batch` consume unacceptable amount of RAM? On the other hand, wouldn't removing peers one-by-one via `delete_peer` harm overall performance in this case?

# How Has This Been Tested?

1. Selected a small test value of 2 minutes of peer expiration timeout
2. Launched Grin and observed that it has collected list of peers, most of which were unavailable for various reasons and marked as defunct
3. After timeout of 2 minutes has been expired, removal of those peers has been observed in log
4. `monitor_peers` reported in log about some number of healthy peers and *zero* defunct peers which were deleted due to timeout

*Testing and feedback is highly appreciated!*

### DIFF
--- a/core/src/global.rs
+++ b/core/src/global.rs
@@ -69,11 +69,11 @@ pub const TESTNET3_INITIAL_DIFFICULTY: u64 = 30000;
 /// we're sure this peer is a stuck node, and we will kick out such kind of stuck peers.
 pub const STUCK_PEER_KICK_TIME: i64 = 2 * 3600 * 1000;
 
-/// If a peer's last seen time is 2 weeks ago we will kick out such kind of 'dead' peers.
+/// If a peer's last seen time is 2 weeks ago we will forget such kind of defunct peers.
 const PEER_EXPIRATION_DAYS: i64 = 7 * 2;
 
-/// Constant that expresses dead peer timeout in milliseconds to be used in checks.
-pub const PEER_EXPIRATION_REMOVE_TIME: i64 = PEER_EXPIRATION_DAYS * 24 * 3600 * 1000;
+/// Constant that expresses defunct peer timeout in seconds to be used in checks.
+pub const PEER_EXPIRATION_REMOVE_TIME: i64 = PEER_EXPIRATION_DAYS * 24 * 3600;
 
 /// Testnet 4 initial block difficulty
 /// 1_000 times natural scale factor for cuckatoo29
--- a/core/src/global.rs
+++ b/core/src/global.rs
@@ -69,6 +69,12 @@ pub const TESTNET3_INITIAL_DIFFICULTY: u64 = 30000;
 /// we're sure this peer is a stuck node, and we will kick out such kind of stuck peers.
 pub const STUCK_PEER_KICK_TIME: i64 = 2 * 3600 * 1000;
 
+/// If a peer's last seen time is 2 weeks ago we will kick out such kind of 'dead' peers.
+const PEER_EXPIRATION_DAYS: i64 = 7 * 2;
+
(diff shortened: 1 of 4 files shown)


ignopeverell:
Feel free to add me as reviewer when you think the PR is ready. By the way, we often use the [WIP] prefix to PRs while we're still working on them, and just remove it when ready. And thanks for the great work!

eupn:
@ignopeverell 
>while we're still working on them

It's pretty much readt unless there are some nits or feedback (and answers to open questions) from other maintainers.

garyyu:
@eupn  to answer your questions:
> Where/when to check for expired peers and remove them (for now it's done in monitor_peers())?

`monitor_peers()` is mainly used to clean peers and to connect peers. It's in a 20s timer loop. If a handling is for low frequency for example 2 weeks in this one, you can add to `seed` thread (in func `connect_and_monitor`) with a new timer checking for example every hour.

> Where/when to update peer's last_connected value to the current one?

It depends on how you will use this `last_connected` value. For example, if you use it in every hour, it's not good to update database in every 5s.

> It is necessary to also remove expired peers from list of peers list besides removing them from the storage?

Yes. But if its last seen date is 2 weeks ago, it's definitely not in the list of `self.peers`, which will be checked and cleaned in every 20s.

> If for some reason there're huge bunch of peers to remove, wouldn't removing them in single batch consume unacceptable amount of RAM? On the other hand, wouldn't removing peers one-by-one via delete_peer harm overall performance in this case?

That's very odd edge case. And single batch is absolutely much faster than removing one-by-one.

eupn:
?r @ignopeverell 
?r @garyyu

ignopeverell:
Sorry, was confused by your addition of [WIP], thought you had discovered more to do. Will review now.

eupn:
@ignopeverell
>Sorry, was confused by your addition of [WIP], thought you had discovered more to do. Will review now

Sorry, I was unable to remove WIP prefix due to today’s Github outage/DB failures

ignopeverell:
This broke peers with an existing peer db as they weren't able to deserialize their peers. I added a quick fix in #1812. But don't let that affect you too much, this is my fault, I should have caught this in the review.

PR #1795, opened 2018-10-19
Expose post_tx in OwnerAPI Post

quentinlesceller:
Fixed https://github.com/mimblewimble/grin/issues/1592. Equivalent of https://github.com/mimblewimble/grin/pull/1595.
Just adding another useful endpoint in Owner API.
Add wallet/owner/post_tx where a user can post a slate.
Transaction can be fluffed by adding the ?fluff param at the end of the query.

### DIFF
--- a/wallet/src/libwallet/controller.rs
+++ b/wallet/src/libwallet/controller.rs
@@ -357,6 +357,32 @@ where
 		}
 	}
 
+	fn post_tx(
+		&self,
+		req: Request<Body>,
+		api: APIOwner<T, C, K>,
+	) -> Box<Future<Item = (), Error = Error> + Send> {
+		let params = match req.uri().query() {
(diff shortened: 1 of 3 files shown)


wallet/src/libwallet/controller.rs
```diff
@@ -391,6 +417,10 @@ where
 				self.cancel_tx(req, api)
 					.and_then(|_| ok(response(StatusCode::OK, ""))),
 			),
+			"post_tx" => Box::new(
+				self.post_tx(req, api)
+					.and_then(|_| ok(response(StatusCode::OK, ""))),
```
hashmap: Perhaps 201 fits better than 200

quentinlesceller: Agree, will update.

PR #1796, opened 2018-10-19
Fix scaling diff in blocks for mining

jaspervdm:
The expected fraction of secondary blocks depends on the height (starts at 90% and goes down 1% every 10080 blocks). This has influence on the difficulty scaling of the secondary PoW.

Currently, the miner will build blocks always assuming a ratio of 90%, which only works for blocks at height <10080. From that height onwards, the expected ratio drops and all blocks that are built will be invalid. This PR fixes the block builder.


geop5:
You

PR #1802, opened 2018-10-21
Fix shift by 1 block in TUI (#1780)

ignopeverell:
* Only use graph_weight
* Fix shift by 1 in block


quentinlesceller:
I just pushed a fix (basically remove the debug logging to T4). This should fix the build error once Github is back.

PR #1809, opened 2018-10-22
full node == fast sync (no full archival guarantees)

antiochp:
We have discussed this a few times - putting this up in a PR to see if there are any concerns around making "fast sync" the default behavior for "full nodes" (and changing the semantics of "archive mode").

* Simplify `Capabilities` -
  * combine `FULL_NODE` and `FAST_SYNC_NODE` (we now only have `FULL_NODE`)
  * peers no longer advertise any kind of "archival" capability
  * no need to filter peers by `more_work_archival` etc.
* change `archive_mode = true` behavior
  * the _only_ thing this does is prevent old blocks in the db from being deleted
  * _all_ nodes now run `check_compact()` and prune/compact the txhashset
  * _all_ nodes now fast-sync via a `txhashset.zip` download
* get rid of our special case "block 1" logic for "full archival nodes"

I think this will go a long way toward setting expectations around exactly what a _FULL NODE_ is in Grin.

For everyone wanting to build tooling (mining pools, blockchain explorers etc.) you will want to run nodes from day 1 to ensure you have full block history (if this is deemed important).
Otherwise it would be possible to obtain a full block history from another party by asking for an export of their database. But this may not be something that we would officially support.

Thoughts? Concerns?

TODO - 

- [x] Update `grin-server.toml` defaults and comments for `7` vs. `6` (all nodes should be a `7` by default now).

What will happen if we deploy this and some existing (legacy FAST_SYNC) nodes are configured with capabilities = `6` (i.e. they are not FULL_NODEs according to these rules?)

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -729,30 +729,11 @@ impl Chain {
 		Ok(())
 	}
 
-	/// Triggers chain compaction, cleaning up some unnecessary historical
-	/// information. We introduce a chain depth called horizon, which is
-	/// typically in the range of a couple days. Before that horizon, this
-	/// method will:
-	///
-	/// * compact the MMRs data files and flushing the corresponding remove logs
-	/// * delete old records from the k/v store (older blocks, indexes, etc.)
-	///
-	/// This operation can be resource intensive and takes some time to execute.
-	/// Meanwhile, the chain will not be able to accept new blocks. It should
-	/// therefore be called judiciously.
-	pub fn compact(&self) -> Result<(), Error> {
-		if self.archive_mode {
-			debug!("Blockchain compaction disabled, node running in archive mode.");
-			return Ok(());
-		}
-
+	fn compact_txhashset(&self) -> Result<(), Error> {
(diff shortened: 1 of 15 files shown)


mably:
Made a quick search on this repo without much success.  Could you point me to the documentation where we describe how are handled deep blockchain reorgs in the case we don't have all the txs available anymore?

antiochp:
A deep re-org like that would kick a "fast sync" off due to the diff between the advertised head and our head.

mably:
Ok, thanks for your explanations @antiochp

antiochp:
Oh I see what you're asking. Yes right now these txs would be lost.
I think we have a couple of issues in github tracking this (in terms of getting them back in the txpool).

mably:
Thanks for clarifying @antiochp.  Could they get lost even with 1-block reorgs?

antiochp:
Unfortunately yes - I'm actually looking at the txpool impl as we speak to see if there is a relatively easy way of getting these txs back in the pool (it drives me crazy sometimes).

PR #1814, opened 2018-10-23
Actually cleanup old blocks if archive_mode=false

antiochp:
🤦‍♀️ 

We were being _extra_ careful with archive_mode with a check here - 

https://github.com/mimblewimble/grin/blob/38cbd6eafba683f1bcd85bff7a338ee8c1df2514/chain/src/chain.rs#L815-L817

And here - 

https://github.com/mimblewimble/grin/blob/38cbd6eafba683f1bcd85bff7a338ee8c1df2514/chain/src/chain.rs#L753-L756

And that 2nd one is incorrect, so we were making sure we _never_ actually removed old blocks from the db...

I want to leave that 2nd check in there (after fixing it) just in case this code gets rewritten and somebody calls `compact_blocks_db()` on an "archive" node.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -752,7 +752,7 @@ impl Chain {
 	/// *Only* runs if we are not in archive mode.
 	fn compact_blocks_db(&self) -> Result<(), Error> {
 		if self.archive_mode {
-			return Ok(())
+			return Ok(());
 		}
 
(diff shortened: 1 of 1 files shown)


antiochp:
Note: this was harmless, archive nodes were _not_ accidentally removing old blocks, no nodes were.

antiochp:
Merging this and pretending this never happened.

PR #1816, opened 2018-10-23
Small style changes in core crate

hashmap:


core/src/consensus.rs
```diff
@@ -296,7 +296,7 @@ where
 }
 
 /// Factor by which the secondary proof of work difficulty will be adjusted
-pub fn secondary_pow_scaling(height: u64, diff_data: &Vec<HeaderInfo>) -> u32 {
+pub fn secondary_pow_scaling(height: u64, diff_data: &[HeaderInfo]) -> u32 {
```
antiochp: Is this a `slice` like this? Rather than a Vec?

hashmap: Yes, slice. https://doc.rust-lang.org/std/vec/struct.Vec.html#slicing
```
In Rust, it's more common to pass slices as arguments rather than vectors when you just want to provide a read access. The same goes for String and &str.
```
It allows to use a function with different datastructures with less overhead.

PR #1819, opened 2018-10-24
Add `nix-shell` to help test/build on NixOS

dpc:


ignopeverell:
I'm wondering if we should add these under some subdirectory, like `etc`. We already have a `Dockerfile` and I could see more of these be needed for other environments. Opinions?

PR #1829, opened 2018-10-24
handle re-orgs in transaction_pool

### LINKED ISSUE #58: [Pool] Outstanding: Recover Previously Invalidated Transactions During a Reorg
This is more of a longer-term project, as it's not necessary for basic operation of the system. However, in order for transaction pools to operate reliably, I believe there will need to be a system to return transactions invalidated on a fork that was previously head to the pool, so that any transactions that conflict between the fork and new head are not lost forever. This requires a bit more machinery than in most existing blockchains because recovering the transaction from the block itself is not possible.

Conceptually, I was thinking that the pool should retain recently invalidated transactions in their entirety for a period of time (some acceptable reorg depth), which can be much less than the cut-through reorg depth because this property is not necessary for correctness, only usability. Even in the worst case, some miners (those who built the blocks that constitute the new head) should have a consistent view of the transaction pool without any loss. Any thoughts?

antiochp:
* introduce a `reorg_cache`
  * store all recent txs added to the txpool in it
  * limit this to `100` recent txs for now (likely too small?)
* during `reconcile_block()` we simply - 
  * attempt to apply all txs from the `reorg_cache` back into the txpool

Resolves #58.
Related #1809.

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -418,6 +418,39 @@ impl Block {
 		Ok(block)
 	}
 
+	/// Extract tx data from this block as a single aggregate tx.
+	pub fn aggregate_transaction(
+		&self,
+		prev_kernel_offset: BlindingFactor,
+	) -> Result<Option<Transaction>, Error> {
+		let inputs = self
+			.inputs()
+			.iter()
+			.cloned()
+			.collect();
+		let outputs = self
+			.outputs()
+			.iter()
+			.filter(|x| !x.features.contains(OutputFeatures::COINBASE_OUTPUT))
+			.cloned()
+			.collect();
+		let kernels = self
+			.kernels()
+			.iter()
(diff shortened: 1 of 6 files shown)


antiochp:
This is some testing on a `release` build with 51 recent txs in the `reorg_cache` - 

```
2018-10-24T16:18:26.178961+01:00 DEBUG grin_pool::transaction_pool - reconcile_reorg_cache: size: 51 ...
2018-10-24T16:18:26.231207+01:00 DEBUG grin_pool::transaction_pool - reconcile_reorg_cache: ... done.
```

antiochp:
I got _really close_ to a working impl based on maintaining a valid transaction pool relative to a block at height `height(head) - n` which I think would have been really neat.
But it started to get really complicated when trying to account for the Dandelion `stempool`...

Might revisit this at some point.

PR #1835, opened 2018-10-25
Pool crate cleanup

hashmap:
Major change is ` retrieve_transaction` also `reconcile_block` was changed - both to remove some extra loops. The rest is just a cleanup.

### DIFF
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -84,7 +84,6 @@ impl Pool {
 			for k in x.tx.kernels() {
 				// rehash each kernel to calculate the block specific short_id
 				let short_id = k.short_id(&hash, nonce);
-				//rehashed.insert(short_id, x.tx.hash());
 				if kern_ids.contains(&short_id) {
 					txs.push(x.tx.clone());
 					found_ids.push(short_id);
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -74,34 +74,35 @@ impl Pool {
 		&self,
 		hash: Hash,
 		nonce: u64,
-		kern_ids: &Vec<ShortId>,
+		kern_ids: &[ShortId],
 	) -> (Vec<Transaction>, Vec<ShortId>) {
-		let mut rehashed = HashMap::new();
+		let mut txs = vec![];
+		let mut found_ids = vec![];
 
 		// Rehash all entries in the pool using short_ids based on provided hash and nonce.
-		for x in &self.entries {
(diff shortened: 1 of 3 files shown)


pool/src/pool.rs
```diff
@@ -74,34 +74,35 @@ impl Pool {
...
+		let mut found_ids = vec![];
 
 		// Rehash all entries in the pool using short_ids based on provided hash and nonce.
-		for x in &self.entries {
+		'outer: for x in &self.entries {
 			for k in x.tx.kernels() {
 				// rehash each kernel to calculate the block specific short_id
 				let short_id = k.short_id(&hash, nonce);
-				rehashed.insert(short_id, x.tx.hash());
+				//rehashed.insert(short_id, x.tx.hash());
```
antiochp: Nit: commented out line of code.

pool/src/pool.rs
```diff
@@ -74,34 +74,35 @@ impl Pool {
...
+				if kern_ids.contains(&short_id) {
+					txs.push(x.tx.clone());
+					found_ids.push(short_id);
+				}
+				if found_ids.len() == kern_ids.len() {
+					break 'outer;
```
antiochp: Oh cool - did not know about this syntax for breaking out of nested loops... :+1:

pool/src/types.rs
```diff
@@ -129,7 +129,7 @@ pub struct PoolEntry {
 }
 
 /// The possible states a pool entry can be in.
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Copy, Debug, PartialEq)]
```
antiochp: I'm never clear on what _exactly_ this does... this allows us to get rid of some calls to `clone()` above but we're not actually gaining anything performance wise?

pool/src/pool.rs
hashmap: thanks, fixed

pool/src/types.rs
hashmap: Correct, it changes assignment semantics from move to copy, so right hand value could be used after assignment (like in normal languages). Usually we achieve that with explicit `.clone()`. `Copy` is implicit and can't be overloaded, it's always bit-wise copy. Rust documentation makes a bit controversial claim `Generally speaking, if your type can implement Copy, it should.` https://doc.rust-lang.org/std/marker/trait.Copy.html which mean that you should have as least Rust as possible (move  assignment semantics is the most Rust-specific behavior, cornerstone of borrow/ownership thing)

PR #1836, opened 2018-10-25
validate root of header MMR when processing headers (during sync and full blocks)

antiochp:
* `header` now contains the following - 
  * `prev_root`
  * `prev_hash` (was `previous`, renamed for consistency)
* introduce `batch.get_previous_header()` to hide the impl
* `pipe::process_block()` now leverages a `header_extension` to generate and validate the new `header_root` before proceeding to process the full block
* lots of cleanup

Edit: we were originally planning on deprecating/removing the `previous` hash from headers but recent discussions are leaning toward leaving it in (renamed to `prev_hash` in this PR).

### DIFF
--- a/chain/src/types.rs
+++ b/chain/src/types.rs
@@ -95,6 +95,17 @@ impl Tip {
 	}
 }
 
+impl Default for Tip {
+	fn default() -> Self {
+		Tip {
+			height: 0,
+			last_block_h: ZERO_HASH,
+			prev_block_h: ZERO_HASH,
+			total_difficulty: Difficulty::one(),
+		}
+	}
+}
+
 /// Serialization of a tip, required to save to datastore.
 impl ser::Writeable for Tip {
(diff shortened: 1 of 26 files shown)


chain/src/types.rs
```diff
@@ -57,30 +57,32 @@ pub struct Tip {
...
+	/// Previous block
 	pub prev_block_h: Hash,
 	/// Total difficulty accumulated on that fork
 	pub total_difficulty: Difficulty,
 }
 
 impl Tip {
-	/// Creates a new tip at height zero and the provided genesis hash.
-	pub fn new(gbh: Hash) -> Tip {
+	/// TODO - why do we have Tip when we could just use a block header?
```
ignopeverell: Block header comes with all sorts of things (mmr roots, PoW, etc) that we don't care about when tracking just a chain tip?

PR #1840, opened 2018-10-25
Use temporary file to save bitmaps

eupn:
Removes `TODO`.

### DIFF
--- a/store/src/leaf_set.rs
+++ b/store/src/leaf_set.rs
@@ -15,8 +15,6 @@
 //! Compact (roaring) bitmap representing the set of leaf positions
 //! that exist and are not currently pruned in the MMR.
 
-use std::fs::{remove_file, rename, File};
-use std::io::{self, BufWriter, Read, Write};
 use std::path::Path;
 
 use croaring::Bitmap;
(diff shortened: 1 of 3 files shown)


store/src/leaf_set.rs
```diff
@@ -168,13 +168,26 @@ impl LeafSet {
 		// First run the optimization step on the bitmap.
 		self.bitmap.run_optimize();
 
-		// TODO - consider writing this to disk in a tmp file and then renaming?
-
 		// Write the updated bitmap file to disk.
 		{
-			let mut file = BufWriter::new(File::create(self.path.clone())?);
+			// Write temporary file
```
hashmap: It seems like a common pattern, can we extract it into function which takes a closure or macros (if we have to)?

ignopeverell: Given that you have it twice here already, I'd second @hashmap. It'll also have the nice consequence of making those 2 functions more readable, all the file manipulations hides a little bit what they're really doing.

eupn: 👌

eupn:
r=@ignopeverell 
r=@hashmap

Issue #1843, opened 2018-10-25
Improve crates publication on crates.io


ignopeverell:
At the moment it's a multiple hours process to push all grin crates to crates.io, and as we start doing more releases it would be nice to have this mostly working painlessly. To get there we need to:

- [x] Rename our `secp256k1zkp` crate to `grin_secp256k1zkp` so we don't name squat upstream.
- [x] Add website, description and license properties to all our `Cargo.toml`
- [x] Make them publishable (they all have `publish = false`)
- [x] Add a version all path-based internal dependencies
- [ ] Remove circular crate dependencies, always involving test dependencies
- [x] Fix `build.rs` to not mandate `BUILT_TIME_UTC` and `DEPENDENCIES_STR` as they aren't set by `cargo publish`
- [ ] Automate publish with a script to do all crates in the right order, once all the above is fixed

pras1507:
nice

garyyu:
👍 I was thinking to do this before 😄

hashmap:
Closing, feel free to reopen if some more work is needed

PR #1844, opened 2018-10-25
Improve crate descriptors to be more cargo-friendly

ignopeverell:
Adds details and fixes our various `Cargo.toml` to be more along the lines of what `cargo` wants and specifically `cargo publish`. Partially addresses #1843


stanjojo:
Nice here. I will try to hit that

PR #1847, opened 2018-10-26
fix: add missing API in list

### LINKED ISSUE #1653: Node API: Multiple Handlers are not displayed
When accessing `/v1/` multiple handers are not visible:

-`/v1/header/`
-`/v1/txhashset/merkleproof`

Also not documented in the code that:
- lastrangeproofs
- lastkernels
accepts the parameters `?n=x` in the query.

Also not documented that `/v1/txhashset/merkleproofs?n=x` does not actually requires a position in the txhashset but an output id.

haarts:
Closes #1653.

Looking at the code if was not clear what the query parameter `n` meant. The numbers used are examples. Would it perhaps be a good idea to replace the value of parameter `n` into something more descriptive? For example:
- `txhashset/lastoutputs?n=10`: `txhashset/lastoutputs?n=<count>`
- `txhashset/outputs?start_index=1&max=100`: `txhashset/outputs?start_index=<index>&max=<amount>`
- `txhashset/merkleproof?n=output_id`: `txhashset/merkleproof?n=<output_id>`
- `peers/<ip>:<port>/ban`

An argument could be made to replace the query param `n` with a clearer name. A discussion for another day. :)


quentinlesceller:
Hi @haarts.

Thank for the PR. Unfortunately this does not fix #1653. It was more on matter of:
- Adding the missing handlers on `/v1` page.
- Adding comments in the code for lastrangeproofs and lastkernels
- Probably replacing `n` by something else (as you mentioned above) in the query parameters when we are not looking for a max number for example.

haarts:
Done!
I think replacing `start_index` with just `start` would be better still but I don't know who/what depends on this API. 
Also the `merkleproof` path accepts a `id` parameter, not a `n` parameter. I've fixed that too.

api/src/handlers.rs
```diff
@@ -820,17 +824,19 @@ pub fn build_router(
...
 		"get pool".to_string(),
 		"post pool/push".to_string(),
-		"post peers/a.b.c.d:p/ban".to_string(),
-		"post peers/a.b.c.d:p/unban".to_string(),
+		"post peers/<ip>:<port>/ban".to_string(),
+		"post peers/<ip>:<port>/unban".to_string(),
 		"get peers/all".to_string(),
 		"get peers/connected".to_string(),
-		"get peers/a.b.c.d".to_string(),
+		"get peers/<ip>".to_string(),
```
hashmap:
```suggestion
		"get peers/<ip>:<port>".to_string(),
```
```
$ curl http://localhost:13413/v1/peers/95.90.223.11
peer address unrecognized: /v1/peers/95.90.223.1
```

PR #1853, opened 2018-10-27
Update README.md

jacqlu:


ignopeverell:
A Spanish translation would be better as its own file, perhaps `README.md.es`? Also can you confirm this isn't just Google translate?

quentinlesceller:
Pretty sure it's Google translate (e.g. "sonrisa" instead of "Grin").

knocte:
I'm a native Spanish speaker, I confirm this is not a good translation.

quentinlesceller:
Closing then.

Issue #1880, opened 2018-10-30
body_sync: cannot sync full blocks earlier than horizon


garyyu:
```
DEBUG grin_chain::chain - check_orphans: 9 blocks accepted since height 10814, remaining # orphans 7
DEBUG grin_p2p::protocol - handle_payload: received block: msg_len: 1255
DEBUG grin_servers::common::adapters - Received block 0538b707 at 10823 from 54.38.70.127:13414, inputs: 0, outputs: 1, kernels: 1, going to process.
DEBUG grin_chain::pipe - pipe: process_block 0538b707 at 10823 with 0 inputs, 1 outputs, 1 kernels
DEBUG grin_servers::grin::sync::body_sync - body_sync: no pending block request, asking more
DEBUG grin_servers::grin::sync::body_sync - body_sync: body_head - 0610a255, 10822, header_head - 06b88088, 17334, sync_head - 06b88088, 17334
DEBUG grin_servers::grin::sync::body_sync - body_sync: cannot sync full blocks earlier than horizon.
```

Looks like an impossible happening! since header_head at 17334 and body_head at 10822.  horizon is 10080.
https://github.com/mimblewimble/grin/blob/b4ee1e14a435c33c3e1e30c4be7a99a87ba07a0a/servers/src/grin/sync/body_sync.rs#L124-L127

Can't explain how it happened at this moment, let's wait to next happening of this problem.

BTW, then it will enter an infinite loop to report same log:
```
2018-10-30T14:57:38.901211+08:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: no pending block request, asking more
2018-10-30T14:57:38.901321+08:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: body_head - 00dbe703, 10828, header_head - 06b88088, 17334, sync_head - 06b88088, 17334
2018-10-30T14:57:38.901371+08:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: cannot sync full blocks earlier than horizon.
2018-10-30T14:57:38.913408+08:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: no pending block request, asking more
2018-10-30T14:57:38.913499+08:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: body_head - 00dbe703, 10828, header_head - 06b88088, 17334, sync_head - 06b88088, 17334
[...]

Issue #1845, opened 2018-10-26
Body sync failed


hashmap:
First it stopped syncing and just logging for 8 mins:
```
2018-10-25T16:22:49.804705157+02:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: 0/0 blocks received in 5s
2018-10-25T16:22:49.804739265+02:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: body_head - 027d28df, 5625, header_head - 0b12a183, 10474, sync_head - 0b12a183, 10474
2018-10-25T16:22:49.804759344+02:00 DEBUG grin_servers::grin::sync::body_sync - body_sync: cannot sync full blocks earlier than horizon.
```
after that the thread crashed:
```
2018-10-25T16:22:49.814945697+02:00 DEBUG grin_servers::common::types - sync_state: sync_status: BodySync { current_height: 5625, highest_height: 10475 } -> HeaderSync { current_height: 10474, highest_height: 10475 }
2018-10-25T16:22:49.814995924+02:00 DEBUG grin_servers::grin::sync::header_sync - sync: locator heights : [10474, 10472, 10468, 10460, 10444, 10412, 10348, 10220, 9964, 9452, 8428, 6380, 2284, 0]
2018-10-25T16:22:49.831846577+02:00 ERROR grin_util::logger -
thread 'sync' panicked at 'index out of bounds: the len is 0 but the index is 0': /checkout/src/libcore/slice/mod.rs:2085stack backtrace:
   0:     0x55c778a5333c - backtrace::backtrace::trace::h0d480c688c6e5983
   1:     0x55c778a530f2 - <backtrace::capture::Backtrace as core::default::Default>::default::hbf23cd498c62c039
   2:     0x55c778a53168 - backtrace::capture::Backtrace::new::hb00f34110e8fb8d2
   3:     0x55c7789e457e - grin_util::logger::send_panic_to_log::{{closure}}::h40696617360cc809
   4:     0x55c778b5fd73 - std::panicking::rust_panic_with_hook::h542dacac3f3a20d8
                        at libstd/panicking.rs:479
   5:     0x55c778b5f8d9 - std::panicking::continue_panic_fmt::hd989cd729c0c5a4b
                        at libstd/panicking.rs:390
   6:     0x55c778b5f7d5 - rust_begin_unwind
                        at libstd/panicking.rs:325
[...]

PR #1885, opened 2018-10-30
be careful resetting sync_head

antiochp:
* make sure we have the header corresponding to header_head
* and that nothing is corrupted from an earlier shutdown
* reset both header_head and sync_head to head if anything looks wrong

This should fix a hard to reproduce issue I saw locally where a node was failing to start up cleanly due to a `header not found`.
I'm not 100% sure I understand _why_ we got into a state where header_head was saved to the db, but the underlying header was not but it was likely a result of a bad shutdown/restart during testing.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1095,8 +1095,18 @@ fn setup_head(
 		Err(e) => return Err(ErrorKind::StoreErr(e, "chain init load head".to_owned()))?,
 	};
 
-	// Reset sync_head to be consistent with current header_head.
-	batch.reset_sync_head()?;
+	// Check we have the header corresponding to the header_head.
+	// If not then something is corrupted and we should reset our header_head.
+	// Either way we want to reset sync_head to match header_head.
+	let header_head = batch.header_head()?;
+	if batch.get_block_header(&header_head.last_block_h).is_ok() {
+		// Reset sync_head to be consistent with current header_head.
+		batch.reset_sync_head()?;
+	} else {
+		// Reset both header_head and sync_head to be consistent with current head.
+		batch.reset_head()?;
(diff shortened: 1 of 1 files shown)


ahusewar6:
Testing careful...

ahusewar6:
Not bad,be reach...

ahusewar6:
So good morality....

chain/src/chain.rs
```diff
@@ -1095,8 +1095,18 @@ fn setup_head(
...
-	batch.reset_sync_head()?;
+	// Check we have the header corresponding to the header_head.
+	// If not then something is corrupted and we should reset our header_head.
+	// Either way we want to reset sync_head to match header_head.
+	let header_head = batch.header_head()?;
+	if batch.get_block_header(&header_head.last_block_h).is_ok() {
+		// Reset sync_head to be consistent with current header_head.
+		batch.reset_sync_head()?;
+	} else {
+		// Reset both header_head and sync_head to be consistent with current head.
```
garyyu: We'd better log (warning) this surprised event, to let's know how frequent it happened.

garyyu:
> I'm not 100% sure I understand why we got into a state where header_head was saved to the db, but the underlying header was not but it was likely a result of a bad shutdown/restart during testing.

This can explain why https://github.com/mimblewimble/grin/issues/1880 and #1845 happened:  `self.chain.get_block_header(&header_head.last_block_h)` could return Error. 

But I'm afraid it's not a result of shutdown/restart, because #1880 happened in a normal bodySync stage which's syncing from scratch.

Before we can understand the root cause of this, We could need add some warning log on these surprised happening place, to remind / help us to approach the root cause.

PR #1886, opened 2018-10-30
cleanup unused archive_mode param

antiochp:


garyyu:
So, I think that configuration item can be removed?
```
#run the node in "full archive" mode (default is fast-sync, pruned node)
archive_mode = false
```

Also the `Bit flags for FULL_NODE`.

ignopeverell:
Might have merged too quickly. Want to add a PR for these @garyyu ?

garyyu:
OK 😄

antiochp:
No don't remove that - we still use it...

And the `bitflags` are not used but I think still useful to have as a placeholder (all nodes are back to being `FULL_NODES` today).

garyyu:
> No don't remove that - we still use it...

@antiochp   Do you mean you want to use this config item only for `compact_blocks_db`?
If so, then perhaps the comments here `#run the node in "full archive" mode (default is fast-sync, pruned node)` is not correct.

antiochp:
Yes that comment is incorrect now.
It should probably be something along the lines of "Preserve historical blocks in the db?".

Issue #1416, opened 2018-08-24
Add a maximum size to the stem transaction pool


ignopeverell:
Currently the test as to whether the pool is full is made by `is_acceptable` which doesn't consider the stem pool size. So as long as the regular pool size is fine, a peer can feed as many stem transactions at it pleases. Doing so may be a little tricky because some of them will get fluffed, but done quickly enough it may be used to crash a given node due to high memory use. No point in risking that.

aupiff:
The current implementation of `total_size` in `pool/src/transaction_pool.rs`:

```rust
/// Get the total size of the pool.
/// Note: we only consider the txpool here as stempool is under embargo.
pub fn total_size(&self) -> usize {
	self.txpool.size()
}
```

Based on the comments above the `total_size` function, I assume you want to add an additional configuration parameter `pub max_stempool_size: usize` to the `PoolConfig` struct.

As opposed to doing something like:

```rust
pub fn total_size(&self) -> usize {
	self.txpool.size() + self.stempool.size()
}
```

 What does the comment "stempool is under embargo" mean?

PR #1895, opened 2018-10-31
Add a maximum size to stempool

quentinlesceller:
Fixes https://github.com/mimblewimble/grin/issues/1416. Introduce a `maximum_stempool_size`, same as `maximum_pool_size` by default.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -241,6 +241,13 @@ fn comments() -> HashMap<String, String> {
 	);
 
 	retval.insert(
+		"max_stempool_size".to_string(),
+		"
+#maximum number of transactions allowed in the stempool
+".to_string(),
+	);
+
+	retval.insert(
 		"[server.stratum_mining_config]".to_string(),
 		"
 ################################################
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -223,7 +223,7 @@ impl TransactionPool {
 
 		// Check that the stempool can accept this transaction
 		if stem {
-			if self.total_stempool_size() > self.config.max_stempool_size {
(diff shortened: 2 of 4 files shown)


antiochp:
What happens if a node refuses to accept a stem tx? Does the originator get notified of this or is it silently dropped?

We want to make sure no info is "leaked" with this.

But if this is silently dropped - what happens to the stem tx itself? An upstream peer timer will expire and the tx gets broadcast?

Either way, practically speaking I'm not sure how big the stempool will ever get for a given size of network and tx volume  (I'm guessing its going to be pretty small...)

quentinlesceller:
Normally, the plan is that the originator does not knows that the pool is full (need to check the code out that).
And yes if it gets silently dropped, the timer expires and the tx is fluffed.,

I'm currently working on removing transaction from the mempool instead of just giving an error. However, I'm not quite sure what we should do with the stempool. We cannot drop stem tx under embargo. I'm worried that someone could just spam a node with high fees stem tx and technically empty the stempool or whatever is the logic to do when stempool is full (but this is the same security trade off as all cryptocurrencies).

antiochp:
:+1:

I think "spam a node with high fees stem tx" has to be treated as valid txs and not really spam (assuming they are valid txs)?

quentinlesceller:
Indeed, but in that case should we evict from stempool or just silently drop the transaction (I'm leaning toward the latter)?

antiochp:
Yeah I'm leaning toward that as well - keep it simple and just do "first in wins" and assume the timers on other stem nodes will handle it correctly.

quentinlesceller:
Regardless of #1898. I think we can merge this one safely.

Issue #1417, opened 2018-08-24
Evict transactions from the pool by order of fee/weight


ignopeverell:
We should evict transactions when the pool gets full (more than `max_pool_size`) instead of plain rejecting them, otherwise one could send a lot of transactions right above the minimum fee and block every other from even entering the pool relatively cheaply. The ordering should be very similar to what's used by `prepare_mineable_transactions`, and we could just eliminate the bottom 10%.

yeastplume:
Been staring at this for a while now, mostly to familarise self with code I haven't thought too hard about.. so please bear with me here.. So a bit confused I am most probably missing something about how this should work.

I'm assuming that we're only referring to the size of the the 'fluffed' tx pool and not the stem pool. `max_pool_size` on its own probably isn't a very good indicator as to how full the pool actually is, as when each transaction is placed in the txpool (i.e. fluffed pool,) they're big franken-transactions that have been cut-through and aggregated with the contents of the current tx pool as they go through each hop in the stem phase. 

Assuming we're looking for the original transactions with fees most suitable to the weight placed on the chain, once something is fluffed it needs to have each individual transaction available in its unmodified form to decide which should be kept and which should be dropped. So this seems to exist, and when `add_to_txpool` is called all of the transactions related to the passed in transaction are deaggregated (using the original transactions, which are kept all along the way). At the moment, this puts back and outputs/inputs that were cut and returns another Franken-transaction that's been do-cut-through.

So now, do we need to go and de-aggregate every transaction that's in the txpool as well as the current, (which essentially means finding all of the original transactions) apply the weighting criteria, then drop inputs/outputs and kernels from the Franken-transactions accordingly?

antiochp:
`max_pool_size` should be based on tx weight I think (if its not already).
The tx weight is based on # inputs, outputs, kernels, so whether its lots of small txs or one big aggregated tx the overall weight should be the same.

As far as the tx_pool is concerned there are no original txs - all it sees is the aggregated fluffed txs.

We have to either accept a whole "franken-tx" to the txpool or evict it as a whole unit, there is no info available to de-aggregate it back into constituent txs.

If we _could_ de-aggregate at this point then the whole tx aggregation thing is broken as we have not gained any privacy here...

yeastplume:
Okay, thanks.. was a bit confused by the `deaggregate` function, was assuming it was trying to deaggregate a single transaction by repopulating it against a list of original transactions, but it's operating on already aggregated transactions.

[...]

PR #1898, opened 2018-10-31
[WIP] Evict transaction from txpool

quentinlesceller:
Fixes https://github.com/mimblewimble/grin/issues/1417.
Not tested yet. Also not entirely sure that ` evicted_tx.contains(&x.tx))` works with `Transaction` object.
Summoning the master of the pool @antiochp.


pool/src/transaction_pool.rs
```diff
@@ -215,10 +215,9 @@ impl TransactionPool {
 
 	/// Whether the transaction is acceptable to the pool, given both how
 	/// full the pool is and the transaction weight.
-	fn is_acceptable(&self, tx: &Transaction) -> Result<(), PoolError> {
+	fn is_acceptable(&mut self, tx: &Transaction) -> Result<(), PoolError> {
```
antiochp: If we're going to evict existing txs from the pool we should rework this and not do it in `is_acceptable()`.

pool/src/pool.rs
```diff
@@ -152,6 +152,35 @@ impl Pool {
 		Ok(Some(tx))
 	}
 
+	/// evict last 10% of the transaction from the pool (based on fees over weight)
+	pub fn evict_from_pool(&mut self, max_pool_size: usize) {
+		let tx_buckets = self.bucket_transactions();
```
antiochp: Not directly related to this PR but I wonder how well `bucket_transactions()` handles the case where a tx in the pool is dependent on multiple prior unconfirmed txs.

pool/src/pool.rs
```diff
@@ -152,6 +152,35 @@ impl Pool {
...
+			for tx in tx_bucket {
+				flat_txs.push(tx);
+			}
+		}
+
+		// sort by fees over weight, multiplying by 1000 to keep some precision
+		// don't think we'll ever see a >max_u64/1000 fee transaction
+		flat_txs.sort_unstable_by_key(|tx| tx.fee() * 1000 / tx.tx_weight() as u64);
+
+		// Get last 10%
```
antiochp: This feels wrong - it would be possible to craft a single well-timed tx and cause nodes to evict 10% of the existing tx pool.

pool/src/pool.rs
antiochp: Just thought of something else - not sure we want to evict the whole "bucket" of txs either - it may make sense to just evict the dependent txs?

antiochp:
If we try and add a single new tx we should evict at most one existing tx (is this correct?)

I think we should go really simple and look for -
a) the lowest fee tx to evict
b) that itself has no dependent txs (its not the root of a "bucket")

Sort the txs in the pool by fee/weight.
Find the cheapest one that is not a bucket root (by bucketing them all up).
Evict it.

ignopeverell:
> If we try and add a single new tx we should evict at most one existing tx (is this correct?)

I don't think so. You could be inserting a humongous high paying transaction while the lowest paying transaction is tiny.

quentinlesceller:
Okay I will adapt the code based on your previous comments.

pool/src/pool.rs
ignopeverell: Right, we should only evict what's beyond the max block weight

pool/src/pool.rs
ignopeverell: Buckets are all dependent transactions. Independent transactions are all on one-buckets and if 5 transactions depend on each other, no matter which way, they'll all end up in the same bucket.

From a transaction selection standpoint, the nice thing is we already have child-pays-for-parent. The bad thing is someone involved in a transaction could try to kick it out of the pools by appending a very large one. Little risky and costly attack but still a current weakness.

pool/src/transaction_pool.rs
quentinlesceller: Fixed.

ignopeverell:
> While mempool_size < MAX_BLOCK_WEIGHT remove the cheapest one that is not a bucket root.

That would be a problem. I could add any very large transaction and be sure it strays there as long as another transaction depends on it.

pool/src/pool.rs
```diff
@@ -159,6 +163,50 @@ impl Pool {
 		Ok(Some(tx))
 	}
 
+	/// evict a transaction from the pool
+	pub fn evict_from_pool(&mut self) {
+		// Check that the new tx is not a bullshit tx in term of fee
+
+		let mut ordered_entries = self.entries.clone();
```
ignopeverell: This cloning makes me uncomfortable. I understand that's how the pool operates generally (cloning a lot) so I think it's fine for this PR. But we should track this and other entries cloning in another issue to optimize all that. It's a lot of allocation and copying (including range proofs, etc) just to order based on weight.

pool/src/pool.rs
```diff
@@ -159,6 +163,50 @@ impl Pool {
...
+			if fw_a > fw_b {
+				Ordering::Greater
+			} else if fw_a == fw_b && entry_a.tx_at < entry_b.tx_at {
+				Ordering::Greater
+			} else {
+				Ordering::Less
+			}
+		});
+
+		// remove first one iff not a bucket root of several txs
```
ignopeverell: This is O(n^2) in number of transactions in the pool. Definitely needs to be optimized.

quentinlesceller: Agree. Looking into this.

pool/src/pool.rs
quentinlesceller: Oups I forgot a comment there ^^. Also yeah definitely agree that this cloning is quite heavy in term of ressources.

quentinlesceller: Also I missing a check where I compare the tx fee/weight ratio with the ones in the pool so we don't remove a higher fee transaction for a low fee one.

quentinlesceller:
Closing this pull request going to open another one.

PR #1899, opened 2018-10-31
Tiny store crate cleanup

hashmap:


store/src/types.rs
```diff
@@ -28,6 +28,16 @@ use libc::{ftruncate64, off64_t};
 use core::core::hash::Hash;
 use core::ser;
 
+macro_rules! read_bitmap {
```
antiochp: (personal bias against macros) Can we do this with a simple fn and avoid having to introduce a macro?

hashmap:
Sorry, tap the wrong button:(. Bitmap is unsized, would have to Box it to return from function, kinda ugly. Also a risk to copy a huge chunk of memory, hence macros

hashmap:
Ah, don‘t merge, forgot to use it in couple places.

antiochp:
But we return a bitmap in a couple of place like this without having to box it up? 

https://github.com/mimblewimble/grin/blob/3b5a39dd42242263f25e031f75e652d122f6a9af/store/src/leaf_set.rs#L92

hashmap:
@antiochp yeah, my bad, Bitmap is sized, not sure why I thought differently. I've replaced it with a function.

Issue #1892, opened 2018-10-31
ask for txhashset at 90% of horizon?


garyyu:
In PR https://github.com/mimblewimble/grin/pull/1721, we have change the `horizon` from 48 hours to 1 week (i.e.  from 2880 to 10080).

So, for initial `BodySync` stage, we need download 90% * horizon  (i.e.  9072 full blocks), but before this change we only download 2592 blocks.

Is it safe to change this `90%` to for example `30%` here?  or something else, for faster sync.

```
DEBUG grin_servers::grin::sync::state_sync - fast_sync: before txhashset request, header head: 18460 / 058c207d, txhashset_head: 9387 / 052fc3fc
```

https://github.com/mimblewimble/grin/blob/b4ee1e14a435c33c3e1e30c4be7a99a87ba07a0a/servers/src/grin/sync/state_sync.rs#L147-L153

ignopeverell:
Possibly, but then we need to think of what happens if there's a long reorg before the node had time to complete its history to get up to horizon.

A just-sync'd node will have a history that's more recent than horizon. It's temporary because as it reaches the head and gets blocks through gossip, it will complete its history fairly quickly. The reason for this is it takes some time to download and validate the state and we don't want our just-downloaded state to be older than horizon by the time we're done.

Now if the newly sync'd node is just missing 50 blocks after horizon, it's not a huge deal, the chances that a week-minus-50-blocks long reorg happens in the next hour is slim-to-none. If we extend that however, the risk increases. So we would need to make the code that decides to go into redoing the full sync including state on reorgs older than horizon use whatever our oldest state is at instead. May not be a bad thing to do regardless.

garyyu:
Close this one since most of related discussions are in PR: #1902

PR #1902, opened 2018-11-01
ask for txhashset at 30% of horizon instead of 90%

garyyu:
Fulfill https://github.com/mimblewimble/grin/issues/1892

### DIFF
--- a/servers/src/grin/sync/state_sync.rs
+++ b/servers/src/grin/sync/state_sync.rs
@@ -144,13 +144,13 @@ impl StateSync {
 		let horizon = global::cut_through_horizon() as u64;
 
 		if let Some(peer) = self.peers.most_work_peer() {
-			// ask for txhashset at 90% of horizon, this still leaves time for download
+			// ask for txhashset at 30% of horizon, this still leaves time for download
 			// and validation to happen and stay within horizon
 			let mut txhashset_head = self
 				.chain
 				.get_block_header(&header_head.prev_block_h)
 				.unwrap();
-			for _ in 0..(horizon - horizon / 10) {
+			for _ in 0..(horizon / 10 * 3) {
 				txhashset_head = self
 					.chain
 					.get_block_header(&txhashset_head.previous)


hashmap:
I'm not sure, the reason of 90% was to make sure that we don't go below horizon while downloading txhashset, not to sync faster. To be fair I was thinking about increasing threshold to 95%

garyyu:
Changed.  
@antiochp  Could you review it again?
@hashmap  Do I need change `90%` here as `95%` as you said?

servers/src/grin/sync/state_sync.rs
```diff
@@ -150,7 +150,7 @@ impl StateSync {
 				.chain
 				.get_block_header(&header_head.prev_block_h)
 				.unwrap();
-			for _ in 0..(horizon - horizon / 10) {
+			for _ in 0..(threshold - threshold / 10) {
```
antiochp: If horizon is `10,080` and threshold is `2,880` then _this_ is already the "percentage" - we are using approx 25% of the `horizon` by using the lower `threshold` value (we don't need to use 90% of threshold, we can just use 100% of it).

ignopeverell:
> We should let nodes that have 2-7 days of history handle forks larger than 2 days

What @antiochp says.

servers/src/grin/sync/state_sync.rs
garyyu: @antiochp   this percentage (`90%` or `95%` as suggested by @hashmap ) seems a different thing from 2,880/10,080. 
As the comments line said `this still leaves time for download and validation to happen and stay within`.

If remove it and just use 100% here, we will always enter a txhashset download because of https://github.com/mimblewimble/grin/blob/42aa7a5d67eea3c7e7d085dd5e98fe918bea10a5/servers/src/grin/sync/state_sync.rs#L66-L72

Could you please think it again? I have no idea how to use 100% here and ensure state_sync.rs#L66-L72 get `need_state_sync` as false, after txhashset download and validation completed.

garyyu:
thanks @antiochp  and @ignopeverell  for your reviewing, the modification is done and please take a look again 😄 

I add a `tail` as suggested by @antiochp , and the core logic becomes:
https://github.com/mimblewimble/grin/blob/42aa7a5d67eea3c7e7d085dd5e98fe918bea10a5/servers/src/grin/sync/state_sync.rs#L66-L72

```
	let need_state_sync = highest_height.saturating_sub(head.height) > cmp::min(
		global::cut_through_horizon() as u64,
		cmp::max(
			head.height.saturating_sub(tail.height),
			global::state_sync_threshold() as u64,
		),
	);
```

servers/src/grin/sync/state_sync.rs
antiochp: Ah ok I see what you're saying. We transition to sync mode based on _either_ `tail` or the `sync_threshold` -

```
cmp::max(
	head.height.saturating_sub(tail.height),
	global::state_sync_threshold() as u64,
),
```

So a question - do we even need a `sync_threshold` if we now track the local `tail`?
If we know our local oldest full block (based on the last time we pruned/compacted the db or the last time we sync'd) then is that sufficient?

What if we drove the `need_state_sync` purely off our oldest full block (the `tail`)`?

antiochp:
If we have headers up to 1,000,000 and full blocks from 999,000 to 1,000,000 (for example).

    Head: 1,000,000
    Tail:   999,000

1) We see a fork from a peer from 999,001 to 1,000,001' then we can safely rewind and request blocks.

2) But if we see a fork from a peer from 998,999 to 1,000,001' then we are unable to rewind 
and we need to go request a new txhashset state.

Basically if the "fork point" is between our `head` and our `tail` then we can rewind and request blocks.

Thinking a bit more about this it gets more complex than our current `check_run()` logic because we would need more context about the fork and not just the `highest_height` from the peer.
We'd need to know the height the peer forked from our chain? Which I think we'd have because we must have all the headers on this fork before we consider requesting the new state.

garyyu:
@antiochp  thanks for further explain, now I understand.
Now I have added this new logic for `check_run()`:
```
		// if gap > horizon, need state sync
		// because nobody has the old full blocks before horizon
		let mut need_state_sync =
			highest_height.saturating_sub(head.height) > global::cut_through_horizon() as u64;

		// turn off state sync when previous sync just completed
		if !need_state_sync && head.height > 0 && tail.last_block_h == head.last_block_h {
			return false;
		}

		// or if tail is not on the current header chain, need state sync
		// because the fork point is beyond the local database so we can't rewind and request blocks
		if !need_state_sync && header_head.height > tail.height {
			let mut header_cursor = self.chain.get_block_header(&header_head.prev_block_h);
			while let Ok(header) = header_cursor {
				if tail.height + 1 == header.height {
					if tail.last_block_h != header.prev_hash {
						need_state_sync = true;
						debug!(
							"fast_sync: tail {} at {} is not on the local chain. state sync needed",
							tail.last_block_h, tail.height,
						);
					}
					break;
				}
				header_cursor = self.chain.get_previous_header(&header);
			}
		}
```

But I'm thinking how to test this a little bit complex logic,  any suggestion?

chain/src/chain.rs
```diff
@@ -712,12 +712,29 @@ impl Chain {
...
+			let tail = self.tail().unwrap_or_else(|_| head.clone());
 			let header_head = self.header_head().unwrap();
-			if header_head.height - head.height < global::cut_through_horizon() as u64 {
-				return Err(ErrorKind::InvalidTxHashSet("not needed".to_owned()).into());
+			if header_head.height - head.height < global::cut_through_horizon() as u64
+				&& tail.last_block_h != head.last_block_h
+				&& header_head.height > tail.height
+			{
+				let mut header_cursor = self.get_block_header(&header_head.prev_block_h);
+				while let Ok(header) = header_cursor {
```
antiochp: We have `is_on_current_chain()` that I think does the same thing as this (or something very similar) - 

https://github.com/mimblewimble/grin/blob/92e41fa571adf4c7a3f3064a6e4d6ffe9486fb2e/chain/src/pipe.rs#L652-L660

```
pub fn rewind_and_apply_fork(b: &Block, ext: &mut txhashset::Extension) -> Result<(), Error> {
	// extending a fork, first identify the block where forking occurred
	// keeping the hashes of blocks along the fork
	let mut fork_hashes = vec![];
	let mut current = ext.batch.get_previous_header(&b.header)?;
	while current.height > 0 && !ext.batch.is_on_current_chain(&current).is_ok() {
		fork_hashes.push(current.hash());
		current = ext.batch.get_previous_header(&current)?;
	}
```

garyyu: Here we're doing a check `whether the tail is on the current header chain`, looks like a different thing from `is_on_current_chain()`.  
The tail is 100% on the current chain, but in case a deep fork, the header chain and the local block chain diverse deeply, the body tail could be not on the current chain.
Am I right?

garyyu:
I’m writing a test about servers fork test, facing a lot of strange issues. Include these but not limited:
```
1. thread 'seed' panicked at 'called `Result::unwrap()` on an `Err` value: LmdbErr(Error::Code(-30783, 'MDB_BAD_RSLOT: Invalid reuse of reader locktable slot'))’
2. thread 'apis' panicked at 'error binding to 127.0.0.1:22102: error creating server listener: Address already in use
3. thread 'dandelion' panicked at 'Failed to upgrade the weak ref to our chain.’
4. thread 'sync' panicked at 'called `Result::unwrap()`  … Store Error: chain head }
```
continuous headache~  the root cause could be `Server::stop()` can’t stop server cleanly.  Anybody has idea on this?

The test is here:  https://github.com/mimblewimble/grin/blob/23938b248b87a1754ec0503ce51371a3a17c919d/servers/tests/simulnet.rs#L378-L675

In case somebody would take a look and suggest.

garyyu:
Finally, the new test case `simulate_long_fork` passed Travis-CI:  https://travis-ci.org/mimblewimble/grin/builds/452882853

For this test case, I have to refactor the code to avoid all kinds of `stop` exceptions, now I use a `pause()` and a `resume` function for p2p test.

Because this new test case `simulate_long_fork` is quite slow (about 3.5 minutes), I will disable it for normal Travis-CI test to avoid slow down the PR test.

BTW, this test case is also quite useful to reproduce some strange `header not found` problems, just repeat running this test case and often see those strange errors. But we can handle those issues in new PRs, since this PR is already too heavy.

garyyu:
It's ready for merge for me.  
@antiochp  @ignopeverell   please take a review again if have time.

PR #1903, opened 2018-11-01
replace cyan and yellow which not easy to read on white background

garyyu:
Cyan color and Yellow color is not easy to read on the white background, giving small color adjustment.

Before this change:
screen shot 2018-11-01 at 12 28 04 pm

After this change:
screen shot 2018-11-01 at 12 29 02 pm

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -346,6 +346,12 @@ fn comments() -> HashMap<String, String> {
 #where to find wallet files (seed, data, etc)
 ".to_string(),
 	);
+	retval.insert(
+		"dark_background_color_scheme".to_string(),
+		"
+#Whether to use the black background color scheme for command line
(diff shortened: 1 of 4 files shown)


hashmap:
Gray is barely visible on dark background. Perhaps we need to support 2 color schemes?

garyyu:
Done with a new configuration item:
```
#Whether to use the black background color scheme for command line
dark_background_color_scheme = false
```

Since the default terminal color scheme is white background, here I also set it as white background as default for grin wallet output. If want to use in dark background, user need to change this config as `true`.

@hashmap   Are you OK with this?

garyyu:
@hashmap  Did you find a way to detect current color scheme? 
If can't, perhaps I can change the default as the dark background to make its default behavior same as before.

PR #1910, opened 2018-11-01
PMMRable cleanup

antiochp:
* HashOnlyPMMRable for hash only header MMR
* FixedLength trait with LEN const
* Hash now implements FixedLength for consistency


antiochp:
Tested against testnet4 - no change to hash values etc in the MMRs, roots still match.

Issue #1701, opened 2018-10-09
feature request: add tests to catch regressions (performance, deadlocks, more?)


sesam:
from recent issues, we probably want to test

- [ ] performance regressions in API:s
- [ ] deadlocks from API calls
- [ ] interrupted sync causing chain to reset to 0

PR #1914, opened 2018-11-02
refactor: refactored the API into related handler files

jeffbarg:
I moved each set of related APIs into their own file within the handlers module.  This should make it more sustainable to document and test each individual API (for example for #1701 ).  I will try to add more docs / tests in follow-up PRs.


hashmap:
Awesome! Thanks.

Issue #1883, opened 2018-10-30
pruned hash malleability


antiochp:
Discussion on gitter - https://gitter.im/grin_community/Lobby?at=5bd83a26ae7be94016dc111c

Recording it here for reference.

__Q.__ Does this matter in practice and do we need to mitigate it?

----

> John Tromp @tromp 11:01
i think the whole idea of utxo only fast sync implies that pruned hashes are malleable

It would be possible for a miner to mine a long fork predating the fast sync "horizon" and changing a pruned (spent) hash in the MMR.
This would change the root of the output MMR in such as way that existing nodes would see it as invalid, but freshly sync'd nodes would see it as valid.

Old nodes would see this block as invalid, eventually running a new "fast sync" to sync with this new fork.

We would have a block accepted as valid that would be _impossible_ to recreate by reapplying all historical transactions.

> Antioch Peverell @antiochp 11:10
@tromp so we end up with fast sync state that cannot be recreated from a full archival set of blocks...
@jaspervdm I don't think there is any reduced security with this
they can rewrite spent output hashes
but nobody cares

Miners could effectively "rewrite the STXO set", but this would be immaterial in terms of the UTXO set which Grin/MW actually cares about.

> John Tromp @tromp 11:15
we can avoid this by having fixed pruned hashes, but that loses other nice MMR properties

> John Tromp @tromp 11:16
all 0 hash for entire pruned subtree
that destroys append only property of MMR

ignopeverell:
Interesting, I'm just not sure how it could matter very much either. If you have a long fork, and get to rewrite that whole section of history, I'm more worried about what one could do to the UTXO set. You can't really "malleate" them as you could with TXOs and lead to the weird situation described here, but the consequences seem a lot worse.

One mitigation solution could be to have a special handling for long forks and force a new state sync not up to horizon but up to the fork point, and then sync all blocks from that fork point. This way we can make sure the headers and the MMRs are consistent along the whole fork.

antiochp:
Closing for now - consensus is that this is possible but not really an issue to be concerned with.

Issue #1733, opened 2018-10-13
UTXO sums from output MMR (and proof of unspentness)


antiochp:
I think we should consider adding "sums" back to our output MMR "hashtree".
We used to have these when we had the original `sumtrees` but these were TXO sums.
But I think if we are clever about it we can sum the UTXO set (not simply the TXO set) and produce a useful set of peaks and overall root (the sum would be our `utxo_sum` that we store in the db in our `block_sums`).

3 files -

``` 
pmmr_hash.bin
pmmr_data.bin
pmmr_sums.bin
```

The `hash` and `data` files would act just like they do today, prunable (compactable) via the `prune_list`, with the `leaf_set` representing the unpruned leaves.

Note: the `leaf_set` represents the list of pos that make up the UTXO set in the case of the output MMR.

TXO vs. UTXO - 
* Leaf nodes represent the TXO set. 
* Leaf nodes that are not removed represent the UTXO set.

The `sums` file would behave very similar to the `hash` file but each parent node would be the sum of its children.
* It would be prunable in the same way the hash file is.
* But it would also be _editable in place_ to support removing leaf nodes while having the sums work as the sum of the UTXO set.

Pruning a leaf node would involve - 
1) Removing the pos from the `leaf_set`.
2) Updating the `sums` file to replace the output commitment with a commitment to zero value and recalculating sums up the path to the peak to adjust the root sum (the UTXO sum).

Compaction would still work - we can remove leaves once both leaves under a parent node are removed, the sums of the parent node would be a commitment to zero and all nodes in the path up to the peak would reflect this.

We would not need to commit to the UTXO set anywhere explicitly.
We commit to the root of the output MMR (TXO set) in each block header.
We also indirectly commit to the UTXO set by committing to the root of kernel MMR in each block header.

A Merkle proof against the current output MMR gives us "proof of inclusion" against the TXO set. We can prove an output exists in the TXO set.
But this does not give us any information about the output being spent or not.

The `sums` file lets us build a Merkle proof against the sums. 
And if spent sums are represented as commitments to zero then I believe we can build a "proof of unspentness" from this.

It would not be a Merkle proof of hashes up to the root hash but a Merkle proof of sums up to the root sum.
[Or do these need to be hashes of the sums, where we can leverage `hash_with_index` to include position information?]

https://petertodd.org/2016/delayed-txo-commitments

[...]

PR #1916, opened 2018-11-03
[DNM] experimental UTXO commitments

antiochp:
Playing around with possible approaches/implementation for "UTXO commitments". Not sure if this would be useful in its current state.

This PR allows a parallel UTXO PMMR to be built alongside the existing output (aka TXO) PMMR.
Too expensive to do this in realtime so hooked it up to run on demand.

Also too complex to maintain in realtime to handle rewind, sync etc.

We can trigger a UTXO PMMR rebuild via the `/compact` API endpoint (easiest place to hook it into) -

```
curl -X POST 127.0.0.1:13413/v1/chain/compact
```

```
...
2018-11-03T13:25:56.755685+00:00 DEBUG grin_chain::chain - Rebuilding UTXO set (experimental).
2018-11-03T13:25:56.755747+00:00 DEBUG grin_chain::txhashset::utxo_set - *** rebuild: rebuilding the utxo_set (rewinding aka truncating to pos 0 first).
2018-11-03T13:25:56.755758+00:00 DEBUG grin_chain::txhashset::utxo_set - Truncating utxo_set extension.
2018-11-03T13:25:56.794249+00:00 WARN grin_chain::chain - ********** UTXO root: 21025ec4 (output root: f0e58c30)
...
```

Approx 40ms to rebuild the UTXO MMR in testnet4 based on the existing output MMR (can be pruned/compacted). This is rebuilding from scratch with no thought put into optimizing this.

The existing output MMR is "append only". spending an output does not change the data in the MMR and will not affect the root of the MMR. Only by appending new outputs to the MMR will the root change.
We can prune and compact the output MMR and leave the root unaffected by removing spent outputs.

To construct the UTXO MMR we simply replace spent outputs with a "zero" sentinel value.
Each leaf is then hashed via `hash_with_index("zero", pos)`.
The interesting thing is these can roll-up to parent nodes in a deterministic way.
For example, if pos 1 and pos 2 are both spent we can hash them as - 

* hash<sub>1</sub> = hash_with_index("zero", 1)
* hash<sub>2</sub> = hash_with_index("zero", 2)

and the parent node at pos 3 can be generated as before -

* hash<sub>3</sub> = hash_with_index((hash<sub>1</sub>, hash<sub>2</sub>), 3)

And we can recreate the hash at pos 3 without needing to know anything about the underlying data beyond them both having been removed/spent.

Related - https://github.com/mimblewimble/grin/issues/1883 (@tromp suggested "zeroing out" spent commitments to fix the spent hash "malleability").
Related - https://github.com/mimblewimble/grin/issues/1733 (proof of "unspentness")

Inspiration from - https://petertodd.org/2016/delayed-txo-commitments (Delayed TXO commitments).

Thinking out loud - 
Maybe we don't need to do this in realtime - maybe there are ways we can do this in a "delayed" fashion. We commit to the UTXO MMR every 60 blocks (hourly) or 1,440 blocks (daily) say.
This would allow us to generate a Merkle proof for any unspent output older than a day, for example.
Nodes could generate proofs for their own outputs by maintaining some subset of the UTXO MMR.
Nodes could verify this without needing to maintain the full UTXO set etc.

Thoughts? @tromp @ignopeverell

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -811,7 +811,11 @@ impl Chain {
 			let mut utxo_set = self.utxo_set.write();
 			utxo_set::extending(&mut utxo_set, |utxo_extension| {
 				utxo_extension.rebuild(&utxo_view)?;
-				warn!("********** UTXO root: {} (output root: {})", utxo_extension.root(), utxo_view.root());
+				warn!(
+					"********** UTXO root: {} (output root: {})",
+					utxo_extension.root(),
+					utxo_view.root()
+				);
 				Ok(())
 			})?;
 
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -36,6 +36,7 @@ use grin_store::Error::NotFoundErr;
 use pipe;
 use store;
 use txhashset;
+use txhashset::utxo_set;
 use types::{ChainAdapter, NoStatus, Options, Tip, TxHashSetRoots, TxHashsetWriteStatus};
 use util::secp::pedersen::{Commitment, RangeProof};
 
@@ -147,6 +148,7 @@ pub struct Chain {
 	adapter: Arc<ChainAdapter>,
 	orphans: Arc<OrphanBlockPool>,
 	txhashset: Arc<RwLock<txhashset::TxHashSet>>,
+	utxo_set: Arc<RwLock<utxo_set::UTXOSet>>,
 	// Recently processed blocks to avoid double-processing
 	block_hashes_cache: Arc<RwLock<LruCache<Hash, bool>>>,
 	verifier_cache: Arc<RwLock<VerifierCache>>,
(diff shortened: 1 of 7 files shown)


antiochp:
@tromp I don't think I fully understand your comment.

The benefit of keeping the zero'd out spent outputs is we maintain the same overall structure between the output and UTXO MMRs. The output at `pos n` is the same output in both MMRs, they are just different views on the same data.

The intent here was to keep the output MMR in its existing state - its append only and very efficient for what we want it to do.
And we can see that it is relatively easy to translate the output MMR into a UTXO MMR (using a zero value for spent outputs).

Pruning/compacting outputs in this structure would not involve reindexing the whole UTXO set, which would be necessary if we used a UTXO MMR that only included unspent leaves.

tromp:
My comment was pointing out that an attacker able to rewrite horizon blocks can go as far as making it look like no tx was ever spent.

I see that preserving txo MMR structure minimizes computation, but It's not append only if every node has to make the spent txo hash changes to verify a UTXO commitment.

Or is this only an optional feature, that no full node is required to support?

antiochp:
Closing this for now - not something we're going to be implementing prior to mainnet.

PR #1926, opened 2018-11-05
Chain send sync

antiochp:
Chain does not need to be explicitly `Send + Sync`.

We can make this far more granular by wrapping `prune_list` and `leaf_set` in `Arc<RwLock<<_>>` and making the `PMMRBackend` itself `Send + Sync`.

`ChainAdaptor` needs to be `Send + Sync` to keep the api router code happy.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -156,9 +156,6 @@ pub struct Chain {
 	genesis: BlockHeader,
 }
 
-unsafe impl Sync for Chain {}
-unsafe impl Send for Chain {}
-
 impl Chain {
 	/// Initializes the blockchain and returns a new Chain instance. Does a
 	/// check on the current chain head to make sure it exists and creates one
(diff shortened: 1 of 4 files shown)


ignopeverell:
Anytime I see 2 locks in the same struct (`PMMRBackend` here) I get nervous, especially when concurrency semantics aren't explicit. For example, I think you might have a deadlock already between `rewind` and `append`.

How about a single lock for these 2 references? Either access them as a pair or introduce a small inner struct for the references that require locking? Looks like they're almost always locked together anyway.

antiochp:
😑 let me take another pass (don't want to introduce a deadlock here...)

antiochp:
Closing this for now - got distracted by the _other_ `Send + Sync` issue #1937.
The more I think about this and the more I read about `Send + Sync` the less confident I am that this PR is the right approach.

PR #1927, opened 2018-11-05
Instantiate wallet once in wallet cli

hashmap:
Also remove extra Box to simplify type


hashmap:
@ignopeverell rebased

PR #1929, opened 2018-11-05
broadcast tx kernel hash (if supported by peer)

antiochp:
* We can now lookup individual txs from the txpool by kernel hash.
* Introduce new `TX_KERNEL_HASH` capability to peers (`FULL_NODE` is now 15 by default)
* 2 new msg types - 
  * `GetTransaction`
  * `TransactionKernel`
* When broadcasting a new tx - 
  * if peer supports it then send the hash of the first kernel in the tx
  * otherwise, fallback to sending full tx 
* If we receive a kernel hash
  * look it up in the txpool, and
  * if not known go request full tx from peer
* tracking_adapter tracks "seen" txs via kernel hash to support this new broadcast logic

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -269,8 +269,7 @@ impl Peer {
 		} else {
 			debug!(
 				"Not sending tx kernel hash {} to {} (already seen)",
-				h,
-				self.info.addr
+				h, self.info.addr
 			);
 			Ok(false)
 		}
@@ -283,7 +282,11 @@ impl Peer {
 	pub fn send_transaction(&self, tx: &core::Transaction) -> Result<bool, Error> {
 		let kernel = &tx.kernels()[0];
 
-		if self.info.capabilities.contains(Capabilities::TX_KERNEL_HASH) {
+		if self
+			.info
+			.capabilities
+			.contains(Capabilities::TX_KERNEL_HASH)
+		{
 			return self.send_tx_kernel_hash(kernel.hash());
 		}
 
@@ -325,7 +328,10 @@ impl Peer {
 	}
 
 	pub fn send_tx_request(&self, h: Hash) -> Result<(), Error> {
-		debug!("Requesting tx (kernel hash) {} from peer {}.", h, self.info.addr);
+		debug!(
+			"Requesting tx (kernel hash) {} from peer {}.",
+			h, self.info.addr
+		);
 		self.connection
 			.as_ref()
 			.unwrap()
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -258,11 +258,37 @@ impl Peer {
 		}
 	}
 
+	pub fn send_tx_kernel_hash(&self, h: Hash) -> Result<bool, Error> {
+		if !self.tracking_adapter.has(h) {
+			debug!("Send tx kernel hash {} to {}", h, self.info.addr);
+			self.connection
+				.as_ref()
+				.unwrap()
+				.send(h, msg::Type::TransactionKernel)?;
+			Ok(true)
+		} else {
+			debug!(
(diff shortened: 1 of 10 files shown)


antiochp:
Logs showing node receiving a "tx kernel hash", requesting and then receiving the full tx - 

```
20181105 13:53:01.939 DEBUG grin_p2p::protocol - handle_payload: received tx kernel: 741f8b00, msg_len: 32
20181105 13:53:01.939 DEBUG grin_p2p::peer - Requesting tx (kernel hash) 741f8b00 from peer 127.0.0.1:13414.
20181105 13:53:01.941 DEBUG grin_p2p::protocol - handle_payload: received tx: msg_len: 1842
20181105 13:53:01.942 DEBUG grin_servers::common::adapters - Received tx 5df257c4, inputs: 7, outputs: 2, kernels: 1, going to process.
20181105 13:53:01.985 DEBUG grin_pool::pool - add_to_pool [txpool]: 5df257c4 (p2p), in/out/kern: 7/2/1, pool: 0 (at block 2d2afe88)
20181105 13:53:01.991 DEBUG grin_p2p::peer - Not sending tx 5df257c4 to 127.0.0.1:13414 (already seen)
```

p2p/src/peer.rs
```diff
@@ -258,11 +258,40 @@ impl Peer {
...
+	}
+
 	/// Sends the provided transaction to the remote peer. The request may be
 	/// dropped if the remote peer is known to already have the transaction.
+	/// We support broadcast of lightweight tx kernel hash
+	/// so track known txs by kernel hash.
 	pub fn send_transaction(&self, tx: &core::Transaction) -> Result<bool, Error> {
-		if !self.tracking_adapter.has(tx.hash()) {
-			debug!("Send tx {} to {}", tx.hash(), self.info.addr);
+		let kernel = &tx.kernels()[0];
```
jaspervdm: What happens if a transaction has multiple kernels? If a node receiving this message already has only the first kernel in their pool, it won't request the full trasaction and miss out on the other kernels (which it might not all have)

antiochp: That's a good point. We were originally thinking of doing this by sending the full set of kernels but thought this was overly complex.

In normal operation we will not see aggregate txs like this because aggregation _should_ only occur during the Dandelion stem phase.

But there is nothing to stop a node from arbitrarily (maliciously or otherwise) aggregating transactions during regular tx propagation.

Say a node aggregates `tx1 (k1)` and `tx2 (k2)` to give `tx3 (k1, k2)`.
Now we have two different scenarios depending on _which_ kernel is chosen here - 
  * if `k1` then `tx3` will be ignored by nodes that are aware of `tx1`
  * if `k2` then `tx3` will be ignored by nodes that are aware of `tx2`

@ignopeverell thoughts?

Today we make an effort to "de-aggregate" aggregated txs like this - 
  * if I see `tx3` after `tx1` then I will recreate `tx2` (the difference) and broadcast `tx2`

This PR will reduce the effectiveness of this significantly.

antiochp: That said - the result of this is a node may send a "kernel hash" of an aggregated tx, only for it to be ignored by all its peers because they have seen this kernel before (and mistakenly think they have seen the full tx before).

Is this any different to a node simply not notifying its peers of the new tx?

To attempt to cause mischief with this it would be easier and cheaper just to drop txs and not notify anyone.

sesam:
:+1:
this PR potentially kills a kind of amplification attack, where variations
of tx cutthroughs could be gossiped needlessly.

also we get less honest nodes gossiping all cutthrough variations, probably
not important

antiochp:
@sesam - that's a god point actually. We're trading the benefits of seeing various aggregated combinations (and being able to deaggregate them) against simply suppressing/ignoring them on the network.

Will merge this today.

PR #1933, opened 2018-11-06
Peer rate limiting

ignopeverell:
Add some basic rate limiting on peers and ban abusers. The idea for now is to start with generous values and only kick out egregious abusers. I think we'll need this for mainnet, otherwise a large network-wide attack could cause some real damage.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -252,7 +252,8 @@ fn poll<H>(
 				let maybe_data = retry_send.or_else(|_| send_rx.try_recv());
 				retry_send = Err(());
 				if let Ok(data) = maybe_data {
-					let written = try_break!(error_tx, conn.write_all(&data[..]).map_err(&From::from));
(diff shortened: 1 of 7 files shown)


antiochp:
Do we want to consider increasing (or even temporarily disabling) the limits during sync?
If we're only connected to a few peers we can request a _lot_ of blocks or headers from a single peer in a relatively short amount of time.
And that risks cascading failures as we reduce our connected peers and pile on more requests to the remaining ones?

ignopeverell:
I think we should set the limits so that they're still higher than sync, we can't know when some other peer is going to sync with us either. So they need to be high enough but not too high (obviously cut it too close the first time). Does that make sense?

ignopeverell:
After fixing the bugs and a couple fast syncs, the max number of messages I've seen was around 200 per minute. So the current limit of 500 seems reasonable.

PR #1938, opened 2018-11-06
Remove unsafe sync/send impl for Peers

### LINKED ISSUE #1937: Peers - Concurrency Issues (Send + Sync)
We do the following in `Peers` and this is almost certainly hiding some potential (and potentially serious) concurrency issues.

```
unsafe impl Send for Peers {}
unsafe impl Sync for Peers {}
```

I assume we did this to make the compiler happy - but I suspect this is just hiding the problem and not actually resolving the underlying issue.

If we comment these two lines out then we get the following build errors -

```
...
error[E0277]: `std::sync::mpsc::Sender<()>` cannot be shared between threads safely
--> p2p/src/peers.rs:476:6  
...
error[E0277]: `std::sync::mpsc::Receiver<types::Error>` cannot be shared between threads safely
   --> p2p/src/peers.rs:476:6
...
```

From the `std::sync::mpsc` docs - 

https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html

> The sending-half of Rust's asynchronous channel type. This half can only be owned by one thread, but it can be cloned to send to other threads.

https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html

> The receiving half of Rust's channel (or sync_channel) type. This half can only be owned by one thread.

So we can fix the first compilation error by simply cloning the sender and sending it to the various peer threads. The sending is designed to be cloneable. By definition `mpsc` is multi-producer, single-consumer).

But the 2nd compilation error regarding the `Receiver` is worrying and we are just hiding it by explicitly making Peers implement `Send + Sync`.

To reiterate the docs -

> This half can only be owned by one thread.

And the error message - 

> `std::sync::mpsc::Receiver<types::Error>` cannot be shared between threads safely

----

https://doc.rust-lang.org/beta/nomicon/send-and-sync.html

> First and foremost, they're unsafe traits. This means that they are unsafe to implement, and other unsafe code can assume that they are correctly implemented.

> Since they're marker traits (they have no associated items like methods), correctly implemented simply means that they have the intrinsic properties an implementor should have. Incorrectly implementing Send or Sync can cause Undefined Behavior.

Implementing `Send + Sync` does not change _anything_ in terms of concurrency or thread safety, all we are doing is telling the rust compiler that we know more about the underlying shared data and that it should ignore any warnings/errors around sharing this data between threads.

This can (and likely will) result in _Undefined Behavior_ because we are passing a `Receiver` around between threads when the docs clearly state this should not be done.

The full compilation error is - 

```
error[E0277]: `std::sync::mpsc::Receiver<types::Error>` cannot be shared between threads safely                     
   --> p2p/src/peers.rs:476:6                                                                                       
    |                                                                                                               
476 | impl ChainAdapter for Peers {                                                                                 
    |      ^^^^^^^^^^^^ `std::sync::mpsc::Receiver<types::Error>` cannot be shared between threads safely           
    |                                                                                                               
    = help: within `peer::Peer`, the trait `std::marker::Sync` is not implemented for `std::sync::mpsc::Receiver<types::Error>`
    = note: required because it appears within the type `conn::Tracker`                                             
    = note: required because it appears within the type `enum_primitive::Option<conn::Tracker>`                     
    = note: required because it appears within the type `peer::Peer`                                                
    = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<peer::Peer>`
    = note: required because it appears within the type `(std::net::SocketAddr, std::sync::Arc<peer::Peer>)`        
    = note: required because it appears within the type `std::marker::PhantomData<(std::net::SocketAddr, std::sync::Arc<peer::Peer>)>`
    = note: required because it appears within the type `std::collections::hash::table::RawTable<std::net::SocketAddr, std::sync::Arc<peer::Peer>>`
    = note: required because it appears within the type `std::collections::HashMap<std::net::SocketAddr, std::sync::Arc<peer::Peer>>`
    = note: required because of the requirements on the impl of `std::marker::Sync` for `lock_api::rwlock::RwLock<parking_lot::raw_rwlock::RawRwLock, std::collections::HashMap<std::net::SocketAddr, std::sync::Arc<peer::Peer>>>`
    = note: required because it appears within the type `peers::Peers` 
```

* Each `Peer` has a `Tracker` 
* The `Tracker` wraps two `Senders` and a `Receiver`.
* The outer `Peers` maintains a map of `Peer` instances keyed by `SocketAddr` (the connected peers).
* The `Adapter` in `Tracker` points back to the outer `Peers`. 

So every `Peer` (and every thread spawned for each peer) points to `Peers` which points to all the other `Peer` instances.
So we have a `Receiver` (the `error_channel`) for each `Peer` shared across all the threads.
And we are not allowed to do this - as dong this is effectively "Undefined Behavior".

Starting to come around to the idea that explicitly implementing `Send + Sync` on any complex type is a pretty serious anti-pattern.

We did something similar in https://github.com/mimblewimble/grin/pull/1694
Here we were using a `RefCell` (which is explicitly not thread-safe) and then telling the compiler to ignore this by implementing `Sync + Send` for `OneTime`.
This manifested itself as some _really weird_ "Undefined Behavior" relating to non-atomic reference counting.

Currently we have two instances of this in the grin codebase - 

```
// Issue described above.
unsafe impl Send for Peers {}
unsafe impl Sync for Peers {}


// Still investigating this one but 99% we should not be doing it here either. 
unsafe impl Sync for Chain {}
unsafe impl Send for Chain {}
```

----

How to solve this?

1) I'm lacking a clear understanding of what the topology of this actually should be - in terms of `Peer` vs. `Peers` and which threads have which responsibilities etc.
2) Should we have a single central `error_channel` responsible for peer connection health?
3) Alternatively can each peer manage connection health internally in the peer specific thread?

hashmap:
This is really unsafe. Fixes #1937

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -14,8 +14,8 @@
 
 use std::fs::File;
 use std::net::{SocketAddr, TcpStream};
-use std::sync::{Arc, Mutex};
-use util::RwLock;
+use std::sync::Arc;
+use util::{Mutex, RwLock};
 
 use chrono::prelude::{DateTime, Utc};
 use conn;
@@ -152,7 +152,7 @@ impl Peer {
 	/// Number of bytes sent to the peer
 	pub fn sent_bytes(&self) -> Option<u64> {
 		if let Some(ref tracker) = self.connection {
-			let conn = tracker.lock().unwrap();
+			let conn = tracker.lock();
 			let sent_bytes = conn.sent_bytes.read();
 			return Some(*sent_bytes);
 		}
@@ -162,7 +162,7 @@ impl Peer {
 	/// Number of bytes received from the peer
 	pub fn received_bytes(&self) -> Option<u64> {
 		if let Some(ref tracker) = self.connection {
-			let conn = tracker.lock().unwrap();
+			let conn = tracker.lock();
 			let received_bytes = conn.received_bytes.read();
 			return Some(*received_bytes);
 		}
@@ -185,7 +185,6 @@ impl Peer {
 			.as_ref()
 			.unwrap()
 			.lock()
-			.unwrap()
 			.send(ping_msg, msg::Type::Ping)
 	}
 
@@ -197,7 +196,6 @@ impl Peer {
 			.as_ref()
 			.unwrap()
 			.lock()
-			.unwrap()
 			.send(ban_reason_msg, msg::Type::BanReason)
 		{
 			Ok(_) => debug!("Sent ban reason {:?} to {}", ban_reason, self.info.addr),
@@ -217,7 +215,6 @@ impl Peer {
 				.as_ref()
 				.unwrap()
 				.lock()
-				.unwrap()
 				.send(b, msg::Type::Block)?;
 			Ok(true)
 		} else {
@@ -237,7 +234,6 @@ impl Peer {
 				.as_ref()
 				.unwrap()
 				.lock()
-				.unwrap()
 				.send(b, msg::Type::CompactBlock)?;
 			Ok(true)
 		} else {
@@ -257,7 +253,6 @@ impl Peer {
 				.as_ref()
 				.unwrap()
 				.lock()
-				.unwrap()
 				.send(bh, msg::Type::Header)?;
 			Ok(true)
 		} else {
@@ -279,7 +274,6 @@ impl Peer {
 				.as_ref()
 				.unwrap()
 				.lock()
-				.unwrap()
 				.send(tx, msg::Type::Transaction)?;
 			Ok(true)
 		} else {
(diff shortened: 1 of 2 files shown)


p2p/src/peer.rs
```diff
@@ -14,7 +14,7 @@
 
 use std::fs::File;
 use std::net::{SocketAddr, TcpStream};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
```
antiochp: We should use the `Mutex` from parking_lot (as we do for `RwLock`) then we can avoid the unwrap() calls etc.

antiochp:
I'm going to tackle our one remaining unsafe send/sync (on `Chain`) with something similar.

p2p/src/peer.rs
hashmap: 👍

Issue #1909, opened 2018-11-01
Grins lost after cancelling transaction


lehnberg:
From user "Arthur" on [Grin-forum](https://www.grin-forum.org/t/lost-5k-grins-because-of-a-bug-t4/1109):

> I believe that I could possibly have lost around 5k grins because of a bug in the test net.
>
>Yesterday I was messing around with grin commands and trying all different things, when I’ve realized >that after canceling a transaction, my locked funds didn’t return to my wallet and were gone forever.
>
> Here is my current wallet txs and wallet info, 12 hours after the bug:
>https://pastebin.com/QBARnzpR 4 (I believe the bug happened at transaction 178 - line 364 of the >pastebin)
>
>And here is all the grin commands I’ve run in the session that the bug occurred:
>https://pastebin.com/69Ge0Pbe 1 (Bug happened after line 743)
>
>Another thing I’ve noticed, is that yesterday some transactions I’ve tried to make returned this error, >even though I had more than enough “currently spendable” grins in my wallet:
>
```
grin wallet send -d "http://108.196.200.233:13415" 20.0
2018-10-31T00:23:43.983935175-03:00 WARN grin - Using wallet configuration file at /home/arthur/.grin/grin-wallet.toml
2018-10-31T00:23:46.214033101-03:00 ERROR grin::cmd::wallet - Tx not created: Error { inner: 

Not enough funds. Required: 20005000000, Available: 0 }
```
>This last behaviour occurred in line 73,119,170 and 214 of the above pastebin.
>
>Have anyone ever experienced anything like this?

hashmap:
It happened to me, after cancelling a tx and some additional operations my balance was quite off. But in my case I was running a malfunctioning (for unrelated test purpose) version of grin so I decided it was my fault. I restored my balance with `grin wallet restore`.

aacoimbra:
After some tests I believe I've managed to replicate the bug.

Let's analyze a transaction where the sender sends 10 grins to himself.

Running `grin wallet send 10 -d "http://0.0.0.0:13415"` will generate two transactions in `grin wallet txs` e.g.: 

```
243  Received Tx          2c33b04d-bd11-4e72-8ffd-05b5806bb8d9  2018-11-02 02:28:03  true        2018-11-02 02:29:23  0            1             10.0             0.0             None   10.0            None 
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

PR #1939, opened 2018-11-06
Wallet self-send

yeastplume:
I think we can avoid quite a few potential issues by providing a 'self' method of wallet sends (which creates a tx within the same wallet,) and discouraging people from trying to send to localhost addresses. This should also pave the way for other direct operations people may wish to perform on the wallet, such as output consolidation or splitting an amount amongst multiple internal accounts.

### DIFF
--- a/doc/wallet/usage.md
+++ b/doc/wallet/usage.md
@@ -189,6 +189,8 @@ or, to send between accounts, use the -a flag to specify the source account:
 [host]$ grin wallet -a "my_source_account" send -m self -d "my_dest_account" 60
 ```
 
+When sending to self, the transaction will be created and posted to the chain in the same operation.
+
 Other flags here are:
 
 * `-m` 'Method', which can be 'http', 'file' or 'self' (described above). If 'http' is specified (default), the transaction will be sent to the IP address which follows the `-d` flag. If 'file' is specified, Grin wallet will generate a partial transaction file under the file name specified in the `-d` flag. This file needs to be signed by the recipient using the `grin wallet receive -i filename` command and finalized by the sender using the `grin wallet finalize -i filename.response` command. To create a partial transaction file, use:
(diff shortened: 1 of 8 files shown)


ignopeverell:
Is there some fundamental issue with sending to localhost?

antiochp:
Maybe I don't understand the problem but sending to 127.0.0.1 works here (I do it all the time).

Something like - 

```
../target/release/grin wallet send 1.5 -d "http://127.0.0.1:13415" --fluff
```

yeastplume:
Sorry, I didn't mean to imply sending to self doesn't work (it does, of course), but there are a few reasons why a 'self' send might be worthwhile:

1) It's awkward to write automated tests for a send-to-self since both sender and receiver need to have their wallets locked during the exchange
2) There are quite a few use cases have come up where doing a self-transaction without having to worry about communication might be desirable. e.g. consolidating outputs, or in the case of a pool where you want to split rewards to a bunch of user accounts in one transaction.
3) We probably want to let users move funds between accounts without having to stop listeners, restart listener with a new account, send funds from old account etc. So say someone has a 'mining' account and a 'funds' account in their wallet, it seems desirable to let them type `grin waller send -m self -a mining -d funds 1000`
4) It can prevent someone doing this: https://github.com/mimblewimble/grin/issues/1909#issuecomment-435255183, i.e. thinking they can cancel a send to self even though the transaction was posted to the chain during the operation. If the log entry is marked as self and was sent, the wallet can refuse to delete it.

Granted there are probably better ways to deal with 4, but I think 1 and 3 are justification enough for having the functionality.

We can detect attempts to send to localhost, but I'm not sure there's a logical places to specify the desired destination account. I'm sure I can think a bit to come up with something that requires minimal changes to current command line functionality.

sesam:
nice! send to self could be detected via a nonce, like hand/shake. nonce
must be independent of tx details I guess. useful? maybe useful to avoid
wallet locking up.

>

PR #1940, opened 2018-11-06
[WIP] Thread-safe access to internal bitmaps in Chain

hashmap:
Externalize all data structures with bitmaps (which are non thread-safe)  and wrap it into RwLock. Bitmaps support RwLock semantics (one writer or multiple readers). Also move unsafe Send/Sync implementation from Chain to this new data structure (FragileData) to decrease the surface of unsafe code. So instead of assuming that Chain is thread-safe we delegate this check to rust compiler and just assume that all non thread-safe code is in Fragile data, guarded by RwLock.

### DIFF
--- a/store/src/pmmr.rs
+++ b/store/src/pmmr.rs
@@ -24,7 +24,7 @@ use core::ser::{self, PMMRable};
 use leaf_set::LeafSet;
 use prune_list::PruneList;
 use types::{prune_noop, AppendOnlyFile, HashFile};
-use util::RwLock;
+use util::Mutex;
 
 const PMMR_HASH_FILE: &str = "pmmr_hash.bin";
 const PMMR_DATA_FILE: &str = "pmmr_data.bin";
@@ -55,7 +55,7 @@ pub struct PMMRBackend<T: PMMRable> {
 	prunable: bool,
 	hash_file: HashFile,
 	data_file: AppendOnlyFile,
-	data: RwLock<FragileData>,
+	data: Mutex<FragileData>,
 	_marker: marker::PhantomData<T>,
 }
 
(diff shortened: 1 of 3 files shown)


antiochp:
From the docs of the underlying croaring lib - 

https://github.com/RoaringBitmap/CRoaring#thread-safety

> Like, for example, STL containers or Java's default data structures, the CRoaring library has no built-in thread support. Thus whenever you modify a bitmap in one thread, it is unsafe to query it in others. It is safe however to query bitmaps (without modifying them) from several distinct threads, as long as you do not use the copy-on-write attribute. For example, you can safely copy a bitmap and use both copies in concurrently. One should probably avoid the use of the copy-on-write attribute in a threaded environment.

So I think we can impl `Send/Sync` in our rust code and wrap the bitmaps in a `Mutex` like you are doing here and we should be safe?

hashmap:
Thanks, sounds great. It would work fine with RwLock in this case.

store/src/pmmr.rs
```diff
@@ -54,20 +55,33 @@ pub struct PMMRBackend<T: PMMRable> {
...
+	data: Mutex<FragileData>,
+	_marker: marker::PhantomData<T>,
+}
+
+struct FragileData {
 	leaf_set: LeafSet,
 	prune_list: PruneList,
-	_marker: marker::PhantomData<T>,
 }
+// FragileData contains multiple Bitmaps inside. "CRoaring library has no built-in thread support. Thus
```
antiochp: :+1:

hashmap:
Let me check why Send/Sync were not implemented for bitmaps durectly

hashmap:
If bitmap is Sync/Send we don't need this fix https://github.com/saulius/croaring-rs/pull/38

antiochp:
@hashmap is that actually true though? I can see how marking bitmap as `Send` makes sense. But I'm not sure it would be safe to mark it as `Sync`.
If we wanted to share access between threads we would still want to wrap a bitmap in a `Mutex` (which would do the "make it Sync" for us).

hashmap:
@antiochp I don’t know yet, will test. From the docs is Sync

>  It is safe however to query bitmaps (without modifying them) from several distinct threads, as long as you do not use the copy-on-write attribute

My pr was merged, will see how it works

hashmap:
@antiochp Sync implies it's thread safe in read-only mode. I'm closing it in favor of https://github.com/mimblewimble/grin/pull/1990 but can bring it back if we see any issues

antiochp:
Ahh ok - that trips me up each time, I keep thinking `Sync` means its thread safe in general.
:+1:

PR #1943, opened 2018-11-07
API server stop

### LINKED ISSUE #1934: REST API Server Need Support Stop
Our REST API server doesn't support `stop` like other servers, which cause an annoying `Address already in use` problem when I want to stop and restart a server (in a test case which is to be added for fork syncing test and so on).

https://github.com/mimblewimble/grin/blob/0af1f13bf9ff98beae94fd7d73e5613a790e8477/servers/src/grin/server.rs#L228-L235

https://github.com/mimblewimble/grin/blob/0af1f13bf9ff98beae94fd7d73e5613a790e8477/api/src/handlers/mod.rs#L89

https://github.com/mimblewimble/grin/blob/master/api/src/rest.rs#L161-L173
```
	/// Starts ApiServer at the provided address.
	/// TODO support stop operation
	pub fn start(
		&mut self,
		addr: SocketAddr,
		router: Router,
		conf: Option<TLSConfig>,
		match conf {
			Some(conf) => self.start_tls(addr, router, conf),
			None => self.start_no_tls(addr, router),
		}
	}
```

Who would like to help to add the `stop` support?
just like https://github.com/mimblewimble/grin/blob/0af1f13bf9ff98beae94fd7d73e5613a790e8477/servers/src/grin/server.rs#L205,
https://github.com/mimblewimble/grin/blob/0af1f13bf9ff98beae94fd7d73e5613a790e8477/servers/src/grin/server.rs#L218,
https://github.com/mimblewimble/grin/blob/0af1f13bf9ff98beae94fd7d73e5613a790e8477/servers/src/grin/server.rs#L242
 and so on.

garyyu:
Close #1934

### DIFF
--- a/api/src/handlers/chain_api.rs
+++ b/api/src/handlers/chain_api.rs
@@ -147,8 +147,7 @@ impl OutputHandler {
 			.filter(|output| commitments.is_empty() || commitments.contains(&output.commit))
 			.map(|output| {
 				OutputPrintable::from_output(output, w(&self.chain), Some(&header), include_proof)
-			})
-			.collect();
+			}).collect();
 
 		Ok(BlockOutputs {
 			header: BlockHeaderInfo::from_header(&header),
--- a/api/src/handlers/mod.rs
+++ b/api/src/handlers/mod.rs
@@ -73,7 +73,7 @@ pub fn start_rest_apis(
 	peers: Arc<p2p::Peers>,
 	api_secret: Option<String>,
 	tls_config: Option<TLSConfig>,
-) -> bool {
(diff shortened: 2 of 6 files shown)


garyyu:
A confusing problem:
```
		Server::bind(&addr)
		.serve(router)
		.with_graceful_shutdown(rx)
```

`with_graceful_shutdown` works in Grin binary, but can't work in `test_p2p`, function return OK but API port not listening. No idea about the root cause of this.

hashmap:
that's where I stopped last time, it also works fine in API tests, I have a loop there to start/stop server twice, works like a charm

PR #1944, opened 2018-11-07
Fix capabilities

antiochp:
So we cannot safely add a new `capability` (and enable it) as old nodes do not recognize this and will not successfully connect as peers.

This PR fixes this problem by not enabling the new capability by default.

### DIFF
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -132,7 +132,7 @@ impl Default for P2PConfig {
 		P2PConfig {
 			host: ipaddr,
 			port: 13414,
-			capabilities: Capabilities::FULL_NODE,
+			capabilities: Capabilities::HEADER_HIST | Capabilities::TXHASHSET_HIST | Capabilities::PEER_LIST,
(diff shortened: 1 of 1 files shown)


antiochp:
I thought I'd tested this fully (#1929) with local nodes running in various combinations of capabilities.

Clearly missed the "does a legacy node safely handle an unexpected capability bits value" test case...

antiochp:
Deserialization causes problems and treats unexpected bits being set as `CorruptedData` - 

```
		let capabilities = Capabilities::from_bits(capab).ok_or(ser::Error::CorruptedData)?;

```

antiochp:
I think there is an easy fix here to handle this.

From the bitflags docs - 

https://docs.rs/bitflags/1.0.4/bitflags/index.html#methods-1

> `from_bits`: convert from underlying bit representation, unless that representation contains bits that do not correspond to a flag
> `from_bits_truncate`: convert from underlying bit representation, dropping any bits that do not correspond to flags

Issue #1947, opened 2018-11-07
cannot fully deploy "kernel first" tx broadcast without network split...


antiochp:
* We deployed the necessary changes to allow "kernel first" tx broadcast in #1929.
* There is a new bit flag on capabilities to enable this on a per-node basis (currently not enabled by default).
* But old nodes do not deserialize capabilities bitflags with the new flag set... (#1945).
  * This is fixed and made more robust in #1946 for future capability changes.

To be clear - this is not a problem specific to "kernel first" changes, it is an underlying issue with how we currently deserialize the peer capabilities bitflags on the `Hand/Shake` msgs.
We cannot add new capabilities (via currently unused bits) as old nodes will refuse to parse the new flags...

How do we roll this out across the network?
We want to enable the `TX_KERNEL_HASH` capability so nodes can advertise their support of this new style tx broadcast, but to do this we need _other_ nodes to upgrade to latest code (to safely support additional capabilities). Currently a node enabling this capability will be split off from the network as peers treat this as `CorruptedData`.

tl;dr New nodes that enable this new capability will not successfully connect to "legacy" nodes.

We just need a majority of nodes on the network with the updated capability handling code, the nodes do not need to have the `TX_KERNEL_HASH` capability enabled.

Q) Do we need to roll out a soft-fork (not at the consensus level, but at the networking level)?
i.e. 
  * Deploy code so that this capability will be automatically enabled at a specific block height
  * Notify network participants via gitter/forum
  * keep nudging everyone to update and redeploy?

antiochp:
Actually this won't even work cleanly with a soft-fork as we only advertise capabilities during the initial `Hand/Shake`...

ignopeverell:
Sounds good. Maybe we can increment the minor version in the user agent to get an idea of how many nodes have upgraded? We should also allow for unknown capabilities so we don't have to go through those shenanigans anytime we want to deploy a new one.

ignopeverell:
Hadn't seen #1944, nice job!

sesam:
breaking change -> major version increase, as per semver?

antiochp:
This is now deployed and enabled by default in config file, so new nodes will broadcast "kernel first" by default.
Existing nodes should update capabilities to `15` to enable this.

antiochp:
Logs from a node with `TX_KERNEL_HASH` enabled- 

```
[...]

PR #1950, opened 2018-11-09
bump 0.4.0 -> 0.4.1 (tx kernel broadcast support)

antiochp:
Related #1947.
Related #1949.
Once we see a majority of peers on 0.4.1 we should be safe to enable `TX_KERNEL_HASH` in our capabilities.


antiochp:
An hour before grincon0 seems as good a time as any to bump the minor version number... 💅

PR #1961, opened 2018-11-12
replace ftruncate/ftruncate64 with a simple set_len() call

antiochp:
Just took another look at our conditional `ftruncate/ftruncate64` stuff and now I'm wondering why we didn't just use `file.set_len()` originally...

Is there some reason that I'm missing that means we cannot just do this?

From the docs - 

https://doc.rust-lang.org/std/fs/struct.File.html#method.set_len

> Truncates or extends the underlying file, updating the size of this file to become size.
If the size is less than the current file's size, then the file will be shrunk. If it is greater than the current file's size, then the file will be extended to size and have all of the intermediate data filled in with 0s.
The file's cursor isn't changed. In particular, if the cursor was at the end and the file is shrunk using this operation, the cursor will now be past the end.

### DIFF
--- a/store/src/types.rs
+++ b/store/src/types.rs
@@ -250,17 +250,6 @@ impl AppendOnlyFile {
 		}
 	}
 
-	// /// Truncates the underlying file to the provided offset
-	// pub fn truncate(&self, offs: usize) -> io::Result<()> {
-	// 	let fd = self.file.as_raw_fd();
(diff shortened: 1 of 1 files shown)


ignopeverell:
The only excuse I could come up with is that it wasn't there yet but apparently it's been in since 1.0.0. So I must have just missed it...

PR #1973, opened 2018-11-13
do not treat txhashset.zip download as abusive behavior

### LINKED ISSUE #1972: Peers banned for "abusive" during fast sync (txhashset.zip)
Seeing peers getting banned for "abusive behavior" when trying to download `txhashset.zip` - 

```
20181113 12:15:52.677 DEBUG grin_p2p::peer - Asking 109.74.202.16:13414 for txhashset archive at 34824 10115b50.
20181113 12:16:32.068 DEBUG grin_p2p::protocol - handle_payload: txhashset archive for 10115b50 at 34824. size=46346754
20181113 12:17:17.411 DEBUG grin_p2p::peers - clean_peers V4(109.74.202.16:13414), abusive (6 sent, 737 recv)
```

Is `txhashset.zip` exceeding the acceptable number of bytes over a given time period?

/cc @ignopeverell 

Edit: looks like we are treating each "chunk" of txhashset.zip as an increment to the counter as bytes increase each time.
I think there is a quick fix here where we can update bytes received during txhashset.zip download without touching the counter itself.

antiochp:
Introduce `inc_quiet()` so we can increment the bytes received without touching the counter itself.

This is used *only* during `txhashset.zip` download so we don't ban the peer as abusive for sending us too many "chunks" of bytes.

Resolves #1972.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -140,7 +140,8 @@ impl<'a> Response<'a> {
 					Ok(0) => break,
 					Ok(n) => {
 						write_all(&mut self.conn, &buf[..n], time::Duration::from_secs(10))?;
-						// Increase sent bytes counter
+						// Increase sent bytes "quietly" without incrementing the counter.
+						// (In a loop here for the single attachment).
 						let mut sent_bytes = sent_bytes.write();
 						sent_bytes.inc_quiet(n as u64);
 					}
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -142,7 +142,7 @@ impl<'a> Response<'a> {
 						write_all(&mut self.conn, &buf[..n], time::Duration::from_secs(10))?;
 						// Increase sent bytes counter
 						let mut sent_bytes = sent_bytes.write();
-						sent_bytes.inc(n as u64);
+						sent_bytes.inc_quiet(n as u64);
 					}
 					Err(e) => return Err(From::from(e)),
 				}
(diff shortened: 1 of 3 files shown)


antiochp:
Hmm. Not sure this actually fixed the problem -

```
20181113 13:38:36.244 DEBUG grin_p2p::peer - Asking 163.172.177.250:13414 for txhashset archive at 34909 055fed5c.
20181113 13:39:10.974 DEBUG grin_p2p::protocol - handle_payload: txhashset archive for 055fed5c at 34909. size=48998720
20181113 13:40:01.126 DEBUG grin_p2p::peers - clean_peers V4(163.172.177.250:13414), abusive (6 sent, 699 recv)
```

antiochp:
Still seeing this intermittently so looks like the other end is also having issues - 

```
20181113 13:58:21.511 ERROR grin_p2p::protocol - handle_payload: txhashset archive save to file fail. err=Connection(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" })
```

ignopeverell:
I think this is going to lead to ever-increasing byte counts as they only get evicted with their time and the current algo expects a 1-for-1.

ignopeverell:
Couple alternative solutions:
* Do no count txhashset bytes at all.
* Insert a 0 time on `inc_quiet` and filter them out to get the `len`. This should fix the eviction problem.

antiochp:
Oh right - I hadn't realized we remove these in pairs.

antiochp:
0 time seems like it might get complicated when working out _which_ to evict (they'd get evicted immediately).

Could we update the most recent entry on `inc_quiet()` rather than adding a new one?

Issue #1968, opened 2018-11-13
Sync Performance: Support requesting kernel MMRs separate from the rest of TxHashSet


DavidBurkett:
Creating this issue to track progress on a subset of the proposals in #952. The current plan is to:
1. Support requesting Kernel MMRs in one request, and outputs + rangeproofs in a separate request. A new (temporary) capability will be added to advertise support for these new requests.
2. Return raw data for the kernel MMR instead of using BZip2, which is unnecessarily slow, and doesn't provide any real compression due to the randomness of our data.
3. Return only the leaves for the kernel MMR, which is all that's necessary for the consumer to rebuild it (since kernels are never pruned).

Related:
#1219 
#952 
#804

ignopeverell:
It would also be interesting to have the rangeproofs be requested separately for a given set of output positions (possibly through a bloom filter). The idea would be to have a client request a random subset of range proofs of its choosing for validation. This would be useful for light clients that could validate range proofs statistically by getting say only 20% of them. We could even have some light clients not request any range proof at all and trust the miners have done that validation for them.

DavidBurkett:
So you're thinking an additional message for that? We would still be required to have a separate message that gets ALL of the non-pruned leaves for the outputs and rangeproofs, correct?

antiochp:
Just reading through this now.

> Also, you can validate each batch as it comes in (as part of a Kernels message), since you know what the merklish root of it should be.

You can validate that what you received is what the sender _says_ they sent, but this won't actually validate that the kernels are "correct". To do this you must validate the kernel MMR root against the block headers.

There are a couple of existing things we can leverage here - 
  * each header commits to the following - 
    * `kernel_mmr_size`
    * `kernel_root`

We can potentially download in parallel from different peers, but to validate the kernels we can _only_ do this sequentially, by appending the kernels to the "one true" kernel MMR and verifying the root at various MMR sizes (corresponding to various headers).

Say header 1 committed to the kernel MMR root for a kernel MMR of size 7.
We could retrieve kernels, maybe all of them, maybe some chunk of them.
And then appending them incrementally to the kernel MMR, stop at size 7 and verify the root matches the kernel_root in the header.
Then apply more kernels until we hit the MMR size for the next header and verify the root again.
This way we verify the root at each header height and confirm the MMR is growing exactly as expected.

In fact we currently do this, but in reverse order as part of the txhashset.zip validation. We "rewind" back through the header chain, verifying the kernel MMR root for each header.
[...]

PR #1987, opened 2018-11-15
[WIP] Perf: Kernel Sync Performance

DavidBurkett:
Sync Performance: Support requesting kernel MMRs separate from the rest of TxHashSet
https://github.com/mimblewimble/grin/issues/1968


chain/src/pipe.rs
```diff
@@ -276,6 +276,45 @@ pub fn sync_block_headers(
...
+			kernel.hash(),
+			first_kernel_index
+		);
+	} else {
+		return Ok(());
+	}
+
+	if ctx.txhashset.num_kernels() < (first_kernel_index + 1) {
+		txhashset::extending(&mut ctx.txhashset, &mut ctx.batch, |extension| {
+			// DAVID: Rewind mmr to correct kernel index just to be safe?
```
ignopeverell: That would make sense.

chain/src/pipe.rs
```diff
@@ -276,6 +276,45 @@ pub fn sync_block_headers(
...
+			for kernel in kernels {
+				// Ensure kernel is self-consistent
+				kernel.verify()?;
+
+				// Apply the kernel to the kernel MMR.
+				extension.apply_kernel(kernel)?;
+
+				// DAVID: If kernel is last in block, validate root.
+				// Use RewindableKernelView? Or modify validate_kernel_history?
+			}
```
ignopeverell: I'd consider pushing that whole loop down in txhashset in some `apply_kernels`.

ignopeverell: My bad, looks like apply does it. But there should be some clearer separation around validation. Here you're verifying each kernel, but the whole state validation will do that again. So we either slip the latter one or this one.

DavidBurkett: Depends what you mean by whole state validation. If you're talking about after the TxHashSet is fully downloaded and processed, sure. But then we'd have to go back and start the kernel validation process all over again. If instead you're referring to the expected call to "validate_kernel_history" on line 309, then I disagree. It doesn't verify the signature of the kernel. It only verifies that the MMR root matches the header.

DavidBurkett: After re-reading, it appears you're talking about verify_kernel_signatures in Extension. Yea, I can probably skip that verify if it's already been done. It'd probably be easiest to store a 'validated' boolean on the kernel. If you're opposed to adding the extra byte (since kernels are all written to disk), I can look into other options.

sesam:
previous approaches have included (BTC example): verify twice, followed by
changing one to an assertation that is optimized out for production
builds, followed
by removing the really working check. From this, maybe keeping a bool is
wise. The bool can be stripped before serialization to disk.

DavidBurkett:
> previous approaches have included (BTC example): verify twice, followed by changing one to an assertation that is optimized out for production builds, followed by removing the really working check. From this, maybe keeping a bool is wise. The bool can be stripped before serialization to disk.

I'm not sure we can strip the bool before serialization to disk. We'd be flushing to disk shortly after first validation I believe, and wouldn't perform the second validation (where boolean is needed) until much later in the sync process when we receive the outputs in rangeproofs as part of TxHashSet download.

DavidBurkett:
@antiochp Requested that this be reviewed and merged in several phases. Closing this one and opening one for the first phase.

PR #1991, opened 2018-11-16
Various minor store crate cleanups

ignopeverell:


core/src/ser.rs
```diff
@@ -211,7 +211,7 @@ pub trait Writeable {
 
 /// Reader that exposes an Iterator interface.
 pub struct IteratingReader<'a, T> {
-	count: u64,
+	count: Option<u64>,
```
antiochp: Is the intent here to have an "open ended" iterating reader?
We would need to be very careful never to use this on anything network based as we will not know when to stop reading.

ignopeverell: Correct. In this particular PR it's used to read from a file. I'll add a note in the comments to pay attention to that particular gotcha.

ignopeverell: The functions that needed this ended up being unused now, so reverted this and removed them.

PR #1975, opened 2018-11-13
Update build.md

faddat:
added LLVM to the dependency one-liner and dependency list for 

https://github.com/saulius/croaring-rs

PR #1994, opened 2018-11-16
Fix build docs

RyanZim:
- Update rust version
- Update Spanish docs, including changes from #1975


quentinlesceller:
Thanks for the PR!

PR #1997, opened 2018-11-19
Cuckaroo Testnet4 hard fork

ignopeverell:
PoW context is now properly picked depending on the chain type,
edge bits and block height. Height const for T4 hard fork leaving
a couple weeks to have miners in place.

Removed now unused Cuckoo context.


ignopeverell:
Agreed, but this is Rust which makes it quite difficult to leak `for` loop variables out of their scope, unlike C, which is why I had initially decided against it. But I just gave it a try, assigning a new variable at each iteration and it may be a smidgen cleaner.

ignopeverell:
Well that was silly of me. Thanks for pointing out what should have been obvious...

ignopeverell:
`max_sols` is used by `CuckatooContext`, so yes. Not sure we really need it practically but that's why I didn't eliminate it entirely. I could still see it being useful in tests.

The `cuckoo_sz` was still provided as a leftover from when it wasn't in the header. Wasn't going to refactor too much in this PR but I guess it's too late now, so did that.

I've also updated the cuckatoo min sizes to 30+ on T4 and 31+ on mainnet.

Issue #1979, opened 2018-11-14
primary / secondary nomencalture


tromp:
these seem to add an unnecessarily layer to understanding our asic friendly and resistant pow,
so lets just use af and ar to denote them. For instance, secondary_scale becomes ar_scale,
also a big savings in typing.

PR #1999, opened 2018-11-20
Rename PoW from primary/secondary to AF/AR

eupn:
As suggested in #1979


garyyu:
@eupn  Could you resolve the conflicting files?

eupn:
Yes, I’ll rebase. There weren’t any conflicts at the moment when PR was created but appeared afterwards.

ignopeverell:
I'm afraid there are more conflicts now... I swear if you rebase I'll try to merge this first.

tromp:
It would make sense to use AT (ASIC Tuned/Tweaked/Targeted) instead of AF since that would match the PoW names cuckAToo and cuckARoo.

sesam:
@eupn will you use AT instead of AF as tromp suggested? It would make documenting grin easier. :+1:

PR #2005, opened 2018-11-21
Turn a Vec of bytes into a BIP39 mnemonic

mcdallas:
Adds a `from_entropy()` function to convert bytes into a BIP39 mnemonic phrase. Useful for easily creating a random valid phrase and for turning pre-existing secrets into a phrase.


yeastplume:
Thanks, looks good. I'll be integrating this into the wallet workflow over the next day or so

PR #2006, opened 2018-11-21
tweak input/output/kernel weights (1/21/3)

### LINKED ISSUE #1977: BLOCK_INPUT_WEIGHT, BLOCK_OUTPUT_WEIGHT,  BLOCK_KERNEL_WEIGHT optimization
Consider changing BLOCK_{INPUT,OUTPUT,KERNEL}_WEIGHT
from 1 / 10 / 2 to more easily motivated values like
1 / 21 / 3 which closely matches their relative sizes.

antiochp:
Resolves #1977.

I know this is tagged as `consensus breaking` but are we good merging this directly, hoping nobody mines a block close to the current limits?


jaspervdm:
We should activate it at the same time as the PoW HF.

Issue #998, opened 2018-04-23
switch commitment discussion


antiochp:
Tim Ruffing resurrected the "switch commitment" discussion on the ML - 
https://lists.launchpad.net/mimblewimble/msg00479.html

tl;dr sounds like there is a "zero cost" way of getting switch commitments in as part of the commitment itself, so we would not need to store and maintain a separate "switch commitment" on each output.

> I saw that switch commitments have been removed for various reasons.
Let me suggest a variant (idea suggested by Pieter Wuille initially):

> The switch commitment is (v*G + b*H), where b = b' + hash(v*G + b'*H,
b'*J). (So this "tweaks" the commitment, in a pay-to-contract / taproot
style).

> Before the switch, this is just used like a normal Pedersen commitment
v*G + b*H.
After the switch, users can reveal (v*G + b'*H, b'*J), and verifiers
check if it's computed correctly and use as if it were the ElGamal
commitment (v*G + b*H, b*J).

> This variant has the following features:
>  - It's as secure as the other one (a security proof will appear in my
thesis).
>  - It does not need additional space (we suggested the other one
because we thought it's anyway useful to store the hash in Grin).
>  - It does not need an additional random value (better than the
previous suggestion).
>  - b' can be derived from the private key (and then b can be derived
via the hash).
>  - It does not allow for the inclusion of arbitrary data. (You can of
course use everything as a preimage of the hash but people can always
use a hash here, no matter whether switch commitments are used or not.)
>  - I don't know the code but I don't see how this should add a lot of
complexity:  generating the commitment is a few lines more and the
verification is not changed at all.
>  - I could see how this adds complexity because the "main secret" b of
the user looks different now. But if this is an issue, it's not
necessary to tweak the commitment itself. For example, it's possible to
tweak EC points in the range proof instead, and Andrew confirmed that
does not introduce any problems or overhead in the range proofs.

> Feel free to ignore this suggestion but I'm resurrecting that
discussion yet again because I think that this switch commitments come
at basically zero cost now, and so it can be wrong to have them in the
future, even if it's unclear if they will be needed.

> Best,
Tim

antiochp:
Hey @real-or-random, tracking your suggestion here to make sure it gets discussed.

quentinlesceller:
Linking the previous PR https://github.com/mimblewimble/grin/pull/841.

antiochp:
Closing. Yay!

PR #2007, opened 2018-11-21
[DNM] Switch commitments

jaspervdm:
Add switch commitments following the scheme defined in https://github.com/mimblewimble/grin/issues/998. Blinding factors are modified to `x' = x + H(xG+vH | xJ)`

If in the future doubts arise about the existence of mechanisms that would break the discrete logarithm between `G` and `H`, we could require revealing the ElGamal commitment `(xG+vH, xJ)` (and a corresponding rangeproof) in order to spend an output.

TODO: check for other places that need to be updated, testing

Note that this will break existing wallets, as they used the unmodified secret keys. How should we handle this case? @ignopeverell @yeastplume

### DIFF
--- a/keychain/src/keychain.rs
+++ b/keychain/src/keychain.rs
@@ -68,9 +68,15 @@ impl Keychain for ExtKeychain {
 		Ok(sk)
 	}
 
+	fn derive_switch_key(&self, amount: u64, id: &Identifier) -> Result<SecretKey, Error> {
+		let ext_key = self.derive_key(id)?;
+		let key = self.secp.blind_switch(amount, ext_key.secret_key)?;
+		Ok(key)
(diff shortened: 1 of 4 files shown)


ignopeverell:
Maybe add a marker on the wallet tx log for the transactions built with switch commitments? Won't survive a restore but better than nothing.

garyyu:
I hope we can complete this PR before mainnet, refer to my post here: https://www.grin-forum.org/t/what-will-happen-if-somebody-found-that-magic-h-h-g-h/1384

jaspervdm:
@garyyu Sorry I have been pre-occupied but I will do my best to finish the PR this week, I agree we should have it before mainnet!

PR #2010, opened 2018-11-22
Translated introduction into Dutch

Borkmywork:
Translated `intro.md` into Dutch plus, I've added the link referring to the Dutch version in the original English introduction.


quentinlesceller:
Thanks for the PIR @Borkmywork ! Could you resolve the merge conflict?

PR #2009, opened 2018-11-22
Remove header cache (remove need to invalidate the cache for orphans)

antiochp:
Resolves #2008.
Resolves #1936.

We tracked the issues in #1936 down to the "header cache" not being invalidated and getting out of sync with the actual header db.
We get limited benefit from the "header cache" in general so proposing we simply get rid of it from our store.

The place we _do_ see a performance benefit with the "header cache" is when we calculate `next_difficulty` via the `DifficultyIter`. Here we iterate over 60mins of recent block headers (approx 60 headers) and having them cached is obviously faster.

I think we should tackle that separately. 
And for the "happy path" during sync or during block propagation when we process headers in order - we can simply maintain a vec of `HeaderInfo` and simply cache data needed to calc the next diff.

Other than that specific piece of processing I don't think the complexity added by maintaining the header cache is worth the performance benefits we get from it.

PR #2013, opened 2018-11-23
Cache "recent headers" for validating header difficulty

antiochp:
Related - #2009.

* get rid of `DifficultyIter` 
  * this was hiding the fact we hit the db excessively during header difficulty validation
* cache "recent headers" during block and header processing
  * leverage this cache when validating header difficulty

We *only* use these cached "recent headers" when validating header difficulty.
We only update the cache if total work has increased, which ensures the cache stays in sync with the header db.


ignopeverell:
I'm surprised LMDB and the operating system don't help us more here. In theory (and often practically) Linux is really good nowadays at figuring out which parts of the mmap'd file you read a lot and keeps all of that in memory (when there is some available).

A few things to check maybe:

* Did you have large differences in total memory use during those tests?
* Maybe the culprit is the deserialization (which the cache skips). Our headers are largish, maybe we have some inefficiencies there? I think my first bit-exact PoW deserialization wasn't super optimized. It could be good to have a header ser/deser benchmark.
* We do read/write a lot during a sync, which could bust most of the recent mmap cache. But I don't think it's the case, unless in very low memory environments.
* Perhaps the Mac OS (I'm assuming that's where you're running this) mmap implementation sucks? That's a wilder guess.

My best hunch would be deserialization though.

ignopeverell:
11 microseconds to deserialize a header. Ignore my hunch...

antiochp:
I think what is biting us here is that even though retrieving a single header from LMDB is _really fast_ this still adds up when we need to retrieve 60 of them to validate the next header.
And we need to retrieve 60 of them for _every_ header during sync.

That said - this PR got out of hand a bit I think.
I broke a chunk of the refactoring out into #2015.

Going to take another look at the rest of it and see if there is a less disruptive way of approaching this.

ignopeverell:
Okay so with 50*600 now it's 300 millisec per chunk we spend on deserializing. Not sure if that's the main slowdown. The reason I'm trying to see the main costs is that the ideal solution is to make it fast enough that we don't need to cache anything.

antiochp:
@ignopeverell I just realized how (_I think_) we can do this without caching anything.

We don't actually care about the 60 previous headers.
We only care about the total change in cumulative difficulty over those 60 headers and the time elapsed.
So we only really care about the single header at `height h-60`.

We do not need to traverse the chain back via `prev_hash` on each header. 
We can indexing into the Header MMR directly as insertion index _is_ header height. For any given leaf pos we can directly find the leaf pos for _any_ previous header.

So we replace 60 sequential db lookups with a single MMR lookup and a single db lookup.

And this should work exactly the same for header sync, block sync and regular block processing as we already validate the header MMR root for new headers in each case.

PR #2024, opened 2018-11-26
Cuckatoo size shift upgrade schedule

### LINKED ISSUE #2001: Cuckatoo size upgrade schedule
Enforce in code a yearly increase in Cuckatoo edge size.

ignopeverell:
Fixes #2001


ignopeverell:
@tromp just pushed an update that moves the schedule into `graph_weight`. It could be written more succinctly or elegantly but I wanted to first have an agreement on that method before finicking.

ignopeverell:
It is clearer. Finished up the PR and have the test passing now. Should be ready for merge.

PR #2030, opened 2018-11-27
Replace "header by height" index with reads into the header MMR 

antiochp:
The header MMR can be used as the "header by height" index.
MMR insertion index _is_ height for the header MMR.

#2013 was the impetus for this - ideally we'd like to be able to read a header at height `h-60` to validate difficulty adjustment. But we need to be able to do this for _both_ the header chain and the main block chain.

We cannot do this with the existing `header_by_height` index but we can do this with the header_mmr (tracking the main block chain) and sync_mmr (tracking the header chain during sync) pair.

Related - #2028 (we now store header hashes in the header MMR).

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -762,11 +762,17 @@ impl Chain {
 			);
 		}
 
+		//
+		// TODO - Investigate finding the "common header" by comparing header_mmr and
+		// sync_mmr (bytes will be identical up to the common header).
+		//
 		while let Ok(header) = current {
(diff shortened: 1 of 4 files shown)


antiochp:
Two servers deployed on testnet4 running code from this branch.
Will let them run over the next 24 hours to see if things hold up.

antiochp:
Both nodes still running, both in sync.
One has a stratum miner attached and mining is still happening.
Tested that chain compaction works successfully.

Just testing a final fast sync from scratch and then merging.

Issue #1980, opened 2018-11-14
secondary_pow_scaling bug and separate DAMP_FACTORs


tromp:
the first line of this function

	let snd_count = 100 * diff_data.iter().filter(|n| n.is_secondary).count() as u64;

has a bug in that the first element of diff_data should be excluded as per the comment
	// DIFFICULTY_ADJUST_WINDOW + 1 (for initial block time bound)

Also, since the count is highly variable across a mere 60 blocks, it needs its own much larger DAMP_FACTOR. Experiments suggest a value around 13 might be appropriate.

antiochp:
Damn - just saw that we count primary vs. secondary across the 60 blocks... which complicates my proposed perf improvements in #2013 (we need all 60 blocks and not just the first and last in that period).

tromp:
No problem if you add number of AR blocks as header field...
You could even store it modulo 64 in 6 bits if you want to optimize for space:-)

antiochp:
I'm really confused by this line of code + comment - 
```
	// Get the secondary count across the window, in pct (100 * 60 * 2nd_pow_fraction)
	let snd_count = 100 * diff_data.iter().filter(|n| n.is_secondary).count() as u64;
```

We count all the header infos that are `is_secondary` and then multiply this by 100.

And this seems to be very different to `in pct (100 * 60 * 2nd_pow_fraction)`.

tromp:
That  comment equates a pct with 0.01, so getting a quantity in units of pct is dividing it by 0.01, or equivalenty, multiplying by 100.
The 2nd _pow_fraction is simply the count of secondaries divided by window size (60).
Feel free to change comment to 2nd_pow_count / 0.01

antiochp:
Oh I get it now 🤦‍♂️ 

For 6 secondary blocks in a 60 block window - 

`100 * 60 * (6 / 60) == 100 * 6`

PR #2037, opened 2018-11-28
Two year secondary pow ratio fix

### LINKED ISSUE #1978: fix secondary_pow_ratio to take 2 years rather than 90 weeks (1.73 years)
90u64.saturating_sub(height / WEEK_HEIGHT)

should read

	90u64.saturating_sub(height  / (2*YEAR_HEIGHT / 90))

Note that DAY_HEIGHT and hence YEAR_HEIGHT are in fact divisible by 90.

antiochp:
Resolves #1978.

Locks the change in to activate as part of the scheduled `T4_CUCKAROO_HARDFORK (height: 64_000)` hard fork.

Q) Do we want to do this as part of this hard fork?


core/src/consensus.rs
```diff
@@ -63,7 +63,11 @@ pub const COINBASE_MATURITY: u64 = DAY_HEIGHT;
 /// function of block height (time). Starts at 90% losing a percent
 /// approximately every week. Represented as an integer between 0 and 100.
 pub fn secondary_pow_ratio(height: u64) -> u64 {
```
ignopeverell: That means we would start mainnet with the same slightly different schedule? Should we check the global chain type as well here?

ignopeverell:
Doing this with the HF is a good idea. But I think we should start with the right calculation right away on mainnet.

antiochp:
I was assuming we would be cleaning up the T4 hardfork code prior to mainnet.
But yeah - let me add the global chaintype here also just to be sure.

Agreed we should start with the correct calc from the first block on mainnet.

ignopeverell:
I don't think we'll be able to do cleanup with T4, at least as long as we keep it around.

tromp:
I don't think we should keep T4 around post launch. We should just clean up the code and launch the real testnet shortly before mainnet, so they run exactly the same code (but different genesis).

ignopeverell:
I guess that's something we need to discuss then.

antiochp:
Ah of course - we want _some_ testnet running in parallel with maninet, and I was forgetting about that.

:+1:

antiochp:
We activate this at `T4_CUCKAROO_HARDFORK` which is in a couple of days. Is this long enough to reasonably expect the network to upgrade by then?

@ignopeverell @tromp

tromp:
It's a little too soon, given that https://github.com/mimblewimble/grin/issues/1980 is not part of the HF yet. I would suggest delaying it by about a week. By that time we might have a cuckaroo cpu miner ready as well...

antiochp:
Hmm ok. I'll add a new hardfork height and activate this (and presumably #1980) at that point.

We'll need to work out what do do with anybody who deployed between this PR getting merged and the new hardfork height being introduced...

tromp:
Right; people who don't update to the new HF height in time will find themselves on a minority fork...

ignopeverell:
From Gitter:

> I wouldn't mind merging mimblewimble/grin#2050 now for the HF tomorrow, creating a little mayhem on T4 wouldn't be a bad thing, and make sure things behave appropriately in the few days after that

Issue #1348, opened 2018-08-12
cleanup pmmr_leaf snapshot files (from fast-sync)


antiochp:
Right now during fast-sync the node we ware sync'ing _from_ generates a "snapshot" of the pmmr_leaf file based on a rewound txhashset.
These get saved to the local filesystem and are never cleaned up.

Probably makes sense to clean these up, maybe as part of `check_compact()` that runs on a periodic basis.
We want to be careful that we don't `rm` anything _during_ a sync to avoid sending corrupted data across, but these never get used again once sync has completed. 

```
ls -Alt .grin/txhashset/output/
total 5932
-rw-r--r-- 1 root root 1989068 Aug 12 20:09 pmmr_data.bin
-rw-r--r-- 1 root root   16408 Aug 12 20:09 pmmr_leaf.bin
-rw-r--r-- 1 root root 3743904 Aug 12 20:09 pmmr_hash.bin
-rw-r--r-- 1 root root   16408 Aug 12 19:28 pmmr_leaf.bin.1084a7f2
-rw-r--r-- 1 root root   16408 Aug 11 12:25 pmmr_leaf.bin.0114b520
-rw-r--r-- 1 root root   16408 Aug  7 14:14 pmmr_leaf.bin.02cdd912
-rw-r--r-- 1 root root   16408 Aug  7 03:16 pmmr_leaf.bin.03437f9a
-rw-r--r-- 1 root root   16408 Aug  6 08:32 pmmr_leaf.bin.09a0b9f6
-rw-r--r-- 1 root root    8208 Jul 21 00:29 pmmr_leaf.bin.0a94f33f
-rw-r--r-- 1 root root    8208 Jul 20 02:20 pmmr_leaf.bin.1089dfac
-rw-r--r-- 1 root root    8208 Jul 19 20:45 pmmr_leaf.bin.089cb892
-rw-r--r-- 1 root root    8208 Jul 19 03:51 pmmr_leaf.bin.325aae69
-rw-r--r-- 1 root root    8208 Jul 18 16:42 pmmr_leaf.bin.149dc380
-rw-r--r-- 1 root root    8208 Jul 18 11:24 pmmr_leaf.bin.1d9ffe1e
-rw-r--r-- 1 root root    8208 Jul 18 09:04 pmmr_leaf.bin.242d170c
-rw-r--r-- 1 root root    8208 Jul 17 07:25 pmmr_leaf.bin.19125da8
[...]

PR #2049, opened 2018-11-29
Issue #1348 - cleanup pmmr_leaf snapshot files (from fast-sync) [WIP]

rentenmark:
This is a WIP PR for #1348 

* Expose some constants and import them
* add function for calculating the path and implement
* add funciton signature for deleting the files

This is my first PR and I know that is probably going to be quite obvious. I wanted to get this put up before I get much further for feedback on general direction before I continue.

All feedback will be well received, I have a thick skin 🙇

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -20,9 +20,11 @@ use std::net::SocketAddr;
 use std::sync::{Arc, Weak};
 use std::thread;
 use std::time::Instant;
+use std::path::{Path, PathBuf}; 
 use util::RwLock;
 
 use chain::{self, BlockStatus, ChainAdapter, Options};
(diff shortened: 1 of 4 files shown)


servers/src/common/adapters.rs
```diff
@@ -531,6 +534,20 @@ impl NetToChainAdapter {
 		}
 	}
 
+	fn clean_rewind_files(&self) {
```
antiochp: Maybe consider moving this to a fn on `txhashset`?

As `adaptor.check_compact()` calls `chain.compact()` which ends up calling `txhashset.compact()`.

That way we don't need to expose `TXHASHSET_SUBDIR` and `OUTPUT_SUBDIR` and we can keep the file related operations confined to `txhashset`.

rentenmark:
closed in favor of https://github.com/mimblewimble/grin/pull/2054

PR #2050, opened 2018-11-29
[WIP] additional hard fork at 65,080 for secondary pow ratio changes

antiochp:
Related - #2037.

As discussed in #2037 it was maybe too aggressive to try and get this into the scheduled hard fork at height `64,000`.

This PR introduces a subsequent hard fork a week later at height `64,000 + 10,080 = 65,080`.

Regarding anyone who deployed the interim code that schedules this changes to activate (prematurely) at `64,000`, are we happy with just warning people via gitter and hoping they update?

PR #2051, opened 2018-11-29
Fix secondary pow scaling bug

### LINKED ISSUE #1980: secondary_pow_scaling bug and separate DAMP_FACTORs

antiochp:
Resolves #1980.

* Add `skip(1)` to exclude the first header_info
* Add `SECONDARY_COUNT_DAMP_FACTOR`

TODO - 
- [x] What height to use for the HF? (Related - #2050). Single HF at 64,000.
- [x] rename
- [x] height dependent on skip()

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -75,14 +75,16 @@ pub fn secondary_pow_ratio(height: u64) -> u64 {
 	}
 }
 
-fn secondary_count_damp_factor(height: u64) -> u64 {
+/// The AR scale damping factor to use. Dependent on block height
+/// to account for pre HF behavior on testnet4.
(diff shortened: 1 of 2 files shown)


antiochp:
Merging. HF here we come.

@ignopeverell shout loudly if you disagree.

ignopeverell:
All good, thanks!

PR #2052, opened 2018-11-29
Keybase wallet plugin

mcdallas:
Adds a new wallet communication adaptor to exchange slates via keybase chat.

`grin wallet send -m keybase -d <recipient> <amt>` for sending (only synchronously for now)
`grin wallet listen -m keybase` initialises a keybase listener


src/bin/grin.rs
```diff
@@ -198,6 +198,13 @@ fn real_main() -> i32 {
 				.short("l")
 				.long("port")
 				.help("Port on which to run the wallet listener")
+				.takes_value(true))
+			.arg(Arg::with_name("method")
+				.short("m")
+				.long("method")
+				.help("Which method to use for communication")
+				.possible_values(&["http", "keybase"])
+				.default_value("http")
```
quentinlesceller: I believe there is something wrong here. Arg `method` is already defined at line 241.

ignopeverell:
So are we merging plug-ins in the wallet now? Isn't the point of a plug-in that it can be plugged-in (or not)? Or is this meant as a demo plug-in of sorts?

yeastplume:
I hadn't intended this quite yet, the refactoring is done to allow plugins to exist, but the infrastructure bits to externalise them aren't. However, this is a good sample use-case and will help ensure the plug-in interface is actually useful, so not against having this as a demo plugin (that comes with 'experimental' warnings)

mcdallas:
@ignopeverell my understanding from the meeting is that it was decided to have two "official" ones for http and keybase and the rest could be external or even seperate crates

mcdallas:
I fixed some issues (thanks @mably and @quentinlesceller for helping me test it) and it seems to work fine. Only issue I am not 100% sure it was fixed is a crash when the keybase channel between the two parties was uninitialised. Will have to test with someone I haven't tested before to make sure.

yeastplume:
Think it looks great and will be a nice little addition (especially now it's using the wallet API directly). Would definitely like to hear @ignopeverell's concerns as he doesn't seem convinced.

To be clear, I think this would be okay as a 'use at your own risk' example. There's an action following from this to externalise plugins somehow, so they (and any dependencies they might want to pull in) don't end up bloating grin's dependency tree.

Also @mcdallas, if we're okay to merge this would you mind documenting it's use somewhere in `docs\usage\wallet.md`?

mably:
One thing I find a bit annoying is that, even if only the keybase listener is required to make a keybase tx, you still need to start the http listener if you want to have access to all the other wallet information like txs or outputs.

It could be interesting to be able to start multiple listeners in one single command.

ignopeverell:
@yeastplume I'm fine with merging this as long as we have an understanding that the purpose is to have a good couple of plugin examples. I would just object to start accumulating plugins under the grin wallet tree.

yeastplume:
All good then, thanks for this @mcdallas

PR #2054, opened 2018-11-30
[WIP] cleanup pmmr_leaf snapshot files (from fast-sync) 

rentenmark:
I took the suggestion of @antiochp in #2049 

* calculate the path on txhashset
* expose a constant for use later

I ran cargo format because I got a message from the pre-commit hook but it went way overboard. Not sure if I have the git-fu to undo the cargo format..

This is still a work in progress.


rentenmark:
closed in favor of #2061

PR #2055, opened 2018-11-30
FIX: Attempt at fixing etc/Dockerfile

npinto:
`docker build -t grin etc/` now works ;-)

### DIFF
--- a/etc/Dockerfile
+++ b/etc/Dockerfile
@@ -1,7 +1,7 @@
 # Multistage docker build, requires docker 17.05
 
 # builder stage
-FROM rust:1.29.0 as builder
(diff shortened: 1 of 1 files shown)


quentinlesceller:
Thanks for the PR @npinto. However, I don't this is the correct solution. Currently it is not allowed by docker to copy file from the parent dir so I just think we should update the instructions with this command instead of cloning again.

```
docker build -t grin -f etc/Dockerfile .
```

I'll open a PR shortly.

quentinlesceller:
Fixed by https://github.com/mimblewimble/grin/pull/2059 closing.

PR #2059, opened 2018-11-30
Fix and update Docker image

quentinlesceller:
- Update rust version to 1.30.1.
- Update build instructions.
- Formatting.

### DIFF
--- a/etc/Dockerfile
+++ b/etc/Dockerfile
@@ -1,18 +1,18 @@
 # Multistage docker build, requires docker 17.05
 
 # builder stage
-FROM rust:1.29.0 as builder
+FROM rust:1.30.1 as builder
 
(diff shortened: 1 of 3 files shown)


etc/Dockerfile
```diff
@@ -1,18 +1,18 @@
...
-        git
+    clang \
+    libclang-dev \
+    llvm-dev \
+    libncurses5 \
+    libncursesw5 \
+    cmake \
+    git
 
 WORKDIR /usr/src/grin
```
kargakis: I don't think the COPY below works as expected now that the Dockerfile has moved in /etc.

quentinlesceller: Actually it just works. Tried it several times.

PR #2065, opened 2018-12-03
[WIP] Move libtx to its own crate

thehouseelf:
The `libtx` module in the `wallet` crate seems much more low-level than the wallet. Many tests depend on it (including `core`) and so I think it might be cleaner to give it its own crate. It would also be necessary to have a build of grin without the wallet. Alternatively `libtx` could be moved into `core` but I was hesitant to increase the scope of the `core` crate.

In addition, should I rename `libtx` or is it the right name?


yeastplume:
We've known about this for a while and agree the situation is not idea. There are quite a few dependencies other than this that may need to be moved around and cleaned up (the types in core may be another example of something that needs to be factored out). 

I'm not 100% sure what the best solutions is to some of these dependency issues, and thus not yet sure creating yet another crate is the right solution. (Also some knock-on effects to creating new crates that have to be considered in the crates.io build) Happy to have the discussion though, so would you mind if we left this here as a discussion point?

thehouseelf:
I am happy to discuss this, I am not sure what the best option is either. As mentioned I can also just move it to `core`, to avoid another crate? I don't see another option than giving it its crate, or moving it to another. It does seem that dependencies would work better with it in `core` than in `wallet`.

ignopeverell:
I have to say I'm a little partial to this PR because it's the first step for 2 things I want very much:

1. Not recompile everything I try to test some development on core (changing a source in core and running tests, which depend on wallet, rebuilds everything as it all depends on core).
2. A lightweight binary that just includes a node.

I might have a slight preference for putting this in core instead of giving it its own crate. Beside some of our internal crates (like chain), I can't think of a use case that would require core but not libtx. And even for our crates, tests end up requiring libtx. I think it'd also be nice for some lower-level transaction building libraries to only have to depend on core to get something going.

yeastplume:
It should slot easily enough into core, given it just has ‘keychain’ and ‘util’ as dependencies (same as core). I think I’d prefer it there as well as opposed to a new crate

thehouseelf:
I will update the PR to move `libtx` under `core` then.

thehouseelf:
Replace by #2080

PR #2067, opened 2018-12-03
Refactor wallet commands

yeastplume:
Effort to split wallet command line parsing from actual functionality.


yeastplume:
Will test a bit more then merge. There may still be quite a few non-ideal error outputs when things go wrong, but argument parsing and error handling for the wallet command line have been made much more consistent so it should be easy to fill those in. Intend to automate testing of the command line + all flags and error conditions shortly.

PR #2072, opened 2018-12-03
cleanup unnecessary reset_head calls

### LINKED ISSUE #2071: Receiving a bad header (invalid pow say) while syncing resets header sync
If we receive a "bad" block header via broadcast we do the following - 

https://github.com/mimblewimble/grin/blob/a24d613b0541adaeeb226532f8d7e30899995d70/servers/src/common/adapters.rs#L216-L223

The call to `chain().reset_head()` resets the header head (and sync head) to the current chain head.

This is pretty disruptive if we happen to be in header sync at the time, we reset the header sync and start syncing headers from genesis again. 😞

antiochp:
Resolves #2071.

Make `reset_header_head()` and `reset_sync_head()` more explicit.
We don't need to call `reset_head()` if we receive a bad header or block, the head will not be updated in this scenario.
Reorder the checks in `validate_chain()` so the fastest one (checking the config) comes first, so we skip chain validation quickly.

### DIFF
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -66,9 +66,6 @@ impl HeaderSync {
 					header_head.height,
 				);
 
-				// Reset sync_head to the same as current header_head.
-				self.chain.reset_sync_head(&header_head).unwrap();
-
 				// Rebuild the sync MMR to match our updated sync_head.
 				self.chain.rebuild_sync_mmr(&header_head).unwrap();
 
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1014,14 +1014,6 @@ impl Chain {
 		self.orphans.len()
 	}
 
-	/// Reset header_head and sync_head to head of current body chain
-	pub fn reset_head(&self) -> Result<(), Error> {
-		let batch = self.store.batch()?;
-		batch.reset_head()?;
-		batch.commit()?;
-		Ok(())
-	}
-
 	/// Tip (head) of the block chain.
 	pub fn head(&self) -> Result<Tip, Error> {
 		self.store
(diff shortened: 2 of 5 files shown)


servers/src/grin/sync/header_sync.rs
```diff
@@ -66,9 +66,6 @@ impl HeaderSync {
 					header_head.height,
 				);
 
-				// Reset sync_head to the same as current header_head.
-				self.chain.reset_sync_head(&header_head).unwrap();
```
garyyu: Why we remove the `reset_sync_head` here?  it broke those syncing logic we have discussed heavily before.

garyyu:
On latest Foonet, I see the following log:
```
20181223 20:56:39.314 DEBUG grin_servers::grin::sync::header_sync - sync: initial transition to HeaderSync. sync_head: 012822c5 at 3681, reset to: 012f1c9b at 3698
20181223 20:56:39.314 DEBUG grin_chain::txhashset::txhashset - About to rebuild header extension from cb272478 to 012f1c9b.
20181223 20:56:39.381 DEBUG grin_chain::txhashset::txhashset - Truncating header extension.
20181223 20:56:39.381 DEBUG grin_chain::txhashset::txhashset - Re-applying 3698 headers to extension, from 0036703e to 012f1c9b.
20181223 20:56:39.507 DEBUG grin_servers::common::types - sync_state: sync_status: NoSync -> HeaderSync { current_height: 3698, highest_height: 3698 }
20181223 20:56:39.533 DEBUG grin_servers::grin::sync::header_sync - sync: locator : [(3681, 012822c5), (3679, 016c2a3a), (3675, 00c81a25), (3667, 021dba31), (3651, 00b63591), (3619, 00137f28), (3555, 01c23db0), (3427, 004f1c63), (3171, 01ef13c7), (2659, 02887317), (1635, 024f1aee), (0, cb272478)]
```

then I realized that `reset to: 012f1c9b at 3698` doesn't work anymore.  @antiochp  could you remind me why we remove the `reset_sync_head` here?

servers/src/grin/sync/header_sync.rs
antiochp: Hmm. This may have been an unintentional change when we were cleaning up the "reset head" code.
We _do_ need something here don't we.

PR #2074, opened 2018-12-04
Fix verify_messages function

quentinlesceller:
Which also fixes `finalize` error such as https://pastebin.com/YMdN6yRm.


jaspervdm:
Feels kind of weird to have to lock the wallet to verify a bunch of signatures. Could we not just create a new `secp` instance and pass that to `verify_messages`?

yeastplume:
@jaspervdm yes, good point

quentinlesceller:
True let me see if I can fix that.

Issue #2053, opened 2018-11-29
Genesis release process and script


ignopeverell:
We need to automate the genesis block release process so it's relatively quick and minimizes last moment errors (done a few of these on testnets). Writing it down in code will also help in normalizing what it should look like and avoid bad surprises.

The process itself should be the following:

1. Prepare a multisig output and kernel owned by the council treasurers for the coinbase. This can be down a few days prior.
2. Grab the latest bitcoin block hash.
3. Build a genesis block with the prepared coinbase and the bitcoin block hash as `prev_root`. The timestamp of the block should be 30 min ahead to leave enough time to run a build and start a node.
4. Mine the block so we have at least a valid Cuckatoo29 cycle. We don't require a given difficulty for that solution.
5. Finalize the block with the proof-of-work, inserting a high-enough difficulty (to be agreed upon separately).
6. Commit the block to github.
7. Tag version 1.0.0 (scary).
8. Dig a hole, hide in it for the next 48h.

Let me know if I missed anything.

antiochp:
We should check the current code will actually handle a non-empty genesis block. 

I suspect there are some assumptions in the code around there being no outputs in there, particularly with the "supply/emission" logic relative to block height.

mcdallas:
> 
> 
> Where do we sign up for one of the juicy genesis grins and/or our name carved into the concrete?

Merkle root of all github contributors!

PR #2079, opened 2018-12-05
Support for genesis block that includes a coinbase

ignopeverell:
Related to #2053


chain/tests/mine_simple_chain.rs
```diff
@@ -60,14 +61,52 @@ fn setup(dir_name: &str, genesis: Block) -> Chain {
...
+#[test]
+fn mine_genesis_reward_chain() {
+	global::set_mining_mode(ChainTypes::AutomatedTesting);
+
+	// add coinbase data from the dev genesis block
+	let mut genesis = genesis::genesis_dev();
+	let keychain = keychain::ExtKeychain::from_random_seed().unwrap();
+	let key_id = keychain::ExtKeychain::derive_key_id(0, 1, 0, 0, 0);
+	let reward = reward::output(&keychain, &key_id, 0, 0).unwrap();
+	genesis.body.outputs.push(reward.0);
```
antiochp: Noticed this and wondered if we could do something similar to the "factory methods" on transaction - 

https://github.com/mimblewimble/grin/blob/8d8f533b8e8d8df77ddb800f7f8366ada4b617fc/core/src/core/transaction.rs#L798-L806

That way we could avoid making `body` public.

That said maybe we don't want to worry too much about it right now as we are still iterating on exactly how this genesis block is going to be constructed in mainnet. 

We haven't really had a need to add arbitrary outputs/kernels to a block before. This will likely be clearer later.

chain/src/chain.rs
```diff
@@ -1247,16 +1249,33 @@ fn setup_head(
...
+				};
+				txhashset::extending(txhashset, &mut batch, |extension| {
+					extension.apply_block(&genesis)?;
+					extension.validate_roots()?;
+					extension.validate_sizes()?;
+					Ok(())
+				})?;
+			} else {
+				// Initialize our header MMR with the genesis header.
+				txhashset::header_extending(txhashset, &mut batch, |extension| {
```
antiochp: Could we just `apply_block()` in both cases (genesis empty or non-empty) and have `validate_roots()` and `validate_sizes()` do the "right thing" for an empty genesis block?

chain/src/txhashset/txhashset.rs
```diff
@@ -1242,8 +1242,9 @@ impl<'a> Extension<'a> {
 	/// from the respective MMRs.
 	/// For a significantly faster way of validating full kernel sums see BlockSums.
 	pub fn validate_kernel_sums(&self) -> Result<((Commitment, Commitment)), Error> {
+		let genesis = self.get_header_by_height(0)?;
 		let (utxo_sum, kernel_sum) = self.verify_kernel_sums(
-			self.header.total_overage(),
+			self.header.total_overage(genesis.kernel_mmr_size > 0),
```
antiochp: This should probably be based on chain type eventually?

chain/tests/mine_simple_chain.rs
ignopeverell: Great idea. I've added a `with_reward` for that purpose. Not that there was already one that was more of a constructor, and renamed it `from_reward` which I think conveys what it does better. Perhaps we should try to use `from_` everywhere when the semantic is closer to a `new`.

chain/src/chain.rs
ignopeverell: Also great idea, looks like using `apply_block()` just works. The 2 validation methods do nothing for the genesis block and I think that's fair for now. This way we can use the root to embed a hash.

chain/src/txhashset/txhashset.rs
ignopeverell: That was my first inkling as well but it makes it a pain for tests. Tests should all be running in AutomatedTesting mode but some use an empty genesis while others won't.

PR #2084, opened 2018-12-05
Convert to Rust 2018 edition

hashmap:
* Remove `extern crate`, leave it just for macros import. It could be done later, but instead of 2 line we need to add `use crate:macro` in each file which uses the macro
* Proper absolute path support
* Get rid of `mod.rs` (have some issue with couple tests)

^ https://rust-lang-nursery.github.io/edition-guide/rust-2018/module-system/path-clarity.html

* Dyn trait for trait objects 
https://rust-lang-nursery.github.io/edition-guide/rust-2018/trait-system/dyn-trait-for-trait-objects.html


hashmap:
Ah, can't keep with updates:) will merge one more time

hashmap:
Yeah, this pr oxidizes very quickly:). If no objection I will merge it after resolving conflicts. 

No problems with stable rust, it was released yesterday as part of 1.31, it’s why I removed WIP and sacrified a good chunk of the night to fix conflicts. It’s green on Travis, I assume we use the latest stable Rust

sesam:
Always be upgrading :). I guess `cargo fix` can help with part of this upgrade. Is there any pre-2018 version or utility that helps with discovering code deprecations?

hashmap:
@sesam 80% of this pr is done by `cargo fix` (twice per crate, plain and idioms). It fails on `tests` crates though, so there is some hand crafted code. 
I find `clippy` useful and `cargo fix` may be used with any rust code (without —edition) if I’m not mistaken

clefru:
@ignopeverell @hashmap 

With all due respect, is this a crypto-currency project or a Rust language demo? I have to ask again to STOP requiring bleeding edge Rust. Rust 1.31.0 was released 2 days ago. I have to drop my general politeness on GitHub issues and say: This is an **insane** requirement.

No, I am not going to use "rustup". It's a waste of time for everyone to not just keep up to date with Grin, but also update with Rust. That's what we have distros for. Also let me remind you that rustup recommends you to do: "curl https://sh.rustup.rs -sSf | sh". I am surprised that from a security point of view that's still a thing in 2018.

Please revert.

sesam:
@clefru Thank you for your concern :). Generally I think you're both right and wrong; older versions have been more tried, but newer versions might contain fixes for yet-to-be-publicly-announced critical bugs. So I think we're stuck with Always be upgrading. Maybe it would suit you to use grin via Docker? That way you can be more certain that grin is built with the same libraries that others use, and with minimal risk of conflicting or tainting your primary OS. Come see us in https://gitter.im/grin_community/dev or https://gitter.im/grin_community/Lobby and we can discuss this more in depth :).

clefru:
@sesam Thanks for your nice words. I have filed a separate issue here: https://github.com/mimblewimble/grin/issues/2103 as it was requested on gitter.

hashmap:
@clefru thanks, I totally understand your point. Our goal in this case is not to blindly update Rust to the latest, but increase dev productivity and make Grin more inclusive for people without Rust background. That's the reason why we need Rust 2018 [1]:

> The theme of Rust 2018 is productivity. Rust 2018 improves upon Rust 2015 through new features, simpler syntax in some cases, a smarter borrow-checker, and a host of other things. These are all in service of the productivity goal. Rust 2015 was a foundation; Rust 2018 smooths off rough edges, makes writing code simpler and easier, and removes some inconsistencies.

Non-lexical lifetimes [2] is a game changer for Rust.  I'd argue it would sufficiently lower the barrier for new contributors.

For sure after the mainnet release we will be much more conservative, so it was our the last chance to switch to Rust 2018. It's why we switched to rust 1.31 as fast as possible -  to have more time to test it. We may be insane but it was a pragmatic decision in this particular case, discussed in September.

[1] https://rust-lang-nursery.github.io/edition-guide/rust-2018/index.html
[2] https://rust-lang-nursery.github.io/edition-guide/rust-2018/ownership-and-lifetimes/non-lexical-lifetimes.html

PR #2093, opened 2018-12-07
Script to publish to crates.io

ignopeverell:


etc/crate-release.sh
```diff
@@ -0,0 +1,31 @@
...
+
+echo
+crates=( util keychain core store chain pool p2p api wallet config servers )
+
+for crate in "${crates[@]}"
+do
+	echo "** Publishing $crate"
+	cd $crate
+	cargo package
+	cargo pubish
```
mcdallas: *publish

ignopeverell: Last min typo, thanks!

PR #2094, opened 2018-12-07
[WIP] Header info total difficulty

antiochp:
This PR reworks `HeaderInfo` to maintain `total_difficulty` and not `difficulty` for consistency with `BlockHeader` itself.

This allows us to build header_infos from headers directly, without needing to iterate over previous/current headers to calculate the difficulty change per header.

Why would this be preferable? 

Right now we can only build a `header_info` from the header and its previous header. This need for pairs of headers complicates the code in various places as we need access to the db to obtain the previous header.

The `DifficultyIter` for example is relatively complex - we need to pass an active `batch` or `store` into the iterator to allow us to iterate back over these previous headers.

Being able to build header_infos from instances of header will allow us to to simply retrieve a vec of headers from the db and then convert them, via `into()` to `header_infos` as part of the difficulty validation step later on. And we will not need an iterator with an active db batch.

It also opens up the possibility of storing `header_infos` (or something similar) in the header MMR (right now we just store header hashes). This would potentially allow us to validate header difficulty looking only at the header MMR and without needing to retrieve headers from the db.

We could use the header MMR effectively as the "cache" of header infos, knowing that it was always consistent with the current chain head.

Related - #2013 (attempt at caching headers for difficulty validation).

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -404,11 +404,11 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext) -> Result<(), E
 		let child_batch = ctx.batch.child()?;
 		let diff_iter = store::DifficultyIter::from_batch(prev.hash(), child_batch);
 		let next_header_info = consensus::next_difficulty(header.height, diff_iter);
-		if target_difficulty != next_header_info.difficulty {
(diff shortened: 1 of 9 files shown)


antiochp:
Closing this. Still doesn't feel right.

PR #2104, opened 2018-12-08
Sign kernel features

### LINKED ISSUE #2060: kernel features
Some notes from grin/dev conversation:

John Tromp @tromp 21:43
in fact, if miner not only has coinbase, but also regular tx in block, then he could combine kernels into one
so notion of coinbase kernel makes no sense?!

John Tromp @tromp
we shld get rid of coinbase flags in kernels though
and include kernel features in kernel_sig_msg
maybe it should be kernel_type rather than just a set of bits
256 different types offers more flexibility than 8 orthogonal features

David Burkett @DavidBurkett 23:53
Yep, I agree

John Tromp @tromp 23:55
we only have two types now; plain and (absolute) height locked; only latter needs to provide lock height
after launch we'll need to add type relative height locked, which will require kernel reference and height offset in message
that would be more awkward with feature bits

antiochp:
For mainnet the message to be signed is `hash(fee || lock_height || features)`.
For testnet4 we have the existing `(fee || lock_height)`.

The `slate` has no way of passing kernel features around between participants, so we are limited to using "default kernel" features, but this is fine for now as this is the only kind of tx kernel other than coinbase today.

Question 1 - does it make sense to hash the message before we sign it to give us a 32 byte message.
Is there a reason _not_ to hash the message?

Question 2 - do we want to HF this in to testnet4 at some block height? Do we really want code that is effectively untested, activating just on mainnet?

Other changes as part of this PR - 
* added `as_bytes()` to `Hash`
* removed `pub` on the internal byte array in `Hash` (see above)
* cleaned up `Readable/Writeable` impl for `TxKernel` and `TxKernelEntry`
* added `Writeable` for `KernelFeatures` (we need this to hash the message to be signed)

Resolves #2060.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -18,6 +18,7 @@ use crate::consensus;
 use crate::core::hash::Hashed;
 use crate::core::verifier_cache::VerifierCache;
 use crate::core::{committed, Committed};
+use crate::global;
 use crate::keychain::{self, BlindingFactor};
 use crate::ser::{
 	self, read_multi, FixedLength, PMMRable, Readable, Reader, VerifySortedAndUnique, Writeable,
@@ -43,6 +44,13 @@ bitflags! {
 		const DEFAULT_KERNEL = 0b00000000;
 		/// Kernel matching a coinbase output
 		const COINBASE_KERNEL = 0b00000001;
+	}
+}
+
+impl Writeable for KernelFeatures {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
+		writer.write_u8(self.bits())?;
+		Ok(())
 	}
 }
 
@@ -163,13 +171,13 @@ impl ::std::hash::Hash for TxKernel {
 
 impl Writeable for TxKernel {
 	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
+		self.features.write(writer)?;
 		ser_multiwrite!(
 			writer,
-			[write_u8, self.features.bits()],
 			[write_u64, self.fee],
-			[write_u64, self.lock_height],
-			[write_fixed_bytes, &self.excess]
+			[write_u64, self.lock_height]
 		);
+		self.excess.write(writer)?;
 		self.excess_sig.write(writer)?;
 		Ok(())
 	}
@@ -207,7 +215,7 @@ impl TxKernel {
 	/// The msg signed as part of the tx kernel.
 	/// Consists of the fee and the lock_height.
 	pub fn msg_to_sign(&self) -> Result<secp::Message, Error> {
-		let msg = kernel_sig_msg(self.fee, self.lock_height)?;
+		let msg = kernel_sig_msg(self.fee, self.lock_height, self.features)?;
 		Ok(msg)
 	}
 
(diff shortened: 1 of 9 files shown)


tromp:
> For mainnet the message to be signed is hash(fee || lock_height || features).

This differs from what I proposed, which is for features to be a enumeration of (up to 256) tx types
rather than a set of (only 8) bits. Also, the lock_height field is only needed for the tx type "height_locked", whereas the vast majority of tx will be of tx type "plain" and can enjoy an 8-byte size reduction.

antiochp:
> This differs from what I proposed, which is for features to be a enumeration of (up to 256) tx types
rather than a set of (only 8) bits.

We already have the existing implementation of kernel features. I'm not sure we gain anything in the short-term by reimplementing this as a richer set of kernel types?

> Also, the lock_height field is only needed for the tx type "height_locked", whereas the vast majority of tx will be of tx type "plain" and can enjoy an 8-byte size reduction.

Technically yes. But we only support "fixed length" serialization into our PMMRs and we store the kernels in the kernel MMR. So we could save 8 bytes "on the wire", but then we'd need to support multiple serialization strategies to handle them in the PMMR.

Just to be clear, not opposing your proposal but this PR was not intended to fully implement everything described there. It is only focused on adding the kernel features to the message being signed, which puts us a step closer to having "relative lock height" support.

antiochp:
> Why the extra hashing?

The secp256k1 lib assumes a msg is 32 bytes on length.
We jumped through some hoops to put the bytes for fee and lockheight into a vec of 32 bytes.

But I'm wondering if this would just be simpler if we hashed our "message" to get 32 bytes that we then pass into secp256k1 to sign. 

https://github.com/mimblewimble/rust-secp256k1-zkp/blob/d26b7640f935a5c34a97f987f823a97a564ea7ed/src/lib.rs#L421

antiochp:
> Won't we need to support variable length kernels in future for kernel-relative height locks?

That's interesting yes.
In my branch we're implementing the "relative kernels" via an additional "extra" MMR file.
The kernels are added to the hash_file + data_file. And the optional relative kernels are added to the extra data_file.
So we handle the fixed length portion of the kernels in the kernel MMR.
And the optional relative kernels in the extra MMR file.
So they are variable length but only in one of two different configurations: kernel or kernel+relative_kernel.
And we do this via two MMR files, both taking fixed length elements.

Your comment makes me wonder if we can somehow generalize this even more though.
We could (not completely sure this is a good idea) implement arbitrary optional fields (like an optional lock height) via separate MMR data files.
We could have a data file that contained only lock heights (and a corresponding leaf_set file, defining which kernels at which pos had an optional lock height entry and which did not).

DavidBurkett:
Why not just store a byte for each leaf in a leaf_size file, indexable by leaf index and taking advantage of prune_list's shifts? Then you can just build a cache similar to how we cache shiftCache and leafShiftCache. We could retain the original data mmr, then.

tromp:
I'm ok with the changes just to get the features signed, and in consideration of secp256k1 lib's limitations.

antiochp:

Does it make sense to target a HF for testnet4 or are we happy with it activating directly on mainnet?

antiochp:
Given the rapidly approaching holiday season and the fact we're down to a few weeks until mainnet I'm leaning toward just activating this on mainnet and not trying to force another HF through for this.

We can revisit this decision if we do plan on an additional HF for any other reason.

core/src/core/transaction.rs
```diff
@@ -1293,12 +1282,29 @@ impl From<Output> for OutputIdentifier {
 	}
 }
 
-/// Construct msg from tx fee and lock_height.
-pub fn kernel_sig_msg(fee: u64, lock_height: u64) -> Result<secp::Message, Error> {
-	let mut bytes = [0; 32];
-	BigEndian::write_u64(&mut bytes[16..24], fee);
-	BigEndian::write_u64(&mut bytes[24..], lock_height);
-	let msg = secp::Message::from_slice(&bytes)?;
+/// TODO - conditional (hard fork?).
```
ignopeverell: Want to clean that up?

antiochp: On it.

tromp:
Closely related PR: https://github.com/mimblewimble/grin/pull/2168

PR #2111, opened 2018-12-10
fix: bzip2 format used is incompatible with the windows bzip2

DavidBurkett:
The txhashset bzip2 archives cannot be unzipped from windows using the standard bzip2 implementation from Julian Seward. Since the pmmr files are nearly indistinguishable from random anyway, we aren't gaining anything by compressing, so it's not worth the performance hit (bzip2 is known for being slow) and the lack of portability. This change is backward compatible, and should cause no disruptions.

### DIFF
--- a/util/src/zip.rs
+++ b/util/src/zip.rs
@@ -13,8 +13,7 @@
 // limitations under the License.
 
 use std::fs::{self, File};
-/// Wrappers around the `zip-rs` library to compress and decompress zip
-/// bzip2 archives.
+/// Wrappers around the `zip-rs` library to compress and decompress zip archives.
 use std::io;
 use std::path::Path;
 use walkdir::WalkDir;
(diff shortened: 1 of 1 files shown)


antiochp:
For completeness can we produce some numbers for file size pre/post change just to compare?

DavidBurkett:
@drizzt Pointed out that we should be removing the dependency on bzip by changing the zip dependency to:
zip = {version = "0.4", default-features = false}
This is true, but shouldn't be done until the majority of the nodes are using this code. New nodes would still need bzip to sync from ones that don't have this code yet. Should I just create an issue to track this?

sesam:
Old clients that receive bzip2 already have support. New clients trying
resync, do they get bzip2 from the old?
Old clients syncing from a new, can they parse non-bzip2 data? Is gzip or
classical pkzip (deflate?) created by new clients?

Issue #2108, opened 2018-12-09
Keep getting Temporary note BIP32 warning


thisconnect:
Hi I am new and created a test wallet in `v0.4.0`. After updating to `v0.4.2` I keep getting the following message, but as far as I see the wallet.seed and wallet.seed.bak have been created.

```bash
$ grin wallet info
Temporary note:
If this is your first time running your wallet since BIP32 (word lists) were implemented, your seed will be converted to the new format. Please ensure the provided password is correct.
If this goes wrong, your old 'wallet.seed' file has been saved as 'wallet.seed.bak' Rename this file to back to `wallet.seed` and try again
Password: 

____ Wallet Summary Info - Account 'default' as of height 74551 ____

 Total                            | 0.000000000 
 Awaiting Confirmation (< 10)     | 0.000000000 
 Locked by previous transaction   | 0.000000000 
 -------------------------------- | ------------- 
 Currently Spendable              | 0.000000000 

Command 'info' completed successfully
```

Code https://github.com/mimblewimble/grin/blob/aedac483f5a116b91a8baf6acffd70e5f980b8cc/src/bin/cmd/wallet_args.rs#L51

NOTE: 0 is the correct balance

PR #2115, opened 2018-12-10
 Remove temporary BIP32 note

quentinlesceller:
Fixes https://github.com/mimblewimble/grin/issues/2108.


yeastplume:
#2108 isn't an issue, it's a misunderstanding, but I think we're okay to remove this now anyhow.

thisconnect:
> #2108 isn't an issue, it's a misunderstanding, but I think we're okay to remove this now anyhow.

sorry I didn't realize that

PR #2116, opened 2018-12-10
fix: Only MAX_BLOCK_HEADERS - 1 (511) headers sent during sync

DavidBurkett:
Only MAX_BLOCK_HEADERS - 1 (511) headers sent during sync, instead of MAX_BLOCK_HEADERS (512). Not a big deal, but figured I'd change it since the code supports receiving the full 512.

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -267,7 +267,7 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 		// looks like we know one, getting as many following headers as allowed
 		let hh = header.height;
 		let mut headers = vec![];
-		for h in (hh + 1)..(hh + (p2p::MAX_BLOCK_HEADERS as u64)) {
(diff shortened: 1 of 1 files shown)


servers/src/common/adapters.rs
```diff
@@ -267,7 +267,7 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 		// looks like we know one, getting as many following headers as allowed
 		let hh = header.height;
 		let mut headers = vec![];
-		for h in (hh + 1)..(hh + (p2p::MAX_BLOCK_HEADERS as u64)) {
+		for h in (hh + 1)..=(hh + (p2p::MAX_BLOCK_HEADERS as u64)) {
```
antiochp: 🎉 
I always wondered why we did it in chunks of 511 headers...

antiochp:
This does makes me surprisingly happy - 

```
20181211 09:42:43.872 DEBUG grin_servers::common::adapters - returning headers: 512
```

PR #2122, opened 2018-12-11
Utility to automatically generate mainnet genesis block

ignopeverell:


ignopeverell:
Still needs some testing and cleanup but I think it's ready for a first review.

etc/gen_gen/src/bin/gen_gen.rs
```diff
@@ -0,0 +1,324 @@
...
+		.status()
+		.expect("git tag failed");
+	process::Command::new("git")
+		.args(&["push", "origin", "v1.0"])
+		.status()
+		.expect("git tag push failed");
+	println!("All done!");
+}
+
+fn update_genesis_rs(gen: &core::core::Block) {
```
antiochp: Took me a while to get my head aorund what this is doing. We're actually going into the `genesis.rs` file and replacing strings? 
Feels kind of fragile with the looking for `REPLACE` logic.

Maybe worth considering using a similar approach to simply write out a `genesis.toml` file and then `genesis.rs` can read it in?
That way its a lot more explicit what params were "replaced" etc.

antiochp:
I guess the downside of a `genesis.toml` file is it is "config" and people can potentially mess around with it without recompiling the code.
But I do think it will make it clearer exactly what is being parameterized - we can point to an explicit list of things in a file.

yeastplume:
I think the genesis.rs replacement is okay, assuming this is only to be run once manually (or for every pointless deflationary-monetary-policy fork) and the result looked-over and commit. They're not really parameters, but hard-coded constants so I don't think there will be a need to consider 'what was changed'. All looks good to me.

etc/gen_gen/src/bin/gen_gen.rs
ignopeverell: I did hesitate in the approach with some external data file (FWIW that's how Ethereum did it). It does feel cleaner and replacing in `genesis.rs` is a tad weird. But in the end I still opted for replacement in `genesis.rs`, even though as you mention it feels more fragile, because distribution is straightforward (git commit+compile), I can rely on the Rust compiler to catch some mistakes and the result is simpler (the genesis definition you'd expect in `genesis.rs`). You'll also see a clean diff on the commit, so saying what was replaced is actually very easy.

With an external file, distribution is more open. Is it downloaded? Or loaded from the repo? It's also another way to define genesis, when so far they've all been defined programmatically. Comparison between the 2 definitions is also harder. And finally, it would eventually be simpler to just update the genesis definition in `genesis.rs` after a few days. So why not dot that in the first place?

etc/gen_gen/src/bin/gen_gen.rs
```diff
@@ -0,0 +1,324 @@
...
+}
+
+fn setup_chain(dir_name: &str, genesis: core::core::Block) -> chain::Chain {
+	util::init_test_logger();
+	let _ = fs::remove_dir_all(dir_name);
+	let verifier_cache = Arc::new(util::RwLock::new(
+		core::core::verifier_cache::LruVerifierCache::new(),
+	));
+	let db_env = Arc::new(store::new_env(dir_name.to_string()));
+	chain::Chain::init(
```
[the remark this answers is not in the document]
ignopeverell: That was done in #2079

etc/gen_gen/src/bin/gen_gen.rs
```diff
@@ -0,0 +1,324 @@
...
+use grin_core::core::hash::Hashed;
+use grin_core::core::verifier_cache::LruVerifierCache;
+use grin_keychain::{BlindingFactor, ExtKeychain, Keychain};
+
+static BCHAIN_INFO_URL: &str = "https://blockchain.info/latestblock";
+static BCYPHER_URL: &str = "https://api.blockcypher.com/v1/btc/main";
+static BCHAIR_URL: &str = "https://api.blockchair.com/bitcoin/blocks?limit=2";
+
+static GENESIS_RS_PATH: &str = "../../core/src/genesis.rs";
+static PLUGIN_PATH: &str = "./cuckaroo_mean_cuda_29.cuckooplugin";
```
kargakis: In the README file you are saying

> Mine the block so we have at least a valid Cuckatoo Cycle.

Here you are using Cuckaroo to mine the block. Not so obvious what is going on here. Is the plan to mine the genesis block using the AR version so the next block that can be mined by everyone has a higher chance to be mined by the AF version? Is README wrong? What am I missing?

Issue #2105, opened 2018-12-08
owner_api panics with refresh argument


tensor5:
```
curl -i http://grin:<apisecret>@localhost:13420/v1/wallet/owner/retrieve_summary_info?refresh
```

panics the `grin wallet owner_api`

```
20181208 22:15:15.953 ERROR grin_util::logger - 
thread 'tokio-runtime-worker-1' panicked at 'nested block_on: EnterError { reason: "attempted to run an executor while another executor is already running" }': libcore/result.rs:1009stack backtrace:
   0:     0x56349592af8d - backtrace::backtrace::trace::h33a8de1a5de4f598
   1:     0x56349592a1a2 - <backtrace::capture::Backtrace as core::default::Default>::default::h13bf884e561085be
   2:     0x56349592a219 - backtrace::capture::Backtrace::new::h810796244ae4b8f2
   3:     0x56349589f720 - grin_util::logger::send_panic_to_log::{{closure}}::h4131e8199612a48b
   4:     0x5634959f6486 - std::panicking::rust_panic_with_hook::h71214e7ce0f7ac01
                        at libstd/panicking.rs:480
   5:     0x5634959f5fd1 - std::panicking::continue_panic_fmt::ha8b8442f4ea9bcac
                        at libstd/panicking.rs:390
   6:     0x5634959f5eb5 - rust_begin_unwind
                        at libstd/panicking.rs:325
   7:     0x563495a3f95c - core::panicking::panic_fmt::h0c93626b89c38af6
                        at libcore/panicking.rs:77
   8:     0x56349555adb6 - core::result::unwrap_failed::hd6d647e647f9d65f
   9:     0x56349555c90a - tokio::runtime::Runtime::block_on::h50c07489d8a34431
  10:     0x563495528fdd - grin_api::client::send_request::h58cd94e37f2486a3
  11:     0x5634953a0e3a - grin_api::client::get::hbbd2dcbfaea3a4b9
  12:     0x56349539976b - <grin_wallet::node_clients::http::HTTPNodeClient as grin_wallet::libwallet::types::NodeClient>::get_chain_height::ha4a82fe83f2fb5e3
  13:     0x5634951a5ed3 - grin_wallet::libwallet::internal::updater::refresh_outputs::he67ca7e885c2c77a
  14:     0x56349514cf02 - <grin_wallet::libwallet::api::APIOwner<W, C, K>>::retrieve_summary_info::h962086bf4a07c49c
[...]

PR #2129, opened 2018-12-11
Pin tokio and hyper versions

hashmap:
This is a tmp fix for  #2105, the current version of tokio doesn't allow
to call `block_on` on Runtime inside another Runtime. In our case we do
it to call API client from API handler. At the moment we switch to old
version of tokio and will implement a proper fix a bit later


quentinlesceller:
Isn't https://github.com/mimblewimble/grin/pull/2128 enough ?

quentinlesceller:
Just merged #2128 feel free to update/reopen this PR if we need to pin hyper version as well.

Issue #1525, opened 2018-09-14
Send API route fails unexpectedly (issue_send_tx)


gavinmcdermott:
Calling the `issue_send_tx` endpoint with a valid body fails unexpectedly.

### The url:
`http://localhost:13420/v1/wallet/owner/issue_send_tx`

### The error:
```sh
Generic error: Invalid request body }
Sep 13 17:31:44.603 ERRO Request Error: Error { inner:
```

### The rust controller expects:
```rs
api.issue_send_tx(
  args.amount,
  args.minimum_confirmations,
  &args.dest,
  args.max_outputs,
  args.num_change_outputs,
  args.selection_strategy_is_use_all,
)
```

### the browser/client sends this in the body:
```js
{
  amount: 1,
  minimum_confirmations: 0,
  dest: 'http://127.0.0.1:13415',
  max_outputs: 500,
  selection_strategy_is_use_all: true,
  num_change_outputs: 1
}
```

### Notes:
I reproduced the error on both the client I'm working on, and the `grin-web-wallet`.

Thanks for your help on this.

quentinlesceller:
It's to be expected, no ? 
Also currently the SendTxArgs are the following: https://github.com/mimblewimble/grin/blob/dd1cef2148babfc4d60478cbb35bd1505453aad5/wallet/src/libwallet/types.rs#L639-L657 (missing the method field). And normally on the web wallet https://github.com/mimblewimble/grin-web-wallet/pull/5 should handle that change.

gavinmcdermott:
Thanks for the note. However, I am still receiving the same error with the updated body. This happens both from the client (also reproduced with Postman).

From the client:
```js
{
    this.amount = 1
    this.minimum_confirmations = 0
    this.method = 'http'
    this.dest = 'http://127.0.0.1:13415'
    this.max_outputs = 500
    this.num_change_outputs = 1
    this.selection_strategy_is_use_all = true
    this.fluff = false
}
```

Postman:
![screen shot 2018-09-14 at 9 59 46 am](https://user-images.githubusercontent.com/302978/45564498-76aa0000-b805-11e8-909d-80695f024b55.png)

Error:
```sh
Sep 14 09:59:37.574 ERRO Request Error: Error { inner:
Generic error: Invalid request body }
```

gavinmcdermott:
Thanks for the note @quentinlesceller, the setup you show does work using curl. But there is an unexpected issue that arises that when I use something like axios or fetch from a client. 

[...]

PR #2131, opened 2018-12-11
Disable authentication for OPTIONS requests

quentinlesceller:
*Should fix* the issue experienced by @gavinmcdermott https://github.com/mimblewimble/grin/issues/1525.
Preflight requests (`OPTIONS` requests) now bypass BasicAuth.

 @gavinmcdermott could you give it a try?

### DIFF
--- a/api/src/auth.rs
+++ b/api/src/auth.rs
@@ -39,6 +39,9 @@ impl Handler for BasicAuthMiddleware {
 		req: Request<Body>,
 		mut handlers: Box<dyn Iterator<Item = HandlerObj>>,
 	) -> ResponseFuture {
+		if req.method().as_str() == "OPTIONS" {
+			return handlers.next().unwrap().call(req, handlers);
+		}
 		if req.headers().contains_key(AUTHORIZATION)
 			&& verify_slices_are_equal(
 				req.headers()[AUTHORIZATION].as_bytes(),


gavinmcdermott:
Thanks @quentinlesceller. So this fixes the first half of the problem—the merged change fixes the issue with the `OPTIONS` preflight request. However, I am still getting the following error: `Request header field Authorization is not allowed by Access-Control-Allow-Headers in preflight response.`.

The reason for this error is because the api server isn't explicitly allowing the `Authorization` header in follow-on requests. I believe you need to explicitly state which headers can be used in requests. 

The fix looks to be a quick one (outlined below). As a quick test, I modified a few functions in `/wallet/controller.rs` to contain the following code: 

`.header("access-control-allow-headers", "Content-Type, Authorization")`

the quick result is a file that contains...

```rs
fn create_error_response(e: Error) -> Response<Body> {
	Response::builder()
		.status(StatusCode::INTERNAL_SERVER_ERROR)
		.header("access-control-allow-origin", "*")
		.header("access-control-allow-headers", "Content-Type, Authorization")
		.body(format!("{}", e).into())
		.unwrap()
}

fn create_ok_response(json: &str) -> Response<Body> {
	Response::builder()
		.status(StatusCode::OK)
		.header("access-control-allow-origin", "*")
		.header("access-control-allow-headers", "Content-Type, Authorization")
		.body(json.to_string().into())
		.unwrap()
}

fn response<T: Into<Body>>(status: StatusCode, text: T) -> Response<Body> {
	Response::builder()
		.status(status)
		.header("access-control-allow-origin", "*")
		.header("access-control-allow-headers", "Content-Type, Authorization")
		.body(text.into())
		.unwrap()
}
```
I recompiled and ran with these changes, and the api works as expected. Screenshots below are from `OPTIONS` and subsequent `GET` requests:

options

get

Here is [a resource](https://stackoverflow.com/questions/32500073/request-header-field-access-control-allow-headers-is-not-allowed-by-itself-in-pr) for questions on this. 

Let me know if you want me to open a PR for this—happy to.

PR #2137, opened 2018-12-12
HeaderEntry for storing subset of header info in the header MMR

antiochp:
~Still a work in progress but it runs against testnet4 successfully right now.~

Tested successfully against testnet4.
Backward compatible - a node will check the header MMR on init, reading the latest hash.
If this does not match or cannot be read successfully then we rebuild the header MMR.
This PR increases the size of an entry in the header MMR, so we rebuild.

We need to retrieve difficulty information for the previous 60 headers to validate the difficulty adjustment on a block header.
So for every header and every block we validate, we pull 60 other block headers from the db.

Headers are stored in lmdb, indexed by header hash. So these 60 retrievals are effectively random access. lmdb is memory mapped, but they will not all be in memory as the db grows.

If we could retrieve the necessary data from the header MMR itself, we can take advantage of the fact that these reads would not all be random - the MMR is structured such that these header entries are all adjacent to each other. If a header entry is x bytes long, the previous 60 headers are stored as the last `60 * x` bytes in the MMR data file.
In theory we can retrieve all 60 header entries via a single read against the header MMR.

This should allow for a significant improvement on sync performance without needing to resort to caching of headers or header_info data (faster difficulty validation not yet implemented).
In effect we can leverage the header MMR as a kind of cache of header data (taking advantage of the fact they are inserted in order).

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -204,8 +204,8 @@ impl TxHashSet {
 		let pos = pmmr::insertion_to_pmmr_index(height + 1);
 		let header_pmmr =
 			ReadonlyPMMR::at(&self.header_pmmr_h.backend, self.header_pmmr_h.last_pos);
-		if let Some(hash) = header_pmmr.get_data(pos) {
-			let header = self.commit_index.get_block_header(&hash)?;
+		if let Some(entry) = header_pmmr.get_data(pos) {
+			let header = self.commit_index.get_block_header(&entry.hash())?;
 			Ok(header)
 		} else {
 			Err(ErrorKind::Other(format!("get header by height")).into())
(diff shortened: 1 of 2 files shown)


antiochp:
> header entries are kept in the MMR data forever just to store the last 60?

Short answer is yes.

Longer answer is yes and we would likely want to keep many more than the last 60 around to handle long forks/reorgs etc. Similar to what we do with the 7 day horizon for chain compaction and deleting old blocks.

We would not need to keep this data forever either, its maintained in a prunable MMR and we can take advantage of the pruning behavior.

One option might be to aggressively prune the header MMR based on a similar 7 day horizon, we would keep the peaks around to validate the `header_root` but we could prune/compact anything beneath these peaks older than 7 days.

That way we could keep something like a rolling 7 day window of header data in the header MMR, providing efficient access to the most recent 60 header difficulties with additional history in there to handle a long re-org.

Maybe 7 days is overly excessive for this but the overall approach would still apply.

PR #2141, opened 2018-12-12
offline build not possible right now closes #2139

### LINKED ISSUE #2139: offline build not possible right now
Looks like something web wallet related causes the build to fail if there is not internet connection.

We see a warning about the webwallet not being available, then it actually errors out and fails the build.

```
   Compiling grin v0.4.2 (/antiochp/grin)                                                                                                                                                    
   Compiling grin_api v0.4.2 (/antiochp/grin/api)                                                                                                                                            
error: failed to run custom build command for `grin v0.4.2 (/antiochp/grin)`                                                                                                                 
process didn't exit successfully: `/antiochp/grin/target/debug/build/grin-c646fa6630b69822/build-script-build` (exit code: 101)
--- stdout
Warning: Failed to download grin-web-wallet. Web wallet will not be available

--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', libcore/result.rs:1009:5

warning: build failed, waiting for other jobs to finish...
error: build failed  
```

rentenmark:
* remove (almost) all unwrap calls
* replace with ? operator
* build script will now work offline and in many other scenarios where the web wallet may not be downloaded

### DIFF
--- a/src/build/build.rs
+++ b/src/build/build.rs
@@ -59,59 +59,61 @@ fn main() {
 		format!("{}{}", env::var("OUT_DIR").unwrap(), "/built.rs"),
 	);
 
-	install_web_wallet();
+	let web_wallet_install = install_web_wallet();
+	match web_wallet_install {
+		Ok(true) => {}
+		_ => println!(
+			"WARNING : Web wallet could not be installed due to {:?}",
+			web_wallet_install
+		),
+	}
 }
 
-fn download_and_decompress(target_file: &str) {
+fn download_and_decompress(target_file: &str) -> Result<bool, Box<std::error::Error>> {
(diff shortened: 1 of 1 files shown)


src/build/build.rs
```diff
@@ -59,59 +59,61 @@ fn main() {
 		format!("{}{}", env::var("OUT_DIR").unwrap(), "/built.rs"),
 	);
 
-	install_web_wallet();
+	let web_wallet_install = install_web_wallet();
+	match web_wallet_install {
+		Ok(true) => {}
+		_ => println!(
```
rentenmark: Note that this message will [probably never be seen](https://github.com/rust-lang/cargo/issues/1106) and thus we should probably consider deleting messaging in any case where the build script will return 0 status code.

PR #2146, opened 2018-12-13
[DNM] prunable header mmrs

antiochp:
Related - #2137.

This PR demonstrates how we can aggressively prune the header MMR and still be able to validate the `prev_root` in each header (against the current root of the header MMR).

But - turns out we had forgotten that we are also using the header MMR as our canonical "header by height" index. And we need to maintain this all the way back to the genesis block so we can let other nodes sync headers from us (via the locator).

We have a couple of competing requirements here - 
* header by height index: maintain _all_ header hashes in the header MMR _forever_
* small header MMR: only maintain _recent_ header entries in the header MMR

I am putting this PR up so we have something to refer to. But closing it as we cannot enable pruning in the header MMR currently due to the locator requirements.

I'm still stating #2137 is a good idea and the trade offs are worth it.
But - we need to think more about how to support a header MMR that needs to maintain header hashes but is flexible around pruning of the associated difficulty data.

I think this bleeds into the discussion and design of the kernel MMR also - in terms of optional kernels and greater flexibility around the handling of MMR data.

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -279,17 +279,13 @@ impl TxHashSet {
 		{
 			let cutoff_pos = pmmr::insertion_to_pmmr_index(horizon_header.height);
 			self.header_pmmr_h.backend.remove_range(0..cutoff_pos);
-			self.header_pmmr_h.backend.check_compact(
-				cutoff_pos,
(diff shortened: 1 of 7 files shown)


antiochp:
Closing. But worth a conversation.

PR #2150, opened 2018-12-14
TUI basis status text align

garyyu:
Before this little change:
screen shot 2018-12-14 at 11 50 14 am

Now:
screen shot 2018-12-14 at 11 45 44 am


antiochp:
:+1:

PR #2151, opened 2018-12-14
TUI mining server status: add blocks found statistics

garyyu:
Add statistics of mining worker's found solutions, and show this found solutions number in TUI:

screen shot 2018-12-14 at 1 38 34 pm

### DIFF
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -80,6 +80,8 @@ pub struct WorkerStats {
 	pub num_rejected: u64,
 	/// number of shares submitted too late
 	pub num_stale: u64,
+	/// number of valid solutions found
+	pub num_solution_found: u64,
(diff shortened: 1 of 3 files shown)


garyyu:
Merging if nobody has concern 😄   and I do want this little improvement~

jaspervdm:
What is "solutions found"? Is it number of graphs solved or number of 42-cycles found? If the latter, isn't it always equal to sum accepted+rejected+stale?

garyyu:
@jaspervdm   it's `Solution Found for block {} - Yay!!!`  😄  not the `shares`.

jaspervdm:
Ah, so maybe we should rename it to `blocks found`?

garyyu:
OK:)  as you want~  will change it tomorrow morning.

Issue #2152, opened 2018-12-14
Header Sync Quite Slow


garyyu:
Don't know when, it seems like a regression for header sync speed, there's a pattern that each loop will pause 10 seconds.

```
20181214 14:09:04.438 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 02ec28af at 39751
20181214 14:09:04.636 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 03e79810 at 39783
20181214 14:09:04.640 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 046fb66d at 39815
20181214 14:09:04.816 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 036cf5a1 at 39847
20181214 14:09:04.820 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 04638eee at 39879
20181214 14:09:04.824 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 028f4753 at 39911
20181214 14:09:04.921 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 00ef8c1f at 39943
20181214 14:09:04.986 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 0d15d58c at 39975
20181214 14:09:04.993 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 01926827 at 40007
20181214 14:09:04.996 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 004ccd6a at 40039
20181214 14:09:05.005 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 00fbcfe6 at 40071
20181214 14:09:05.015 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 04748db1 at 40103
20181214 14:09:05.025 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 00674b9a at 40135
20181214 14:09:05.029 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 02da4e8e at 40167
20181214 14:09:05.103 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 0726f8ff at 40199


[...]

PR #2154, opened 2018-12-14
disconnect slow peer to avoid multiple parallel headers download

garyyu:
Fix for problem found in https://github.com/mimblewimble/grin/issues/2152#issuecomment-447237343

We should avoid 2 parallel downloading of Headers, if one peer is on a very slow network speed, that will make thing a mess for header sync.

The "mess" here means something like this:
```
20181214 14:07:14.107 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 0aa40871 at 31670
20181214 14:07:14.151 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 032ca7a0 at 30520
...
20181214 14:07:38.671 DEBUG grin_chain::pipe - pipe: sync_block_headers: 31 headers from 069e9341 at 36046
20181214 14:07:39.649 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 02c9f847 at 34608
20181214 14:07:40.307 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 022b2a84 at 36077
...
20181214 14:08:06.261 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 01dd7ca6 at 38697
20181214 14:08:06.524 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 0158050d at 37770
20181214 14:08:06.908 DEBUG grin_chain::pipe - pipe: sync_block_headers: 32 headers from 01e78670 at 38729
...
```

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -418,6 +418,10 @@ impl Peer {
 	/// Stops the peer, closing its connection
 	pub fn stop(&self) {
 		stop_with_connection(&self.connection.as_ref().unwrap().lock());
+		let mut state = self.state.write();
+		if State::Connected == *state {
+			*state = State::Disconnected;
+		}
 	}
 
 	fn check_connection(&self) -> bool {
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -28,6 +28,8 @@ pub struct HeaderSync {
 
 	history_locator: Vec<(u64, Hash)>,
 	prev_header_sync: (DateTime<Utc>, u64, u64),
+
+	syncing_peer: Option<Arc<Peer>>,
 }
 
 impl HeaderSync {
@@ -42,6 +44,7 @@ impl HeaderSync {
 			chain,
 			history_locator: vec![],
 			prev_header_sync: (Utc::now(), 0, 0),
+			syncing_peer: None,
 		}
 	}
 
@@ -81,7 +84,7 @@ impl HeaderSync {
 				highest_height: highest_height,
 			});
 
-			self.header_sync();
+			self.syncing_peer = self.header_sync();
 			return true;
 		}
 		false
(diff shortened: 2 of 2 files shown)


hashmap:
I'd prefer to download some headers twice but have more peers. This peer might be overloaded at the moment, but fast and reliable after that.

garyyu:
@hashmap  more peers? 😄   don't worry, the peers connection is dynamic, here just disconnect it if slow, it will automatically increase to configured peers number even connecting same peers again (perhaps after header sync).

antiochp:
I'm concerned the complexity of the sync process is starting to creep up again.

I'm not convinced we should be dropping peer connections if the peers are too slow - I don't think we're tracking this at a granular enough level. And we're not tracking this on both ends of the connection.
We drop a peer because its too slow, but its still trying to serve us data.
I think we run the risk of getting into a bad cycle where we're continually trying to request headers, dropping and reconnecting, then immediately asking for the same headers again.
And the peer is repeatedly sending us headers, just not quite fast enough.

We got rid of the "header cache" a while ago and its quite possible this had a negative impact on the header sync performance.
But - its more robust and simpler without the cache and the trade off is likely worth it.

Unless the header sync performance is a serious bottleneck I'm not sure the increased complexity of dropping connections and relying on reconnection here is worth it?

antiochp:
Just to add another observation - we originally added the "allow parallel header requests" as an optimization but I'm not sure this makes sense? We need these requests to be serialized as we cannot request another 512 headers until the we have finished processing the previous 512 headers.

What does parallel requests here buy us? Seems like all it will do is complicate our processing when we do eventually receive the 2nd set of duplicate headers?

ignopeverell:
We shouldn't drop slow peers at all. This would mean that slow peers will get dropped left and right by the network, which wouldn't be nice at all for them.

garyyu:
to make it clear:  the `slow` peer means 10 seconds timeout without receiving anything after a Headers request.

and more explains here: https://github.com/mimblewimble/grin/issues/2152#issuecomment-447512057

> Just to add another observation - we originally added the "allow parallel header requests" as an optimization but I'm not sure this makes sense? We need these requests to be serialized as we cannot request another 512 headers until the we have finished processing the previous 512 headers.
> 
> What does parallel requests here buy us? Seems like all it will do is complicate our processing when we do eventually receive the 2nd set of duplicate headers?

@antiochp  I write this PR exactly for avoiding the parallel requests for Headers:  before a new request is sending, I disconnect the possible ongoing downloading on slow peer.

antiochp:
@garyyu I commented in #2152.
Proposing we solve the underlying issue causing the multiple headers requests to be initiated in parallel rather than attempting to stop them after being initiated by closing the connection.

garyyu:
OK @antiochp , hope we can find a better way out 😄 
So let's close this one for now.

PR #2157, opened 2018-12-14
[Floonet] Switch commitments

jaspervdm:
Re-opening this because I messed #2007 up.

This PR adds switch commitments following the scheme defined in #998. Blinding factors are modified to `x' = x + H(xG+vH | xJ)`

If in the future doubts arise about the existence of mechanisms that would break the discrete logarithm between `G` and `H`, we could require revealing the ElGamal commitment `(xG+vH, xJ)` (and a corresponding rangeproof) in order to spend an output.

Wallets that want to use switch commitments should add a configuration option `use_switch_commits=true`. The default value for now is `false`, we could consider changing it to `true` before mainnet, if we feel comfortable enough with this PR.

I have constructed this PR in a way that wallets can either use switch commitments, or not use them. I do not think it is a good idea to let wallets have both types (or possibly even more than 2) at the same time. This is because if we want to use switch commitments, _everyone_ should use them. In that case we don't want outputs that don't have the switch commitment lingering around. The same holds if we ever want to change to a different type of switch commitment. In those cases the users should create a new wallet and transfer their funds to it.

TODO:
- [ ] do more testing
- [ ] check built-in tests
- [ ] change `util/Cargo.toml` to the latest version of libsecp with switch commitments

PRs in underlying libraries: mimblewimble/secp256k1-zkp#34, mimblewimble/rust-secp256k1-zkp#38

Comments welcome!

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -406,6 +406,13 @@ fn comments() -> HashMap<String, String> {
 "
 		.to_string(),
 	);
+	retval.insert(
+		"use_switch_commitments".to_string(),
+		"
+#Whether to use switch commitments for this wallet
+"
+		.to_string(),
+	);
 
 	retval.insert(
 		"[logging]".to_string(),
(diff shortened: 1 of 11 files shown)


yeastplume:
I think if we're going to be merging this into Floonet directly, we shouldn't need a flag to switch them on or off. We can just assume switch commits are the way forward and merge them into the forthcoming branch. 

So I'd suggest:

1) Focus on getting the functions/tests into the underlying secp and rust secp libs
2) Once we're happy with those, merge and tag them,
3) Included newly tagged secp libs into this PR
4) Remove flag and just assume switch commits
5) Ensure all automated tests are working
6) Merge into Floonet branch

yeastplume:
Going to merge into floonet now so I can include tagged libsecp and exercise it a bit

PR #2158, opened 2018-12-14
fix: stratum server handles any request ID

### LINKED ISSUE #2149: Stratum server accepts only string ids

NaN-git:
fixes #2149 to enable communication with a JSON RPC 2.0 compliant client. The only drawback of this fix is that the server accepts Bool, Array and Object values as ID, too.

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -42,8 +42,16 @@ use crate::util;
 // RPC Methods
 
 #[derive(Serialize, Deserialize, Debug)]
+#[serde(untagged)]
+enum RpcId {
+    Null,
+    Number(serde_json::Number),
+    String(String),
+}
+
+#[derive(Serialize, Deserialize, Debug)]
 struct RpcRequest {
-	id: Value,
+	id: RpcId,
 	jsonrpc: String,
 	method: String,
 	params: Option<Value>,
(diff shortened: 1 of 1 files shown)


rentenmark:
You mentioned a mining pool but I wonder if this impacts grin-miner and/or grin-gold-miner as well?

NaN-git:
grin-gold-miner seems to be fine because it seems to never use the ID in the code. But grin-miner uses the same data structures as grin and it should therefore be fixed.

rentenmark:
@mcdallas mentioned that this isn't to spec either because it accepts arrays and maps.

NaN-git:
Update: Now IDs are rejected that are Bool, Array or Object values

NaN-git:
@rentenmark Do you have any comments on this PR? It would be helpful, if this issue would be fixed.

waynenilsen:
My only real remaining question is, is this code covered by any tests? If so it would be great to upgrade the tests to check that valid json is serialized and invalid json is rejected.

rentenmark:
Is this code covered by tests? If so can we improve the tests to test accept/reject cases? Also, not sure what is going on with Travis but it will need to build before this can be merged

rentenmark:
Is this code covered by tests? If so can we improve the tests to test accept/reject cases? Also, not sure what is going on with Travis but it will need to build before this can be merged

ignopeverell:
This PR looks valuable to me but some tests would be great, especially to ensure it doesn't break backward compatibility.

yeastplume:
I think we can close this for the time being since the stratum layer is being rewritten anyhow and this PR hasn't moved for a while.

PR #2161, opened 2018-12-15
take lock once in get_header_for_output

### LINKED ISSUE #2160: Wallet restore is slow
From gitter - 
> François @mably 10:06
After 10 hours of wallet restoring, I stupidly restarted my grin instance.
Looks like I'm back to restoring from scratch :(

My guess is this is related to the changes we made around the "header by height" index.

We used to maintain a height index in the db.
We replaced this with reads into the header MMR (insertion index in the MMR maps directly to height).
(This significantly simplified the code, reducing the potential for cache invalidation problems).

But `get_header_by_height()` now involves getting a read lock on the txhashset so we can access the header MMR.

Looks like `chain.get_header_for_output()` does some heavy looping based on header heights, with many calls to `get_header_by_height()`. 
And we use this as part of wallet restore.

https://github.com/mimblewimble/grin/blob/197d4f9575d7f97b080ff7eea3b34a0c81c469dc/chain/src/chain.rs#L1138-L1141

I think we should be able to rework this part of the wallet restore to reduce the number of calls to `get_header_by_height()` involved without too much effort.
It may even make more sense to do the header traversal explicitly via `store.get_previous_header()` and avoid hitting the MMR based header by height "index".

`get_header_by_height()` is useful for going _directly_ to a single header somewhere in the chain. It makes far less sense as a mechanism for iterating back over headers in the chain. It is _much_ faster to recursively get the previous header via `store.get_previous_header()`.

Edit: Hmm, not just iterating over headers in a loop here though, we're doing a binary search to find the matching header. We may need to think a bit more about this (assuming this is the reason for slowdown).
Maybe the search itself can be done within a single lock on the txhashset, rather than within each call to `get_header_by_height()`.

Wallet restore would appear to have slipped through the cracks when we were testing performance with the header MMR based impl of the header by height index. Sorry about that...

antiochp:
Resolves #2160.

```
> time ../target/release/grin wallet restore
20181215 16:51:26.501 INFO grin_wallet::libwallet::internal::restore - Starting restore.
20181215 16:51:26.824 INFO grin_wallet::libwallet::internal::restore - Retrieved 1000 outputs, up to index 217479. (Highest index: 6891)
20181215 16:51:26.824 INFO grin_wallet::libwallet::internal::restore - Scanning 1000 outputs in the current Grin utxo set
20181215 16:51:27.206 INFO grin_wallet::libwallet::internal::restore - Retrieved 1000 outputs, up to index 217479. (Highest index: 9335)
20181215 16:51:27.206 INFO grin_wallet::libwallet::internal::restore - Scanning 1000 outputs in the current Grin utxo set
...
20181215 16:51:43.641 INFO grin_wallet::libwallet::internal::restore - Retrieved 1000 outputs, up to index 217479. (Highest index: 216681)
20181215 16:51:43.641 INFO grin_wallet::libwallet::internal::restore - Scanning 1000 outputs in the current Grin utxo set
20181215 16:51:43.948 INFO grin_wallet::libwallet::internal::restore - Retrieved 793 outputs, up to index 217479. (Highest index: 217479)
20181215 16:51:43.949 INFO grin_wallet::libwallet::internal::restore - Scanning 793 outputs in the current Grin utxo set
...
20181215 16:51:43.987 INFO grin_wallet::command - Wallet restore complete
Command 'restore' completed successfully
../target/release/grin wallet restore  3.31s user 0.31s system 16% cpu 21.785 total
```

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1123,11 +1123,9 @@ impl Chain {
 		&self,
 		output_ref: &OutputIdentifier,
 	) -> Result<BlockHeader, Error> {
-		let pos = {
-			let txhashset = self.txhashset.read();
-			let (_, pos) = txhashset.is_unspent(output_ref)?;
-			pos
-		};
+		let txhashset = self.txhashset.read();
+
+		let (_, pos) = txhashset.is_unspent(output_ref)?;
 
 		let mut min = 1;
 		let mut max = {
@@ -1137,8 +1135,8 @@ impl Chain {
 
 		loop {
 			let search_height = max - (max - min) / 2;
-			let h = self.get_header_by_height(search_height)?;
-			let h_prev = self.get_header_by_height(search_height - 1)?;
+			let h = txhashset.get_header_by_height(search_height)?;
+			let h_prev = txhashset.get_header_by_height(search_height - 1)?;
 			if pos > h.output_mmr_size {
 				min = search_height;
 			} else if pos < h_prev.output_mmr_size {


mably:
Tested and approved 👍

Issue #2169, opened 2018-12-16
"wallet send" failing with "transaction does not exist"


antiochp:
Not sure what is happening here. 

Tx appears to be built successfully but then fails during final validation step with `Wallet command failed: LibWallet Error: Transaction 28c02fa3-868b-4cbb-9cc7-94dca007dee3 doesn't exist` - 

```
20181216 22:00:09.965 DEBUG grin_wallet::adapters::http - Posting transaction slate to http://127.0.0.1:13415/v1/wallet/foreign/receive_tx
20181216 22:00:10.092 DEBUG grin_wallet::types - Using wallet seed file at: /grin/node1/wallet_data/wallet.seed
20181216 22:00:13.156 DEBUG grin_wallet::types - Using wallet seed file at: /grin/node1/wallet_data/wallet.seed
20181216 22:00:13.224 DEBUG grin_core::libtx::slate - Validating final transaction
Wallet command failed: LibWallet Error: Transaction 28c02fa3-868b-4cbb-9cc7-94dca007dee3 doesn't exist
```

yeastplume:
This is likely happening when trying to update the stored transaction with the final version (there have been some changes recently but tests all pass), what are the steps to reproduce? Is this with every transaction?

antiochp:
Not sure if its all transactions.

But just realized - this was a send to myself via 127.0.0.1 -

```
../target/release/grin wallet send 1.7 -d "http://127.0.0.1:13415" --fluff
```

And it was a transaction with a couple hundred inputs in it.

I'm guessing the error may be related to the send to 127.0.0.1?

yeastplume:
yes, I have it, will push a fix shortly

PR #2170, opened 2018-12-17
Fix for #2169

yeastplume:
fixes https://github.com/mimblewimble/grin/issues/2169, (only affecting self-send via the http api)

### DIFF
--- a/wallet/src/libwallet/internal/updater.rs
+++ b/wallet/src/libwallet/internal/updater.rs
@@ -96,7 +96,10 @@ where
 	let mut txs = if let Some(id) = tx_id {
 		wallet.tx_log_iter().filter(|t| t.id == id).collect()
 	} else if tx_slate_id.is_some() {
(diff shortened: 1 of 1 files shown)


wallet/src/libwallet/internal/updater.rs
```diff
@@ -94,19 +94,12 @@ where
 {
 	// just read the wallet here, no need for a write lock
 	let mut txs = if let Some(id) = tx_id {
-		let tx = wallet.tx_log_iter().find(|t| t.id == id);
-		if let Some(t) = tx {
-			vec![t]
-		} else {
-			vec![]
-		}
+		wallet.tx_log_iter().filter(|t| t.id == id).collect()
```
antiochp: Oh... there's 2 txs in the tx log here for a self send right?
:+1:

yeastplume: Yes. The receive tx ends up with a lower ID in the self-send via http case, and previous short circuit find was just returning the first matching TX it found, which is a recieve tx. The finalize is explicitly looking for a send tx.

PR #2173, opened 2018-12-18
Enhance share rejection messages to include nonce and hash

bladedoyle:
This change enhances the stale and invalid share log messages:
When possible share rejection messages now include nonce and hash
This will make it easier to audit the pool records against the server logs. Without that info its not clear which exact share the message is for.


yeastplume:
:+1:

Issue #2176, opened 2018-12-18
Dandelion++ Algorithm Changes


antiochp:
The Dandelion++ paper talks about nodes choosing stem vs. fluff _per epoch_ and not per tx.
i.e. every 600s (in Grin) we choose a new outbound Dandelion relay and set our own relay to either stem or fluff for the duration of this "epoch".

This might allow us to remove the "patience timer" on all relay nodes set in stem mode.
i.e. We would _immediately_ forward a stem tx to the next relay node.
Only on nodes currently set as "fluff" would the patience timer be applicable.

These would then act as "sinks", txs would flow into them from various stem paths and would enable aggregation until the patience timer fired.
This would provide (temporary) places in the network for unaggregated txs to collect - so we could have a longer patience timer, and intuitively, a better chance of achieving tx aggregation as other txs would not be delayed in prior stem phase nodes on the way to the fluff node.

Right now in Grin we delay txs for 10-15s at _every_relay node. And I think this is actually detrimental to tx aggregation as it makes it less likely for 2 txs traveling along the same stem path to actually meet at the same node at the same point in time. One is _always_ delayed behind the other and will never catch up to the earlier one.

At the start of every "epoch" (every 600s) - 
* select an outbound peer as the next stem relay
* decide if we are acting as "stem" or "fluff" for duration of the epoch

Then for every stem tx received (from a peer) -
* if we are in stem mode - _immediately_ relay it to the next stem relay
* if we are in fluff mode - aggregate txs up until our next patience time expires, then broadcast it

This seems to be simpler than our current approach and reduces unnecessary latency while allowing a longer patience timer, so potentially allowing more aggregation.

Related - #1982.

antiochp:
I believe we went with a simplified "one outbound relay" originally after discussions with Giulia Fanti.
This was to increase the chances of aggregation I believe with some trade-offs in terms of the anonymity guarantees in Dandelion, @quentinlesceller probably knows more about this.
But I do wonder if the reference to "linkability of transactions" in the Dandelion++ paper is a reference to these earlier conversations with us.

lehnberg:
> But I do wonder if the reference to "linkability of transactions" in the Dandelion++ paper is a reference to these earlier conversations with us.

I believe it has to do with the forwarding rule of forwarding multiple transaction received from a node in on the same path. 

From the paper (§4.2):
[...]

PR #2177, opened 2018-12-18
[DNM] increase dandelion embargo timer from 3 mins to 10 mins

### LINKED ISSUE #1982: Dandelion: Reduce likelihood of a tx reaching the embargo threshold
It appears that in the current dandelion config, 15% of all transactions can be expected not to fluff prior to the embargo timer. This means that these transactions would then be broadcasted the traditional way from the sender, making it straight forward to determine which IPs they originate from.

15% seems like a high number for this likelihood, and it may make sense to find ways to reduce it.

we see the embargo timer expire pretty frequently
what we do need to do is investigate this appears to expire more often than not
specifically - we see it expire on the originating node (say I send a stem tx, I see my node's embargo timer expire)
>
>lehnberg @lehnberg 14:55
https://github.com/mimblewimble/grin/blob/fb3f967a860e184a87753b00b6281d10e1c2d4cc/pool/src/types.rs#L35
So with
`patience = 10s`,
`embargo = 180s`,
it would mean that there’s a maximum of `180/10=18 hops` that could occur before embargo clocks out and the originating node broadcasts the transaction the traditional way.
Then the question becomes: What’s the probability of hitting 18 stems in a row when `stem=90%`?
Which (I think) can be expressed as
`0.9^18 ~ 0.15`, i.e. 15%
Am I completely off base here?
>
> ...
>
>valdok @valdok 14:59
@lehnberg you seem to be right. If you have a constant "embargo" timer, as well as constant "patience" timer, then it seems that it'll be activated in 15% of the cases. BTW, in the paper of Dandelion++ they tell that the "embargo" timer should be random, otherwise the origin of the transaction will always disclose itself.
>
> lehnberg @lehnberg 14:59
But if what I calculate above is correct, it would mean that you would expect 15% of all transactions to belong to the IP that broadcasts it, and that doesn’t sound right.
Or good, at least.

antiochp:
Resolves #1982.

Lets discuss though (see #1982) - are we happy delaying txs for up to 10 mins when stemming them?
This should only happen rarely but its not a great user experience given that this is pretty much opaque once the tx is first sent.

Arguably we _need_ to delay them this long to give our Dandelion implementation sufficient time to operate correctly.

Also see #2176 for some early thoughts around basing our impl on Dandelion++ (rather than Dandelion) which may give us a way to _increase_ the patience timer while _reducing_ overall latency and potentially increasing aggregation.

### DIFF
--- a/pool/src/types.rs
+++ b/pool/src/types.rs
@@ -30,7 +30,7 @@ use grin_keychain as keychain;
 const DANDELION_RELAY_SECS: u64 = 600;
 
 /// Dandelion embargo timer
-const DANDELION_EMBARGO_SECS: u64 = 180;
+const DANDELION_EMBARGO_SECS: u64 = 600;
 
 /// Dandelion patience timer
 const DANDELION_PATIENCE_SECS: u64 = 10;


lehnberg:
Is the notion of getting #2176 in ahead of mainnet completely ruled out as an option?

antiochp:
Closing. Replaced with #2185 which should fix the issue properly.

PR #2178, opened 2018-12-18
[Floonet] Use switch commits for all blinding factors

yeastplume:
@jaspervdm this just removes the option to use switch commits and enforces it in all cases, have a look when you can.

Also fixed a few tests.


jaspervdm:
Yes, looks good to me!

PR #2191, opened 2018-12-20
verify features in one place; comment on needed Features enum type

tromp:


core/src/core/transaction.rs
```diff
@@ -253,10 +261,6 @@ impl TxKernel {
 	/// as a public key and checking the signature verifies with the fee as
 	/// message.
 	pub fn verify(&self) -> Result<(), Error> {
-		if self.is_coinbase() && self.fee != 0 || !self.is_height_locked() && self.lock_height != 0
```
ignopeverell: I'm slightly confused. Why remove this check? And it looks like we still serialize the fee and lock height, regardless of the type of kernel?

tromp: The check for InvalidKernelFeatures remains in kernel_sig_msg (line 1362), which is called from verify() in line 272:
if !secp::aggsig::verify_single(..., ..., &self.msg_to_sign()?,

In line 1371 we pick the fields to serialize in the hash with
	let hash = match features {
		KernelFeatures::COINBASE => features.hash(),
		KernelFeatures::PLAIN => (features, fee).hash(),
		_ => (features, fee, lock_height).hash(),
	};

ignopeverell:
Had to revert due to Floonet being captured by `is_testnet`. I think that's the right behavior but in this case the change to what gets signed should have been applied _before_ Floonet.

tromp:
> Floonet being captured by is_testnet. I think that's the right behavior

I thought that is_testnet is what should have been called is_testnetN.

PR #2200, opened 2018-12-21
stick to e=H(R|P|m) when use schnorr signature

garyyu:
happen to see the message signature when wallet sending is still using the old form with `e=Hash(R|m)`, better to stick to the formal form everywhere.

*Note*:  it's not a consensus breaking issue, but once it's modified and some nodes not updated, the wallet sending between current version and new version will fail because of message signature verification fail.

DNM it because I can ONLY test this PR after my coinbase mature (tomorrow afternoon).

### DIFF
--- a/core/src/libtx/slate.rs
+++ b/core/src/libtx/slate.rs
@@ -369,10 +369,8 @@ impl Slate {
 						"Optional participant messages do not match signatures".to_owned(),
 					))?;
 				} else {
-					info!(
-						"verify_messages - signature verified ok. Participant message: \"{}\"",
-						String::from_utf8_lossy(&msg.as_bytes()[..])
-					);
+					info!("verify_messages - signature verified ok. Participant message: \"{}\"",
+						  String::from_utf8_lossy(&msg.as_bytes()[..]));
 				}
 			}
 		}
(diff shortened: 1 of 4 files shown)


garyyu:
Tested on Floonet. Ready to review.

It will impact the wallet communication methods `-m file` (and `-m keybase` also I guess), the wallet receive / finalize will fail between the old version and this new version, if user use `-g message` optional parameter.

garyyu:
to make thing simple, I split those modification for message signature verification into an independent PR: https://github.com/mimblewimble/grin/pull/2203, and make this PR pure for the schnorr signature fixing.

garyyu:
I feel nobody is using the optional message for the wallet send, so perhaps it's good time to merge this, before some use cases start using it, and avoid future compatibility problem.

yeastplume:
To be honest, I'm not really convinced this is necessary here...  embedding the public key in the message is needed at transaction time to support the particulars of a Grin exchange, but when it's just simply a single signer signing a simple message there's no need to do anything special with it, hence the simple signature. Why does this need to be more complicated?

garyyu:
I believe confirming to *Gregory Maxwell,Andrew Poelstra,Yannick Seurin, and Peter Wuille* 's  [multi-signature scheme](https://eprint.iacr.org/2018/068.pdf) is our best choice, since we already switched to it in our 2-of-2 signature for transaction kernel since `Testnet4`.

And I'm afraid that `more complicated` could not a good reason to leave this message signature behind, even we don't know yet what benefits it will bring. For example, an "Adaptor" Signature will need this form.

Two words modification in this PR should not be a complicated change, right? 😄 

So, if no any bad point in this change (I believe no), I would prefer this should be taken before mainnet.

yeastplume:
It might mean that any other services that want to verify the message would likely need to use our aggsig libs as opposed to other implementations, but might enable multi-signer messages. But not particularly worried either way in the absence of further info, so happy enough to merge this (will do it into re-floonet so it's picked up there)

Issue #2174, opened 2018-12-18
Wallet seed recovery process leaves recovery phrase in shell history


yeastplume:
The current wallet seed recovery process involves typing or pasting the recovery phrase on the command line, which leaves it lying around in the shell history. We need a better way of doing this to ensure it's not stored, in a manner that balances a bit of usability with the inevitable challenges in doing it all through a command line.

hashmap:
Smth like pinentry https://www.gnupg.org/related_software/pinentry/index.html

mrpqzm:
https://www.ostechnix.com/how-to-clear-command-line-history-in-linux/

yeastplume:
Tried the stdin approach.. it's pants. Make one mistake and you have to enter everything again, backspace doesn't work, you can't populate it with the previous attempt to let the user fix errors.

We could go something like:
Enter word 1) this
enter word 2) is

Then give a menu for editing your word list... might work but also sounds pants from a UI perspective.

yeastplume:
Unfortunately, I don't think wiping your bash history is something a well-behaved member of the executable community should be doing.

PR #2201, opened 2018-12-21
Minor wallet recover fixes

ignopeverell:
Couple of those are security sensitive:

1. Do not log the recovery phrase on error.
2. Prompt for the phrase instead of passing it as a command arg so it doesn't appear in any command line history.

### DIFF
--- a/src/bin/cmd/wallet_args.rs
+++ b/src/bin/cmd/wallet_args.rs
@@ -23,6 +23,7 @@ use grin_keychain as keychain;
 use grin_wallet::{command, instantiate_wallet, NodeClient, WalletConfig, WalletInst, WalletSeed};
 use grin_wallet::{Error, ErrorKind};
 use rpassword;
+use std::io::{self, BufRead};
 use std::path::Path;
 use std::sync::Arc;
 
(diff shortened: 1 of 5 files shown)


yeastplume:
Good catch on log. I tried to fix the recovery phrase issue earlier in the same way, but found it a bit problematic from a UX experience, see: https://github.com/mimblewimble/grin/issues/2174

Basically, (at least on my system) stdin ignores any keymaps, so any attempt to erase or hit arrow keys as you're typing just creates a bigger mess. Agreed doing it this way is better than leaving it in your shell history, but still.. hmm..

yeastplume:
1) ____ 2) ____ 3) _____ (print to 12 or 24)
(Q to quit, 1-24 to edit a word)

Type word 1 > pants

1) pants 2) ____ 3) _____ (print to 12 or 24)
(Q to quit, 1-24 to edit a word)

Type word 2 > yeast


Something like this? And bonus points if we can figure out the equivalent of typing ‘stty erase <backspace>’, to ensure delete works

ignopeverell:
Ah, missed #2174. Not sure what's this UX thing you're talking about though.

Btw, hate to keep adding more dependencies but isn't that what readline is for? We could pick a Rust port (like https://github.com/murarth/linefeed).

yeastplume:
Only problem with that linefeed crate is that I didn’t know it existed until now.

ignopeverell:
Superseded by #2276

PR #2205, opened 2018-12-22
Fix secondary scaling bugs; add global::is_mainnet(); use it to chang…

tromp:
…e pre-genesis pow type

### DIFF
--- a/core/src/global.rs
+++ b/core/src/global.rs
@@ -266,6 +266,12 @@ pub fn is_testnet() -> bool {
 	ChainTypes::Floonet == *param_ref
 }
 
+/// Are we for real?
+pub fn is_mainnet() -> bool {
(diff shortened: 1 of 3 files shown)


core/src/global.rs
```diff
@@ -266,6 +266,12 @@ pub fn is_testnet() -> bool {
 	ChainTypes::Floonet == *param_ref
 }
 
+/// Are we for real?
+pub fn is_mainnet() -> bool {
```
antiochp: We added `is_mainnet()` back in? I thought we had removed this and replaced it with `!is_testnet()`.

The reason being we now have the majority of the tests running in an env that differs from mainnet.

Majority of the tests will return false here for `is_mainnet()` *and* for `is_testnet()`.

antiochp:
:+1: on `is_floonet()`

Issue #2202, opened 2018-12-22
Improve error thrown when compiling on ARM systems


liamaharon:
First off, huge thank you to all the Grin developers working on this project, it's really something special.

I ran into trouble compiling Grin on my Raspberry Pi 3 B+. 

The build fails when compiling `grin_store`, it seems the RPI architecture can't handle a usize type as big as specified in `store/src/lmdb` ln 75 (`env_builder.set_mapsize(549755813888).unwrap_or_else(|e| {`).

This is the error thrown:
![screenshot from 2018-12-22 15-45-09](https://user-images.githubusercontent.com/16665596/50370757-aa64a500-0600-11e9-9313-471921ea28ae.png)

a bit confusing for a newb like me who isn't familiar with system architecture limitations, or Rust. 

I suggest a check (or similar) is added before that line to ensure `usize::max_value()` is large enough to handle that value. If it's not large enough, we can throw a more friendly error to the user.

Even better, the check could be made before compilation starts so the user doesn't need to wait for a few hrs on a low power system before realising it isn't compatible.

If something like this makes sense, let me know and I'll have a go opening a PR to address it.

edit: 
could also add a line to https://github.com/mimblewimble/grin/blob/master/doc/build.md similar to 
"Not Windows 10 yet [grin kind-of builds. No mining yet. Help wanted!]" 
but for RPi / ARM

sesam:
Is it possible to lower the map_size value to something small enough to work on ARM?

kravens:
On the 1.0.0 release this is still an issue, hope a workaround is found soon.

> pi@raspberrypi:~/grin $ cargo build --release
   Compiling reqwest v0.9.5
   Compiling grin_store v1.0.0 (/home/pi/grin/store)
warning: unused import: `lmdb_zero`
  --> store/src/lib.rs:23:5
   |
23 | use lmdb_zero;
   |     ^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default

>warning: unused import: `memmap`
  --> store/src/lib.rs:24:5
   |
>24 | use memmap;
   |     ^^^^^^

>error: literal out of range for usize
  --> store/src/lmdb.rs:77:16
   |
77 |         .set_mapsize(549_755_813_888)
   |                      ^^^^^^^^^^^^^^^
   |
   = note: #[deny(overflowing_literals)] on by default

>error: aborting due to previous error

>error: Could not compile `grin_store`.
>warning: build failed, waiting for other jobs to finish...
[...]

Issue #1681, opened 2018-10-07
Autodetect lmdb storage allocation based on the env


kargakis:
The hardcoded allocation in [here](https://github.com/mimblewimble/grin/blob/3eb64af1ef2fc85622f42e82d9b56fd7a123a413/store/src/lmdb.rs#L75) is causing trouble in low-end devices such as ARM 64bit boards (eg. odroid c2). It should either be autodetected based on the environment where grin runs or at the very least driven by a config option.

kargakis:
I saw this issue on an odroid c2, another occurence of it can be found in https://www.grin-forum.org/t/problems-building-on-windows-linux/184/13

i1skn:
It also causes an issue with `cargo lipo` for iOS device. So far I manage to run it with `/8192`

mcdallas:
I tried to compile from source on a raspberry pi 3 (1GB RAM) with 300GB HD and 3GB swap.
I changed this value to 1GB 
https://github.com/mimblewimble/grin/blob/fcec70d1e37354697aadcf47ff86395f5e7d788e/store/src/lmdb.rs#L75
and I got this error when I try to run grin:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error::Code(12, 'Cannot allocate memory')': libcore/result.rs:1009stack backtrace:
   0:   0xccac97 - backtrace::backtrace::trace::h1e6b3cbff53b4eca
   1:   0xccaa43 - <backtrace::capture::Backtrace as core::default::Default>::default::h8ede5ca42cdc5af6
   2:   0xccaa9f - backtrace::capture::Backtrace::new::hee8cfb25433837a0
   3:   0xc41077 - grin_util::logger::send_panic_to_log::{{closure}}::hec8b726050880d32
   4:   0xd6c34b - std::panicking::rust_panic_with_hook::hb4c40b3721cfbfb0
                at libstd/panicking.rs:480
   5:   0xd6bf2b - std::panicking::continue_panic_fmt::h7e84a0a182ce9c2d
                at libstd/panicking.rs:390
   6:   0xd6be13 - rust_begin_unwind
                at libstd/panicking.rs:325
   7:   0xdc2ccf - core::panicking::panic_fmt::h0c93626b89c38af6
                at libcore/panicking.rs:77
   8:   0xbdd1b7 - core::result::unwrap_failed::hd301dc9e6e84f390
[...]

PR #2210, opened 2018-12-23
docs: be explicit about ARM-based systems not being supported

liamaharon:
From what I can gather, Grin in its current state doesn't build on ARM-based systems.

For newbies like myself, it isn't clear that `Linux x86\_64` excludes ARM-based systems. 
This PR adds a 1 liner making it explicit.

Related: https://github.com/mimblewimble/grin/issues/1681, https://github.com/mimblewimble/grin/issues/2202


liamaharon:
Whoops, just saw this
>For typos and other small changes, either contact one of the developers, or if you think it's a significant enough error to cause problems for other users, please feel free to open an issue.

PR #2211, opened 2018-12-23
Keybase notice on receiving grin transaction

garyyu:
1. Keybase notification message (from self) on receive transaction
2. A new configuration item in `grin-wallet.tom` to message exploding TTL:
```
#The exploding lifetime for keybase notification on coins received.
#Unit: Minute. Default value 1440 minutes for one day.
#Refer to https://keybase.io/blog/keybase-exploding-messages for detail.
#To disable this notification, set it as 0.
keybase_notify_ttl = 1440
```

### DIFF
--- a/wallet/src/adapters/keybase.rs
+++ b/wallet/src/adapters/keybase.rs
@@ -302,8 +302,8 @@ impl WalletCommAdapter for KeybaseWalletCommAdapter {
 
 												let msg = format!(
 													"[grin wallet notice]: \
-												 you could have some coins received from @{}\n\
-												 Transaction Id: {}",
+													 you could have some coins received from @{}\n\
+													 Transaction Id: {}",
 													sender, tx_uuid
 												);
 												notify(&msg, &receiver, config.keybase_notify_ttl);
--- a/wallet/src/adapters/keybase.rs
+++ b/wallet/src/adapters/keybase.rs
@@ -393,9 +393,9 @@ fn notify_on_receive(keybase_notify_ttl: u16, channel: String, tx_uuid: String) 
 					sender, tx_uuid
 				);
 				notify(&msg, &receiver, keybase_notify_ttl);
-				info!("tx from @{} is done, please check on grin wallet. tx uuid: {}",
-					sender,
-					tx_uuid,
+				info!(
+					"tx from @{} is done, please check on grin wallet. tx uuid: {}",
+					sender, tx_uuid,
 				);
 			}
 		} else {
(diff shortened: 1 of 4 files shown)


garyyu:
The message in keybase desktop looks like this:


screen shot 2018-12-23 at 11 49 33 am

wallet/src/adapters/keybase.rs
```diff
@@ -261,3 +305,35 @@ impl WalletCommAdapter for KeybaseWalletCommAdapter {
...
+fn notify_on_receive(keybase_notify_ttl: u16, channel: String, tx_uuid: String) {
+	if keybase_notify_ttl > 0 {
+		let my_username = whoami();
+		if let Some(username) = my_username {
+			let split = channel.split(",");
+			let vec: Vec<&str> = split.collect();
+			if vec.len() > 1 {
+				let receiver = username;
+				let sender = if vec[0] == receiver {
+					vec[1]
```
mcdallas: Keybase allows for multi person channels i.e `me,alice,bob` or `bob,alice,me`. If that's the case the above will fail or even show the wrong sender. One workaround is just to output the channel name, for example: You might have received some grins from {channel}

wallet/src/adapters/keybase.rs
```diff
@@ -242,6 +281,11 @@ impl WalletCommAdapter for KeybaseWalletCommAdapter {
 							// Reply to the same channel with topic SLATE_SIGNED
 							Ok(_) => match send(slate, channel, SLATE_SIGNED, TTL) {
 								true => {
+									notify_on_receive(
+										config.keybase_notify_ttl,
+										channel.to_string(),
+										tx_uuid.to_string(),
+									);
 									println!("Returned slate to {}", channel);
```
mcdallas: nit: put the println before the notification in case the notification fails but the slate is still returned

wallet/src/adapters/keybase.rs
garyyu: thanks @mcdallas for your review, not sure I understand,  are you talking about 3 or more party payment?  current grin wallet only support 2-party transaction, so the channel will always be `someone,me`, yes?

and if leaving it as the whole `{channel}`, that will also include `me`, it will make user confused which one is the real payer.

wallet/src/adapters/keybase.rs
mably: It would be interesting to replace all `println!` by standard logging macros: `debug!`, `info!`, etc.

wallet/src/adapters/keybase.rs
garyyu: 👍  updated for the limitation of 2 party.

wallet/src/adapters/keybase.rs
garyyu: both accepted and updated, thanks for the review.

garyyu:
All review comments are taken and code has been updated.

PR #2215, opened 2018-12-23
 Fix secondary scaling bugs;  rename is_testnet -> is_floonet

tromp:


tromp:
Before relaunch of floonet we must eliminate the first branch of kernel_sig_msg hashing...

tromp:
This code was tested to sync and run fine on current floonet.

PR #2222, opened 2018-12-24
remember to reset sync head on first transition to header sync

antiochp:
Related https://github.com/mimblewimble/grin/pull/2072

I think we were overly aggressive cleaning up code in #2072.

This puts the "reset sync head" back on initial transition to header sync (which is needed).

### DIFF
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -57,7 +57,7 @@ impl HeaderSync {
 			SyncStatus::NoSync | SyncStatus::Initial | SyncStatus::AwaitingPeers(_) => {
 				let sync_head = self.chain.get_sync_head().unwrap();
 				debug!(
-					"sync: initial transition to HeaderSync. sync_head: {} at {}, reset to: {} at {}",
+					"sync: initial transition to HeaderSync. sync_head: {} at {}, resetting to: {} at {}",
 					sync_head.hash(),
 					sync_head.height,
 					header_head.hash(),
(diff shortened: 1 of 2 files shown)


servers/src/grin/sync/header_sync.rs
```diff
@@ -66,6 +64,15 @@ impl HeaderSync {
 					header_head.height,
 				);
 
+				// Reset sync_head to header_head on transition to HeaderSync,
```
ignopeverell: Can you add a couple lines of comments as to what cases this covers? Maybe just an example? Helps to not forget yet again why this is here, a few months from now :-)

servers/src/grin/sync/header_sync.rs
```diff
@@ -66,6 +64,15 @@ impl HeaderSync {
...
+				// but ONLY on initial transition to HeaderSync state.
+				{
+					let store = self.chain.store();
+					let batch = store.batch().unwrap();
+					batch.reset_sync_head().unwrap();
+					batch.commit().unwrap();
+				}
+
```
garyyu: Need move that debug log down (after this reset), because it says `reset to: {} at {}`:

```
				debug!(
					"sync: initial transition to HeaderSync. sync_head: {} at {}, reset to: {} at {}",
					sync_head.hash(),
					sync_head.height,
					header_head.hash(),
					header_head.height,
				);
```

antiochp: Doesn't make sense to log this _after_ updating the `sync_head` as we log `header_head` and `sync_head` (and they will be identical after the reset).

antiochp:
Cleaned the code up - put `reset_sync_head()` back in chain.
Should be good to go now. Will merge if I'm around after travis passes but feel free to merge.

PR #2229, opened 2018-12-25
Implement zeroing memory for wallet password

hashmap:
Testing ergonomics, if this approach is acceptable I'll implement it for
other sensitive parts.

### DIFF
--- a/wallet/src/lmdb_wallet.rs
+++ b/wallet/src/lmdb_wallet.rs
@@ -36,6 +36,7 @@ use crate::libwallet::{internal, Error, ErrorKind};
 use crate::types::{WalletConfig, WalletSeed};
 use crate::util;
 use crate::util::secp::pedersen;
+use crate::util::ZeroingString;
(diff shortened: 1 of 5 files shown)


wallet/src/lmdb_wallet.rs
```diff
@@ -66,7 +67,7 @@ pub struct LMDBBackend<C, K> {
 	db: store::Store,
 	config: WalletConfig,
 	/// passphrase: TODO better ways of dealing with this other than storing
-	passphrase: String,
+	passphrase: ZeroingString,
 	/// Keychain
```
garyyu: Should we change from the 1st door instead of here?  such as in the `inst_wallet()`:
```
	let passphrase = prompt_password(&g_args.password);
```
So, in the `prompt_password()`, `prompt_password_confirm()`, and so on.

ignopeverell:
Nicely structured, thanks!

PR #2230, opened 2018-12-26
Fix couple floonet loose ends

### LINKED ISSUE #2216: Make genesis coinbase easier to restore and spend
Currently it looks like the restore process gets the derivation wrong, leading the an output being properly detected but then setup with the wrong key.

ignopeverell:
Fixes #2216

/cc @tromp @yeastplume

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -1366,20 +1366,12 @@ pub fn kernel_sig_msg(
 	if features.is_coinbase() && fee != 0 || !features.is_height_locked() && lock_height != 0 {
 		return Err(Error::InvalidKernelFeatures);
 	}
-	let msg = if global::is_floonet() {
-		let mut bytes = [0; 32];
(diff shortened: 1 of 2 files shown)


core/src/core/transaction.rs
```diff
@@ -1366,20 +1366,12 @@ pub fn kernel_sig_msg(
...
-		BigEndian::write_u64(&mut bytes[24..], lock_height);
-		secp::Message::from_slice(&bytes)?
-	} else {
-		let hash = match features {
-			KernelFeatures::COINBASE => (features).hash(),
-			KernelFeatures::PLAIN => (features, fee).hash(),
-			_ => (features, fee, lock_height).hash(),
-		};
-		secp::Message::from_slice(&hash.as_bytes())?
+	let hash = match features {
```
mably: References to testnet4 in the function description comment should probably be removed as the corresponding code is being removed for the function body.

ignopeverell: Good catch, fixed.

Issue #2233, opened 2018-12-26
TUI stay on "Downloading headers: 6%, step 1/4"


garyyu:
Already saw several reports on similar issue such as "4%", "6%", etc, and stuck here.

So I look into an old log and find this:
```
20181223 20:56:39.533 DEBUG grin_servers::common::types - sync_state: sync_status: HeaderSync { current_height: 3698, highest_height: 3698 } -> BodySync { current_height: 3698, highest_height: 3698 }
...
20181223 20:56:49.317 DEBUG grin_servers::common::types - sync_state: sync_status: BodySync { current_height: 3698, highest_height: 3698 } -> HeaderSync { current_height: 3698, highest_height: 94148 }
...
20181223 20:56:53.744 DEBUG grin_servers::common::types - sync_state: sync_status: HeaderSync { current_height: 3698, highest_height: 94148 } -> HeaderSync { current_height: 3699, highest_height: 94148 }
```

That means the Floonet network see a highest height at `94,148`!   that's why this `3699` show  `4%`. It definitely is a problem for us, since the current highest height is `7,566`!

I check those old Testnets:
- Testnet4, the highest height stay in `90,900` since 5 days ago
- Testnet3 stay in `142,700` since 69 days ago
- Testnet2 `134,400` at 170 days ago
- Testnet1 `130,000` at 274 days ago

So, it could not be caused by an un-upgraded node which stay on an old testnet, but since https://github.com/mimblewimble/grin/pull/2208 gave different magic code for the new Floonet and Mainnet, it will not be problem in the future to get a peer connected from TestnetN.

But let's give a further think about a possibility of a fraud peer, if this node is broadcasting a highest height far bigger than any other nodes (and a far bigger total_difficulty), in current implementation, it will draw all other nodes into `syncing` state from `NoSync` state, and because current `grin-miner` depends on the node state to start mining, that will cause all other `grin-miner` paused, and only that fraud node can continue mining.

IIRC, this fraud peer issue had been discussed before, but I can't remember what we did to get the immunity on this.  Could somebody remind me?

Anyway, since the issue is still there, looks like we still need a fix to resist such kind of fraud peers.

ignopeverell:
#1139 has a lot of background about this.

And I *think* the check on the miner's block submission has been removed. Or are you talking about startup?

ignopeverell:
Excellent idea!

garyyu:
We could kill a batch of bugs here 😄 
[...]

PR #2237, opened 2018-12-27
Ban a fraud peer which claims a fake height and total difficulty

garyyu:
Fix for #2 problem: https://github.com/mimblewimble/grin/issues/2233#issuecomment-450059401
> No.2 - If a new node is syncing but with this fraud peer connected, it will stuck at Header Syncing stage for ever.

### DIFF
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -132,7 +132,10 @@ impl HeaderSync {
 				self.stalling_ts = None;
 			}
 
-			if !all_headers_received {
+			if all_headers_received {
+				// reset the stalling start time if syncing goes well
+				self.stalling_ts = None;
+			} else {
 				if let Some(ref stalling_ts) = self.stalling_ts {
 					if let Some(ref peer) = self.syncing_peer {
 						match self.sync_state.status() {
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -28,6 +28,9 @@ pub struct HeaderSync {
 
 	history_locator: Vec<(u64, Hash)>,
 	prev_header_sync: (DateTime<Utc>, u64, u64),
+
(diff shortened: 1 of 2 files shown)


servers/src/grin/sync/header_sync.rs
```diff
@@ -117,6 +122,43 @@ impl HeaderSync {
...
+				if let Some(ref stalling_ts) = self.stalling_ts {
+					if let Some(ref peer) = self.syncing_peer {
+						match self.sync_state.status() {
+							SyncStatus::HeaderSync {
+								current_height: _,
+								highest_height,
+							} => {
+								// Ban this fraud peer which claims a higher work but can't send us the real headers
+								if now > *stalling_ts + Duration::seconds(120)
+									&& highest_height == peer.info.height()
```
DavidBurkett: I still find myself struggling with Rust from time to time, but If I understand this code correctly, you're saving stalling_ts the first time we stall during header sync. Then, if 120 seconds pass, and we stall again (potentially with a different peer), we ban the peer that we most recently requested from? That doesn't seem like the right thing to do.

garyyu: Damn! You're right, I completely forgot to clear `self.stalling_ts` when syncing goes well.
thanks @DavidBurkett

garyyu: Updated for reset `self.stalling_ts` when syncing goes well.

garyyu:
The banned state can't save into the database, that's driving me crazy!
```
	/// Ban a peer, disconnecting it if we're currently connected
	pub fn ban_peer(&self, peer_addr: &SocketAddr, ban_reason: ReasonForBan) {
		if let Err(e) = self.update_state(*peer_addr, State::Banned) {
			error!("Couldn't ban {}: {:?}", peer_addr, e);
		}
```

This `ban_peer` function is called successfully, but the banned peer always can connect again. I dig into the raw lmdb data file and confirmed indeed it doesn't change the peer state in database.

What I have missed?

PR #2240, opened 2018-12-27
[DNM] fix: new mined block need broadcast

garyyu:
Regarding to fraud peer issue https://github.com/mimblewimble/grin/issues/2233#issuecomment-450005782

@ignopeverell 
> And I think the check on the miner's block submission has been removed.

I believe the block submission on node syncing state do have been removed, but without broadcasting 😸 

I'm not sure removing the check here (in this PR) is a good idea, because this check is absolutely needed for the real initial syncing stage of a new node.

Want some suggestions on this PR.  @ignopeverell   @antiochp

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -635,9 +635,9 @@ impl ChainAdapter for ChainToPoolAndNetAdapter {
 			}
 		}
 
-		if self.sync_state.is_syncing() {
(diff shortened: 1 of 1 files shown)


garyyu:
Let's close this one, since we already fixed the fraud peer (with fake height and total difficulty) banning  issue, then we should never come to this case.

PR #2257, opened 2018-12-29
Allow passing multiple options to grin in docker

kim0:
Docker best practice entrypoint should be the main binary, while cmd should be the default (but overrideable options). Using this patch enables using the docker image as a replacement to the `grin` binary. For example this command now works `docker run --rm -it grin wallet --help` .. Also `alias grin=docker run --rm -it grin` can be useful

### DIFF
--- a/p2p/src/handshake.rs
+++ b/p2p/src/handshake.rs
@@ -14,7 +14,7 @@
 
 use crate::util::RwLock;
 use std::collections::VecDeque;
-use std::net::{Shutdown, SocketAddr, TcpStream};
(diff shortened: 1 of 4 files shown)


kargakis:
You also need to be passing --floonet in `grin server config` at L39 which is probably broken atm since grin now needs to create a directory for each network config and data.

kargakis:
Also mind adding the extra mainnet ports in the EXPOSE command? I think they are going to be 3413, 3414, 3415, 3416

Issue #1250, opened 2018-07-11
[dandelion] specify node address for dandelion stem phase


antiochp:
Right now its *really* difficult to test/demonstrate the aggregation behavior via Dandelion.
You can send txs from a couple of nodes but the chances of them actually crossing paths during the stem phase is low.

* we only send txs during stem phase to a single outbound node
* we randomly choose that stem node every 10 mins
* we need to wait for a confirmation before sending a subsequent tx (spend all outputs by default)

It would be useful if we could either - 
1. configure a single outbound node in `grin.toml` to always forward stems txs to, or
1. pass an optional (first hop) node address when sending a tx from the grin command line

This would allow us to get a couple of txs into the same node within a short period of time and have them aggregate on their way through the stem phase with reasonably high success rate.

beaver-tooth:
Hey I am interested in giving this a whack this weekend.

sesam:
To get warmed up, you could start with adding a command line option and
print it out like "Dandelion stem added: (partially masked IP)" so it will
show in the logs, but not fully give away where the stem is. I.e. logs
should not give away too much details if grin wants to claim the anonymous
niche.
And on Monday there's a dev meeting on gitter chat, please come join there!

beaver-tooth:
Gave this a start. @antiochp are you partial to option 1 or 2?
> 
>    1. configure a single outbound node in grin.toml to always forward stems txs to, or
>    2. pass an optional (first hop) node address when sending a tx from the grin command line
> 

Or would you like to see a PR with both?

beaver-tooth:
Thanks @sesam , I will go down this route. I will also try to tune in monday!

beaver-tooth:
Just wanted to update -- I spent last couple days getting grin up and running and also figuring out how to get a proper local net up and working to play around.

With this up and running, I am able to give this a serious start.

beaver-tooth:
Giving this up, swamped with outside work stuff right now.

TTD1993:
grin.toml

svechinsky:
Is this still necessary?
If so I would like to give it a try

sesam:
yes I think so. There might be other things more urgent though until mainnet. Once you got grin building and synched up, come into the dev chat for a better update.

PR #2263, opened 2018-12-31
feature: Enable setting a node address for a dandelion peer

svechinsky:
So this should address https://github.com/mimblewimble/grin/issues/1250
I don't have too much experience with rust so not sure if this is the right way to do it.
I also don't have much idea as to the best way to test this, what's the best way to go about it?

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -230,6 +230,10 @@ fn comments() -> HashMap<String, String> {
 
 # 15 = Bit flags for FULL_NODE
 #This structure needs to be changed internally, to make it more configurable
+
+# A prefered dandelion_peer, mainly used for testing dandelion
+# dandelion_peer = \"10.0.0.1:13144\"
+
 "
 		.to_string(),
 	);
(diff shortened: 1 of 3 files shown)


config/src/comments.rs
```diff
@@ -230,6 +230,10 @@ fn comments() -> HashMap<String, String> {
 
 # 15 = Bit flags for FULL_NODE
 #This structure needs to be changed internally, to make it more configurable
+
+# A prefered dandelion_peer, mainly used for testing dandelion
```
kargakis: typo: preferred

p2p/src/peers.rs
```diff
@@ -83,24 +83,38 @@ impl Peers {
...
-				let dandelion_relay = &self.dandelion_relay;
-				dandelion_relay.write().clear();
-				dandelion_relay
-					.write()
-					.insert(Utc::now().timestamp(), peer.clone());
-				debug!(
-					"Successfully updated Dandelion relay to: {}",
-					peer.info.addr
-				);
+		match &self.config.dandelion_peer {
```
ignopeverell: This would likely look nicer (and more idiomatic) if you first looked for and assigned the peer your want to set as relay, and then did the set on that. Look for the `find` method from `Iterator`.

PR #2265, opened 2018-12-31
support floonet, separate entrypoint from cmd, add docker volume

kim0:
* Support floonet as a buildtime ARG
* Separate entrypoint from cmd, allowing the use of different grin sub-commands on `docker run`
* Adding a docker volume to hold state

### DIFF
--- a/doc/build.md
+++ b/doc/build.md
@@ -96,15 +96,15 @@ grin client help
 ```sh
 docker build -t grin -f etc/Dockerfile .
 ```
-for floonet, use `etc/Dockerfile.floonet` instead
+For floonet, use `etc/Dockerfile.floonet` instead
 
 You can bind-mount your grin cache to run inside the container.
 
 ```sh
 docker run -it -d -v $HOME/.grin:/root/.grin grin
 ```
-if you prefer to use a docker named volume, you can pass `-v dotgrin:/root/.grin` instead.
-Using a named volume copies a default grin-server.toml
+If you prefer to use a docker named volume, you can pass `-v dotgrin:/root/.grin` instead.
+Using a named volume copies default configurations upon volume creation
 
 ## Cross-platform builds
 
--- a/doc/build.md
+++ b/doc/build.md
@@ -96,12 +96,15 @@ grin client help
 ```sh
 docker build -t grin -f etc/Dockerfile .
 ```
+for floonet, use `etc/Dockerfile.floonet` instead
 
 You can bind-mount your grin cache to run inside the container.
 
 ```sh
 docker run -it -d -v $HOME/.grin:/root/.grin grin
 ```
+if you prefer to use a docker named volume, you can pass `-v dotgrin:/root/.grin` instead.
+Using a named volume copies a default grin-server.toml
 
 ## Cross-platform builds
 
--- a/etc/Dockerfile
+++ b/etc/Dockerfile
@@ -36,9 +36,18 @@ ENV LANG en_US.UTF-8
 COPY --from=builder /usr/src/grin/target/release/grin /usr/local/bin/grin
 
 WORKDIR /root/.grin
-RUN grin server config && \
+
+ARG grinnet=""
+
+RUN grin ${grinnet} server config && \
     sed -i -e 's/run_tui = true/run_tui = false/' grin-server.toml
+
+VOLUME ["/root/.grin"]
 
 EXPOSE 13413 13414 13415 13416
 
-ENTRYPOINT ["grin", "server", "run"]
+EXPOSE 3413 3414 3415 3416
+
+ENTRYPOINT ["grin"]
+
+CMD ["server", "run"]
--- a/etc/Dockerfile.floonet
+++ b/etc/Dockerfile.floonet
@@ -37,13 +37,13 @@ COPY --from=builder /usr/src/grin/target/release/grin /usr/local/bin/grin
 
 WORKDIR /root/.grin
 
-RUN grin server config && \
+RUN grin --floonet server config && \
     sed -i -e 's/run_tui = true/run_tui = false/' grin-server.toml
 
 VOLUME ["/root/.grin"]
 
-EXPOSE 3413 3414 3415 3416
+EXPOSE 13413 13414 13415 13416
 
-ENTRYPOINT ["grin"]
+ENTRYPOINT ["grin", "--floonet"]
 
 CMD ["server", "run"]


kim0:
@kargakis .. This is a continuation of https://github.com/mimblewimble/grin/pull/2257 which it seems I closed .. Anyway, this is a bigger revamp.

ignopeverell:
This doesn't look like it's going to run Floonet (no `--floonet` option) but exposes future mainnet ports?

kim0:
@ignopeverell .. I default to mainnet .. To build `--floonet` use the below

`docker build -t grin -f etc/Dockerfile . --build-arg grinnet=--floonet`

Ofc, grinnet could be anything else supported by grin. If not mentioned (default), it should default to mainnet

etc/Dockerfile
```diff
@@ -36,9 +36,18 @@ ENV LANG en_US.UTF-8
...
 
 WORKDIR /root/.grin
-RUN grin server config && \
+
+ARG grinnet=""
+
+RUN grin ${grinnet} server config && \
     sed -i -e 's/run_tui = true/run_tui = false/' grin-server.toml
 
+VOLUME ["/root/.grin"]
```
kargakis: The volume cmd is unnecessary?

etc/Dockerfile
```diff
@@ -36,9 +36,18 @@ ENV LANG en_US.UTF-8
...
+
 EXPOSE 13413 13414 13415 13416
 
-ENTRYPOINT ["grin", "server", "run"]
+EXPOSE 3413 3414 3415 3416
```
kargakis: Add a comment that these are the mainnet ports.

etc/Dockerfile
```diff
@@ -36,9 +36,18 @@ ENV LANG en_US.UTF-8
...
+
+ENTRYPOINT ["grin"]
```
kargakis: Shouldn't you be using grinnet here?

etc/Dockerfile
```diff
@@ -36,9 +36,18 @@ ENV LANG en_US.UTF-8
 COPY --from=builder /usr/src/grin/target/release/grin /usr/local/bin/grin
 
 WORKDIR /root/.grin
-RUN grin server config && \
+
+ARG grinnet=""
+
+RUN grin ${grinnet} server config && \
     sed -i -e 's/run_tui = true/run_tui = false/' grin-server.toml
```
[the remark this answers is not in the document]
kim0: That was my impression too .. but when I tested, it worked .. I see the below inside the container
```
root@27fa945a58d6:~/.grin# ls
chain_data  floo  grin-server.log  grin-server.toml
root@27fa945a58d6:~/.grin# ls floo/
root@27fa945a58d6:~/.grin# ls -la floo/
total 10
drwxr-xr-x 2 root root  3 Dec 31 19:31 .
drwxr-xr-x 4 root root  6 Dec 31 19:31 ..
-rw-r--r-- 1 root root 20 Dec 31 19:31 .api_secret
```

etc/Dockerfile
kim0: Not really .. since ARG is only available at buildtime. In run time, passing `--floonet` should just be regarded as extra cli args, which are now supported. So for running a floonet container, I use:
`docker run --rm -it -p 13413-13416:13413-13416 -v dotgrin:/root/.grin --name grin grin --floonet server run`

etc/Dockerfile
kim0: 👍

etc/Dockerfile
kim0: Nope, it's useful .. When it's not marked as a volume .. docker does not copy data from inside the image to that location .. so the grin-server.toml is **not** copied to that external location. While when marked as a volume, it is. When it's marked as a volume, we should pass a volume name like `docker run ... -v dotgrin:/root/.grin` .. and see the volume with `docker volume ls`

AFAICT, this is how docker should be used! And it works well. When bind mounting a volume -v .:/root/.grin .. docker does NOT copy the grin-server.toml file.

kim0:
@kargakis .. I see 2 paths ... either the one above .. and I simply add a couple of lines of docs on how to build and run in both cases (floonet and mainnet) .. or we simply 2 dockerfiles (Dockerfile for mainnet, and Dockerfile.floonet for floonet) and we basically avoid the 2 extra pieces of config when using floonet (--build-arg when building, and --floonet arg when running)

Let me know which path makes more sense to you, and I'll adjust and update the readme
thanks!

kim0:
In my opinion, I think the 2 files solution would be slightly better. Also, ideally grin would accept some configs from environment variables `export GRINNET=floonet` or so! Like all buzzword compliant apps (cloud-native ..etc) should :)

kargakis:
I think I prefer the separate Dockerfile approach too, it keeps the
Dockerfile simpler. So maybe rename the present Dockerfile to
Dockerfile.floonet?

Regarding the VOLUME I don't feel strong with adding it since I don't
expect most people are going to use Docker directly to deploy Grin but will
use something like Kubernetes which handles volume management separately.
Maybe it's fine to have it for local development?

kim0:
@kargakis .. Does that look better? I'm still awaiting the build on my box, which takes 24 mins :) Is there any interest to try improving performance by caching some docker layers? Let me know, and I can work on it .. it will add a few lines to the dockerfile though

doc/build.md
```diff
@@ -96,12 +96,15 @@ grin client help
 ```sh
 docker build -t grin -f etc/Dockerfile .
 ```
+for floonet, use `etc/Dockerfile.floonet` instead
```
kargakis: nit: "For floonet, use `etc/Dockerfile.floonet` instead."

doc/build.md
```diff
@@ -96,12 +96,15 @@ grin client help
...
 
 You can bind-mount your grin cache to run inside the container.
 
 ```sh
 docker run -it -d -v $HOME/.grin:/root/.grin grin
 ```
+if you prefer to use a docker named volume, you can pass `-v dotgrin:/root/.grin` instead.
```
kargakis: If

doc/build.md
```diff
@@ -96,12 +96,15 @@ grin client help
...
+Using a named volume copies a default grin-server.toml
```
kargakis: The grin cache does not contain only the server config. Also, is this true for named volumes that already contain cached data?

kargakis:
> @kargakis .. Does that look better? I'm still awaiting the build on my box, which takes 24 mins :) Is there any interest to try improving performance by caching some docker layers? Let me know, and I can work on it .. it will add a few lines to the dockerfile though

Yes, it looks good to me, had just a couple of comments in the docs. Regarding speeding up the docker build, we could tear down the whole build step process and bind-mount the binary from the host, so the flow for building the image would be something like this:
```
cargo build --release
docker build -t grin -f etc/Dockerfile
```
and inside the container we would
```
COPY target/release/grin /usr/bin/
```
The downside would be that we would have to keep the dependencies on the host but I think most folks who build it have the necessary deps installed already.

doc/build.md
kim0: Yeah, I know it contains other data. When docker creates a new volume .. it will copy everything in the image under /root/.grin to that volume. But it will not for a bind mounted directory. When using a named volume `-v dotgrin:/root/.grin` .. if the volume contains data, it will not be overwritten

kim0:
@kargakis .. I fixed the typos .. Let's get this merged if you like :)

I don't really like having the build deps on the host for reproducibility at least .. let's look at improving build performance in a separate PR .. Thanks!

kargakis:
Thanls, lgtm

Issue #2241, opened 2018-12-27
Peer ban doesn't work as expected


garyyu:
A banned peer continue appearing on my peers list, check log and find this:
```
20181227 17:19:12.838 DEBUG grin_p2p::protocol - handler: consume: peer V4(35.227.48.174:13414) banned, received: GetPeerAddrs, dropping.
20181227 17:19:19.956 DEBUG grin_p2p::peers - clean_peers V4(35.227.48.174:13414), peer banned
20181227 17:19:52.979 DEBUG grin_p2p::peers - Saving newly connected peer 35.227.48.174:13414.
20181227 17:19:52.979 DEBUG grin_p2p::store - save_peer: V4(35.227.48.174:13414) marked Healthy
20181227 17:19:52.982 DEBUG grin_p2p::peers - Saving newly connected peer 35.227.48.174:13414.
20181227 17:19:52.982 DEBUG grin_p2p::store - save_peer: V4(35.227.48.174:13414) marked Healthy
```

That means when I ban a peer, it works. But after 30 seconds, if this banned peer request connect again, it will connect to my node again!

It will enter an infinite loop here: banned, connected, banned, connected, ...
```
20181227 17:20:12.843 DEBUG grin_p2p::protocol - handler: consume: peer V4(35.227.48.174:13414) banned, received: GetPeerAddrs, dropping.
20181227 17:20:20.151 DEBUG grin_p2p::peers - clean_peers V4(35.227.48.174:13414), peer banned
20181227 17:20:52.978 DEBUG grin_p2p::peers - Saving newly connected peer 35.227.48.174:13414.
20181227 17:20:52.978 DEBUG grin_p2p::store - save_peer: V4(35.227.48.174:13414) marked Healthy
```

garyyu:
It proves that this is caused by a wrong usage.
I called `peer.set_banned();` but actually this function call will never write ban state into database.  Instead, I should call `ban_peer::ban_peer()`.

Sorry for noise.

garyyu:
Even after switching to `peers::ban_peer()`, the banned peer still always can connect.

Finally I find the root cause, it looks like a deep buried bug here which was committed on  **2018-03-28** since the `is_banned()` function was written:
```
------------------------------- p2p/src/peers.rs -------------------------------
index c09dfb00..0e8ccb9b 100644
@@ -638,4 +638,13 @@ impl NetAdapter for Peers {
 			}
 		}
 	}
+
[...]

PR #2267, opened 2019-01-01
fix: check_banned must care about ip address only, and ignore the peer port info

garyyu:
To close https://github.com/mimblewimble/grin/issues/2241 again.

Please refer to https://github.com/mimblewimble/grin/issues/2241#issuecomment-450652666 and https://github.com/mimblewimble/grin/issues/2241#issuecomment-450704362 for the detail info.

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -208,9 +208,22 @@ impl Peers {
 	}
 
 	pub fn is_banned(&self, peer_addr: SocketAddr) -> bool {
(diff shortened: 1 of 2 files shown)


antiochp:
Do we _really_ want to ignore the port here? 
Somebody mentioned nodes behind a NAT in another conversation - this seems applicable here.
And what about for testing (say locally) - ideally we would not ban all nodes on a given ip address (say 127.0.0.1) if a single node did something bad.

PR #2277, opened 2019-01-02
Wallet crate cleanup and idiomatic

elichai:
Pretty similiar to this(https://github.com/mimblewimble/grin/pull/2269) but for the wallet crate.

### DIFF
--- a/servers/tests/framework.rs
+++ b/servers/tests/framework.rs
@@ -14,9 +14,7 @@
 
 use self::keychain::Keychain;
 use self::util::Mutex;
-use self::wallet::{
-	HTTPNodeClient, HTTPWalletCommAdapter, LMDBBackend, WalletCommAdapter, WalletConfig,
-};
+use self::wallet::{HTTPNodeClient, HTTPWalletCommAdapter, LMDBBackend, WalletConfig};
 use blake2_rfc as blake2;
 use grin_api as api;
 use grin_core as core;
@@ -127,6 +125,7 @@ impl Default for LocalServerContainerConfig {
 /// A top-level container to hold everything that might be running
 /// on a server, i.e. server, wallet in send or receive mode
 
+#[allow(dead_code)]
 pub struct LocalServerContainer {
 	// Configuration
 	config: LocalServerContainerConfig,
@@ -307,6 +306,7 @@ impl LocalServerContainer {
 	}
 
 	/// Starts a wallet owner daemon
+	#[allow(dead_code)]
 	pub fn run_owner(&mut self) {
 		let wallet = self.make_wallet_for_tests();
 
@@ -329,6 +329,7 @@ impl LocalServerContainer {
 		});
 	}
 
+	#[allow(dead_code)]
 	pub fn get_wallet_seed(config: &WalletConfig) -> wallet::WalletSeed {
 		let _ = fs::create_dir_all(config.clone().data_file_dir);
 		wallet::WalletSeed::init_file(config, 32, None, "").unwrap();
@@ -337,6 +338,7 @@ impl LocalServerContainer {
 		wallet_seed
 	}
 
+	#[allow(dead_code)]
 	pub fn get_wallet_info(
 		config: &WalletConfig,
 		wallet_seed: &wallet::WalletSeed,
@@ -354,6 +356,7 @@ impl LocalServerContainer {
 		wallet::libwallet::internal::updater::retrieve_info(&mut wallet, &parent_id, 1).unwrap()
 	}
 
+	#[allow(dead_code)]
 	pub fn send_amount_to(
 		config: &WalletConfig,
 		amount: &str,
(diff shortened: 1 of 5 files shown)


elichai:
hmm, it looks like github doesn't handle well the squash commits.
maybe I'll need to force push from upstream before I continue working so the old commits will get replaced by the squashed ones.

elichai:
Ok, after reading it seems that the right solution to this is creating a new branch for every PR/feature and keeping my master aligned with upstream

servers/tests/framework.rs
```diff
@@ -264,7 +264,8 @@ impl LocalServerContainer {
 		let _ = fs::create_dir_all(self.wallet_config.clone().data_file_dir);
 		let r = wallet::WalletSeed::init_file(&self.wallet_config, 32, "");
 
-		let client_n = HTTPNodeClient::new(&self.wallet_config.check_node_api_http_addr, None);
+		let client_n =
+			HTTPNodeClient::new(self.wallet_config.check_node_api_http_addr.clone(), None);
```
hashmap: what is the reason to pass `String`?

wallet/src/adapters/http.rs
```diff
@@ -45,12 +45,12 @@ impl WalletCommAdapter for HTTPWalletCommAdapter {
 			error!("{}", err_str,);
 			Err(ErrorKind::Uri)?
 		}
-		let url = format!("{}/v1/wallet/foreign/receive_tx", dest);
+		let url = dest.to_owned() + "/v1/wallet/foreign/receive_tx";
```
hashmap: for my education - why do you prefer `+`?

servers/tests/framework.rs
elichai: Because `HTTPNodeClient::new` needs an ownership of this String,
before it received an `&str` and then cloned it using `to_owned()`, but this can cause unnecessary clone if the caller of the function doesn't need the String any more.
The more idiomatic thing to do is to try and not make any decisions for the caller, if you need ownership you should take it and let the owner decide if he should clone or not, and if you don't need ownership get a refrence.

wallet/src/adapters/http.rs
elichai: Actually in this case there shouldn't be any meaningful difference except for readability.
But in general, the `format!` macro is almost always less efficient than pushing, it dismisses any pre allocated memory and just copies everything to a new memory location.

But here because `dest` is `&str` and not `String` it's not really less efficient and I probably did it out of habit(I use format only on non string variables), but I shouldn't have changed someone elses code for no good enough reason. 

btw, even with `&str`s `format!` is just a little slower than anything else https://github.com/hoodie/concatenation_benchmarks-rs

mably: Seems to be some kind of premature optimization to me at this stage.

elichai: @mably you're right. If it's already a String I think you shouldn't use `format!`, you should use `push` or `+`,  because that's less idiomatic and multiple redundant allocations.
But if it's a `&str` anyway that's really a premature optimization and I won't change that in the future.

servers/tests/framework.rs
garyyu: Looks like not beautiful as it was, adding a lot of `clone()` here and there.

elichai: Hmm I agree that in some situations it can look less nice, 
but it can accumulate to a lot of memory if you pass a lot of `&Vec<T>` and `String` to functions and clone them inside and they can pass it on to another function that will clone it too etc.

wallet/src/adapters/http.rs
garyyu: If no performance improvement, or simply the code, I prefer not changing those trivial things.

And agree with @mably , better to focus on functional issues at this stage.

elichai: I can try and amend it out if you prefer

garyyu:
Thanks @elichai for your 2nd PR 😄 

Generally comments:   Could you split into small PRs?  hard to review for so many changes. And better to focus on one issue only in one PR.  No need to change some personal taste things, such as `string.len() > 0` or `is_empty()`, etc.  If not obvious, we'd better explain the change reason(s) in PR description.

hashmap:
@garyyu  I wouldn't call `len` vs `is_empty` a personal taste thing, it's a clippy warning `len-zero` (https://rust-lang.github.io/rust-clippy/current/index.html):

> Why is this bad
Some structures can answer .is_empty() much faster than calculating their length. Notably, for slices, getting the length requires a subtraction whereas .is_empty() is just a comparison. So it is good to get into the habit of using .is_empty(), and having it is cheap. Besides, it makes the intent clearer than a manual comparison.

garyyu:
@hashmap Agree!  Looks like I listed a bad example there :P
So, pls ignore that example in my comments.

garyyu:
@elichai appreciated if you can split, thanks~

PR #2278, opened 2019-01-02
fix keybase slate return

bladedoyle:
dont drop the slate after send_tx_sync, and also call finalize here f……oloowing the same pattern as other sync method

Issue #2266, opened 2018-12-31
LibTx Error when finalizing a slate returned by issue_send_tx using keybase


bladedoyle:
I am not able to get a keybase tx to finalize using the API.  This is the base error message I can find:

```
20181231 13:26:33.790 ERROR grin_wallet::controller - finalize_tx: failed with error: LibTx Error
20181231 13:26:33.790 ERROR grin_wallet::controller - Request Error: Error { inner: 

LibTx Error }
```

Please use the following steps to reproduce:

run a grin server on address 0.0.0.0:13413
  grin --floonet server run

set the owner_api key to "key"

Run a owner_api listener:
  grin --floonet wallet -p dummy owner_api

Have a local wallet with some grin in it

have keybase server running and a user logged in

Run the test script:
  ./keybase_send_tx.py

where  ./keybase_send_tx.py contains:
```
#!/usr/bin/env python3

import requests
import json
import sys

keybase_user = "qsaszs"
grin_owner_api_url = "http://localhost:13420"
wallet_auth = ("grin", "key")

def main():
    tx_send_url = grin_owner_api_url + "/v1/wallet/owner/issue_send_tx"
    data = {
            'method': 'keybase',
            'amount': 1230000000,
            'minimum_confirmations': 10,
            'dest': keybase_user,
            'max_outputs': 10,
            'num_change_outputs': 1,
            'selection_strategy_is_use_all': False,
            'fluff': False,
        }
    print("Command: {} {}".format(tx_send_url, data))
    r = requests.post(
            url=tx_send_url,
            data=json.dumps(data),
            auth=wallet_auth,
        )
    if not (r.status_code >= 200 and r.status_code < 300):
        print("Send failed with error {} : {}".format(r.status_code, r.reason))
        sys.exit(1)
    print("Sent OK: {} : {}".format(r.status_code, r.json()))


    # WARNING:  keybase send does not FINALIZE or POST the tx after, we need to do that manually

    tx_finalize_url = grin_owner_api_url + "/v1/wallet/owner/finalize_tx"
    print("Command: {}".format(tx_finalize_url))
    r = requests.post(
            url=tx_finalize_url,
            data=r.text,
            auth=wallet_auth,
        )
    if not (r.status_code >= 200 and r.status_code < 300):
        print("Finalize failed with error {} : {}".format(r.status_code, r.reason))
        sys.exit(1)
    print("Finalized OK: {} : {}".format(r.status_code, r.json()))


[...]

PR #2280, opened 2019-01-03
a bit refactoring on wallet controller for issue_send_tx

garyyu:
- a little bit refactoring based on https://github.com/mimblewimble/grin/pull/2278
- leave a TODO item there for the service hanging issue when using keybase (according to @mcdallas 's concern: https://github.com/mimblewimble/grin/issues/2266#issuecomment-450702339)
- clean 2 minor build warnings

### DIFF
--- a/src/bin/cmd/wallet.rs
+++ b/src/bin/cmd/wallet.rs
@@ -14,7 +14,9 @@
 
 use crate::cmd::wallet_args;
 use crate::config::GlobalWalletConfig;
+/* Turned off for the time being
 use crate::servers::start_webwallet_server;
+*/
 use clap::ArgMatches;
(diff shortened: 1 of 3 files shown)


src/bin/cmd/wallet.rs
```diff
@@ -14,7 +14,9 @@
 
 use crate::cmd::wallet_args;
 use crate::config::GlobalWalletConfig;
+/* Turned off for the time being
```
hashmap: should we just remove it?

garyyu: 😄   I just copy all those web wallet related switching off.
Ok to remove it, since just one `use`.

PR #2283, opened 2019-01-03
add @hendi and community DNS seed

hendi:
This adds two more DNS seeds for floonet and mainnet:

`(main|floo)net.grin-seed.prokapi.com` are at least 3 seeds that power www.grinexplorer.net and that I update on a rolling basis.

`accio.(main|floo)net.grin-seed.prokapi.com` uses the summoning charm to get as many community nodes as possible. I'll monitor the [forum post](https://www.grin-forum.org/t/grin-mainnet-seeds-needed/1580/7) to keep it updated.

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -36,12 +36,14 @@ const MAINNET_DNS_SEEDS: &'static [&'static str] = &[
 	"mainnet.seed.grin.icu",           // gary.peverell@protonmail.com
 	"mainnet.seed.713.mw",             // jasper@713.mw
 	"mainnet.seed.grin.lesceller.com", // q.lesceller@gmail.com
+	"mainnet.seed.grin.prokapi.com",   // hendi@prokapi.com
 ];
 const FLOONET_DNS_SEEDS: &'static [&'static str] = &[
 	"floonet.seed.grin-tech.org",      // igno.peverell@protonmail.com
 	"floonet.seed.grin.icu",           // gary.peverell@protonmail.com
 	"floonet.seed.713.mw",             // jasper@713.mw
 	"floonet.seed.grin.lesceller.com", // q.lesceller@gmail.com
+	"floonet.seed.grin.prokapi.com",   // hendi@prokapi.com
 ];
 
 pub fn connect_and_monitor(


servers/src/grin/seed.rs
```diff
@@ -32,14 +32,18 @@ use crate::util::{Mutex, StopState};
...
 // DNS Seeds with contact email associated
 const MAINNET_DNS_SEEDS: &'static [&'static str] = &[
-	"mainnet.seed.grin-tech.org", // igno.peverell@protonmail.com
-	"mainnet.seed.grin.icu",      // gary.peverell@protonmail.com
-	"mainnet.seed.713.mw",        // jasper@713.mw
+	"mainnet.seed.grin-tech.org",          // igno.peverell@protonmail.com
+	"mainnet.seed.grin.icu",               // gary.peverell@protonmail.com
+	"mainnet.seed.713.mw",                 // jasper@713.mw
+	"mainnet.grin-seed.prokapi.com",       // hendi@prokapi.com
+	"accio.mainnet.grin-seed.prokapi.com", // hendi@prokapi.com (email to add your IP)
```
garyyu: perhaps merge your 6 nodes into one dns? and can make the code looks more clean here.
the `grin-seed` can be simple like `seed`?
Sorry for my nitpicking, just want a beautiful code 😄

hendi: > perhaps merge your 6 nodes into one dns?
I've put these into two separate records because the first are my nodes that I commit to keep running, and use for the explorer etc. The other ones are separate from these, not run by me, but sourced from the forum thread. Igno asked me to do this.

garyyu: better to follow the `mainnet.seed`, `floonet.seed` style, then no exception for all seeds, including any new seeds to be added. does it make sense?

and a little bit worry about the `email to add your IP` part, because we can't know if it's a fraud node, without a period to check its honestness. Shall we remove it?

In the future, we can reply on some additional auto collection DNS seeds, based on a long term honestness checking, such as the banning happening counter, online time ratio, and so on. IIRC, there's a prototype somewhere.

hendi: Sorry, I don't quite understand you, I think we're talking past each other. Let me try to be more verbose:

(And now that I'm about to write this, I probably should've done two separate PRs).

1. `(main|floo)net.grin-seed.prokapi.com` are my peers, I run them, I vouch for them, I commit to them being updated and running for the next 2 years. These should be uncontroversial and should be added.

2. I've posted these in the Grin forum, and after that had a quick chat with @ignopeverell who asked me to also add the IPs that got posted by people as well. Since this is a list of DNS seeds, we need to add domain names, not IPs. Some people don't have a domain, but just a server with a static IP running grin. This is what `accio.(main|floo)net.grin-seed.prokapi.com` is for.

I don't think we should have a uniform format auf domain names as a requirement. Same as you I want clean and beautiful code, but IMHO the DNS seeds are not related to this. Neither are the non-uniform email addresses.

garyyu: Sorry, I have to insist to not taking `email to add your IP`. Currently, we're not smart enough to quickly address the fraud peer(s) and ban them, even with a few fixes already done on the banning, but my current implementation is quite naive and every connected fraud peer will still get things stuck for 2 minutes before it find and ban. This PR will definitely put salt on that wound.

We indeed need improve our banning algorithm, but before we complete that, we have to be careful to add trusted dns seeds.

I agree with @hendi that it's good to split these 2 dns seeds and we can merge @hendi 's self nodes.

ignopeverell:
Can we get a rebase? And regarding "add the IPs that got posted by people", I didn't mean any random people. I meant seeds that have been vouched for by us and may already be advertised by other DNS names.

hendi:
Oops, sorry for the confusion! I've rebased and changed the PR to just add my seeds, not random IPs.

PR #2285, opened 2019-01-03
Use ZeroingString for password and passpharse in wallet

hashmap:
There are a few usage of sensitive data as `&str`, it's safe.


ignopeverell:
@hashmap looks like that breaks a test now, care to take a look?

hashmap:
@ignopeverell fixed

PR #2287, opened 2019-01-04
Clean existing sockets when accept new p2p connection with same address

garyyu:
Still suffering on the full connection problem on seed servers, which force me restart grin server several times every day (I don't want to schedule a periodic auto-restart).

Examples:
```
tcp        0      0 10.142.0.2:13414        40.114.250.29:2944      SYN_RECV   
tcp        0      0 10.142.0.2:13414        40.114.250.29:2946      SYN_RECV   
tcp        0      0 10.142.0.2:13414        40.114.250.29:2947      SYN_RECV   
tcp        0      0 10.142.0.2:13414        40.114.250.29:2948      SYN_RECV   
```
```
tcp        0      0 10.142.0.2:13414        35.185.101.157:39166    SYN_RECV   
tcp        0      0 10.142.0.2:13414        35.185.101.157:39182    SYN_RECV   
tcp        0      0 10.142.0.2:13414        35.185.101.157:39246    SYN_RECV   
```
```
tcp        0      0 10.142.0.2:13414        117.50.56.131:45802     SYN_RECV   
tcp        0      0 10.142.0.2:13414        117.50.56.131:45922     SYN_RECV   
tcp        0      0 10.142.0.2:13414        117.50.56.131:45992     SYN_RECV   
... 
tcp        0      0 10.142.0.2:13414        117.50.56.131:46136     SYN_RECV   
```
Or even in 'ESTABLISHED` state:
```
tcp        0    102 45.118.135.254:51034    45.76.0.106:13414       ESTABLISHED
tcp        0    102 45.118.135.254:51182    45.76.0.106:13414       ESTABLISHED
tcp        0    102 45.118.135.254:53060    45.32.188.238:13414     ESTABLISHED
tcp        0    102 45.118.135.254:53110    45.32.188.238:13414     ESTABLISHED
```

I test with this PR, and finally it's clean, no more wasted tcp connections, the `13414` port connections number is always same (or very close) as the peers number.

The basic idea is putting a clean-up function on every peer connection accept, if same IP address is found in the saved list, then we close the old stream before accept new connection.

And another clean-up for the saved socket addresses: if it's not in the peers list, we clean it also to close tcp connection, since it will never be used. Somehow, we leaked some connections, but I still don't know where.

PR #2298, opened 2019-01-06
clean the unused p2p sockets (i.e. not in peers list)

garyyu:
Draw half of https://github.com/mimblewimble/grin/pull/2287 to here:

When a connected socket is not in our active peers list, it becomes a lost peer and we never have chance to use it for sending, and it brings us a security risk that we connected an "invisible" peer (or peers), this "invisible" peer still can send us something!

This PR will kill this risk.

### DIFF
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -201,6 +201,14 @@ impl Server {
 		false
 	}
 
+	/// For all kinds of exception cases, the node could accepted / initiated a peer connection successfully but
+	/// failed on the Handshake protocol communication, or a connected peer was closed but without a successful
+	/// clean-up on its socket, that will cause this connected (on TcpStream) peer becomes so-called "invisible" peer!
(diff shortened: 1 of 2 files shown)


p2p/src/serv.rs
```diff
@@ -191,6 +201,25 @@ impl Server {
 		false
 	}
 
+	fn clean_lost_sockets(&self, sockets: &mut HashMap<SocketAddr, TcpStream>) {
```
ignopeverell: Perhaps add a code comment here outlining the NAT issue and why this direct connection tracking and cleanup is necessary? This way we won't forget why this is required right now, and what sort of improvements will be needed to make this unnecessary.

garyyu: ok, added.

garyyu:
If no more comments on this PR, i will merge it soon, since I'm still suffering on that full connections problem :)

Issue #2296, opened 2019-01-06
No grins after restore, strange accounts


phooton:
Happened when mining and sending grins via faucet. Faucet transactions started to fail, saw strangely grouped recent three coinbases and I restored without thinking too much about it.

```gringod@gringod:~/grin/target/release$ ./grin --floonet  wallet  restore
20190106 00:42:28.635 INFO grin_util::logger - log4rs is initialized, file level: Error, stdout level: Debug, min. level: Debug
20190106 00:42:28.635 INFO grin - Using wallet configuration file at /home/gringod/.grin/floo/grin-wallet.toml
20190106 00:42:28.635 INFO grin - This is Grin version 0.5.0 (git v0.5.0-7-g5e86cc8), built for x86_64-unknown-linux-gnu by rustc 1.33.0-nightly (09d6ab90e 2018-12-20).
20190106 00:42:28.635 DEBUG grin - Built with profile "release", features "".
Password:
20190106 00:42:30.523 DEBUG grin_wallet::types - Using wallet seed file at: /home/gringod/.grin/floo/wallet_data/wallet.seed
20190106 00:42:30.548 INFO grin_wallet - Using LMDB Backend for wallet
20190106 00:42:30.548 DEBUG grin_wallet::types - Using wallet seed file at: /home/gringod/.grin/floo/wallet_data/wallet.seed
20190106 00:42:30.572 INFO grin_wallet::libwallet::internal::restore - Starting restore.
20190106 00:42:30.830 INFO grin_wallet::libwallet::internal::restore - Retrieved 1000 outputs, up to index 35709. (Highest index: 3476)
20190106 00:42:30.831 INFO grin_wallet::libwallet::internal::restore - Scanning 1000 outputs in the current Grin utxo set
20190106 00:42:31.419 INFO grin_wallet::libwallet::internal::restore - Retrieved 1000 outputs, up to index 35709. (Highest index: 7003)
20190106 00:42:31.420 INFO grin_wallet::libwallet::internal::restore - Scanning 1000 outputs in the current Grin utxo set
20190106 00:42:31.974 INFO grin_wallet::libwallet::internal::restore - Retrieved 1000 outputs, up to index 35709. (Highest index: 10840)
20190106 00:42:31.975 INFO grin_wallet::libwallet::internal::restore - Scanning 1000 outputs in the current Grin utxo set
[...]

PR #2304, opened 2019-01-06
Update libsecp version and a few trivial enhancements

yeastplume:
Important update to rust-libsecp (to address #2296 and others), and a few small QOL changes to some output.


yeastplume:
Heh, indeed yes. Thought that was included, thanks for catching.

PR #2305, opened 2019-01-06
feat: Configuration option to include foreign API endpoints in the owner API

ducembarr:
This pull request adds an `owner_api_include_foreign` wallet configuration option to make endpoints from the [Foreign API](https://github.com/mimblewimble/grin/blob/master/doc/api/wallet_foreign_api.md) available over the same port as the [Owner API](https://github.com/mimblewimble/grin/blob/master/doc/api/wallet_owner_api.md).

A few notes on the actual change:
- This feature is useful when running the wallet in environments that make it difficult to access multiple ports on the same container.  In particular, Amazon's ECS and the various Amazon load balancers make it very difficult to have multiple ports open on the same container.
- The feature defaults to false.
- There shouldn't be any overlapping paths because the Foreign API endpoints start with `/v1/wallet/foreign` and the Owner API endpoints start with `/v1/wallet/owner/`.
- I do not believe there are any security concerns about setting the configuration option, because presumably anyone trusted to access the Owner API (which can perform arbitrary sends) is trusted to access the Foreign API.

I have added a basic integration test to make sure setting the configuration option on/off works as expected.  This involved making some changes to the test harness, so I've separated the changes for the tests into a separate commit in case those changes are problematic.

This is my first PR to this repository, so as usual please let me know if I'm missing anything major from the PR or description.  Thanks!

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -394,6 +394,15 @@ fn comments() -> HashMap<String, String> {
 		.to_string(),
 	);
 	retval.insert(
+		"owner_api_include_foreign".to_string(),
+		"
+#include the foreign API endpoints on the same port as the owner
+#API. Useful for networking environments like AWS ECS that make
+#it difficult to access multiple ports on a single service.
+"
+		.to_string(),
+	);
+	retval.insert(
 		"data_file_dir".to_string(),
 		"
 #where to find wallet files (seed, data, etc)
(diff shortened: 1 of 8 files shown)


yeastplume:
Looks good to me, and thanks for submitting this. Appreciate very much that you've also included tests and additions to documentation where necessary. Thanks!

PR #2312, opened 2019-01-08
Implement kernel and output features as enums

antiochp:
Longer term we may want to consider being able to maintain the various feature specific data (lock heights for a height locked kernel, fee for coinbase kernel etc.) on the enum itself. 
But to do this we would need to implement per enum type serialization and we would no longer be able to represent the enum as a simple u8.

This PR gets us 90% of the way there in terms of simplifying how we deal with features.

TODO - 
- [x] test serialization against `floonet` (nothing should change)
  - [x] fast sync
  - [x] broadcast (blocks and txs)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -771,11 +771,11 @@ dependencies = [
 name = "grin_core"
 version = "0.5.1"
 dependencies = [
- "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
  "blake2-rfc 0.2.18 (registry+https://github.com/rust-lang/crates.io-index)",
  "byteorder 1.2.7 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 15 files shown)


antiochp:
@tromp Rust enums matching _must_ be exhaustive (it won't compile otherwise). We don't need a fallback here because we have covered all the possibilities.

tromp:
Then what happens if the deserializer reads a value 3? Is that guaranteed to result in error?

antiochp:
> Then what happens if the deserializer reads a value 3? Is that guaranteed to result in error?

Yes we will catch this during deserialization here - 

https://github.com/mimblewimble/grin/pull/2312/files#diff-404c77cf79ebda67cdbd58c1c99daa8cR63

```
KernelFeatures::from_u8(reader.read_u8()?).ok_or(ser::Error::CorruptedData)?;
```

tromp:
So the KernelFeatures::from_u8 throws an error given a value of 3?
Then let's add a test for that and we're good...

PR #2315, opened 2019-01-08
Enforce zeroing of serialized proof excess bits

ignopeverell:


ignopeverell:
And you are right of course... fixing.

PR #2319, opened 2019-01-09
Automate wallet seed recovery process

yeastplume:
Automates the wallet recover process:

* If wallet.seed file exists, warn the user that it will be backed up if they continue
* back up the wallet.seed file as "wallet.seed.bak" "wallet.seed.bak.1" etc
* Create new wallet.seed from recovery

### DIFF
--- a/src/bin/cmd/wallet_args.rs
+++ b/src/bin/cmd/wallet_args.rs
@@ -74,6 +74,33 @@ fn prompt_password_confirm() -> String {
 	first
 }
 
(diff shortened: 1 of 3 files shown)


Onefox:
Is a grin wallet init still needed before the recovery?

yeastplume:
You may want to recover into an existing wallet simply because you forgot the password, however I think adding an ‘init from recovery phrase’ to the init command might be an option

PR #2320, opened 2019-01-09
Ban peers that can't complete a handshake

ignopeverell:
On my seed, all hanging connections that aren't represented by a peer seem to originate from a failed receiving handshake in the listener accept loop. This isn't covered by @garyyu suspender fix of tracking connections.

The Rust docs say that a `TcpStream` will be shutdown on `Drop`. Empirically this doesn't seem to be happening here. Perhaps it doesn't apply to listener-initated connections. Either way, this forced immediate shudown of error'd handshake connections.

In addition, some peers (I'm guessing old ones) try to initiate a connection over and over every 30 secs, always failing. We should ban them right away. I can't think of a reasonable reason to try a connection again 30 secs later from a peer that can't handshake.

### DIFF
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -86,9 +86,6 @@ impl Server {
 					if !self.check_banned(&stream) {
 						let sc = stream.try_clone();
 						if let Err(e) = self.handle_new_peer(stream) {
-							// in theory, should be shutdown when stream gets dropped,
-							// practically doesn't seem to be happening
-							stream.shutdown(Shutdown::Both);
 							warn!("Error accepting peer {}: {:?}", peer_addr.to_string(), e);
 						} else if let Ok(s) = sc {
 							connected_sockets.insert(peer_addr, s);
@@ -176,13 +173,21 @@ impl Server {
 		let total_diff = self.peers.total_difficulty();
 
 		// accept the peer and add it to the server map
-		let mut peer = Peer::accept(
+		let mut peer = match Peer::accept(
 			&mut stream,
 			self.capabilities,
 			total_diff,
 			&self.handshake,
 			self.peers.clone(),
-		)?;
+		) {
+			Ok(p) => p,
+			Err(e) => {
+				// in theory, should be shutdown when stream gets dropped,
+				// practically doesn't seem to be happening
+				let _ = stream.shutdown(Shutdown::Both);
+				return Err(e);
+			}
+		};
 		peer.start(stream);
 		self.peers.add_connected(Arc::new(peer))?;
 		Ok(())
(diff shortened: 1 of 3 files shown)


DavidBurkett:
I can think of one reason, at least on testnets: other node developers. Each testnet relaunch, I always failed 5 or 6 times at least before I could get the right params for handshakes to work. Its annoying to run out of peers who will connect to u when you're trying to test. Probably doesn't justify not banning, just throwing that out there.

ignopeverell:
Thought about it too, but remember that ban is temporary now (I think 2h?). So you'd have to exhaust all peers in 2h, which seems a bit excessive.

p2p/src/serv.rs
```diff
@@ -87,10 +87,9 @@ impl Server {
 						let sc = stream.try_clone();
 						if let Err(e) = self.handle_new_peer(stream) {
 							warn!("Error accepting peer {}: {:?}", peer_addr.to_string(), e);
-						} else {
-							if let Ok(s) = sc {
-								connected_sockets.insert(peer_addr, s);
-							}
+							let _ = self.peers.update_state(peer_addr, State::Banned);
```
garyyu: 👍 the idea to ban the handshake failed peer should be the killer for this annoying connection exhaustion problem!

but a quick question here, I think we don't add this peer address into the peers list and database if handshake fail, so the `update_state` here should be a `save_peer` call?

p2p/src/serv.rs
```diff
@@ -175,13 +174,21 @@ impl Server {
...
-		)?;
+		) {
+			Ok(p) => p,
+			Err(e) => {
+				// in theory, should be shutdown when stream gets dropped,
+				// practically doesn't seem to be happening
+				let _ = stream.shutdown(Shutdown::Both);
+				return Err(e);
+			}
+		};
```
garyyu: this seems not necessary, since we already have the error handling in function `accept` and did that `shutdown`.

ignopeverell: Looks like you're right, good catch! I'm still confused because after a while I end up with some peers having 20+ connections and not showing at all in my peer list. I'll try to track that down some more.

p2p/src/serv.rs
ignopeverell: You're right, I thought `update_state` would still save the peer address with the state (as other other pieces of data are unknown) but looks like it'll fail instead. Will fix.

garyyu:
I have thought this one has been merged into 0.5.2 :) but actually not.
Is there any remaining problem in this PR?  It can be merged and tested.

ignopeverell:
Been running this on my seed for a day or two. Behaves well, although it doesn't entirely alleviate the connection refusal issue it's an improvement.

p2p/src/serv.rs
yankai-wang: .

PR #2323, opened 2019-01-09
Cleanup compiler warnings

sesam:


core/tests/common.rs
```diff
@@ -14,17 +14,16 @@
...
 	block::{Block, BlockHeader},
 	Transaction,
 };
-use crate::core::libtx::{
+use grin_core::libtx::{
 	build::{self, input, output, with_fee},
 	reward,
 };
-use crate::core::pow::Difficulty;
-use crate::keychain::{Identifier, Keychain};
```
sesam: (rustfmt moved this line 3 lines up)

servers/tests/framework.rs
```diff
@@ -127,6 +125,7 @@ impl Default for LocalServerContainerConfig {
 /// A top-level container to hold everything that might be running
 /// on a server, i.e. server, wallet in send or receive mode
 
+#[allow(dead_code)]
```
sesam: allow(dead_code) could theoretically be added up top with ! to avoid repetition, but that kind of defeats it purpose completely. The allow markings are actually not dead. Possibly there's another way by marking `pub` also on the module, but I didn't figure that out. If anyone knows or has an idea, say now or stay silent until some later date.

I'll be happy to clean this up further also later on. My first goal now is to have Travis build and test log outputs get a cleaner look.

Issue #1139, opened 2018-06-05
Identify and ban total proof of work cheats


ignopeverell:
During header sync, we trust the total difficulty given by peers during the handshake to select who to get headers from. This is easy to lie about. We need to identify when a node fails to send us headers to match that total difficulty and ban them.

ignopeverell:
Thinking some more about this and validating my current understanding by going through the code, we're a lot less vulnerable than I thought to a "fake difficulty" attack. The reasons are quite simple:

1. A fake high difficulty will set the node back in sync mode. But this doesn't operate regular gossip operations, we would still process blocks on the currently best valid chain.
2. Only the `sync_head` get directly impacted by a sync reset. As long as the total proof of work on the `sync_head` is less than what's the on the `header_head`, there is no impact on the more authoritative chain records: `header_head` and `body_head`.

With that in mind, the following would happen if a fake high difficulty peer showed up:
1. Nodes would reset their `sync_head` and be ready to follow new block headers being sent to them.
2. Nodes would send locator requests. Note that this is rather cheap as they're now cached.
3. From there 2 possibilities:
    * We start receiving some headers and add them to the `sync_head`. This is a fairly cheap operation, especially given that those headers still have to have a valid PoW against current difficulty.
    * Denial of service angle: the fake peer doesn't send us anything. This could happen after the previous point occurred as well. This would cause the node to repeat 2 every 10 seconds until the fake peer either disappears to actually provides a most worked chain.

It seems the first thing that'll happen overall is that all peers connected to the cheating peer will start polling it, with little effort from honest peers and that's about it. Note that the each locator request isn't very large.

The main impact here is that this could prevent honest peers from triggering what could be a real sync because of the masking of the cheating peer. But then the faking would have to be going on for a while with unclear benefits. It may still be worth detecting very long non-resolving header syncs and banning the peer that initiated it.

sesam:
Maybe @garyyu @antiochp already have covered this recently? What is still to be done for this issue to be closeable?

antiochp:
@garyyu Please correct me if I'm wrong but my understanding of the impact here is -
 
1) new nodes try and sync against the "fraud peer" and fail to sync successfully against any alternative honest peers (we sync headers from the "most work" peer).
2) all mining nodes stop broadcasting new blocks due to a mistaken belief that they are out of sync and their new block is "stale"

garyyu:
@antiochp 
The two impacts above listed are correct.
[...]

PR #2324, opened 2019-01-10
new mined block should be broadcasted if a node was synced

garyyu:
Re-pickup #2240 

When a well synced node is mining, the new mined block(s) need broadcast, no matter what pulled this node into a sync state again, possibly because of a connected fraud peer with a fake difficulty and height, and not banned yet.

This will protect the grin network from such kind of attacking.

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -635,10 +635,6 @@ impl ChainAdapter for ChainToPoolAndNetAdapter {
 			}
 		}
 
-		if !self.sync_state.is_synced_once() {
-			return;
-		}
-
 		// If we mined the block then we want to broadcast the compact block.
 		// If we received the block from another node then broadcast "header first"
 		// to minimize network traffic.
@@ -647,6 +643,9 @@ impl ChainAdapter for ChainToPoolAndNetAdapter {
 			let cb: CompactBlock = b.clone().into();
 			self.peers().broadcast_compact_block(&cb);
 		} else {
+			if !self.sync_state.is_synced_once() {
+				return;
+			}
 			// "header first" propagation if we are not the originator of this block
 			self.peers().broadcast_header(&b.header);
 		}
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -635,7 +635,7 @@ impl ChainAdapter for ChainToPoolAndNetAdapter {
 			}
 		}
 
-		if self.sync_state.is_first_syncing() {
+		if !self.sync_state.is_synced_once() {
 			return;
 		}
 
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -635,7 +635,7 @@ impl ChainAdapter for ChainToPoolAndNetAdapter {
 			}
 		}
 
-		if self.sync_state.is_syncing() {
+		if self.sync_state.is_first_syncing() {
 			return;
 		}
 
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -643,11 +643,10 @@ impl ChainAdapter for ChainToPoolAndNetAdapter {
 			let cb: CompactBlock = b.clone().into();
 			self.peers().broadcast_compact_block(&cb);
 		} else {
-			if !self.sync_state.is_synced_once() {
-				return;
+			if self.sync_state.is_synced_once() {
+				// "header first" propagation if we are not the originator of this block
+				self.peers().broadcast_header(&b.header);
 			}
-			// "header first" propagation if we are not the originator of this block
-			self.peers().broadcast_header(&b.header);
 		}
 
 		// Reconcile the txpool against the new block *after* we have broadcast it too our peers.
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -282,6 +282,8 @@ pub enum SyncStatus {
 /// Current sync state. Encapsulates the current SyncStatus.
 pub struct SyncState {
 	current: RwLock<SyncStatus>,
+	/// Flag to indicate whether we already synced once
+	synced_once: RwLock<bool>,
 	sync_error: Arc<RwLock<Option<Error>>>,
 }
 
@@ -290,6 +292,7 @@ impl SyncState {
 	pub fn new() -> SyncState {
 		SyncState {
 			current: RwLock::new(SyncStatus::Initial),
+			synced_once: RwLock::new(false),
 			sync_error: Arc::new(RwLock::new(None)),
 		}
 	}
@@ -298,6 +301,11 @@ impl SyncState {
 	/// Note: This includes our "initial" state.
 	pub fn is_syncing(&self) -> bool {
 		*self.current.read() != SyncStatus::NoSync
+	}
+
+	/// Same as 'is_syncing' except the flag 'synced_once' is true
+	pub fn is_first_syncing(&self) -> bool {
+		*self.current.read() != SyncStatus::NoSync && *self.synced_once.read() == false
 	}
 
 	/// Current syncing status
@@ -312,10 +320,13 @@ impl SyncState {
 		}
 
 		let mut status = self.current.write();
+		*status = new_status;
+
+		if let SyncStatus::NoSync = new_status {
+			*self.synced_once.write() = true;
+		}
 
 		debug!("sync_state: sync_status: {:?} -> {:?}", *status, new_status,);
-
-		*status = new_status;
 	}
 
 	/// Update txhashset downloading progress
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -303,9 +303,9 @@ impl SyncState {
 		*self.current.read() != SyncStatus::NoSync
 	}
 
-	/// Same as 'is_syncing' except the flag 'synced_once' is true
-	pub fn is_first_syncing(&self) -> bool {
-		*self.current.read() != SyncStatus::NoSync && *self.synced_once.read() == false
+	/// Whether it was synced once after grin running
+	pub fn is_synced_once(&self) -> bool {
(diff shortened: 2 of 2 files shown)


antiochp:
Is there really a difference between the first sync state and subsequent sync states?

I played with doing something similar a while back but came to the conclusion that this just complicates the state handling.

Scenario 1 -
* A mining node is restarted after having been shutdown overnight.
* Needs to sync to catch up with the network.

Scenario 2 - 
* A mining node suddenly sees a peer advertise more work.
* Needs to sync to catch up with the network.

These scenarios should both be handled the same way I think, particularly the logic in `adapters.rs`, particularly the txpool reconciliation etc.
With these changes a mining node will effectively not know it is syncing against the network durig block processing unless it has just been restarted.

antiochp:
> prefer to ignore this case to make the thing simple. otherwise, we need record the last time and status into the database.

I suspect this may be a very common case, at least initially.
There are going to be lots of people trying Grin out early on and I would expect many of them to only run it intermittently (some hobby miners included).

> we need record the last time and status into the database.
I don't follow - I'm pushing for _less_ state, not more state.

In the adapter code for accepting a new block we already have - 
* syncing vs. not_syncing
* mining vs. regular node
* next vs. fork vs. reorg (block status)

My gut feeling here is adding another state risks adding significant complexity here in ways we do not yet have a good handle on.
* first time sync vs. regular sync vs. not_syncing

Higher level question - Why does a mining node care if it is syncing or not? Why are we not just building blocks against whatever chain state we currently know about?

Is the question then about if/when to broadcast these potentially stale blocks?
And if so - why not just broadcast the compact block regardless and leave it up to our peers to handle the block or reject it.

A miner can see a peer advertise more work and transition to sync mode in two different scenarios - 
* on startup
* or while already running

In either case I think we should handle it the same way.

> When a well synced node is mining, the new mined block(s) need broadcast, no matter what pulled this node into a sync state again, possibly because of a connected fraud peer with a fake difficulty and height, and not banned yet.

I'm arguing a node cannot be both `well synced` and `in sync state` simultaneously - its either syncing or not syncing. We don't know (or care) why its syncing other than a peer advertises more work. And for a period of time (during header sync) we don't know if this is legitimate or fraudulent work.

garyyu:
>I suspect this may be a very common case, at least initially.
There are going to be lots of people trying Grin out early on and I would expect many of them to only run it intermittently (some hobby miners included).

I suppose there's no problem for this, people just wait the initial sync when restart. Same as now.


>Higher level question - Why does a mining node care if it is syncing or not? Why are we not just building blocks against whatever chain state we currently know about?
>Is the question then about if/when to broadcast these potentially stale blocks?
And if so - why not just broadcast the compact block regardless and leave it up to our peers to handle the block or reject it.

Come back to this PR, if we can simply avoid the case to broadcast stale blocks for an initial running node which is not synced yet, why not? as you see, just one line of code in adapter.rs.

>I'm arguing a node cannot be both well synced and in sync state simultaneously - its either syncing or not syncing. We don't know (or care) why its syncing other than a peer advertises more work. And for a period of time (during header sync) we don't know if this is legitimate or fraudulent work.

Again, I remind that this PR is used to handle the case of https://github.com/mimblewimble/grin/issues/1139#issuecomment-451665099, please let's use one PR to handle one fix, much simple and easy to discuss :)

The idea is to make the fraud peers have no real impact on the grin nodes, the mining nodes can continue mining and broadcasting, even it was forced (by fraud peers) into sync state.

Without this PR, an attacker can easily find a way to attack Grin network by fake work.

antiochp:
Let me restate the problem to make sure I understand it correctly and to make sure we are on the same page. Please correct me if I am misunderstanding the problem.

1. A mining node will currently suspend mining and will not broadcast blocks if the node is in "syncing" state.
1. A node is in "syncing" state if it sees a peer advertise more work, without verifying this to be correct.
1. A node can trivially put _all_ connected peers into "syncing" state by sending out a hand/shake or ping/pong msg with a large total_difficulty value in it.


Now let me summarize your proposed fix to make sure I understand it correctly.

1. Prior to a node syncing the first time, do not broadcast any blocks out, regardless of mining or not.
1. After a node has sync'd once, always broadcast all blocks, regardless of mining or syncing state.

----

The proposed fix changes the behavior such that a node will broadcast blocks during a subsequent sync state, but it does not differentiate between blocks mined by the node and blocks it received via broadcast.
The proposed fix also changes this behavior for _all_ nodes, not limited to just mining nodes.

We only want to do this for mining nodes and only for blocks mined by that node.

_Any_ node finding itself lagging behind the network _for any reason_ will transition to "syncing" to catch up.  But will now broadcast out _all_ blocks during sync.

For example, a node on a laptop waking after being closed overnight will now broadcast out to all its peers all blocks that it missed as it re-syncs.

So to summarize - 

* The proposal fixes the "issue".
* But it introduces undesirable behavior unrelated to the issue itself.
  * Non-mining nodes are affected (this should only affect mining nodes).
  * Blocks received via broadcast are affected (this should only affect blocks mined by the miner).
* And does this at the cost of introducing additional complexity.

----

> Come back to this PR, if we can simply avoid the case to broadcast stale blocks for an initial running node which is not synced yet, why not? as you see, just one line of code in adapter.rs.

The concern around complexity is not in terms of LoC. It is the growing number of possible states the system can be in at any given time and the multiple state permutations we need to be aware of.

We are not proposing to add an additional SyncState here, but adding a boolean flag _on top_ of the existing set of sync states. This adds significant complexity across the system. For example the stratum server itself is not aware of this new introduced state flag.

We also now have two separate fields within SyncState (current state and synced_once) wrapped in individual RwLock instances. We can no longer guarantee we are updating the sync state atomically and we potentially have new and exciting edge cases here to handle.

----

> Again, I remind that this PR is used to handle the case of #1139 (comment), please let's use one PR to handle one fix, much simple and easy to discuss :)

Strongly disagree here. We need to approach this in terms of fixing the underlying issue. We need to minimize any impact on unrelated parts of the system. We need to do this while minimizing complexity across the system as a whole. We cannot do this if we try to handle each fix individually, without considering the bigger picture.

ignopeverell:
> We only want to do this for mining nodes and only for blocks mined by that node.

Agreed. And we can't tell whether a node is mining or not. So it means we can't do this. In my original suggestion to differentiate initial syncing state from subsequent syncs, I mean initial syncing _from zero_. AKA IBD. Not any start.

> We cannot do this if we try to handle each fix individually, without considering the bigger picture.

Also agreed. But I don't think Gary meant ignoring the bigger picture either.

The more I think about this, the more I think we shouldn't do anything special while a node switches to sync node, other than trying to sync. Perhaps this check made some sense on testnet(s) but a lot less so now. So I think the simpler, minimal thing that removes the possible exploit is just to keep mining and keep relaying no matter what.

garyyu:
Updated for most simple logic - "keep mining and keep relaying no matter what" :)

But keeping the `is_synced_once` state to avoid :
```
"header first" propagation if we are not the originator of this block
```
for the case of node restart and new node syncing. Otherwise, there will be too much old headers broadcasting from the syncing.   

Does this make sense?  @antiochp  @ignopeverell

servers/src/common/adapters.rs
```diff
@@ -647,8 +643,10 @@ impl ChainAdapter for ChainToPoolAndNetAdapter {
 			let cb: CompactBlock = b.clone().into();
 			self.peers().broadcast_compact_block(&cb);
 		} else {
-			// "header first" propagation if we are not the originator of this block
-			self.peers().broadcast_header(&b.header);
+			if self.sync_state.is_synced_once() {
```
antiochp: Still not convinced the added complexity of `is_synced_once` is worth it. There is no difference between the 1st and 2nd sync for a node.
Can we not just use the existing `is_syncing()` here?

garyyu: If using existing `is_syncing`, the fraudulent most work peer will make the gossip block headers propagation pause (because node is in syncing state), until the node leave syncing mode by a fraud detection and banning.

sesam:
:+1: always broadcast new blocks whenever we don't know yet if it is valid, for now.

Later and if needed we could avoid

1. broadcasting far-future blocks ignore any block that looks like it is increasing hashing power by a magnitude.
2. wait with broadcasting near-future blocks, instead first syncing up and validating, and only then broadcast

I can create 1. and 2. as issues if wanted.

servers/src/common/adapters.rs
antiochp: 🤦‍♂️ Of course, forgot about that.

We should just _always_ broadcast out "header first" here, regardless of syncing or not, regardless of synced_first or not. 
Keep it simple and let peers handle these redundant headers as necessary.

antiochp:
@sesam - neither of these are applicably here. We only ever broadcast _accepted_ blocks, i.e. we added them successfully to our chain locally. So nothing "far future" or "near future" will ever make it this far.

servers/src/common/adapters.rs
garyyu: I worry there will be too much redundant headers then, and not sure whether the current grin version can handle these flood redundant headers, we don't have much time to test it since only less 4 days before launching.

Also here's a solution to kill that fraudulent work case, I don't understand why simple design is more important than defending from attack.

BTW, does this `is_synced_once` indeed make design become a very complex one?

garyyu: Look like we couldn't agree each other between us about whether `always` broadcast out "header first" including initial syncing stage, perhaps it's good to invite more pair of eyes on this review 😄   @ignopeverell  what's your opinion on this?

ignopeverell:
Reviewing the code once more, I don't think we ever want to stop block propagation based on whether we're syncing or not. So all these checked should be removed.

What we *do* want to block is the propagation of a header or block we received through sync. And that's an information we have in `block_accepted` with the chain::Options argument.

garyyu:
Use new #2349 to implement this improvement, and obsolete this one.

PR #2325, opened 2019-01-10
Replace default fmt::Display on core::Hash

### LINKED ISSUE #2314: [TUI] block hashes in TUI will need to be larger than first 8 chars for mainnet
> John Tromp @tromp 15:30
btw, the TUI only shows 8 hex digits of hash

> jaspervdm @jaspervdm 15:31
or 5 even

> John Tromp @tromp 15:31
that won't cut it on mainnet, where most will be 0

eupn:
Resolves #2314 

Visually:
image

?r @antiochp

### DIFF
--- a/core/src/core/hash.rs
+++ b/core/src/core/hash.rs
@@ -54,11 +54,7 @@ impl fmt::Display for Hash {
 		let starting = &hash_hex[..NUM_SHOW];
 		let ending = &hash_hex[(len - NUM_SHOW)..];
 
-		write!(f,
-			"{}...{}",
-			starting,
-			ending
-		)
+		write!(f, "{}...{}", starting, ending)
 	}
 }
 
--- a/core/src/core/hash.rs
+++ b/core/src/core/hash.rs
@@ -47,7 +47,18 @@ impl fmt::Debug for Hash {
 
 impl fmt::Display for Hash {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-		fmt::Debug::fmt(self, f)
+		let hash_hex = self.to_hex();
+		let len = hash_hex.len();
+		const NUM_SHOW: usize = 8;
+
+		let starting = &hash_hex[..NUM_SHOW];
+		let ending = &hash_hex[(len - NUM_SHOW)..];
+
+		write!(f,
+			"{}...{}",
+			starting,
+			ending
+		)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


src/bin/tui/mining.rs
```diff
@@ -137,7 +137,18 @@ impl TableViewItem<DiffColumn> for DiffBlock {
...
+			DiffColumn::Hash => {
+				// Prints part of the hash's hex
+				let hash_hex = self.block_hash.to_hex();
+				let len = hash_hex.len();
+				const NUM_SHOW: usize = 8; // Number of characters to show
+
+				format!(
+					"{}...{}",
+					&hash_hex[..NUM_SHOW / 2],
+					&hash_hex[(len - NUM_SHOW)..]
```
garyyu:
```suggestion
					&hash_hex[(len - NUM_SHOW / 2)..]
```

antiochp:
Actually I suspect this goes beyond the tui and we should scope this to include the api and the logs and everywhere else we see a block hash displayed.

eupn:
So basically I'll add a method `.show()` to `core::Hash`. 

Right now this is a `.to_hex()` which strips the middle of a hex representation of the hash. 
Is showing hash [like this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9a6094424f2d05c8a2bd7baa202795a3) is good enough?

core/src/core/hash.rs
```diff
@@ -47,7 +47,10 @@ impl fmt::Debug for Hash {
 
 impl fmt::Display for Hash {
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-		fmt::Debug::fmt(self, f)
```
antiochp: Want to do this in `Debug::fmt()` as well in case we're logging it this way anywhere?

eupn: Should we actually log a full hash hex string since we're not limited in number of characters on screen in this case?

antiochp: I suspect logging the full hex string will make the logs significantly harder to read as they will start wrapping lines.

eupn: Understood.

garyyu:
Just a note here, we didn't realize this PR will change the txhashset name, and it caused the old version (all before 11th Jan. 2019, including 0.5.1) in-compatible with 0.5.2 (and after).

In old version, the file name is like:
```
pmmr_leaf.bin.04199c70
```
In new version, the file name becomes:
```
pmmr_leaf.bin.04199c70d1ad...
```

Do remind users to upgrade nodes to latest version if anyone report this problem.  Or try restarting multiple times to download state from a new version node.

There's no problem in mainnet, since all nodes will use new version.

DavidBurkett:
The filename ends in `...`?? This is terrible. Windows doesn't support filenames ending with a `.`. Can we please revert this change?

PR #2326, opened 2019-01-10
wait_for_min_peers should continue waiting if none connected

garyyu:
It doesn't make sense to allow grin enter `running` state with `0` connected peers, better to let it wait forever here to connect peers.


garyyu:
Yes, Travis-CI test not happy on this and failed  :)
OK to close this.

ignopeverell:
You could have kept it but testing for floonet and mainnet?

PR #2329, opened 2019-01-10
Fix doc dump_stored_tx -> retrieve_stored_tx

icook:
It appears the method call in https://github.com/mimblewimble/grin/blob/master/wallet/src/controller.rs got renamed for consistency and this got missed.


quentinlesceller:
Thanks for the PR @icook!

PR #2333, opened 2019-01-10
Load config file from CLI flag

### LINKED ISSUE #2317: Grin server does not read config file when passing `--config_file` flag
When starting grin with: `./target/debug/grin --floonet server --config_file /tmp/grin-server.toml run` the config file pointed to with the `--config_file` flag is ignored. Instead the on in `~/.grin/floonet/grin-server.toml` is read.

Curiously when changing directory to `/tmp` and re-running the command the file *is* correctly read.

## versions

Rust: 1.31.1
Grin version: 27801f6a93f8a4bf5defc9eacd86b6bfb1464923
OS: Ubuntu 18.10

haarts:
Fixes #2317 in the least intrusive way. I believe that given a slightly more
rigorous refactor could improve the readability quite a bit.

Currently the structure of the code is such that either `wallet_config` OR `node_config` is read. Then either the wallet is kicked off OR the server. In pseudo code:

```
wallet_config = None
node_config = None

// load config
match wallet {
  wallet_config = Some
}
match server {
  node_config = Some
}

// start service
match wallet {
  start_wallet
}
match server {
  start_server
}
```

I believe having only one match would do and extracting some functionality into helper functions (like configuration loading, logging setup).

I also feel slightly uneasy about the `initial_setup_server` function accepting a chain type argument. But that is just a feeling ATM.

### DIFF
--- a/src/bin/grin.rs
+++ b/src/bin/grin.rs
@@ -114,6 +114,7 @@ fn real_main() -> i32 {
 		_ => {}
 	}
 
+	// Load relevant config
 	match args.subcommand() {
 		// If it's a wallet command, try and load a wallet config file
 		("wallet", Some(wallet_args)) => {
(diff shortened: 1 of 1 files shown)


ignopeverell:
Did you have a chance to try all the possible configurations locally?

haarts:
I did. Not really a chance though. I was intentional. ;)

sesam:
nice. Looks like a good PR to include in a shortlived floonet-prim testnet.

For testing, did you have to add the --floonet flag just to get past the floonet check?
And any change to make --floonet more flexible would probably have to be a separate PR, in case Igno needs to rebase "floonet2-pre-mainnet" ahead of the 15th

haarts:
I indeed added the `--floonet` flag to get passed that.
I believe the `--{main,use,floo}net` flag should not be treated any different from the other flags. Perhaps the config/flag setup should get a more comprehensive shake up.

PR #2340, opened 2019-01-11
fix: wait peers stage should not wait when all peers are same work

garyyu:
1. In case of mainnet launching practice (`Fake mainnet` test), even set `skip_sync_wait=true`, when the seed node only have genesis block, all connected nodes will wait forever on `waiting more peers...`, and nobody can enter `running` state to mine 1st block. This can be thought as a regression of PR https://github.com/mimblewimble/grin/pull/2326.

2. Actually, the original logic here doesn't make sense, say a well synced grin node is restarted by admin, before this PR the node need wait until get enough `more_work` peers, but actually all connected peers could be in same work same height at this time, then the new restarted node have to wait till next block time (or 30s timeout) to pass this `waiting peers` stage, that's obviously not necessary.

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -183,6 +183,22 @@ impl Peers {
 		max_peers
 	}
 
+	// Return number of connected peers that currently advertise more/same work
+	// (total_difficulty) than/as we do.
+	pub fn more_or_same_work_peers(&self) -> usize {
+		let peers = self.connected_peers();
+		if peers.len() == 0 {
+			return 0;
+		}
+
+		let total_difficulty = self.total_difficulty();
+
+		peers
+			.iter()
+			.filter(|x| x.info.total_difficulty() >= total_difficulty)
+			.count()
+	}
+
 	/// Returns single random peer with more work than us.
 	pub fn more_work_peer(&self) -> Option<Arc<Peer>> {
 		self.more_work_peers().pop()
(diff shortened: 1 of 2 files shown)


antiochp:
Can we make sure any sync related PRs are reviewed by at least one other dev going forward?

garyyu:
Review it please @antiochp, I waited 12 hours and thought no body have time at this moment :-)
And if nobody come to review it for an urgent bug fix, I prefer to merge it so at least having chance to test on floonet before mainnet launching.

> Can we make sure any sync related PRs are reviewed by at least one other dev going forward?

***Why*** your coment only applied `any sync related PRs` here? I'm a little bit curious. Should not be any PR? or at least any non-trivial PR?  @antiochp

antiochp:
Can we make sure any ~sync related~ PRs are reviewed by at least one other dev going forward?

The point was that PRs should be reviewed and this is sync related, which is a critical and complex part of the system, not that we should _only_ review sync related PRs. But I think you knew this.

lehnberg:
https://github.com/mimblewimble/grin-pm/blob/master/notes/20181127-meeting-development.md#6-pull-request-review-process

> Agreed new process:
> 1. Trivial PRs can be directly merged
> 1. Non-trivial PRs need at least one review
> 1. Identified consensus-touching PRs need 2 reviews
> 1. Igno stops committing straight in the repo, even once in a while

garyyu:
Easy to revert and wait for review on this PR.

ignopeverell:
The PR in itself is fine. But we have too many subtle issues and regressions right now (like that `...` one) to keep going without more scrutiny.

PR #2347, opened 2019-01-11
Refuse duplicate peer IPs

ignopeverell:
This builds on @garyyu #2287 but takes a stronger stance by entirely refusing to accept 2 peers with the same IP, rather than just cleaning up unknown ones. It also simplifies the handling of duplicate connections quite a bit.

The main objection against this would be people who are behind NAT and share the same public IPs. My objections to this objection are the following:

1. There should be plenty of peers out there. If no different ones can be found, maybe that should be the problem.
2. This change increases network distribution and security. For example, a peer that has 4 connections to 4 peers behind the same NAT is less secure than one that's connected to all distinct IPs.

### DIFF
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -187,7 +187,7 @@ impl Server {
 	/// expired yet.
 	/// 2. We're already connected to a peer at the same IP. While there are
 	/// many reasons multiple peers can legitimately share identical IP
-	/// addresses, network distribution is improved if they choose different
+	/// addresses (NAT), network distribution is improved if they choose
 	/// different sets of peers themselves. In addition, it prevent potential
 	/// duplicate connections, malicious or not.
 	fn check_undesirable(&self, stream: &TcpStream) -> bool {
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -29,7 +29,8 @@ use crate::peer::Peer;
 use crate::peers::Peers;
 use crate::store::PeerStore;
 use crate::types::{
-	Capabilities, ChainAdapter, Error, NetAdapter, P2PConfig, ReasonForBan, TxHashSetRead,
+	Capabilities, ChainAdapter, Error, NetAdapter, P2PConfig, ReasonForBan,
+	TxHashSetRead, Seeding,
 };
 use crate::util::{Mutex, StopState};
 use chrono::prelude::{DateTime, Utc};
@@ -194,7 +195,8 @@ impl Server {
 		// peer has been banned, go away!
 		if let Ok(peer_addr) = stream.peer_addr() {
 			let banned = self.peers.is_banned(peer_addr);
-			let known_ip = self.peers.is_known_ip(&peer_addr);
+			let known_ip = self.peers.is_known_ip(&peer_addr)
+				&& self.config.seeding_type == Seeding::DNSSeed;
 			if banned || known_ip {
 				debug!("Peer {} banned or known, refusing connection.", peer_addr);
 				if let Err(e) = stream.shutdown(Shutdown::Both) {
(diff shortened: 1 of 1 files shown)


mably:
May be you could add your explanation about NAT adresses as a comment in the code?

p2p/src/serv.rs
```diff
@@ -188,11 +181,23 @@ impl Server {
...
+	/// different sets of peers themselves. In addition, it prevent potential
+	/// duplicate connections, malicious or not.
+	fn check_undesirable(&self, stream: &TcpStream) -> bool {
 		// peer has been banned, go away!
 		if let Ok(peer_addr) = stream.peer_addr() {
-			if self.peers.is_banned(peer_addr) {
-				debug!("Peer {} banned, refusing connection.", peer_addr);
+			let banned = self.peers.is_banned(peer_addr);
+			let known_ip =
+				self.peers.is_known_ip(&peer_addr) && self.config.seeding_type == Seeding::DNSSeed;
```
garyyu: Perhaps call a `global::is_production_mode()` here for `known_ip` checking.

antiochp:
I wonder if this is causing some unexpected issues.

Dump of a conversation on gitter earlier -
 
> umma08 @umma08 13:36
ah i understand better now. thanks
how many connections are other people's nodes getting? mine seems to always settle on four
Antioch Peverell @antiochp 13:44
there are considerably more nodes that are not accessible (behind NAT, non static IP, whatever) than publicly available
I'm still not sure exactly why we're not seeing healthier numbers of peers but this is definitely related
Peter Hilton @mi1b 13:44
I only see 4 by default
u can remove comment peer_min_preferred_count = 8
and get more
in grin-server.toml
Antioch Peverell @antiochp 13:45
I wonder if we're being overly aggressive cleaning up stale peer connections or closing existing connections if other peers connect, even temporarily
David Burkett @DavidBurkett 13:46
@antiochp There’s definitely a code problem. Grin++ consistently sees more peers than my Grin node
Grin++ is much quicker to close stale connections though, so I don’t believe that’s the problem. Idk though
Antioch Peverell @antiochp 13:47
yeah I don't really know
there does seem to be too much "churn" and turnover of connections without a node settling down with a decent number of long lived ones
David Burkett @DavidBurkett 13:52
I agree, but I’m not entirely sure why. We also tend to hold onto dead peers for far longer than we should
I’ve seen peers that haven’t ponged my pings (technical phrase) in quite some time that Grin still considers “active"
Peter Hilton @mi1b 14:24
I've notice that behind NAT with UPnP enabled peers always drop off to 4 - port 3414 needs to be forwarded to get more
Antioch Peverell @antiochp 14:26
hmm - I wonder if we have some code path that tries to connect to peers, regardless of if we're currently connected, fails and ends up cleaning up the connection (by ip, ignoring the port)
there was some talk of that recently (looking at ip and ignoring the port)
that might explain us dropping outbound connections (other end closes them)
Peter Hilton @mi1b 14:30
sounds like a good theory - even with port 3414 forwarded and peer_min_preferred_count = 25 I only get around 20 peers (only 4 or 5 outbound)
David Burkett @DavidBurkett 14:36
Ahh possibly. After closer inspection, Grin++ only has more connections because it’s faster at connecting to peers. In reality, there appears to be just as much churn. Your theory could explain that, @antiochp.
amazarache @amazarache 14:43
I got even more than 30 peers with 12 outbound connections with port forwarded
for short time 40 peers
Peter Hilton @mi1b 14:46
I get lots of connections when the node first starts..
but then they drop off fairly sharpish

Is it possible that something like the following is happening - 
* node connects "outbound" to a peer
* node also receives an "inbound" connection from that peer
* causing it to drop the original connection as "undesirable"?

Or something along those lines?
It feels like we do not keep connections alive under certain circumstances and trying to investigate the root of the problem.

Say our node is behind a NAT. We can connect outbound to other nodes, but if they attempt to connect to us they fail and drop the connection attempt _and any existing_ outbound connections from us?

PR #2348, opened 2019-01-11
Recover outputs from multiple wallets using same seed

yeastplume:
This covers cases where multiple wallets accidentally use the same seed without knowledge of each other. 
Wallets generate their BIP32 key ids sequentially, with a new wallet starting from 0/0/0, and increment each new output 0/0/1, 0/0/2 etc. If a new wallet is created with the same seed that has no knowledge of the last index, it will start reusing the same BIP32 ids from 0 again.

This can happen when someone recovers their seed without restoring, generates some outputs and then restores. Or when someone accidentally or intentionally use the same seed for 2 different sets of wallet data. This PR ensures that the `wallet check` and `wallet restore` commands can always handle these cases, restoring all outputs belonging to a given seed to a wallet no matter what identifiers are stored in the UTXO set.

To support this:

* A u64 `mmr_index` is added to all outputs returned from the api, and the data stored in the wallet is extended to add an Option<u64> `mmr_index` (which should ensure it's backwards compatible)
* When a wallet output is created locally and added to the wallet database, the `mmr_index` is None, and the key is simply the id.
* When a wallet output is restored from the UTXO set via the `restore` or `check` processes, the `mmr_index` is Some and the mmr_index is appended to the key.
* During the `restore` and `check` processes, outputs are matched by commit instead of ID, meaning they're uniquely identified even when they have duplicate IDs and regardless of whether their pmmr index is stored locally. 

Multiple wallets using the same seed can still potentially create duplicate commits if they use the same ids and amounts, but those cases are handled at the protocol level (and there isn't much we can do to guard against that). This could still be allowed if the duplicate has already been spent, and there's even less we can do about that.

Also includes a few tests that try to replicate the conditions under which UTXOs can have the same IDs.

Update: I've also removed commits from wallet storage, as it was just leading to unnecessary complexity. I can't think why they still need to be stored when they can be reproduced easily from the amount and identifier, (I think previously the key was being derived every time it was being used, hence the need for a stored commit to compare against). Not storing them directly is surely a privacy plus given that you need your seed/password to reproduce them?

### DIFF
--- a/api/src/handlers/transactions_api.rs
+++ b/api/src/handlers/transactions_api.rs
@@ -94,7 +94,8 @@ impl TxHashSetHandler {
 			id
 		)))?;
 		let commit = Commitment::from_vec(c);
-		let output_pos = w(&self.chain).get_output_pos(&commit)
+		let output_pos = w(&self.chain)
+			.get_output_pos(&commit)
 			.context(ErrorKind::NotFound)?;
 		let merkle_proof = chain::Chain::get_merkle_proof_for_pos(&w(&self.chain), commit)
 			.map_err(|_| ErrorKind::NotFound)?;
--- a/api/src/handlers/transactions_api.rs
+++ b/api/src/handlers/transactions_api.rs
@@ -94,6 +94,8 @@ impl TxHashSetHandler {
 			id
 		)))?;
 		let commit = Commitment::from_vec(c);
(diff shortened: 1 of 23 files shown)


yeastplume:
Ready for review if anyone could take a look, as per our new review guidelines, won't merge this until after review.

PR #2353, opened 2019-01-12
Revert 2340 wait for min peers

garyyu:
Regarding to comments in #2340 , let's revert it and wait for review.

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -199,22 +199,6 @@ impl Peers {
 		max_peers
 	}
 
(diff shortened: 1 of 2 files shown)


garyyu:
@antiochp  @lehnberg  Please feel free to review and merge this revert, and let's wait for review of #2340.

ignopeverell:
#2340 in itself is fine, and it's already in the source tree now. No point in reviewing the revert to re-apply. But let's just make sure we still wait for proper review on future changes.

Issue #2351, opened 2019-01-12
txhashet filenames currently ending in ...


antiochp:
Related - #2325 

Looks like we inadvertently included the nice pretty `...` in the block hash string itself... (and not just in the UI).

* displayed/used block hash should be 12 digits long
* UI is free to add `...` to the end, but this should not be used in the filename (obviously)

Lets be more careful with seemingly innocuous changes like this in the future. We should have considered any and _all_ usage of block.hash() here and clearly missed one, and probably other uses.

garyyu:
Closed by #2355

PR #2355, opened 2019-01-12
Fix/serializing hash

svechinsky:
Just fixing https://github.com/mimblewimble/grin/issues/2351

### DIFF
--- a/store/src/pmmr.rs
+++ b/store/src/pmmr.rs
@@ -196,7 +196,7 @@ impl<T: PMMRable> PMMRBackend<T> {
 			// Check for a ... (3 dot) ending version of the file - could probably be removed after mainnet
 			let compatible_snapshot_path = PathBuf::from(leaf_snapshot_path.clone() + "...");
 			if compatible_snapshot_path.exists() {
-				LeafSet::copy_snapshot(&leaf_set_path, &compaitble_snapshot_path)?;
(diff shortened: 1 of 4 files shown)


ignopeverell:
This fixes the issue but doesn't consider backward compatibility. As-is it's going to make syncing on floonet very difficult until everyone updates.

store/src/pmmr.rs
```diff
@@ -193,7 +193,13 @@ impl<T: PMMRable> PMMRBackend<T> {
 				data_dir.join(PMMR_LEAF_FILE).to_str().unwrap(),
 				header.hash()
 			);
-			LeafSet::copy_snapshot(&leaf_set_path, &PathBuf::from(leaf_snapshot_path))?;
+			// Check for a ... (3 dot) ending version of the file - could probably be removed after mainnet
+			let compaitble_snapshot_path = PathBuf::from(leaf_snapshot_path.clone() + "...");
```
ignopeverell: Small typo here: compatible. To your defense, I have a hard time typing too...

svechinsky: The pitfalls of relying on autocorrect too much

ignopeverell:
Tested fast sync a few times and it behaves as expected. Note that sync against pre-0.5.2 nodes will still very likely fail (different hash length in leaf file), but hopefully people will all be updating soon.

PR #2356, opened 2019-01-12
Add secondary scaling in JobTemplate

quentinlesceller:
WARNING: might break mining pools and grin-miner.
I recognised that it might be a little bit late to add this in Grin Stratum server.
This allows miners to compute their share difficulty so they don't send low diff shares.

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -85,6 +85,7 @@ pub struct JobTemplate {
 	height: u64,
 	job_id: u64,
 	difficulty: u64,
+	secondary_scaling: u32,
 	pre_pow: String,
 }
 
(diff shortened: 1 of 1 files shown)


ignopeverell:
Aren't most JSON deserialization libraries supposed to ignore a property they don't know anyway?

quentinlesceller:
Normally yes.

ignopeverell:
mwgrinpool uses Python IIRC, which definitely does. Can you confirm @bladedoyle?

Then a couple other pools I seem to remember mentioned Go, I'm not positive but I think it does ignore new property. Anyone else care to comment?

quentinlesceller:
We use Go and I'm pretty sure this is okay for us.
Let's wait for some more comments and merge.

kargakis:
Yeah, unknown fields are ignored in Go by default unless you [explicitly choose otherwise](https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields).

quentinlesceller:
Just pushed the change in grin-miner. Btw we should probably do a min difficulty pre check before submit but that's another PR.

AvaGlobalgithub:
Thanx ! Very helpful!

Issue #2359, opened 2019-01-12
Large number of CLOSE_WAIT connections


ignopeverell:
Seems to indicate we have cases where the other side of the connection closed it but we aren't releasing it properly.

quentinlesceller:
v1.x.x issue. Closing.

PR #2361, opened 2019-01-13
fix: avoid duplicate connection

garyyu:
I see the following logs on one server:
```
20190112 07:20:30.800 DEBUG grin_p2p::peer - connect: handshaking with Ok(V4(157.230.48.217:13414))
20190112 07:20:30.804 DEBUG grin_p2p::peer - connect: handshaking with Ok(V4(157.230.48.217:13414))
20190112 07:20:30.877 DEBUG grin_p2p::handshake - Connected! Cumulative 732 offered from V4(157.230.48.217:13414) "MW/Grin 0.5.2" HEADER_HIST | TXHASHSET_HIST | PEER_LIST | TX_KERNEL_HASH | FULL_NODE
20190112 07:20:30.877 DEBUG grin_p2p::peers - Saving newly connected peer 157.230.48.217:13414.
20190112 07:20:30.877 DEBUG grin_p2p::store - save_peer: V4(157.230.48.217:13414) marked Healthy
20190112 07:20:30.925 DEBUG grin_p2p::handshake - Connected! Cumulative 732 offered from V4(157.230.48.217:13414) "MW/Grin 0.5.2" HEADER_HIST | TXHASHSET_HIST | PEER_LIST | TX_KERNEL_HASH | FULL_NODE
20190112 07:20:30.926 DEBUG grin_p2p::peers - Saving newly connected peer 157.230.48.217:13414.
20190112 07:20:30.926 DEBUG grin_p2p::store - save_peer: V4(157.230.48.217:13414) marked Healthy
```
That means  two parallel connection requests on same peer will succeed, and save same peer to the peers database twice (the later will overwrite former), only one will appear on the active peers list, but with 2 TcpStream connected. This could be the root cause why I often see a lot of duplicated connections from same peer.

And when this connection need to be closed, since we only know one peer (in our active peers list), I'm not sure both TcpStream can be closed, so it could also be the root cause of https://github.com/mimblewimble/grin/issues/2359.

Need long test to confirm whether this PR can 100% kill the connection exhausted problem.

And I'm still not very clear why we have the case of these parallel connection requests on same peer.

### DIFF
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -147,6 +147,16 @@ impl Server {
 					&self.handshake,
 					self.peers.clone(),
 				)?;
+
+				// to avoid duplicate connection if we're already connected to the addr
+				let addr = peer.info.addr.clone();
+				if let Some(_) = self.peers.get_connected_peer(&addr) {
+					debug!("connect_peer: already connected {}", addr);
+					if let Err(e) = stream.shutdown(Shutdown::Both) {
+						debug!("Error shutting down conn: {:?}", e);
+					}
+					return Err(Error::DuplicateConnection);
+				}
 				peer.start(stream);
 				let peer = Arc::new(peer);
 				self.peers.add_connected(peer.clone())?;
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -147,16 +147,6 @@ impl Server {
 					&self.handshake,
 					self.peers.clone(),
 				)?;
-
-				// to avoid duplicate connection if we're already connected to the addr
-				let addr = peer.info.addr.clone();
-				if let Some(_) = self.peers.get_connected_peer(&addr) {
-					debug!("connect_peer: already connected {}", addr);
-					if let Err(e) = stream.shutdown(Shutdown::Both) {
-						debug!("Error shutting down conn: {:?}", e);
-					}
-					return Err(Error::DuplicateConnection);
-				}
 				peer.start(stream);
 				let peer = Arc::new(peer);
 				self.peers.add_connected(peer.clone())?;
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -351,7 +351,7 @@ fn listen_for_addrs(
 	if connecting_history.len() > 100 {
 		let old: Vec<_> = connecting_history
 			.iter()
-			.filter(|&(_, t)| *t + Duration::seconds(60) > now)
+			.filter(|&(_, t)| *t + Duration::seconds(10) < now)
 			.map(|(s, _)| s.clone())
 			.collect();
 		for addr in old {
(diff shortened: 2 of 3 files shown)


ignopeverell:
This seem to duplicate the test done 30 lines above... I think on the client side the issue is `seed.rs` spawns a thread for each connect, so they can happen in parallel. It makes sense to connect quickly and not be hindered by a peer that has a firewall blocking, but here it means we can also get the same peer twice. I think a lock right above the `get_peer_connected` test would solve most of these.

Now I'm still looking at why the server end doesn't terminate all of these properly and some still leak...

ignopeverell:
There are still 3 lines of code after the add during which we could have double connections. It's a concurrency problem so only a lock or an atomic will work, unless we disconnect after the add. Does that make sense?

garyyu:
> It's a concurrency problem so only a lock or an atomic will work

Agree.
And looks like it's not easy to implement a per-IP lock here, since the ip is dynamic.

I change to another way, use a `connecting history` to save recent connecting addresses, and ignore a new connecting request on same address if it just happened within 10 seconds.

Please take a review on this new implementation.

p2p/src/serv.rs
```diff
@@ -147,6 +147,16 @@ impl Server {
 					&self.handshake,
 					self.peers.clone(),
 				)?;
+
```
ignopeverell: You can probably get rid of all this now?

servers/src/grin/seed.rs
```diff
@@ -297,14 +305,29 @@ fn listen_for_addrs(
...
 ) {
 	if peers.peer_count() >= p2p.config.peer_max_count() {
 		// clean the rx messages to avoid accumulating
 		for _ in rx.try_iter() {}
 		return;
 	}
 
+	let now = Utc::now();
 	for addr in rx.try_iter() {
+		// ignore the duplicate connecting to same peer within 10 seconds
```
ignopeverell: From this section and the code below, looks like a HashMap would make all this a lot easier?

p2p/src/serv.rs
garyyu: OK

servers/src/grin/seed.rs
garyyu: Vector is easy to `truncate`, find a position and call one function.
hashmap is easy to use but hard to truncate. Is my understanding correct ?

Pls let me try to refactor with hashmap and compare it with this vector implementation.

servers/src/grin/seed.rs
```diff
@@ -305,6 +315,22 @@ fn listen_for_addrs(
 	}
 
 	for addr in rx.try_iter() {
+		// ignore the duplicate connecting to same peer within 10 seconds
+		let now = Utc::now();
+		if let Some(last_connect_time) = connecting_history.get(&addr) {
+			if *last_connect_time + Duration::seconds(10) > now {
```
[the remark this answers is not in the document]
garyyu: Ok, make sense. Bigger value here means better protection.

PR #2365, opened 2019-01-13
fix: foreign_api to handle OPTIONS call on POSTs

gavinmcdermott:
This PR enables the `foreign_api` to handle the pre-flight `OPTIONS` call that is made on authenticated POST requests from the browser. As it stands today, HTTP clients cannot make authenticated post requests to the `foreign_api` (e.g.: `receive_tx`).

Related changes were originally made to the API for the `owner_api` by @quentinlesceller. Initial dialogue for context to this change can be found in [this PR](https://github.com/mimblewimble/grin/pull/2184) and [this one](https://github.com/mimblewimble/grin/pull/2131/files).

### DIFF
--- a/wallet/src/controller.rs
+++ b/wallet/src/controller.rs
@@ -627,6 +627,10 @@ where
 			ok(create_error_response(e))
 		}))
 	}
+
+	fn options(&self, _req: Request<Body>) -> ResponseFuture {
+		Box::new(ok(create_ok_response("{}")))
+	}
 }
 
 // Utility to serialize a struct into JSON and produce a sensible Response


wallet/src/controller.rs
```diff
@@ -627,6 +627,10 @@ where
 			ok(create_error_response(e))
 		}))
 	}
+
```
ignopeverell: So with the OPTIONS HTTP method is supposed to return the list of allowed HTTP methods on that endpoint in a `Allow` header. Perhaps add that?

gavinmcdermott: Would it be ok if I opened a separate PR for adding the appropriate HTTP methods in the header? I ask because the `foreign_api` is currently unusable, and this patch brings it in to working condition and parity with the `owner_api` behavior (which also doesn't return the methods). 

Note: I'm in the process of learning Rust right now, so it's slow going to patch even the basics. But I'm also working to get a client wallet in to a good place for the protocol launch this week, so I was prioritizing getting the API up to a basic functioning state. I'll happily open a PR to address your comment in when I have a moment .

Thoughts?

ignopeverell: Seems reasonable, given how small a change this it.

PR #2368, opened 2019-01-13
Cancel tx should set output status to Unspent

yeastplume:
Instead of 'Unconfirmed', as it was doing.

### DIFF
--- a/wallet/src/libwallet/internal/updater.rs
+++ b/wallet/src/libwallet/internal/updater.rs
@@ -202,7 +202,7 @@ where
 			batch.delete(&o.key_id, &o.mmr_index)?;
 		}
 		if o.status == OutputStatus::Locked {
-			o.status = OutputStatus::Unconfirmed;
+			o.status = OutputStatus::Unspent;
 			batch.save(o)?;
 		}
 	}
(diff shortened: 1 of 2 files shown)


wallet/src/libwallet/internal/updater.rs
```diff
@@ -202,7 +202,7 @@ where
 			batch.delete(&o.key_id, &o.mmr_index)?;
 		}
 		if o.status == OutputStatus::Locked {
-			o.status = OutputStatus::Unconfirmed;
+			o.status = OutputStatus::Unspent;
```
mably: Could it be problematic the day we re-enable zero-conf send?

yeastplume: Zero conf-send isn't disabled at the moment. 

A small bug came up as a result of the fact we no longer update all outputs from the node on every operation. Previously on cancel we were setting locked outputs to unconfirmed and assuming the status would be fixed on the next wallet operation, but since we're not updating all outputs every time now (just ones involved in a transaction,) this was causing freshly-unlocked outputs to show up as unconfirmed. 

This just resets outputs back to unspent immediately, which should be fine since they had to have been unspent to become locked in the first place.

mably: I remember having a discussion about zero-conf instability some time ago.  Does that mean that the problem has been fixed?  That would be nice.

PR #2372, opened 2019-01-14
Prevent reward overflow

ignopeverell:
Without this, a miner could cause a crash by including a kernel with an insane fee directly in the block. Of course, a crashed node can't relay a block, so it would have limited impact, but still.

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -62,7 +62,7 @@ pub const COINBASE_MATURITY: u64 = DAY_HEIGHT;
 /// function of block height (time). Starts at 90% losing a percent
 /// approximately every week. Represented as an integer between 0 and 100.
 pub fn secondary_pow_ratio(height: u64) -> u64 {
-	90u64.saturating_sub(height / (2 * YEAR_HEIGHT / 90))
+	90u64.saturating_add(height / (2 * YEAR_HEIGHT / 90))
 }
 
 /// The AR scale damping factor to use. Dependent on block height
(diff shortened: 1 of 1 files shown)


core/src/consensus.rs
```diff
@@ -43,7 +43,7 @@ pub const REWARD: u64 = BLOCK_TIME_SEC * GRIN_BASE;
 
 /// Actual block reward for a given total fee amount
 pub fn reward(fee: u64) -> u64 {
-	REWARD + fee
+	REWARD.saturating_sub(fee)
```
eupn: Shouldn't it be `.saturating_add(fee)` since there were an addition (`+`) operator in deleted line?

ignopeverell: Ouch, obviously I should sleep some more. Thanks for checking.

core/src/consensus.rs
```diff
@@ -62,7 +62,7 @@ pub const COINBASE_MATURITY: u64 = DAY_HEIGHT;
 /// function of block height (time). Starts at 90% losing a percent
 /// approximately every week. Represented as an integer between 0 and 100.
 pub fn secondary_pow_ratio(height: u64) -> u64 {
-	90u64.saturating_sub(height / (2 * YEAR_HEIGHT / 90))
+	90u64.saturating_add(height / (2 * YEAR_HEIGHT / 90))
```
eupn: Haven't you just replaced the wrong line? The `saturating_sub` on the reward function is still there.

ignopeverell: Oh wow. Can't touch code anymore. Just fixed, now out to digging a hole for myself.

antiochp:
Somebody should lock @ignopeverell computer away for a few hours.

ignopeverell:
Merging, now that I can see with fresh eyes that I got the one line correct after 3 attempts.

ignopeverell:
Same as here:

https://github.com/mimblewimble/grin/blob/master/doc/build.md#requirements

PR #2373, opened 2019-01-14
More range proof tests

ignopeverell:
Trying to cheat with range proofs to make sure it still fails at
the block level. Defense in depth, belt and suspenders and all that
good stuff.

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -62,7 +62,7 @@ pub const COINBASE_MATURITY: u64 = DAY_HEIGHT;
 /// function of block height (time). Starts at 90% losing a percent
 /// approximately every week. Represented as an integer between 0 and 100.
 pub fn secondary_pow_ratio(height: u64) -> u64 {
-	90u64.saturating_sub(height / (2 * YEAR_HEIGHT / 90))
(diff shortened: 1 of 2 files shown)


core/src/consensus.rs
```diff
@@ -62,7 +62,7 @@ pub const COINBASE_MATURITY: u64 = DAY_HEIGHT;
 /// function of block height (time). Starts at 90% losing a percent
 /// approximately every week. Represented as an integer between 0 and 100.
 pub fn secondary_pow_ratio(height: u64) -> u64 {
-	90u64.saturating_sub(height / (2 * YEAR_HEIGHT / 90))
+	90u64.saturating_add(height / (2 * YEAR_HEIGHT / 90))
```
eupn: ⚠️

antiochp: Github is pretty confusing with stale/outdated changes like this - just to be clear - this change is _not_ actually part of the PR?

ignopeverell: No, absolutely not part of this PR. Github actually gives you a little "Outdated" yellow tag to warn you. But always better to look at the full PR, yes. And sorry for the confusion.

antiochp: I saw the outdated tag but just wanted to be 100% sure.

Issue #2336, opened 2019-01-10
Mechanism to prove UTXO ownership to third-party


seibelj:
Businesses that use Grin for transactions will have to prove to auditors their current balance, incoming transactions, and outgoing transactions. This is crucial for taxpaying entities to use Grin or any other currency. Other privacy coins such as Monero have this capability through view-only keys and send proofs.

In the gitter chat, @jaspervdm said:

> you can prove you are in posession (ie you own UTXOs) of a specific amount, yes, by signing a message with your blinding factor(s). it is however currently not possible with the wallet

Having a wallet feature do this would be ideal, but even without that, the mathematics of how to accomplish this documented would be the bare minimum requirement for a business.

For a third-party auditor with the full blockchain archive synced, the business needs a way to prove these (without giving the third-party the private key):

- Prove current balance (which unspent UTXOs we own)
- Prove incoming transactions (receives)
- Prove outgoing transactions (sends)

Obviously the addresses we receive from or send to don't exist, but the change of UTXO ownership happens in a block height which itself has a timestamp.

ignopeverell:
To expand a bit on @jaspervdm one line answer, if you recall from the [Grin intro doc](https://github.com/mimblewimble/grin/blob/master/doc/intro.md), a commitment has the form:

`C = r*G + v*H`

Here, `v` is the value, `r` the blinding factor and `G` and `H` two elliptic curve point generators. So `r*G` is the public key derived from the private key `r`.

The Grin wallet uses a BIP32 key derivation, so you can find all the `r` used by the wallet easily. Then for each output, sign a message with those `r` and share it with an auditor, along with the amounts and either each of the `r*G` or the root pubkey so they can derive them. From there the auditor can check that:

1. You know `r` that corresponds to `r*G`, as you were able to produce a valid signature.
2. There is a commitment on chain that's `v*H + r*G`.

Note that what you sign doesn't matter, as long as you show you can produce a valid signature, it's all good (practically would likely be the empty string).

Now did I mention we gladly accept PRs? :-)

GandalfThePink:
It does matter what you sign. If you can choose the string to sign it is possible to forge a signature. 
For simplicity you should probably sign the hash of the Pedersen commitment?

jaspervdm:
The verifier should pick the message, to avoid re-use of the signature

tromp:
> Prove outgoing transactions (sends)

[...]

PR #2374, opened 2019-01-14
[WIP] [wallet] UTXO proof of ownership and verification commands

eupn:
As discussed in #2336, this PR provides basic means of signing and verifying output's ownership based on digital signatures with keys derived from blinding factors of outputs.

## Example of UTXO proof generation:
```bash
$ grin --usernet wallet outputs

--------------------------------------------------------------------------------------------------------------------------------------------------------------
 Output Commitment                                                   MMR Index  Block Height  Locked Until  Status   Coinbase?  # Confirms  Value         Tx 
==============================================================================================================================================================
 081a14f7e693ac74a9741d3297b101f5862fb28eadd7506f250fef7fc9b026f2cf  None       49            52            Unspent  true       409         60.000000000  2 
```

Take `081a14f7e693ac74a9741d3297b101f5862fb28eadd7506f250fef7fc9b026f2cf` UTXO's Pedersen Commitment and use it to identify output that we're willing to create proof of ownership for.

Create a hash for message:
```
Blake2b("eupn") = 065294beaa22c6a8c399adaa2388f813425e774da1d14160e48da6b923bcdb15
```

The message hash should be provided by the verifying party to ensure that no signature reuse from proving party is possible.

Generate a proof using commitment and hash:

```bash
$ grin --usernet wallet sign_utxo --utxo 081a14f7e693ac74a9741d3297b101f5862fb28eadd7506f250fef7fc9b026f2cf --msghash 065294beaa22c6a8c399adaa2388f813425e774da1d14160e48da6b923bcdb15

____ Output Proof of Possession ____

Output amount: 60.000000000
Message hash: 065294beaa22c6a8c399adaa2388f813425e774da1d14160e48da6b923bcdb15
Public key (compressed): 03f6ed0a25d5c3f701034cb05a458d06b0312604e6d443de123e962591fe4d14d1
Message signature (compact): 3666059884f673e14aae4e8c79ee26afdd007122aa42b83b0fbe899e8a75f35019fb25bb79b059832a6f171c79bd13c7a4e2ececd533a603140654ceb5c3821a

Command 'sign_utxo' completed successfully
```

Now proving party can send this proof to verifying party and they will be able to check it with `verify_utxo` command.

Furthermore, by having exact amount and a public key, it's possible to check whether there is a commitment to that amount in the blockchain.

## Example of UTXO ownership proof verification:

```bash
$ grin --usernet wallet verify_utxo --amount 60000000000 --msghash 065294beaa22c6a8c399adaa2388f813425e774da1d14160e48da6b923bcdb15 --pubkey 03f6ed0a25d5c3f701034cb05a458d06b0312604e6d443de123e962591fe4d14d1 --signature 3666059884f673e14aae4e8c79ee26afdd007122aa42b83b0fbe899e8a75f35019fb25bb79b059832a6f171c79bd13c7a4e2ececd533a603140654ceb5c3821a

____ Output Proof of Possession ____

        VERIFIED

Command 'verify_utxo' completed successfully
```

## To Do

- [ ] Implement unit tests in `src/bin/cmd/wallet_tests.rs` and `wallet/tests`
- [ ] Determine scope of use cases

Feedback is highly appreciated, hopefully I didn't totally missed the point.

### DIFF
--- a/src/bin/cmd/wallet_args.rs
+++ b/src/bin/cmd/wallet_args.rs
@@ -502,7 +502,7 @@ pub fn parse_sign_utxo_args(args: &ArgMatches) -> Result<command::SignUtxoArgs, 
 
 	Ok(command::SignUtxoArgs {
 		utxo_commitment_string: utxo_commitment_string.to_owned(),
-		message_hash_to_sign: message_hash.to_owned()
+		message_hash_to_sign: message_hash.to_owned(),
 	})
 }
 
--- a/src/bin/cmd/wallet_args.rs
+++ b/src/bin/cmd/wallet_args.rs
@@ -496,6 +496,31 @@ pub fn parse_cancel_args(args: &ArgMatches) -> Result<command::CancelArgs, Parse
 	})
 }
 
+pub fn parse_sign_utxo_args(args: &ArgMatches) -> Result<command::SignUtxoArgs, ParseError> {
+	let utxo_commitment_string = parse_required(args, "utxo")?;
+	let message_hash = parse_required(args, "msghash")?;
+
+	Ok(command::SignUtxoArgs {
+		utxo_commitment_string: utxo_commitment_string.to_owned(),
+		message_hash_to_sign: message_hash.to_owned()
+	})
+}
+
+pub fn parse_verify_utxo_args(args: &ArgMatches) -> Result<command::VerifyUtxoArgs, ParseError> {
+	let amount = parse_required(args, "amount")?;
+	let amount = parse_u64(amount, "amount")?;
+	let message_hash = parse_required(args, "msghash")?.to_owned();
+	let pubkey = parse_required(args, "pubkey")?.to_owned();
+	let signature = parse_required(args, "signature")?.to_owned();
+
+	Ok(command::VerifyUtxoArgs {
+		amount,
+		message_hash,
+		pubkey,
+		signature,
+	})
+}
+
 pub fn wallet_command(
 	wallet_args: &ArgMatches,
 	mut wallet_config: WalletConfig,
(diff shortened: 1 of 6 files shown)


yeastplume:
Looks good so far, but I can't really give it a full review today and I think we're a bit late in the day to merge before mainnet tag, but can look at it shortly after.

We'd generally want to see some unit test as well that exercise the functionality, see `src/bin/cmd/wallet_tests.rs` as well as the tests in `wallet/tests`

yeastplume:
Also, @tromp's last comment in the issue https://github.com/mimblewimble/grin/issues/2336#issuecomment-453975879 is correct and should be taken into consideration, perhaps a bit more thought might be required

ManfredKarrer:
Bisq require the ability to provide a proof of the transfer (from sender) to a 3rd party if requested (arbitrator in dispute case). So I hope that we will see soon such a feature otherwise we might need to deactivate Grin until its there as otherwise security of a trade is not guaranteed.

tromp:
Does said 3rd part have a known public key?

ManfredKarrer:
@tromp How and what he gets is not so important, best the minimum needed to proof the payment. In Monero they have a tx key to achieve that. In beam they just added something but I did not look closer so far. So preferred is to have a proof for one the scope of one tx.

ManfredKarrer:
@tromp I am not very familiar with the details of Grin and that proposal. Bisq arbitrators are not part of the trade by default but only get contacted if there is a problem. The Grin trade is done pure P2P between the traders (their grin wallets). If the sender can proof he sent x amount of grin to wallet URL y that would be enough.

tromp:
Are both trading parties trusting Bisq to pass them the right wallet URL?
What do wallet URLs look like?  Are they like IP address and port?
I'm not sure how you could prove that you transacted with a particular IP address and port.
Seems you still need some authentication mechanism behind it...

ManfredKarrer:
We support both IP:port and grinbox addresses. In a dispute case both traders reveal some privacy to the arbitrator so that he learns about their wallet URL should not be any issue. 
I don't know how it can work in Beam to provide that proof but it will be for sure not only for us a mandatory feature. I assume any case of a transfer of non-trusted peers will require that. Otherwise the receiver can always claim to never have received anything....

PR #2375, opened 2019-01-14
Wallet Performance enhancements - cache commit + transaction fetching logic

yeastplume:
Some performance enhancements that should greatly improve speed of larger wallet sets:

* I'd removed the cached commit from the database as it was making the logic more complicated, however it also ends up slowing everything down if all commits have to be re-calculated all the time (such as during a `grin wallet outputs`. To add this 'cache' back in a more sensible way, I've simply added it to the stored WalletOutput struct as an optional field, that can be populated when an output is being created.

* Because this leads to slightly worse privacy by having the commit in the database, (though no worse then when I removed the cache earlier,) I've added a flag to the wallet config to turn off commit caching (i.e. store None in the commit field of the WalletOutput'. You can elect to turn off commit caching at the expense of a performance hit.

* Also simplified transaction retrieval logic to avoid looping through iterator many times and speeding up the check for whether an output needs to be updated from the server (via checking the TX log)

### DIFF
--- a/wallet/src/lmdb_wallet.rs
+++ b/wallet/src/lmdb_wallet.rs
@@ -200,7 +200,11 @@ where
 	}
 
 	/// return the version of the commit for caching
-	fn calc_commit_for_cache(&mut self, amount: u64, id: &Identifier) -> Result<Option<String>, Error> {
+	fn calc_commit_for_cache(
+		&mut self,
+		amount: u64,
+		id: &Identifier,
+	) -> Result<Option<String>, Error> {
 		if self.config.no_commit_cache == Some(true) {
 			Ok(None)
 		} else {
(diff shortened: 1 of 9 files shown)


antiochp:
Taking a look now.

@yeastplume how comfortable are you with these changes going in pre-mainnet vs. post-mainnet?
What risks are there in terms of introducing something unexpected here?
i.e. Is it worth the risk to do it now vs. in a few days?

wallet/src/lmdb_wallet.rs
```diff
@@ -199,6 +199,17 @@ where
 		&mut self.w2n_client
 	}
 
+	/// return the commit, if enabled, None otherwise
+	fn cached_commit(&mut self, amount: u64, id: &Identifier) -> Result<Option<String>, Error> {
```
antiochp: Is this maybe mis-named or am I misunderstanding what is going on here?
Its not actually cached in here?

PR #2377, opened 2019-01-14
Only create single TXLog entry on restore

yeastplume:
Futher improvement (that shouldn't affect anything other than restore,) on restore, instead of outputting a new transaction log entry for every output, just create one received TX log entry for the entire amount. The new TX only includes non-coinbase transactions; coinbase transactions get their own TX log entry.


yeastplume:
going to redo PR

Issue #2251, opened 2018-12-29
github release questions


garyyu:
I leave an issue here to remind checking with github why the `release` is always tagged with the wrong  author.

And a task to give a separated script in `/etc` to generate release note manually.

garyyu:
Update for the issue of released by a wrong tagged author:
https://github.community/t5/How-to-use-Git-and-GitHub/When-Alice-making-a-release-sometimes-it-s-tagged-with-Bob/m-p/17214#M5244

Somebody mentioned that "possible that the person who saved the first draft gets their name on the release", I'm not sure whether it indeed is, but before Github supporter check the problem, we can try it, just remove the 1st release (since it's a release test there).

garyyu:
Wrong "released by" problem will be solved by https://github.com/mimblewimble/grin/pull/2384

And then we don't need a separated script to manually generate change log :-)

garyyu:
Closed by #2384

PR #2384, opened 2019-01-15
use travis-ci to update a release instead of creating new release

garyyu:
This PR solve the problem in https://github.com/mimblewimble/grin/issues/2251:  "released by" is wrong.

According to Github staff's feedback:
> Please note, the person who creates the release can be different to the user who originally created the tag.

And after reading Github API docs: https://developer.github.com/v3/repos/releases/#edit-a-release
We can use `edit a release` instead of `create a release`, then we will never have such kind of problem, the member who created a release in github will never be touched and correctly tagged with "released by".

So, we just need add `--request PATCH ` on our request, and put the `release id` as a request parameter.

### DIFF
--- a/.auto-release.sh
+++ b/.auto-release.sh
@@ -2,6 +2,7 @@
 
 repo_slug="mimblewimble/grin"
 token="$GITHUB_TOKEN"
+export CHANGELOG_GITHUB_TOKEN="$token"
 
 tagname=`git describe --tags --exact-match 2>/dev/null || git symbolic-ref -q --short HEAD`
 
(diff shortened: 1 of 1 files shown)


.auto-release.sh
```diff
@@ -62,7 +62,8 @@ jq -n \
     prerelease: false
   }' > CHANGELOG.md
 
-echo "Create release $version for repo: $repo_slug, branch: $branch"
-curl -H "Authorization: token $token" --data @CHANGELOG.md "https://api.github.com/repos/$repo_slug/releases"
+release_id="$(curl -0 -XGET -H "Authorization: token $token" https://api.github.com/repos/garyyu/grin/releases/tags/$tagname 2>/dev/null | grep id | head -n 1 | sed 's/ *"id": *\(.*\),/\1/')"
```
jaspervdm: i think you forgot to replace `garyyu/grin` with `$repo_slug`

PR #2387, opened 2019-01-15
Temp fix on port seed

quentinlesceller:


jasondavies:
Slightly better (retains support for floonet):

```rust
addr.set_port(if global::is_floonet() { 13414 } else { 3414 });
```

PR #2395, opened 2019-01-16
Update dependencies

dragononcrypto:
Include cargo for Debian-based systems, required for building.


antiochp:
If you install rust via rustup (recommended) you should not need to install cargo separately.

ignopeverell:
What @antiochp said. Running debian's `cargo` is likely to still be a problem.

PR #2398, opened 2019-01-16
Fix: Recovering from phrase gives IO Error

polac:
If you try to recover your wallet from the phrase on a fresh "install" of the grin it will always fail with IO Error. This happens because the directory for the seed file to be generated doesn't exist. 

If you check the functions init_file and from_file in this same file, you can see that these exact lines can be found there.

My first lines of rust and grin. Very exiting project.

### DIFF
--- a/wallet/src/types.rs
+++ b/wallet/src/types.rs
@@ -191,6 +191,9 @@ impl WalletSeed {
 		word_list: &str,
 		password: &str,
 	) -> Result<(), Error> {
+		// create directory if it doesn't exist
+		fs::create_dir_all(&wallet_config.data_file_dir).context(ErrorKind::IO)?;
+		
 		let seed_file_path = &format!(
 			"{}{}{}",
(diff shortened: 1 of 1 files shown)


yeastplume:
Thanks for getting involved! There is definitely an issue if you go `grin wallet recover` right off the bat without initializing a wallet (running `grin wallet init` first), but unfortunately this change will leave the user unable to run `grin wallet init` if they call `recover` first (it will fail since the wallet seed file already exists.

Users should really be going `grin wallet init -r` to init and recover at the same time. Will have to think about the best fix, but it should be more a matter of checking whether wallet data exists first before allowing the user to recover.

sesam:
Maybe it would be most safe and functional to catch the error, say `.unwrap_or(_)` and show a warning to instead do `grin wallet init -r` ? Or maybe even immediately launch the `init -r` workflow, maybe with some Info text about that and why it happens that way. Thougts?

yeastplume:
Going to close this for now, as I don't think the proposed fix should be merged.

PR #2405, opened 2019-01-17
Change owner API port for MainNet

### LINKED ISSUE #2389: Owner api port in floonet range on mainnet.
When switching to mainnet. I expected all api endpoint to drop 10k in portnumber. Which has happened for all ports, except the ownerapi. 

Hardcoded here:

https://github.com/mimblewimble/grin/blob/5cf4d7acfe04717ec345e197e9dc69a9ebbe2dee/wallet/src/command.rs#L148

This is a problem in the situation where you run floonet and mainnet on the same machine. Configuration like #2328 would be nice. But setting a default inline with the other apis would be more pressing imo. 

Related discussions: 
https://github.com/mimblewimble/grin/issues/2328
https://github.com/mimblewimble/grin/pull/1867

thedevworks:
Fixes #2389 ?

### DIFF
--- a/wallet/src/command.rs
+++ b/wallet/src/command.rs
@@ -145,7 +145,7 @@ pub fn owner_api(
 ) -> Result<(), Error> {
 	let res = controller::owner_listener(
 		wallet,
-		"127.0.0.1:13420",
+		"127.0.0.1:3420",
 		g_args.node_api_secret.clone(),
 		g_args.tls_conf.clone(),
 		config.owner_api_include_foreign.clone(),
--- a/wallet/src/command.rs
+++ b/wallet/src/command.rs
@@ -26,6 +26,7 @@ use uuid::Uuid;
 
 use crate::api::TLSConfig;
 use crate::core::core;
+use crate::core::global;
 use crate::keychain;
 
 use crate::error::{Error, ErrorKind};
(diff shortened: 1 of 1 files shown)


thedevworks:
I could make the port configurable (But not the address, since that looks to be part of a wider discussion around HTTPS and running owner API outside of localhost) which solves this issue for now?

antiochp:
Oh I see - normally we configure the whole thing (ip+port). But here we need to limit it to the port.

I think we should do this yes, following the existing config approach as closely as possible but for the port only.

Otherwise we fix it for mainnet and immediately break it for floonet for anyone running both on the same system (and having floonet pointing at hard-coded mainnet port is arguably worse).

ignopeverell:
Right, so a basic `if` for floonet to just hardcode it to a different port?

antiochp:
> Right, so a basic if for floonet to just hardcode it to a different port?

That would be the pragmatic approach for `v1.0.1`. :+1:

sesam:
@thedevworks Good initiative! Find `is_floo` in the code and you'll soon be done :)

RaghavSood:
I do think that the owner API port should be configurable, and not hard coded to `3420`. A very simple usecase we are encountering at an exchange is to be able to run the owner API for two separate wallets (not separate accounts, but entirely separate wallets - different seeds and everything) on the same machine. This is currently not possible with the hard coded ports, and requires that we spin up containers/machines for each wallet, which is not ideal.

PR #2409, opened 2019-01-17
Added support for saving message

alesito85:
This PR intends to add support for saving messages that are sent with Grin transactions. In Grin's current state it is possible to store messages but nothing is done with them. In order to be able to use messages and streamline implementation into 3rd party services (merchant or exchange) the team at ChainRift exchange has come up with these basic changes.

What the PR does is save all messages from participants and stores them into the local database, separated by pipe, only extending existing features.

The messages are currently only shown when fetching transactions through owner API. If necessary the txs command can be extended to show messages as well.

Since Grin designers may have intended to save/show messages in a different way I welcome any and all comments in order to meet their intent. Since this was my first touch with Rust feel free to make code suggestions if necessary.

### DIFF
--- a/wallet/src/libwallet/internal/selection.rs
+++ b/wallet/src/libwallet/internal/selection.rs
@@ -71,6 +71,15 @@ where
 	slate.lock_height = lock_height;
 	slate.fee = fee;
 	let slate_id = slate.id.clone();
+	let mut message = "".to_string();
+	if slate.participant_data.len() > 0 {
+		for data in slate.participant_data.iter() {
+			if message.len() > 0 {
+				message.push_str(&"|".to_string());
+			}
+			message.push_str(&data.message.clone().unwrap_or("".to_string()));
+		}
+	}
 
 	let keychain = wallet.keychain().clone();
 
@@ -109,6 +118,7 @@ where
 			let log_id = batch.next_tx_log_id(&parent_key_id)?;
 			let mut t = TxLogEntry::new(parent_key_id.clone(), TxLogEntryType::TxSent, log_id);
 			t.tx_slate_id = Some(slate_id);
+			t.message = Some(message);
 			let filename = format!("{}.grintx", slate_id);
 			t.stored_tx = Some(filename);
 			t.fee = Some(fee);
@@ -183,6 +193,16 @@ where
 	let height = slate.height;
 
 	let slate_id = slate.id.clone();
+	let mut message = "".to_string();
+	if slate.participant_data.len() > 0 {
+		for data in slate.participant_data.iter() {
+			if message.len() > 0 {
+				message.push_str(&"|".to_string());
+			}
+			message.push_str(&data.message.clone().unwrap_or("".to_string()));
+		}
+	}
+
 	let blinding =
 		slate.add_transaction_elements(&keychain, vec![build::output(amount, key_id.clone())])?;
 
@@ -205,6 +225,7 @@ where
 		let mut t = TxLogEntry::new(parent_key_id.clone(), TxLogEntryType::TxReceived, log_id);
 		t.tx_slate_id = Some(slate_id);
 		t.amount_credited = amount;
+		t.message = Some(message);
 		t.num_outputs = 1;
 		batch.save(OutputData {
 			root_key_id: parent_key_id.clone(),
--- a/wallet/src/controller.rs
+++ b/wallet/src/controller.rs
@@ -581,7 +581,8 @@ where
 					error!("Error validating participant messages: {}", e);
 					err(e)
 				} else {
-					match api.receive_tx(&mut slate, None, None) {
+					let msg = slate.participant_data[0].message.clone();
+					match api.receive_tx(&mut slate, None, msg) {
 						Ok(_) => ok(slate.clone()),
 						Err(e) => {
 							error!("receive_tx: failed with error: {}", e);
--- a/wallet/src/libwallet/types.rs
+++ b/wallet/src/libwallet/types.rs
@@ -612,6 +612,8 @@ pub struct TxLogEntry {
 	pub fee: Option<u64>,
 	/// Location of the store transaction, (reference or resending)
 	pub stored_tx: Option<String>,
+	/// Message
+	pub message: Option<String>,
 }
 
 impl ser::Writeable for TxLogEntry {
@@ -644,6 +646,7 @@ impl TxLogEntry {
 			num_outputs: 0,
 			fee: None,
 			stored_tx: None,
+			message: None,
 		}
 	}
 


mmitech:
This also will allow for a standard and unified implementation for exchanges payment processor and merchants where users need to be uniquely identified.
Right now it is complete chaos, each exchange seems to have their own way of dealing with deposits and withdrawals that are in many cases not compatible with other exchanges, which makes it very hard and confusing to users.

alesito85:
@garyyu @yeastplume @ignopeverell Could you review my PR?

CD0x23:
Works great :clap:
screenshot at jan 18 16-13-43

wallet/src/libwallet/internal/selection.rs
```diff
@@ -71,6 +71,15 @@ where
 	slate.lock_height = lock_height;
 	slate.fee = fee;
 	let slate_id = slate.id.clone();
+	let mut message = "".to_string();
+	if slate.participant_data.len() > 0 {
+		for data in slate.participant_data.iter() {
+			if message.len() > 0 {
+				message.push_str(&"|".to_string());
```
yeastplume: What if the message contains a pipe?

wallet/src/libwallet/internal/selection.rs
```diff
@@ -109,6 +118,7 @@ where
 			let log_id = batch.next_tx_log_id(&parent_key_id)?;
 			let mut t = TxLogEntry::new(parent_key_id.clone(), TxLogEntryType::TxSent, log_id);
 			t.tx_slate_id = Some(slate_id);
+			t.message = Some(message);
```
yeastplume: This should be None if there are no messages rather than an empty string.

wallet/src/libwallet/internal/selection.rs
```diff
@@ -205,6 +225,7 @@ where
 		let mut t = TxLogEntry::new(parent_key_id.clone(), TxLogEntryType::TxReceived, log_id);
 		t.tx_slate_id = Some(slate_id);
 		t.amount_credited = amount;
+		t.message = Some(message);
```
yeastplume: Same, should be None if there's no message

wallet/src/libwallet/internal/selection.rs
alesito85: I don't know the details of how it all works. What I've tested until now there were always 2 participants. In this case there will always be a senders message only so I don't expect (could be wrong) that there will be two messages (senders and recipients). Could there be 3 participants and thus 3 messages?

About message containing pipe, I didn't account for that since it's not my use case but since message supports any utf8 character there could also be newlines, correct? I'm open to suggestions for a better delimiter.

Alternatively I could even expand the tx slate and store messages separately if there can only be two participants (as in sender's message and recipient's message to respond).

yeastplume:
Sorry, just one other thing, this only seems to store the message when outputs are being created. I think it would also need to update the stored messages with all participant messages at each part of the exchange (so during finalize as well, which would update the sender's database with all of the participant messages)

wallet/src/libwallet/internal/selection.rs
yeastplume: Yes, and there will ultimately need multiple participants. A JSON structure would work, the best way to do that would be to serialise a struct that contains an id, public key, signature and the message itself

alesito85:
Would this be necessary in order to merge this PR or could it be done separately? Above I've suggested a JSON encoded message store in order not to over complicate things due to the possibility of multiple participants. I'm just not sure what would be the identifier of each participant. Any suggestions?

yeastplume:
Since mainnet has been launched, we need to be very careful about what we add in now since we'll now be required to support it indefinitely. Unfortunately this means we need to move a bit more slowly and thoughtfully than we did before launch.

So I think something like this could work:

* Create a serializable ParticipantMessage struct that contains an id, public_key, signature, and the plaintext message. I'm also noting that the ParticipantData struct in core/src/libtx/slate.rs already has this info, and I wish now the message portion had been split into a separate struct, that would have made things easier here, (but if we change this then slates won't be compatible between minor versions' So I'd probably add this ParticipantMessage struct into slate.rs and add a `to_participant_message` method to ParticipantData that returns a ParticipantMessage
* Similarly, add a method to the Slate that returns a vec of ParticipantMessages.
* Then TxLogEntry gets an Option<Vec<ParticipantMessage>> field
* Ensure this field is updated at every point in the process, including sender creation, recipient, and finalization.

And again, this isn't intended to be definitive proof of participation.

@ignopeverell @antiochp Does this seem reasonable to you

mmitech:
> Since mainnet has been launched, we need to be very careful about what we add in now since we'll now be required to support it indefinitely. Unfortunately this means we need to move a bit more slowly and thoughtfully than we did before launch.
> 
> So I think something like this could work:
> 
> * Create a serializable ParticipantMessage struct that contains an id, public_key, signature, and the plaintext message. I'm also noting that the ParticipantData struct in core/src/libtx/slate.rs already has this info, and I wish now the message portion had been split into a separate struct, that would have made things easier here, (but if we change this then slates won't be compatible between minor versions' So I'd probably add this ParticipantMessage struct into slate.rs and add a `to_participant_message` method to ParticipantData that returns a ParticipantMessage
> * Similarly, add a method to the Slate that returns a vec of ParticipantMessages.
> * Then TxLogEntry gets an Option<Vec> field
> * Ensure this field is updated at every point in the process, including sender creation, recipient, and finalization.
> 
> And again, this isn't intended to be definitive proof of participation.
> 
> @ignopeverell @antiochp Does this seem reasonable to you

How not? we don't need a cryptographical proof, if a user, pool or an exchange provide a unique message with their tx it would be enough for us to help us identify to whom it is intended! because right now exchanges and pools are doing this by including a unique string in their URLs, not sure if that is remotely better than this not to mention the security implications, I would rather see this message packed in a tx, what do you think?

yeastplume:
@mmitech it already is in the TX and exchanges can already use it. This PR is just about how to sensibly store it in the wallet alongside the rest of the transaction info.

mmitech:
yes, chainrift.com is using this feature, I was just replying to the part about it not being definitive proof of participation.
Thanks for taking the time to look at this.

alesito85:
@yeastplume please help me understand the use case of multiple participants. Currently since there are always two one is obviously the sender and one participant. Is it possible, even in this case, that they would both be sender and recipient via multiple slate exchanges in one tx?

In case of 3 or more participants is there always one sender or can there be multiple senders? If this even makes sense.

How would message be set on recipients side. Currently there's no way to do this, correct?

The amount in a slate represents the total transferred amount, even in case of 3 or more participants?

When there are multiple participants, is it intended to be possible to send a different message to each recipient?

ignopeverell:
Agree with @yeastplume comments. Not sure if that's already there, but as always when accepting user inputs, the content must be checked. In this particular case, length limit?

alesito85:
> Agree with @yeastplume comments. Not sure if that's already there, but as always when accepting user inputs, the content must be checked. In this particular case, length limit?

Input limit should be checked where it's entered, I don't think this PR should have anything to do with it since it's just saving the message.

sesam:
@alesito85 Unfortunately I think it will be necessary to check length before saving, because it will be data that comes in from internet, potentially directly into an open IP:port, meaning someone can try to fill up your disk, spam with unfriendly unicode, put in data that is later read by non-rust code and cause that code to crash (ask Bobby Tables). Are you using this on an internet-accessible listener, i.e. the`grin wallet listen`?

If chainrift is already using the message feature, we must take care deserializing any message, or only deserialize string to json (struct) if the message begins with something, say "{".

The immediate next step seems to be adding a ParticipantMessage struct.in slate.rs

yeastplume:
Thanks for getting this going, but I'm going to close this in favor of https://github.com/mimblewimble/grin/pull/2441.. please feel free to review there and/or continue the discussion.

Issue #2394, opened 2019-01-16
Nodes should maintain healthy mix of inbound/outbound peers


antiochp:
Node running overnight configured with `peer_max_count = 100`.
* 1 outbound peer
* 99 inbound peers

We need a healthy number of outbound peers for Dandelion to function correctly as we choose a relay from our outbound peers.

I'm assuming that given the number of nodes on the network, all creating connections we have an imbalance here - 
* we (a single node) try to create outbound connections
* many other nodes try to create inbound connections to us

And as connections drop over time (for a variety of reasons) chances are our connection "slots" get grabbed by other nodes connecting to us and not by us connection to other nodes.

So our number of outbound connections grows smaller over time.

Is this because new nodes are coming online over time?

Is this because many nodes are behind NAT and not able to accept inbound connections leading to this imbalance on "open" nodes?

antiochp:
We have the following config by default -
```
peer_min_preferred_count = 8
```

I wonder if this is as simple as checking the number of outbound peer connection in `monitor_peers` and keeping _this_ number above `peer_min_preferred_count` (instead of the total number of connected peers currently).

https://github.com/mimblewimble/grin/blob/8badb58665a9f9056599d59a5d76d3a92660706f/servers/src/grin/seed.rs#L189-L192

ignopeverell:
Seems like a nice enough solution. Keep in mind your stats may be really skewed due to the recent release however.

antiochp:
Will watch nodes over the next few days to see how they behave.

Pretty consistently seeing nodes lose all their outbound connections over a few hours though right now, with them being replaced by a ready stream of inbound connections.

Same node went from 18 outbound, 24 inbound after a restart, to 1 outbound, 99 inbound after approx 3 hours.

lumirl:
The same happens with us. Is there any IP list of preferred "main" peers to keep them in config and to have both inbound and outboud connections?

antiochp:
Not currently no - we have "preferred" peers but this only applies during initial node startup. Once we've peered we just (currently) keep the total between the min and max values.

I'll plan to work on making these proposed changes.

antiochp:
Thinking a bit more about this, maybe we can be less invasive here by roughly attempting to maintain a 50/50 split between inbound/outbound peers.

Look for more peers if - 
* `connected_peers < peer_min_preferred_count`, or
* `connected_outbound_peers < peer_min_preferred_count / 2`

[...]

PR #2417, opened 2019-01-18
Maintain min outbound peers (min_preferred_peers / 2)

antiochp:
Resolves: #2394.

Large number of nodes on mainnet and a growing network size means nodes are inundated with `inbound` connections (if they are public nodes, not behind NAT).
This means we are losing outbound connections over time as we do not prioritize these at all.

See #2394 for more context.

This PR changes the "do we need to dial more peers" behavior so it is aware of our `outbound` peer count in addition to the total connected count.

If `peer_min_preferred_count = 8` (default config) we attempt to open more outbound connections if - 
* we have less than 8 connected peers, or
* we have less than 4 connected outbound peers

TODO (in progress) - 
- [x] test against mainnet with node that accepts inbound connections
Edit: Node running happily on mainnet for ~12 hours and connections staying healthy.

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -171,12 +171,21 @@ impl Peers {
 		self.peers.read().get(addr).map(|p| p.clone())
 	}
 
-	/// Number of peers we're currently connected to.
+	/// Number of peers currently connected to.
 	pub fn peer_count(&self) -> u32 {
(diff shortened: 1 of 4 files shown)


antiochp:
Merging this to master for inclusion in `v1.0.1`.

PR #2420, opened 2019-01-18
default log level DEBUG -> INFO

### LINKED ISSUE #2419: Default DEBUG log level is probably too verbose
We may want to consider changing this to `INFO` as this is going to get pretty verbose on _every_ node.

https://github.com/mimblewimble/grin/blob/bb933dcd694aef8a3856c37c35978916a869ac9d/util/src/types.rs#L53-L60

antiochp:
Resolves #2419.

We may want to keep it at DEBUG for a bit but its filling everyones disks up, mainly with peer connection messages as the network grows and hopefully stabilizes.

Thoughts?

### DIFF
--- a/util/src/types.rs
+++ b/util/src/types.rs
@@ -56,7 +56,7 @@ impl Default for LoggingConfig {
 			log_to_stdout: true,
 			stdout_log_level: LogLevel::Warning,
 			log_to_file: true,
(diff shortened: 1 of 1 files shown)


ignopeverell:
Agreed. We can easily change to debug or ask people to change whenever we want to dig in some issue.

PR #2424, opened 2019-01-19
Improve error output in cases when db_root / logfile is inaccesible 

canselcik:
I hit this situation earlier today when the server had been previously run as root and now the same db_root is being used. Instead of a silent failure, this makes it simpler to fix the issue and move on for the person running the node.

The same situation applies to the logfile, so addressed that in a second commit.

Minor things like this seem to go a long way in giving it polish. Next I am thinking of looking into implementing the `grin server stop` command however the `/tmp/grin.pid` is empty because potentially Daemonize crate isn't flushing the data written to the file.


ignopeverell:
Nice, we need a lot of small quality-of-lie improvements like these. Thanks!

PR #2432, opened 2019-01-20
Use Option::unwrap_or instead of is_some followed by unwrap in start_server_tui

canselcik:
Functionally equivalent however this makes it more clear that the default is `false`.


antiochp:
Feel like this is basically a subjective preference with some preferring one approach and others preferring the other approach.

https://users.rust-lang.org/t/is-it-bad-style-to-match-a-bool/14359

I think clippy may take issue with you matching on a bool here as well.

We may want to do something like - 

```
if config.run_tui.unwrap_or(false) {
    ...
}
```

canselcik:
Very true, I like the `config.run_tui.unwrap_or(false)` you suggested. I can update the PR if there are no other suggestions?

PR #2433, opened 2019-01-20
Implementing stop and status verbs for the server command

canselcik:
This PR implements the `stop` and `status` verbs for the `server` command. In order to do so, it removes reliance on the `Daemonize` crate's pidfile since it seems to not flush the contents of the pidfile, leading to an empty `/tmp/grin.pid`.

Quick demo:
```
$ ./target/release/grin server status
'grin-server' is not running
$ ./target/release/grin server stop  
'grin-server' is not running

$ ./target/release/grin server start
$ ./target/release/grin server status
'grin-server' running with PID 1687

$ ./target/release/grin server stop  
'grin-server' found running with PID 1687
Successfully stopped
```
One concern I have here is that `proc.kill()` sends a `SIGKILL` however we probably want a `SIGQUIT` here first, which is simple but it would require adding a direct dependency to the `libc` crate. If that's okay, I can take care of that.


ignopeverell:
Wouldn't it be better to take that to the `Daemonize` crate first to see if they would merge a fix?

canselcik:
Definitely and I am all for that however we will still need to address
reading these pidfiles like in this PR even if Daemonize flushed them at
the right time.

I will look into how Daemonize implements things tonight and see if a quick
fix is likely.

canselcik:
Looks like the `daemonize` crate has the https://github.com/knsd/daemonize/issues/15 issue opened about it approximately 2 years ago. It comes down to patching that crate like https://github.com/mikhail-m1/daemonize/commit/3c27a97291c1fa086df415cf4399fee0c874d999. However I don't think they'll be addressing it anytime soon and it doesn't make sense for us to fork that crate in order to implement this feature. I am actually surprised this is an issue given how many projects depend on that crate.

However the `parity-daemonize` crate is used by Parity's nodes and they are also pretty time-tested. I have modified this PR to use that crate.

Let me know what you think. I have tested this implementation and seems to work very well on my hosts. If you don't mind one more crate dependency, I could send a SIGTERM instead of a SIGKILL by using the libc crate.

kargakis:
I would be in favor of removing the ability to run grin as a daemon via the `grin` cli altogether, it does not add much utility (what happens if grin crashes while it runs?) and I expect people are going to use other tools to daemonize grin (eg. systemd, kubernetes). Is anyone running a business using `grin server start`?

canselcik:
I do think the ability to spawn the `grin-server` as a daemon from the `grin-cli` is a nice perk to have. It is indeed true that there is no remediation in case the daemon crashes and for that one would need to use `upstart` or `systemd` or some service acting as a watchdog.

However the reason why I don't see that as a limitation that makes the ability to spawn it from the cli obsolete. After all, if we were using the the `SysVinit` style init scripts and shipping with an init script that could be put into `/etc/init.d/grinserver`, you wouldn't have remediation baked into those scripts either.

kargakis:
> you wouldn't have remediation baked into those scripts either.

Most Linux distros have moved to systemd. Fwiw, I haven't seen any production deployment of any app yet where a supervisor process is not used.

canselcik:
Exactly my point. For people who want the simple daemonization, I think `grin server start` is incredibly useful. Basically if I want to do development work on something that interfaces with the `grin-server`, I simply want to do `grin server run &>/dev/null &; disown` and we get just that and a little bit more with this.

Beyond that if anyone wants to use systemd, upstart, etc., they can just leverage `grin server run`.

And if they want a simple watchdog process, they can simply roll out their own solution and execute the following intuitive and easy-to-implement logic:
```
  pid = readfile("/tmp/grin.pid")
  match get_process(pid) {
    Some(p) => {
      if !some_healthcheck() {
        kill(p)
        start_grin()
      }
    }
    None => start_grin()
  }
```

ignopeverell:
What's the systemd equivalent on Windows? Is there one?

- Someone who stopped using Microsoft products after Windows 2000.

canselcik:
@ignopeverell Same here, I stopped using Windows after XP and only went back to it on occasion for IDA Pro on Windows 7.

As far as I know there are two ways to get the server process out of the user's way in Windows:

(1) Spawn a new process, hide its window. So effectively a `CreateProcessA`  with `DETACHED_PROCESS` creation flag set followed by a `ShowWindow(processHandle, SW_HIDE)`.

(2) Create a proper Windows Service with the following command, however the executable would need to implement `VOID WINAPI ServiceMain( _In_ DWORD dwArgc, _In_ LPTSTR *lpszArgv )` as explained [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms685138(v=vs.85).aspx).
```
sc.exe create <new_service_name> binPath= "<path_to_the_service_executable>"
```

kargakis:
@ignopeverell can be closed now that https://github.com/mimblewimble/grin/pull/2535 got merged

yeastplume:
Agreed I think we can close this for now given daemonize was removed for now. It's not like there was tons of code here and we can revisit later if windows needs any customization to run as a service.

PR #2435, opened 2019-01-20
[DNM] fix overweight bucket issue

antiochp:
This PR fixes a potential issue where we can build an internal "bucket" of txs that exceeds the max allowed tx/block weight, when preparing mineable txs from the txpool.

We build buckets of dependent txs (0-conf txs spending outputs from other txs in the pool).
We then take each bucket and aggregate the txs, checking we do not exceed the block weight as we go.
_But_ we do not check any weights or limits when building the buckets themselves.

Getting into a txpool state where we had a single bucket exceeding the max block weight would be bad - we have no way of recovering from this and the txs would be stuck in the pool forever.

This PR adds a very simple check against a MAX_BUCKET_SIZE (defined currently as 64).
i.e. We allow a bucket to be 64 txs in size (64 dependent 0-conf txs).

Note it is still possible for a bucket to exceed the max allowed block weight but it would need to occur within 64 txs.

Q) Do we want to go further here and actually check the block weight limit during bucket construction.
I was initially thinking no, but now I'm leaning toward doing this properly.

 Edit: `MAX_BUCKET_SIZE` will not actually prevent this happening. A tx at max `weight_as_block` followed by a 0-conf spending one of its outputs would be sufficient to produce an invalid bucket of txs.

So we need to be tracking the `weight_as_block` of the buckets as we build them.
Ideally by aggregating the txs to ensure this weight is accurate, but maybe it would be enough to calculate the approximate `weight_as_block` of the bucket (to avoid the relatively expensive aggregation step).

### DIFF
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -27,6 +27,9 @@ use grin_util as util;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
+/// Max size of a "bucket" of dependent txs in the pool.
+const MAX_BUCKET_SIZE: usize = 64;
+
 pub struct Pool {
 	/// Entries in the pool (tx + info + timer) in simple insertion order.
 	pub entries: Vec<PoolEntry>,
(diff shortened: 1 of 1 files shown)


antiochp:
Closing this. Will reopen a different approach (and with corresponding issue).

Issue #2134, opened 2018-12-11
serialize various slate elements as hex (and not arrays of ints)


antiochp:
I wonder how feasible/realistic it would be to update the wallet slate serialization before mainnet to serialize the various bytes slices as hex rather than arrays of integers?

Could we potentially do this as a v2 of the slate format? 

```
{
    "amount": 4206000000,
    "height": 77907,
    "tx": {
        "body": {
            "outputs": [
                {
                    "proof": [
                        82,
                        154,
                        122,
                        171,
                        25,
                        9,
                        93,
                        192,
                        208,
                        133,
                        80,
                        123,
                        178,
                        45,
                        170,
                        96,
                        89,
                        171,
                        96,
                        149,
                        138,
                        98,
                        168,
                        169,
                        200,
                        221,
                        236,
                        104,
                        131,
                        203,
                        104,
                        100,
                        40,
                        61,
                        199,
                        103,
                        51,
                        65,
                        230,
                        183,
                        244,
                        229,
                        253,
                        70,
                        224,
                        250,
                        182,
                        228,
                        30,
                        163,
                        248,
                        187,
                        255,
                        106,
                        196,
                        175,
                        167,
                        119,
                        237,
                        114,
                        210,
                        151,
                        250,
                        36,
                        12,
                        28,
                        57,
                        63,
                        88,
                        241,
                        127,
                        86,
                        238,
                        241,
                        191,
                        144,
                        48,
                        204,
                        45,
                        60,
                        158,
                        233,
                        75,
                        200,
                        81,
                        27,
                        69,
                        223,
                        242,
                        245,
                        219,
                        45,
                        42,
                        217,
                        152,
                        198,
                        20,
                        239,
                        253,
                        188,
                        38,
                        114,
                        180,
                        208,
                        84,
                        230,
                        119,
                        142,
                        109,
                        236,
                        249,
                        212,
                        5,
                        83,
                        125,
                        220,
                        61,
                        255,
                        96,
                        103,
                        41,
                        197,
                        152,
                        115,
                        172,
                        186,
                        83,
                        127,
                        173,
                        42,
                        52,
                        130,
                        150,
                        216,
                        158,
                        246,
                        162,
                        36,
                        255,
                        179,
                        216,
                        127,
                        168,
                        147,
                        152,
                        17,
                        122,
                        209,
                        63,
                        228,
                        8,
                        72,
                        101,
                        22,
                        28,
                        32,
                        181,
                        247,
                        211,
[...]

PR #2436, opened 2019-01-20
Add hex serialization functions

svechinsky:
Addressing https://github.com/mimblewimble/grin/issues/2134
Comments about everything are welcome.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,31 @@
+---
+name: Bug report
+about: Create a report to help us improve Grin
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Run '...'
+2. Expect '....'
+3. Scroll down to '....'
+4. See error
+
+**Relevant Information**
+An excerpt from the logs or a detailed error message.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
+ - OS: [e.g. iOS]
+ - Version [e.g. 1.0.1]
+
+**Additional context**
+Add any other context about the problem here.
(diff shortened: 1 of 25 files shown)


jaspervdm:
Note that this will break sending/receiving transactions between upgraded and non-upgraded wallets

svechinsky:
hmmm overlooked this, will hopefully add checks tomorrow

yeastplume:
Just looking through this in some more detail, as I think I'd like this in place before merging https://github.com/mimblewimble/grin/pull/2441 since I'd like to use the same serialization features there as well.

Can I suggest moving all of the serialization methods into their own separate module rather than spreading them around, perhaps in core/src/libtx for now (until we decide the best place to put it). I think it will be useful to have all these methods to sanely serialise libsecp primitives somewhere common. We may even want to split this PR into 2, with one to add the serialisation module and another to turn them on in a backwards compatible way. That would also help me move on with #2441

svechinsky:
Ok, so:
1. Reorganize the serialization modules into a separate module.
2. Comment out the serde directives for activation at a later date

yeastplume:
Yes, sounds like a good plan, thanks!

svechinsky:
@yeastplume Just pinging for a review

antiochp:
I'm not to enthusiastic about merging in commented out code like this.

We should either agree to keep this PR open and track it against master (not a huge fan of this either) or push forward on getting this done in a way that preserves backward compatibility.

ignopeverell:
You already have custom serializers now, why not make them revert to the old representation on hex errors?

svechinsky:
The problem is not with deserializing binary/hex but rather with serializing it so that old clients can accept them.

yeastplume:
@svechinsky Right, we're going to have to take a decision at some point as to what to do about older wallets, obviously they won't be able to handle the new format and will choke rather gracelessly. Also, we should probably take the opportunity here to add a version ID to the slate to help us mitigate future problems, and ensure the wallet takes it into consideration (and keep a minimum version it can handle).

Unfortunately this has gone a little bit worm-canny. Can I suggest the following:

1) Remove the commented out directives, and just commit the serialization lib in this PR. (This will let me get on with #2441 as well)
2) Another PR for an optional Slate ID (and have to consider how this remains backwards compatible, given it would adds a field that doesn't exist in older clients)
3) Another PR to handle the integration of new slate + backwards compatibility concerns there

svechinsky:
Ok, will sort out today/tomorrow, will def do 1 today so it won't block #2441

yeastplume:
Okay, great thank you. Lib functions looks good as they are, so I'll merge.

PR #2437, opened 2019-01-20
Cleanup compiler warnings

sesam:
Clean away all compiler warnings, including those that have been showing up in pastes in Lobby / support.

I'm also trying to get the cargo-clippy warnings to become manageable. Currently clippy can't even finish its checking, but I'm slowly getting there.

When allowing:

- clippy::style (as in not showin warning for no-idiomatic style, which is otherwise a default)
- and some other things

then there are still a lot of details clippy complains about, but there are also some interesting nags worth to take a longer look t. And if there is some halfway there equilibrium, then periodical clippy checks could help keep code style follow best practice. Some clippy lints will likely also have to be permanently disabled (="allowed") in order to have clippy outputs stay useful over time.

Currently I'm using clippy like this:
`cargo clippy -- -A clippy::precedence -A clippy::match_bool -A clippy::cast_lossless -A clippy::redundant_field_names -A clippy::trivially_copy_pass_by_ref -A clippy:clone_on_copy -A clippy::style` and pipe stderr output to a file.

I think this might be worth merging already anyway, to get a small bit ahead.

Thoughts?

### DIFF
--- a/api/src/handlers/blocks_api.rs
+++ b/api/src/handlers/blocks_api.rs
@@ -140,7 +140,7 @@ impl Handler for BlockHandler {
 				return response(
 					StatusCode::BAD_REQUEST,
 					format!("failed to parse input: {}", e),
-				)
+				);
 			}
 			Ok(h) => h,
 		};
--- a/api/src/handlers/blocks_api.rs
+++ b/api/src/handlers/blocks_api.rs
@@ -68,7 +68,7 @@ impl HeaderHandler {
 
 impl Handler for HeaderHandler {
 	fn get(&self, req: Request<Body>) -> ResponseFuture {
-		let el = match req.uri().path().trim_right_matches("/").rsplit("/").next() {
+		let el = match req.uri().path().trim_right_matches('/').rsplit('/').next() {
 			None => return response(StatusCode::BAD_REQUEST, "invalid url"),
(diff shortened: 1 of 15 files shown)


api/src/handlers/blocks_api.rs
```diff
@@ -68,7 +68,7 @@ impl HeaderHandler {
 
 impl Handler for HeaderHandler {
 	fn get(&self, req: Request<Body>) -> ResponseFuture {
-		let el = match req.uri().path().trim_right_matches("/").rsplit("/").next() {
+		let el = match req.uri().path().trim_right_matches('/').rsplit('/').next() {
```
ignopeverell: Doesn't look like this has anything to do with compiler warnings?

core/src/consensus.rs
```diff
@@ -250,7 +250,7 @@ impl HeaderInfo {
 
 /// Move value linearly toward a goal
 pub fn damp(actual: u64, goal: u64, damp_factor: u64) -> u64 {
-	(1 * actual + (damp_factor - 1) * goal) / damp_factor
+	(actual + (damp_factor - 1) * goal) / damp_factor
```
ignopeverell: Neither does this.

sesam:
Closed in favor of #2462 and #2465

PR #2440, opened 2019-01-21
Contributing update from wiki

sesam:
Update /CONTRIBUTING.md to talk about:

- stable rustfmt rather than preview.
- The need for tests, and to remember running `cargo test --all`
- rustfmt and current grin best practice how it should be used

Closes https://github.com/mimblewimble/docs/issues/3


CONTRIBUTING.md
```diff
@@ -2,22 +2,39 @@
 
 Find an area you can help with and do it. Open source is about collaboration and open participation. Try to make your code look like what already exists and submit a pull request.
 
-The [list of issues](https://github.com/mimblewimble/grin/issues) is a good place to start, especially the ones tagged as "help wanted" (but don't let that stop you from looking at others). If you're looking for additional ideas, the code includes `TODO` comments for minor to major improvements. Grep is your friend.
+The [list of issues](https://github.com/mimblewimble/grin/issues) is a good place to start, especially the ones tagged as "help wanted" (but don't let that stop you from looking at others). If you're looking for additional ideas, the code includes `TODO` comments for minor to major improvements. Use _Search in files_ in your code editor, or `grep "TODO" -r src/`.
```
ignopeverell: This grep will only cover the binary `src`, which is a very small part of the overall codebase.

sesam: Good catch. This looks like the best pure grep incantation: grep "TODO" -r --exclude-dir=target --exclude-dir=.git .

PR #2441, opened 2019-01-21
Save slate participant messages in database

yeastplume:
Alternative to https://github.com/mimblewimble/grin/pull/2409. Note that this stores messages within a struct in the database, and the actual formatting (if you want in in JSON, for instance,) can be applied somewhere higher up in the pipeline.

Messages + keys + sig for an output are displayed when showing details of a particular transaction, e.g. `grin wallet txs -i 32`.

Shouldn't break backwards compat, as the new stored message field in the TxLogEntry is Optional. 

Messages are now limited to 256 (unicode) characters and will are truncated at the API level.

Todos:
- [x] Limit message length
- [x] Test backwards compat isn't broken
- [x] Check/fix serialization of message elements (or we're going to end up with long byte arrays on JSON output)


core/Cargo.toml
```diff
@@ -30,3 +30,6 @@ chrono = "0.4.4"
 
 grin_keychain = { path = "../keychain", version = "1.0.0" }
 grin_util = { path = "../util", version = "1.0.0" }
+
+[dev-dependencies]
```
yeastplume: Anyone have any major problems with this dev-dependency? Would rather avoid but seems best way to run serialization tests.

yeastplume:
Think this is ready to begin review now.

wallet/src/libwallet/internal/tx.rs
```diff
@@ -224,6 +225,26 @@ where
...
+	C: NodeClient,
+	K: Keychain,
+{
+	let tx_vec = updater::retrieve_txs(wallet, None, Some(slate.id), None, false)?;
+	if tx_vec.is_empty() {
+		return Err(ErrorKind::TransactionDoesntExist(slate.id.to_string()))?;
+	}
+	let mut batch = wallet.batch()?;
+	for tx in tx_vec.into_iter() {
+		let mut t = tx.clone();
```
hashmap: why do we need `t` here? we consume `tx_vec`, could we use `tx`?

wallet/src/libwallet/internal/tx.rs
```diff
@@ -224,6 +225,26 @@ where
...
+		t.messages = Some(slate.participant_messages());
+		batch.save_tx_log_entry(t.clone(), &t.parent_key_id.clone())?;
```
hashmap: do we need to clone `t` here?

wallet/src/libwallet/internal/tx.rs
```diff
@@ -224,6 +225,26 @@ where
...
+	}
+	batch.commit()?;
```
hashmap: I'd replace 2 lines with just `batch.commit();`

hashmap:
Actually enforcing string length using type system may not be that great idea after all, too many underlying levels must be changed

yeastplume:
Thanks for review, tightened up loop a bit. I'd started trying to enforce the message length a bit further down, but ran into issues as you say. So long as it's enforced at the API level (which is what most people will use, I think it should be okay).

alesito85:
@yeastplume is there a way to identify which message belongs to the sender? Or is it always the first one in the array?

PR #2442, opened 2019-01-21
fix owner_api docs: wrong port number

sesam:
Discussed here https://gitter.im/grin_community/Lobby?at=5c460cfaf780a1521f480831
Code only has 13420 and not 3420 anywhere, so this fix makes things consistent.

### DIFF
--- a/doc/api/api.md
+++ b/doc/api/api.md
@@ -17,9 +17,13 @@ To learn about what specific calls can be made read the [wallet foreign API doc]
 
 ### Wallet Owner API
 
-The wallet owner API is an endpoint to manage the user wallet: broadcast transaction, sign transaction, see the current balance... This REST API can be started with the `grin wallet owner_api` command and will listen on `localhost:13420` for --floonet or `localhost:3420` for the default mainnet.
+The wallet owner API is an endpoint to manage the user wallet: broadcast transaction, sign transaction, see the current balance... This REST API can be started with the `grin wallet owner_api` command and will listen on `localhost:3420`.
 
 __This endpoint must **never** be exposed to the outside world.__
 
 This endpoint requires, by default, Basic Authentication. The username is `grin` and the password can be found in the `.api_secret` file.
 To learn about what specific calls can be made read the [wallet owner API doc](wallet_owner_api.md).
+
+## Ports above 10000?
+
+All ports should be below 10000 when running with default settings on mainnet. If your grin owner_api is using the 13420 port but is on mainnet, then you're using an outdated version of grin.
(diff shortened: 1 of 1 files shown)


sesam:
See also #2389 
Maybe the correct fix would actually be to list both floo and main port numbers in every place they are listed? My text editor said there were only a few of this one, but there might be other ports that were debated and should also be updated and have `if is_floo` checks on them.

doc/api/api.md
```diff
@@ -17,6 +17,6 @@ To learn about what specific calls can be made read the [wallet foreign API doc]
 
 ### Wallet Owner API
 
-The wallet owner API is an endpoint to manage the user wallet: broadcast transaction, sign transaction, see the current balance... This REST API can be started with the `grin wallet owner_api` command and will listen on `localhost:3420`. This endpoint must **never** be exposed to the outside world.
+The wallet owner API is an endpoint to manage the user wallet: broadcast transaction, sign transaction, see the current balance... This REST API can be started with the `grin wallet owner_api` command and will listen on `localhost:13420`. This endpoint must **never** be exposed to the outside world.
```
kargakis: Seems the docs are using mainnet ports now?

kargakis: ah, this is actually an issue in grin - it should be fairly easy to fix the issue and avoid amending the docs?

kargakis: should be fixed in https://github.com/mimblewimble/grin/pull/2405

antiochp:
Related - #2405. This was an issue with grin itself.

sesam:
@antiochp I made it clear which ports are for what. Someone reading dos for the owner_api is likely doing testing and reasonably needs to know about both ports and what they're for. Good?

doc/api/api.md
```diff
@@ -17,6 +17,9 @@ To learn about what specific calls can be made read the [wallet foreign API doc]
 
 ### Wallet Owner API
 
-The wallet owner API is an endpoint to manage the user wallet: broadcast transaction, sign transaction, see the current balance... This REST API can be started with the `grin wallet owner_api` command and will listen on `localhost:3420`. This endpoint must **never** be exposed to the outside world.
+The wallet owner API is an endpoint to manage the user wallet: broadcast transaction, sign transaction, see the current balance... This REST API can be started with the `grin wallet owner_api` command and will listen on `localhost:13420` for --floonet or `localhost:3420` for --mainnet.
```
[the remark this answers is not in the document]
sesam: Thank you, fixed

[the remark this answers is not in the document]
sesam: And the reason for keeping the distinction here is because there might be people using almost-updated mainnet code releases that are using 13420 instead of 3420 as the default.

Do you think it's best to add a comment or "errata" about this also in the documentation? Something like: "If your grin owner_api is using the 13420 port but is on mainnet, then you're using an outdated version of grin. Please upgrade to get this and other bugfixes"

PR #2452, opened 2019-01-22
Update api.md to use correct port

eddiewang:
It appears that Grin uses 13420 for the owner_api port, not 3420. This is for both floonet and mainet.
`20190122 10:22:12.140 INFO grin_wallet::controller - Starting HTTP Owner API server at 127.0.0.1:13420.`


antiochp:
This should be `3420` in both the code and the docs.

Closing.

PR #2465, opened 2019-01-24
Catching clippy

sesam:
(This is the second, and extended, part of now superceded https://github.com/mimblewimble/grin/pull/2437)

Initial cleanups with the goal of making `cargo clippy` usable.

Currently I'm using clippy like this:
cargo clippy -- -A clippy::precedence -A clippy::match_bool -A clippy::cast_lossless -A clippy::redundant_field_names -A clippy::trivially_copy_pass_by_ref -A clippy:clone_on_copy -A clippy::style and pipe stderr output to a file.

With this command, I get a few interesting warnings. Also there might be more interesting warnings that clippy due to an error can't yet reveal, since clippy functions by extending the rustc compiler.

(The flag -A means Allow and makes clippy ignore a linting rule. The last -A clippy:style makes clippy ignore _all_ style linting rules, and a future goal of mine is to not need to be so drastic. There also exists a clippy::pedantic which I didn't yet dare to try.)

Cue debate which clippy complaints are relevant or not?
I think this might be worth merging to get a small bit ahead.

### DIFF
--- a/api/src/handlers/blocks_api.rs
+++ b/api/src/handlers/blocks_api.rs
@@ -60,8 +60,8 @@ impl HeaderHandler {
 	fn get_header_for_output(&self, commit_id: String) -> Result<BlockHeaderPrintable, Error> {
 		let oid = get_output(&self.chain, &commit_id)?.1;
 		match w(&self.chain).get_header_for_output(&oid) {
-			Ok(header) => return Ok(BlockHeaderPrintable::from_header(&header)),
-			Err(_) => return Err(ErrorKind::NotFound)?,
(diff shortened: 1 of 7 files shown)


kargakis:
Might be worth adding in travis for validation?

sesam:
@kargakis In time, it would be possible, but currently, clippy can not complete, and the path to get past all clippy lint rules is long. Also bikeshedding. So I'm heading that way, step by step, only beginning with what's likely easiest to agree on, with nearest goal to get clippy to compile so I can see all warnings, and pick some that might need more scrutiny and/or tests to make sure code does what we expect it to do.

PR #2466, opened 2019-01-24
Pool tx weight verification

antiochp:
So it turns out treating the txpool as "one big tx" for validation purposes meant we were verifying the max weight of the whole _transaction pool_ and checking it did not exceed `MAX_BLOCK_WEIGHT`...

This is clearly sub-optimal for a bunch of reasons.

Existing `verify_weight()` logic takes a `with_reward` bool flag to allow txs and blocks to both be weight verified. A block will have an additional output and kernel to cover the coinbase reward, so txs have to be slightly less weighty to account for this.

```
max tx weight + coinbase reward = max block weight
```

This PR introduces a `WeightVerificationType` enum and replaces the `with_reward` bool flag with the enum.

```
pub enum WeightVerificationType {
	/// Tx represents a tx (max weight).
	AsTransaction,
	/// Tx represents a block (max weight + additional coinbase reward).
	AsBlock,
	/// Tx represents the txpool (no max weight).
	AsPool,
}
```

So now we can robustly validate the contents of the txpool by treating it as "one big tx" after aggregation, but skipping the weight verification step.

This is all slightly confusing because we - 
* validate blocks (and the txpool) as if they were txs
* but then we verify weight against MAX_BLOCK_WEIGHT (as if they were blocks)

So in the context of weight we validate txs as if they were pretending to be blocks.
And blocks as if they were pretending to be txs pretending to be blocks.

We validate the txpool as if it was "one big tx", but for weight purposes we treat it as having no weight limit.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -603,7 +603,8 @@ impl TransactionBody {
 			.unwrap_or(0)
 	}
 
-	// Verify the body is not too big in terms of number of inputs|outputs|kernels.
+	/// Verify the body is not too big in terms of number of inputs|outputs|kernels.
+	/// Weight rules vary depending on the "weight type" (block or tx or pool).
 	fn verify_weight(&self, weight_type: WeightVerificationType) -> Result<(), Error> {
 		// If "tx" body then remember to reduce the max_block_weight for block requirements.
 		// If "block" body then verify weight based on full set of inputs|outputs|kernels.
@@ -616,6 +617,7 @@ impl TransactionBody {
 			WeightVerificationType::AsTransaction => 1,
 			WeightVerificationType::AsBlock => 0,
 			WeightVerificationType::AsPool => {
+				// We do not verify "tx as pool" weight so we are done here.
 				return Ok(());
 			}
 		};
@@ -624,7 +626,7 @@ impl TransactionBody {
 		// allowed in the body itself.
 		// A block is allowed to be slightly weightier than a tx to account for
 		// the additional coinbase reward output and kernel.
-		//
+		// i.e. We need to reserve space for coinbase reward when verifying tx weight.
 		// max_block = max_tx + coinbase
 		//
 		let tx_block_weight = TransactionBody::weight_as_block(
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -365,6 +365,22 @@ impl FixedLength for TxKernelEntry {
 		+ secp::constants::AGG_SIGNATURE_SIZE;
 }
 
+/// Enum of possible tx weight verification options -
+///
+/// * As "transaction" checks tx (as block) weight does not exceed max_block_weight.
+/// * As "block" same as above but allow for additional coinbase reward (1 output, 1 kernel).
+/// * As "pool" skips weight check (pool as one big tx, but without max weight).
+///
+#[derive(Clone, Copy)]
+pub enum WeightVerificationType {
+	/// Tx represents a tx (max weight).
+	AsTransaction,
+	/// Tx represents a block (max weight + additional coinbase reward).
+	AsBlock,
+	/// Tx represents the txpool (no max weight).
+	AsPool,
+}
+
 /// TransactionBody is a common abstraction for transaction and block
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub struct TransactionBody {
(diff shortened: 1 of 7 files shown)


pool/src/pool.rs
```diff
@@ -218,12 +239,18 @@ impl Pool {
 			txs.push(entry.tx.clone());
 
 			let tx = transaction::aggregate(txs)?;
-			tx.validate(self.verifier_cache.clone())?;
+
+			// TODO - Is this necessary? We validate_raw_tx below.
```
antiochp: TODO - Need to revisit this and cleanup.

pool/src/pool.rs
```diff
@@ -218,12 +239,18 @@ impl Pool {
...
+			// tx.validate(WeightVerificationType::AsPool, self.verifier_cache.clone())?;
+
 			tx
 		};
 
-		// Validate aggregated tx against a known chain state.
-		self.validate_raw_tx(&agg_tx, header)?;
+		// Validate aggregated tx (existing pool + new tx), ignoring tx weight limits.
+		// Validate against known chain state at the provided header.
+		self.validate_raw_tx(&agg_tx, header, WeightVerificationType::AsPool)?;
```
antiochp: This is the critical change here for the PR. 
When we add a new tx to the pool we want to aggregate the entire contents of the pool and check the resulting tx is still valid.
But we do _not_ want to verify that the weight here is below `MAX_BLOCK_WEIGHT`.

pool/src/transaction_pool.rs
```diff
@@ -145,8 +151,12 @@ impl TransactionPool {
 		self.is_acceptable(&tx, stem)?;
 
 		// Make sure the transaction is valid before anything else.
-		tx.validate(self.verifier_cache.clone())
-			.map_err(PoolError::InvalidTx)?;
+		// Validate tx accounting for max tx weight.
+		tx.validate(
```
antiochp: But when adding a tx to the pool we _also_ want to validate the tx "in isolation" and when we do this we do want to verify it is below the max weight limit.

So we're using tx validation for two different purposes here - 
* is the tx itself valid
* does it produce a valid aggregate tx when added to the txpool (might be _huge_)

yeastplume:
I think the approach makes a lot of sense (I was trying to find the issue where we were discussing this, over the possibility that someone could post a transaction with a fee and have no say over what other transactions they get lumped with).

Just thinking through consensus issues. We're relaxing requirements for the block tx, and adding a constraint to individual transactions that wasn't there before. So an earlier client might have rejected a huge individual tx via the block (throwing out other transactions with it), whereas a new client would throw it out at the individual TX level. I'd guess the worst that can happen here is a brief fork?

antiochp:
> We're relaxing requirements for the block tx, and adding a constraint to individual transactions that wasn't there before.

No - this does not change the tx or block validation logic. It just rewrites it to make it more flexible.

The only thing we're affecting is the "can I add this valid tx to the txpool" logic.
And this relaxes the rules - we now allow the txpool to accept more txs and to grow larger than `MAX_BLOCK_WEIGHT` (which was incorrect behavior).

core/src/core/transaction.rs
```diff
@@ -365,6 +365,22 @@ impl FixedLength for TxKernelEntry {
...
+/// * As "block" same as above but allow for additional coinbase reward (1 output, 1 kernel).
+/// * As "pool" skips weight check (pool as one big tx, but without max weight).
+///
+#[derive(Clone, Copy)]
+pub enum WeightVerificationType {
+	/// Tx represents a tx (max block weight, accounting for additional coinbase reward).
+	AsTransaction,
+	/// Tx represents a block (max block weight).
+	AsBlock,
+	/// Tx represents the txpool (no max weight).
```
ignopeverell: Perhaps a nit but as the pool isn't really a core entity, I would name this more along the lines of `Skip` or `Ignore`. We may have other validation sources that want that as well.

core/src/core/transaction.rs
```diff
@@ -365,6 +365,22 @@ impl FixedLength for TxKernelEntry {
...
 }
 
+/// Enum of possible tx weight verification options -
+///
+/// * As "transaction" checks tx (as block) weight does not exceed max_block_weight.
+/// * As "block" same as above but allow for additional coinbase reward (1 output, 1 kernel).
+/// * As "pool" skips weight check (pool as one big tx, but without max weight).
+///
+#[derive(Clone, Copy)]
+pub enum WeightVerificationType {
```
ignopeverell: Would just `Weighting` work?

pool/src/transaction_pool.rs
```diff
@@ -108,7 +108,13 @@ impl TransactionPool {
 			let txs = self.txpool.find_matching_transactions(entry.tx.kernels());
 			if !txs.is_empty() {
 				let tx = transaction::deaggregate(entry.tx, txs)?;
-				tx.validate(self.verifier_cache.clone())?;
+
+				// Validate this deaggregated tx "as tx", subject to regulat tx weight limits.
```
ignopeverell: Small type: regular

core/src/core/transaction.rs
```diff
@@ -587,11 +603,32 @@ impl TransactionBody {
...
+				// We do not verify "tx as pool" weight so we are done here.
+				return Ok(());
+			}
+		};
+
+		// Note: If we add a reserve then we are reducing the number of outputs and kernels
+		// allowed in the body itself.
+		// A block is allowed to be slightly weightier than a tx to account for
+		// the additional coinbase reward output and kernel.
+		// i.e. We need to reserve space for coinbase reward when verifying tx weight.
```
ignopeverell: In effect, we need to verify that a tx *can* be included in a block. Otherwise it's a rather useless tx.

core/src/core/transaction.rs
antiochp: `NoLimit` to make this explicit. `Ignore` is maybe ambiguous here, are we skipping the check or are we skipping this tx somehow?

antiochp:
Haven't merged this yet but its good to go.
Just wanted to check that we're good with this going into 1.0.1 before merging.

@ignopeverell

ignopeverell:
I'm all good with that.

Issue #2328, opened 2019-01-10
Owner API port should be configurable


alesito85:
Currently it only listens on 13420 without the possibility to change unless one changes source and compiles.

quentinlesceller:
You mean you'd like to be able to change the owner api from local 13420 to local another port?

yeastplume:
This is intentional, see https://github.com/mimblewimble/grin/pull/1867

quentinlesceller:
@yeastplume It seems like @alesito85 wanted to change *only* the API port and not the address? Which should be fine?

yeastplume:
Yes, sorry, I read too hastily.

alesito85:
As @quentinlesceller said, I'd like to configure just the port.

ignopeverell:
Fixed by #2405

PR #2475, opened 2019-01-27
wallet/owner_api: allow owner API port to be configurable

RaghavSood:
Adds an `owner_api_listen_port` to the grin-wallet.toml config to allow the port for the owner API to be changed on a per-wallet basis.

Related issues/PRs:

#2328, #2405

### DIFF
--- a/wallet/src/types.rs
+++ b/wallet/src/types.rs
@@ -41,6 +41,8 @@ pub struct WalletConfig {
 	pub api_listen_interface: String,
 	// The port this wallet will run on
 	pub api_listen_port: u16,
+	// The port this wallet's owner API will run on
+	pub owner_api_listen_port: u16,
 	/// Location of the secret for basic auth on the Owner API
 	pub api_secret_path: Option<String>,
 	/// Location of the node api secret for basic auth on the Grin API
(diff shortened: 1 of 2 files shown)


wallet/src/types.rs
```diff
@@ -72,6 +74,7 @@ impl Default for WalletConfig {
 			chain_type: Some(ChainTypes::Floonet),
```
kargakis: Unrelated to your changes but shouldn't this be `ChainTypes::Mainnet`?

RaghavSood:
I suspected that too but I wasn't sure if there was a reason for it so I left it as-is. I think there's a few floonet remnants floating around the code, and will likely do another pr to resolve those.

wallet/src/types.rs
ignopeverell: Looks like it should, yes.

sesam:
@ignopeverell Anything still blocking merging this?
It is a feature that many have asked for. Would be nice if we can ensure it also gets listed in the release notes.

garyyu:
>Mike Dallas @mcdallas 22:02
mimblewimble/grin#2475 needs to handle the missing value case otherwise everyone's client will break with the next release

>Gary Yu @garyyu 22:09
@mcdallas indeed! and need to add the default config file item.

Issue #2479, opened 2019-01-28
feat: set default floonet port automatically for wallet


sesam:
From #2475 so we don't forget.
 
We currently need to set floonet port manually, so to simplify floonet testing, we want to add a default port also for floonet.

Implement it similar to this:

> In the node global config we override mainnet global config with floonet specific defaults -
https://github.com/mimblewimble/grin/blob/92cbfa20a66e125f18d3fd7385783435de112a79/config/src/config.rs#L216-L218

PR #2484, opened 2019-01-28
default to sane value for missing owner_api_listen_port

### LINKED ISSUE #2482: owner_api_listen_port missing field in grin-wallet.toml
Related #2475.

A new config item was added in #2475 - 
```
owner_api_listen_port
```

We need to be able to safely handle the case where this is missing (i.e. all existing wallets).

Otherwise the wallet fails to startup successfully with a `missing field` error.

antiochp:
Resolves #2482.

Safely default to `3420` if `owner_api_listen_port` is not explicitly set in `grin-wallet.toml`

- [x] Tested `grin wallet info` with and without `owner_api_listen_port` set.
- [x] Tested `grin wallet init` generates a new grin-wallet.toml successfully (confirmed default is set correctly).

### DIFF
--- a/wallet/src/types.rs
+++ b/wallet/src/types.rs
@@ -42,7 +42,7 @@ pub struct WalletConfig {
 	// The port this wallet will run on
 	pub api_listen_port: u16,
 	// The port this wallet's owner API will run on
-	pub owner_api_listen_port: u16,
+	pub owner_api_listen_port: Option<u16>,
 	/// Location of the secret for basic auth on the Owner API
 	pub api_secret_path: Option<String>,
 	/// Location of the node api secret for basic auth on the Grin API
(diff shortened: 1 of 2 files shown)


wallet/src/types.rs
```diff
@@ -94,8 +94,19 @@ impl WalletConfig {
 		format!("{}:{}", self.api_listen_interface, self.api_listen_port)
 	}
 
+	// TODO - This should default to 3420 for mainnent and 13420 for floonet.
```
garyyu: should we leave this todo? :-)
perhaps just one line `if global::is_mainnet()`?

garyyu:
and please don't forget `config/src/comments.rs` for this new config item.

wallet/src/types.rs
antiochp: Unfortunately not that simple - we haven't read the config in yet... so we can't yet access any global config.

antiochp:
Generated config file looks like this - 

```
#path of TLS certificate file, self-signed certificates are not supported
#tls_certificate_file = ""
#private key for the TLS certificate
#tls_certificate_key = ""

#port for wallet listener
api_listen_port = 3415

#port for wallet owner api
owner_api_listen_port = 3420
```

wallet/src/types.rs
antiochp: Tracking the underlying mainnet/floonet issue here - #2479.

PR #2487, opened 2019-01-29
Rework "bucket transactions" logic (buckets are now weight limited)

### LINKED ISSUE #2456: Transaction aggregation in bucket_transactions is not weight aware
When preparing mineable transactions via the txpool we add transactions to buckets and then aggregate each bucket. The buckets allow us to group dependent transactions (0-conf txs) together to maximize aggregation in the candidate block.

But - We do not take max block weight into account when building these buckets.
If we take a transaction at the limit of max block weight and then bucket it along with a 0-conf transaction that spends one of its outputs then we end up producing a transaction that exceed the max block weight.

Proposal:

Every time we add a transaction to an existing bucket (it spends an output that exists in a bucket) we need to construct the aggregate tx and determine its weight. If this exceeds the limit then we cannot proceed with the bucketing, so we should start a new bucket.
This will ensure _all_ buckets are below the max block weight.

Then in the final step (as we do today) we then combine multiple buckets into the final set of aggregate transactions, ensuring we remain below the max block weight.

antiochp:
Resolves #2456.

This PR fixes 2 issues present in our current `prepare_mineable_transactions` implementation.
Specifically around how we "bucket" dependent transactions together.

1) It is currently possible to produce a bucket containing two transactions that together exceed MAX_BLOCK_WEIGHT. (See #2456).
2) If a tx depends on two txs in two _different_ buckets it is added to the last dependent bucket. The subsequent sorting of buckets and filtering based on weight can produce an invalid ordering of buckets and an invalid set of final txs (if one bucket is included and one excluded).

Proposed Implementation - 
* bucket_transactions now aggregates buckets into aggregated txs (each bucket is a single aggregated tx)
* these aggregated txs are "weight aware" and we only aggregate txs if this results ina valid aggregated tx
* if a tx depends on multiple buckets we simply reject it (and let the subsequent block pick it up)
  * The alternative here would be to introduce logic to "merge" multiple existing buckets but this is more complex and covers a rare edge-case. It is simpler to reject the tx and include it in the next block.

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -1232,6 +1232,9 @@ impl<'a> Extension<'a> {
 			self.verify_kernel_signatures(status)?;
 		}
 
+		// Log some stats for the output MMR.
+		self.dump_output_pmmr();
+
 		Ok((output_sum, kernel_sum))
 	}
 
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -1232,9 +1232,6 @@ impl<'a> Extension<'a> {
 			self.verify_kernel_signatures(status)?;
 		}
 
-		// Log some stats for the output MMR.
-		self.dump_output_pmmr();
-
 		Ok((output_sum, kernel_sum))
 	}
 
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -14,7 +14,6 @@
 
 //! Transactions
 
-use crate::consensus;
 use crate::core::hash::Hashed;
 use crate::core::verifier_cache::VerifierCache;
 use crate::core::{committed, Committed};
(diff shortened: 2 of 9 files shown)


antiochp:
@ignopeverell Reworked this a bit more.

Added a new `Weighting::AsLimitedTransaction` because miners can mine artificially small blocks based on their `mineable_max_weight` config.

Also added some test coverage for mining blocks at the max_block_weight limit from the txpool.

So "preparing mineable transactions" from the txpool now takes the miners max block weight config into account when bucketing and aggregating the txs.

This PR got bigger than I originally anticipated but there's no clear way of splitting this up in a meaningful way.

Thoughts?

ignopeverell:
Still looks good to me but looks like the new test is failing?

antiochp:
Some tests in the wallet cli are failing - I'm guessing its to do with the lower block weight limit in the test chain_type.
Let me try and get this fixed up before the weekend.

antiochp:
Final testing on usernet and things look good.
Going to merge.

PR #2491, opened 2019-01-30
Feature/slate version

svechinsky:
Add a version field for slate to symbolize slate format.
Doesn't do much yet but seems to be backwards compatible.

Part of addressing https://github.com/mimblewimble/grin/issues/2134


yeastplume:
When deserializing a slate, it would probably be useful to ensure the slate version is set to 0 (i.e. pre-versioning) to assist any further checks that need to happen.

PR #2502, opened 2019-01-31
Refactor tx lib to be more Sender-Recipient agnostic

yeastplume:
As a prelude to allowing transactions to be created by the recipient, some changes to make the underlying libs less dependent on who the sender is and who the receiver is. From an API perspective the process should become more granular, something like:

To create a slate as sender:

create_new_slate()
add_inputs()
send()
receive()
finalize()

And to create a slate as recipient
create_new_slate()
add_outputs()
send()

In either case the other party is simply calling the same `add_inputs` or `add_outputs` function depending on the order in which the transaction is being sent around.

This just rejigs internals for now without actually changing the API. While doing this, I realized that I think we want to do the grin/wallet split first, to make it easier to manage wallet API changes and additions without having to worry about grin versioning.

### DIFF
--- a/wallet/src/libwallet/internal/tx.rs
+++ b/wallet/src/libwallet/internal/tx.rs
@@ -107,7 +107,7 @@ pub fn add_output_to_slate<T: ?Sized, C, K>(
 	parent_key_id: &Identifier,
 	participant_id: usize,
 	message: Option<String>,
-) -> Result<impl FnOnce(&mut T) -> Result<(), Error>, Error>
+) -> Result<(Context, impl FnOnce(&mut T, &Transaction) -> Result<(), Error>), Error>
 where
 	T: WalletBackend<C, K>,
 	C: NodeClient,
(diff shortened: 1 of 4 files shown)


wallet/src/libwallet/internal/tx.rs
```diff
@@ -117,25 +86,68 @@ where
...
+		message,
+	)?;
+
+	// perform partial sig
+	let _ = slate.fill_round_2(
+		wallet.keychain(),
+		&context.sec_key,
+		&context.sec_nonce,
+		participant_id,
+	)?;
```
e-max: > And to create a slate as recipient
> create_new_slate()
> add_outputs()

Wouldn't `fill_round_2` return an error in this case, because it checks fee inside, which doesn't make sense for receipt on this stage?

yeastplume: Yes, however I haven't gone as far as actually implementing the recipient-initiated transactions yet. Should be easy enough to fix once I get there. (Most likely going to do the wallet split first)

PR #2506, opened 2019-01-31
Set max number of connections for peer db

hashmap:
Use max_peers value for that if set. By default LMDB limit is 126, which is not enough for less than 200 peers on my node.

### DIFF
--- a/servers/src/grin/server.rs
+++ b/servers/src/grin/server.rs
@@ -161,7 +161,11 @@ impl Server {
 			config.clone(),
 		));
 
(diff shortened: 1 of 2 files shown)


antiochp:
Is this happening even after disabling thread local storage for the connections?

hashmap:
@antiochp yes, starting closer to 180 peers (which is improvement anyway), it seems we process more than 126 requests in parallel, but with this fix I have 226 peers with no panics.

antiochp:
Related - https://twitter.com/armon/status/534867803426533376 (And I trust the terraform folks to know what they are talking about).

PR #2516, opened 2019-02-01
Add ability to compare selection strategies

i1skn:
Before tx creation user can estimate fee and locked amount
with different selection strategies by providing `-e` flag for
`wallet send` command.

### DIFF
--- a/src/bin/cmd/wallet_args.rs
+++ b/src/bin/cmd/wallet_args.rs
@@ -349,6 +349,9 @@ pub fn parse_send_args(args: &ArgMatches) -> Result<command::SendArgs, ParseErro
 	// selection_strategy
 	let selection_strategy = parse_required(args, "selection_strategy")?;
 
+	// estimate_selection_strategies
+	let estimate_selection_strategies = args.is_present("estimate_selection_strategies");
+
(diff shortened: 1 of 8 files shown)


src/bin/cmd/wallet_args.rs
```diff
@@ -562,7 +574,11 @@ pub fn wallet_command(
 		}
 		("send", Some(args)) => {
 			let a = arg_parse!(parse_send_args(&args));
-			command::send(inst_wallet(), a)
+			command::send(
+				inst_wallet(),
+				a,
+				wallet_config.dark_background_color_scheme.unwrap_or(true),
```
hashmap: Is it unrelated change?

kargakis:
Why not use `--dry-run` which is clearer than `-e`?

i1skn:
@kargakis short version `-d` is already taken for a `--destination`. `-e` states for `--estimate-selection` which, I think, also makes sense.

yeastplume:
Looks good for now. Might want to explore refactoring some of the underlying coin selection calls to eliminate a bit of redundancy, but can leave that until after the wallet is split off.

PR #2517, opened 2019-02-02
fix wallet --data-dir flag

e-max:
In the grin.yml  a flag which allows to pass a path to a directory with wallet data defined as 
```
        - data_dir:
            help: Directory in which to store wallet files
            short: dd
            long: data_dir
            takes_value: true
```
However in code we expected this flag to be named `dir`.


casey:
Thank you for the fix! ^_^

PR #2526, opened 2019-02-04
[docs] Add switch commitment documentation

johnzweng:
## This PR adds documentation about switch commitments and how they are used in grin.

Please feel free to change, correct or reword the text. I am a programmer with some cryptographic knowledge currently working through grin to understand the details, but I am not a cryptographer. So please bare with me (and correct me) if I got something wrong.

It also changes a comment in `core/src/core/transaction.rs` and two md-files where the switch commitment hash was still referenced (although it was already removed in https://github.com/mimblewimble/grin/pull/2157).

### DIFF
--- a/doc/switch_commitment.md
+++ b/doc/switch_commitment.md
@@ -1,0 +1,293 @@
+# Introduction to Switch Commitments
+
+## General introduction
+
+In cryptography a _Commitment_ (or _commitment scheme_) refers to a concept which can be imagined
+like a box with a lock. You can put something into the box (for example a piece of a paper with a
+secret number written on it), lock it and give it to another person (or the public).
+
+The other person doesn't know yet what's the secret number in the box, but if you decide to publish
+your secret number later in time and want to prove that this really is the secret which you came
+up with in the first place (and not a different one) you can prove this simply by giving the
+key of the box to the other person.
+
+They can unlock the box, compare the secret within the box with the secret you just published
+and can be sure that you didn't change your secret since you locked it. You "**commited**"
+to the secret number beforehand, meaning you cannot change it between the time of
+commitment and time of revealing.
+
+
+## Examples
+
+### Hash Commitment
+
+A simple commitment scheme can be realized with a cryptographic hash function. For example: Alice and Bob
+want to play _"Guess my number"_ and Alice comes up with with her really secret number `29` which
+Bob has to guess in the game, then before the game starts, Alice calculates:
+
+    hash( 29 + r )
+
+an publishes the result to Bob. The `r` is a randomly chosen  _Blinding Factor_ which is
+needed because otherwise Bob could just try hashing all the possible numbers for the game and
+compare the hashes.
+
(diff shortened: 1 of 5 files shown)


johnzweng:
Update: added another commit with some minor formatting changes and a little rewording (points vs curves) in `intro.md`

doc/switch_commitment.md
```diff
@@ -0,0 +1,211 @@
...
+
+For now grin uses and verifies Pedersen Commitments, like other cryptocurrencies do as well, 
+with the only difference that the random Blinding factor `r` is choosen by some special means.
+
+If some time in the future the bindningness of Pedersen Commitments might no longer be assumed
+secure then grin could simply come up with a Soft-Fork which adds additional 
+validation rules which then require that also the computation of `r` must be validated 
+(i.e. the ElGamal commitment must also be revealed) at the additional costs that the 
+range proofs for ElGamal are more costly (in size and computation) and that the hidingness 
+property is no longer perfectly hiding, but only computational hiding. But for a crypto currency
```
ignopeverell: This last point is mistaken I think. And it might be good to clarify in a section above as well. We could change our commitment scheme to be already perfectly binding and computationally hiding. But that would be bad. The argument is that a privacy-oriented chain needs to stay private *forever*. Because one could go back to the history of the chain and de-anonymize all transactions if the commitment scheme was broken.

On the other hand, the commitment scheme only needs to be perfectly binding at the time the scheme is broken. By using switch commitments, in case of disaster (the intent being still that nothing is ever broken), at least people get a choice: they can compromise the privacy on their *last* UTXOs, or just leave the money alone and lose it. There are many cases where a privacy leak is much more dangerous to one's life than some cryptocurrency.

johnzweng: Hi Ignotus! Thanks for your feedback! 🙂 

I see, I had a wrong understanding of some points here: In my naive thinking I didn't realise that activating the switch commitment validation wouldn't immediately affect all users, but instead give each single user the privilege of having a choice.

So not the coin (or its developers) dictates which of the two mutually exclusive properties (perfectly bindingness or perfectly hidingness) should be prioritised but each user itself decides (in the unfortunate case if the scheme really gets broken).

I will try to reword the text (especially the conclusion) and try to include these considerations.

johnzweng: Hi, I reworked parts of the document and tried to give a better understanding of the reasoning behind in 1c88d73. 
Please have a look, when you have some time and let me know if I shall change something. 🙂

johnzweng:
I still was not happy with the wording of the conclusion and pushed another commit with a small update. I didn't want to force push, for easier readability here.

If you decide to merge this PR I will squash the commits and force push here again before you merge them back.

johnzweng:
Hi @ignopeverell, I've squashed the several edits of the switch commitment document together into one commit.

Please take a final look if you agree with the content.

johnzweng:
Thank you for merging. 🙂 

Also thank you for valuable feedback which helped me a lot to understand better which contributions switch commitments add to the system.

PR #2528, opened 2019-02-04
Fix handling of IO WouldBlock error in Stratum server

### LINKED ISSUE #2524: In some cases Stratum server reads part of the message and ignores it
```
20190204 13:28:07.478 WARN grin_servers::mining::stratumserver - (Server ID: 0) Failed to parse JSONRpc: JSON error - [106, 115, 111, 110, 114, 112, 99, 34, 58, 34, 50, 46, 48, 34, 125, 10] 20190204 13:28:07.478 WARN grin_servers::mining::stratumserver - (Server ID: 0) Dropping worker: 6120
```
At this moment server reads `jsonrpc " : " 2 . 0 " }`. It drops the first part of the message here  https://github.com/mimblewimble/grin/blob/master/servers/src/mining/stratumserver.rs#L179
if read operation is about to block.

hashmap:
If WouldBlock error happens Stratum server drops part of a message to
read or write. This PR introduces a worker's buffer to store partially
read message which will be completed next time. For write the existing
util function is used.
Fixes #2524

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -152,6 +152,7 @@ pub struct Worker {
 	stream: BufStream<TcpStream>,
 	error: bool,
 	authenticated: bool,
+	buffer: String,
 }
 
 impl Worker {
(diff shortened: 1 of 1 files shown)


hendi:
@hashmap 
Blade and I can't actually find any code in the fix that handles a partial read or completes the next time
It just seems to add an extra buffer

Could you explain it a bit please? :-)

hashmap:
@hendi that’s the purpose of the buffer. It’s a long lived object, if we read the entire line we clear the buffer and return content. If we read a part of the line and about to block we exit and read the rest to the same buffer on the next round. Does it make sense?

hashmap:
Worth mentioning that read_line appends to the provided buffer

PR #2533, opened 2019-02-05
Move slate code from core/libtx to wallet/libwallet

yeastplume:
In anticipation of splitting the wallet off from grin, move Slate code from core to wallet crate. The functionality remaining in 'core/libtx' should be things that are unlikely to change much without consensus changes, while the Slate will need more flexibility.


svechinsky:
mmmm I'm almost done with working on backwards compatible serailzation, what do you think is the best way to combine the changes?

yeastplume:
This is just moving stuff around so going to merge.

yeastplume:
Sorry didn’t see that comment before merging... unfortunately I stuck you with the merge now, but hopefully shouldn’t be too bad

PR #2534, opened 2019-02-05
Feature/slate serialization

svechinsky:
This change enables hex serialization for slate objects in a backwards compatible way by introducing enum based versioning for the slate.
This was the nicest way I could find/think of to do this using serde.

It will enable an upgraded client to receive grin from an old client but not the other way around (this would require passing another argument down to the send function).

~~Probably shouldn't be merged until after https://github.com/mimblewimble/grin/pull/2533 is merged, my branch is updated and some more testing is done.
If anyone wants to help with testing it would be lovely since I'm a bit swamped with outside stuff atm.
I'll try and get to it as soon as I can, have never used keybase so would especially appreciate help here.~~

Testing is done, ready for CR and merge.

- [x] File transactions
- [x] Http transactions
- [x] Keybase transactions


yeastplume:
Looks good so far! Let me know when it's ready for a full review and I'll look in more detail as well as perform some testing.

svechinsky:
I'm really getting quite swamped with outside work, if anyone can help a bit with testing, especially the keybase part would be a huge help

mcdallas:
Deserialization via keybase fails, you need to change:

https://github.com/mimblewimble/grin/blob/200f87b8bcf111a64bba43c7b3c0fde4522ab786/wallet/src/adapters/keybase.rs#L257

also line 348 in the same file

mcdallas:
Sending from old to new doesn't work either, the new node gets the slate, responds but the old node fails to deserialize and doesnt finalize the transaction

svechinsky:
Thanks Mike!!!

Will look into it asap, probably Tuesday but maybe tomorrow.

mcdallas:
@svechinsky I sent a fix to your branch.

This PR breaks receiving grins for anyone who doesn't upgrade. Should we save it for the hard fork ?

svechinsky:
I think the plan is to enable upgraded nodes to receive grins from non-upgraded nodes.
If someone wants to receive grins they can upgrade.

Thanks for the keybase fix!!!

svechinsky:
Ok @yeastplume, tested for http should be ready now

yeastplume:
Sorry, merge conflict introduced.

I think this looks great, but I'm a little bit unsure as to how we should go about merging this. (I want to get this in one way or another so I can start splitting the wallet out).

Either we:

* Merge now and tell anyone running wallets < 1.0.2 to upgrade
* Merge into 1.1 branch

I think given this is non-consensus breaking and there will be some upgrade pain anyhow due to the upcoming wallet split, I'm not against merging this into master as is. Thoughts @antiochp @ignopeverell ?

ignopeverell:
To keep sane versioning I think we should introduce that with `1.1.0` and then point out in release notes and various configuration that wallet and server should be upgraded at the same time.

ignopeverell:
Targeting 1.1.0 introduced a conflict, sorry about that. Also not really a WIP anymore, right?

svechinsky:
Fixed both :)

garyyu:
@svechinsky 
Did you use `rebase` on your branch?  it has imported several other PRs into this PR.
Normally I use `merge` instead of `rebase`, to avoid this problem. 

And unfortunately, each time when I faced this problem, I have to create another new branch (from your target branch, in your case, it could be `milestone/1.1.0`), and cherry-pick the real own commits into it and submit a new PR.  Perhaps others can propose a better solution to fix this.

svechinsky:
I fetched the new branch (milestone/1.1.0) and merged (with merge not rebase) it into mine.
I'm not sure what went wrong.

svechinsky:
Oh I understand what happened. It also wants to merge all the commits since 1.1.0 was seperated from master.
I'm not sure if this is such a bad thing but if it is then I'll cherry pick as @garyyu suggested a bit later today.

yeastplume:
I don't think this is a bad thing, the other commits that have been picked up are stuff that's needed anyhow, windows support and http repost.

I'm going to merge this as is then sort out any issues manually.

Issue #2525, opened 2019-02-04
(GrinWin) - Windows 10 Support Meta-Issue


yeastplume:
This is a meta-issue to keep track of the tasks required to release a native Windows 10 x64 build. I've managed to get GrinWin (node with test-miner and wallet) mostly working, and there really shouldn't be too much standing in the way of a proper release (bar some testing and time in the field). 

Some notes on building/env (nothing special here):

* Use windows rustup msi to install
* Use MSVC toolchain as advocated by the rustup installer (download and install the VC2017 installer package)
* Clang dlls for windows need to be installed: http://releases.llvm.org/download.html
* ~~Disable TUI for now in `grin-server.toml` when running~~

Known Tasks: (More may come out as we get these done)

- [x] Update `croaring-rs` to 0.3.8. This is currently hardwired to 0.3 in the code, (check if there's a reason for this?))
- [x] Remove `daemonize` or ensure it's only compiled on linux/macos builds. Remove/make conditional the `grin wallet start` daemonize command
- [x] LMDB on windows allocates ALL requested space on DB initialization, which is currently set to 500GB in Grin (intended as a maximum). This should check platform before doing this, and on windows it should start with a small file and reallocate as necessary. Apparently monero already does this, so we have a reference. Edit: PR for this working, but still a few discussion points: https://github.com/mimblewimble/grin/pull/2605 Edit: Done
- [x] What may have been thought to be memmap issues are actually windows-specific problems with file permissions (mostly more aggressive windows file locking). I've fixed several issues around this but they'll of course need to be tested on linux as well. I'll submit a PR for this shortly.
- [x] txhashset unzip during fast sync fails to unzip for some reason, (Edit: was fixed, but appears to be an issue again after 1.0.2 (which was merged into 1.1.0 as well). Edit: fixed again #2641 
- [ ] Check txhashset is properly zipped and sent to peers that request it
- [x] The TUI backend needs to be changed (via rust features) to use pdcurses backend to run on windows. This needs to be a dynamic selection based on build platform.
- [x] The TUI appears when selected in `grin-miner.toml`, but appears quite small and freezes instantly 
[...]

PR #2535, opened 2019-02-06
Windows Compatibility Fixes #1

yeastplume:
Meta Issue: #2525 

After this PR, Grin should compile, run and sync on windows (with remaining problems, see #2525 for instructions and remaining issues). In particular, this:

* Updates `croaring-rs` from 0.3 to 0.3.8
* Remove `daemonize` and `grin server stop|start` commands
* Update inclusion of `cursive` to be platform conditional, change terminal backend for Windows to pancurses
* Updates to PMMR (particularly file rewind) to allow for explicit dropping of file handles. Required due to stricter locking on windows
* Modifications to store and wallet tests (which also need to explicitly drop file handles in many cases)
* Temporarily reduce initial LMDB file storage on windows (more complete solution needed, see #2525)
* Add functions to explicitly drop underlying file handles in TxHashSet. Not doing so was preventing fast-sync archives from unzipping.

These changes definitely need to be reviewed and tested to ensure they don't break anything on linux


store/src/pmmr.rs
```diff
@@ -53,8 +53,8 @@ pub const PMMR_FILES: [&str; 4] = [
 pub struct PMMRBackend<T: PMMRable> {
 	data_dir: PathBuf,
 	prunable: bool,
-	hash_file: DataFile<Hash>,
-	data_file: DataFile<T::E>,
+	hash_file: Option<DataFile<Hash>>,
+	data_file: Option<DataFile<T::E>>,
```
yeastplume: Options so they can be dropped before renaming operations

store/src/types.rs
```diff
@@ -193,21 +193,37 @@ impl AppendOnlyFile {
...
+					.create(true)
+					.write(true)
+					.open(&self.path)?;
+				file.set_len(self.buffer_start as u64)?;
+			}
+			let file = OpenOptions::new()
+				.read(true)
+				.create(true)
+				.append(true)
+				.open(&self.path)?;
```
yeastplume: File permissions for `set_len` on windows need to be as set here or it fails. Hence the drop and reopen with original permissions after the `set_len`.

store/src/lmdb.rs
```diff
@@ -73,11 +73,19 @@ pub fn new_named_env(path: String, name: String, max_readers: Option<u32>) -> lm
...
-		});
+
+	#[cfg(not(target_os = "windows"))]
+	env_builder.set_mapsize(5_368_709_120).unwrap_or_else(|e| {
+		panic!("Unable to allocate LMDB space: {:?}", e);
+	});
+	//TODO: This is temporary to support (beta) windows support
+	//Windows allocates the entire file at once, so this needs to
+	//be changed to allocate as little as possible and increase as needed
+	#[cfg(target_os = "windows")]
```
i1skn: Can we add `cfg!(target_os = "ios")` here as well?

yeastplume: I'd rather just keep this PR to windows issues for now, and this is just temporary. The real fix shouldn't involve OS specific code, but it should allocate the file minimally and increase it as necessary.

yeastplume:
@ignopeverell Thanks, cleaned that up and encapsulated the file rename/replace. However, needs slightly more review given that I had to add functionality to drop underlying PMMR files from the txhashset during fast sync archive unzip. Marked above.

chain/src/chain.rs
```diff
@@ -857,6 +857,14 @@ impl Chain {
 		}
 
 		let header = self.get_block_header(&h)?;
+
+		{
+			let mut txhashset_ref = self.txhashset.write();
+			// Drop file handles in underlying txhashset
+			txhashset_ref.release_backend_files();
```
yeastplume: Seems to work, would just like a bit of reassurance that this isn't going to have any odd side effects.

store/src/lmdb.rs
i1skn: ok, got it! let me know if I can help somehow with this upcoming fix!

yeastplume:
Also, removed Daemonize altogether, think I agree it's far easier to use external commands to run something as a service in most cases

chain/src/chain.rs
ignopeverell: Not that I can think of, seems fair.

PR #2536, opened 2019-02-06
refactor: use less memory in verifier_cache by storing zero-sized type instead of bool

JeremyRubin:
Because the presence of the entry is sufficient to show that it verified true, we no longer need to store a bool type.

A node in the LRU Cache's Map is 2 pointers, the key, and then the value. Via padding this works out to cost 4 bytes for the bool. This represents a total memory savings of 0.4 MB, which is not terribly much, but nice to do.

This PR also removes the use of `.unwrap_or(&mut false)` when checking for entries.


antiochp:
\o/

PR #2537, opened 2019-02-06
Allow to have more than min_peers/2 outbound peers

hashmap:
Default value is 8, so we allow only 4 outbound peers, unfortunately if a node is behind NAT it can't get inbound peers and it stucks with 4 peers as maximum.
This PR allows a number of outbound peers to be between min_peers and max_peers/2. Node behind NAT will eventually have just min_peers number of peers.

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -513,7 +513,13 @@ impl Peers {
 	}
 
 	pub fn enough_peers(&self) -> bool {
-		self.connected_peers().len() >= self.config.peer_min_preferred_count() as usize
+		self.peer_count() >= self.config.peer_min_preferred_count()
+	}
+
+	/// We have enough peers, both total connected and outbound connected
+	pub fn healthy_peers_mix(&self) -> bool {
+		self.enough_peers()
+			&& self.peer_outbound_count() >= self.config.peer_min_preferred_count() / 2
 	}
 
 	/// Removes those peers that seem to have expired
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -186,10 +186,7 @@ fn monitor_peers(
 	// maintenance step first, clean up p2p server peers
 	peers.clean_peers(config.peer_max_count() as usize);
 
-	// We have enough peers, both total connected and outbound connected so we are good.
-	if peers.peer_count() >= config.peer_min_preferred_count()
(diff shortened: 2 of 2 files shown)


hashmap:
Another option is to keep the second condition as `peers.peer_outbound_count() >= p2p.config.peer_min_preferred_count() / 2`, as I said a node behind NAT will eventually have min_peers anyway, but in this case we can save a few slots for inbound peers of a node with public ip.

antiochp:
We do something similar to what you propose in `monitor_peers` - 

https://github.com/mimblewimble/grin/blob/1f7ea4930e43d19efc0a45a2f4b71b400036188e/servers/src/grin/seed.rs#L189-L194

But here we keep the "total peer count" below `peer_min_preferred_count` and the "total outbound peer count" below `peer_min_preferred_count / 2`.

I wonder if we should look at both "total peer count" and "total outbound peer count" in `listen_for_addrs`?

i.e. We go try and connect to more outbound peers if - 
* we have less than `peer_min_preferred_count / 2` _outbound_ peers
* __or__, we have less than `peer_min_preferred_count` _total_ peers

That way we handle nodes behind NAT by just looking at total peers (we will never have inbound peers)
but for public nodes we default to looking at outbound peer counts.

I think it may make sense to keep the `listen_for_addrs` and `monitor_peers` peer counting logic consistent.

antiochp:
Good catch by the way - I don't think the NAT case was fully considered here when we split out default 8 slots into inbound/outbound.

hashmap:
I tested the latest version of this PR - after 20 mins a node behind NAT oscillates around 8 peers in total (all outbound), a node with a public IP has 4 outbounds (55 in the first seconds) and 100+ inbounds.

hashmap:
Ah, I pushed to the origin by mistake, removed the branch but Travis noticed it and now tries to build this removed branch too, please ignore `push` build

p2p/src/peers.rs
```diff
@@ -513,7 +513,13 @@ impl Peers {
 	}
 
 	pub fn enough_peers(&self) -> bool {
-		self.connected_peers().len() >= self.config.peer_min_preferred_count() as usize
+		self.peer_count() >= self.config.peer_min_preferred_count()
+	}
+
+	/// We have enough peers, both total connected and outbound connected
+	pub fn healthy_peers_mix(&self) -> bool {
```
antiochp: Nice.

PR #2539, opened 2019-02-07
fix: false positive wallet tests

i1skn:
Two tests in the wallet were actually failing, we were just ignoring them. This fixes it.

### DIFF
--- a/wallet/tests/restore.rs
+++ b/wallet/tests/restore.rs
@@ -366,10 +366,10 @@ fn perform_restore(test_dir: &str) -> Result<(), libwallet::Error> {
 fn wallet_restore() {
 	let test_dir = "test_output/wallet_restore";
 	if let Err(e) = setup_restore(test_dir) {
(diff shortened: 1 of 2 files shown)


wallet/tests/restore.rs
```diff
@@ -366,10 +366,10 @@ fn perform_restore(test_dir: &str) -> Result<(), libwallet::Error> {
 fn wallet_restore() {
 	let test_dir = "test_output/wallet_restore";
 	if let Err(e) = setup_restore(test_dir) {
```
hashmap: Why not `setup_restore(test_dir).expect("Libwallet Error")`?  It is shorter and does almost the same, panic will print out the stacktrace. Also below

hashmap:
Hm, please ignore my comment, I see that this error handling pattern is common, perhaps we can fix it in a separate pr (and keep backtrace if we like it)

yeastplume:
Ahh, good catch, the position of finalize would have changed recently. Thanks!

PR #2541, opened 2019-02-08
add content disposition headers to node responses

rentenmark:
Content disposition headers were not being set in node responses. This should fix that.

### DIFF
--- a/wallet/src/controller.rs
+++ b/wallet/src/controller.rs
@@ -309,7 +309,7 @@ where
 				"node_height" => json_response(&self.node_height(req, api)?),
 				"retrieve_txs" => json_response(&self.retrieve_txs(req, api)?),
 				"retrieve_stored_tx" => json_response(&self.retrieve_stored_tx(req, api)?),
-				_ => response(StatusCode::BAD_REQUEST, ""),
+				_ => error_response_with_description(StatusCode::BAD_REQUEST, "unknown url"),
 			},
 		)
 	}
@@ -487,11 +487,11 @@ where
 			),
 			"cancel_tx" => Box::new(
 				self.cancel_tx(req, api)
-					.and_then(|_| ok(response(StatusCode::OK, ""))),
+					.and_then(|_| ok(response(StatusCode::OK, "{}"))),
 			),
 			"post_tx" => Box::new(
 				self.post_tx(req, api)
-					.and_then(|_| ok(response(StatusCode::OK, ""))),
+					.and_then(|_| ok(response(StatusCode::OK, "{}"))),
 			),
 			_ => Box::new(err(ErrorKind::GenericError(
 				"Unknown error handling post request".to_owned(),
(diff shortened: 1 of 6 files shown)


wallet/src/controller.rs
```diff
@@ -660,35 +660,38 @@ where
...
 	Response::builder()
 		.status(StatusCode::INTERNAL_SERVER_ERROR)
-		.header("access-control-allow-origin", "*")
+		.header(hyper::header::ACCESS_CONTROL_ALLOW_ORIGIN, "*")
 		.header(
-			"access-control-allow-headers",
+			hyper::header::ACCESS_CONTROL_ALLOW_HEADERS,
 			"Content-Type, Authorization",
 		)
+		.header(hyper::header::CONTENT_TYPE, "application/json")
```
antiochp: Is this safe to do? I suspect the error itself is not being serialized out as json in the body.

wallet/src/controller.rs
```diff
@@ -660,35 +660,38 @@ where
...
 	Response::builder()
 		.status(status)
-		.header("access-control-allow-origin", "*")
+		.header(hyper::header::ACCESS_CONTROL_ALLOW_ORIGIN, "*")
 		.header(
-			"access-control-allow-headers",
+			hyper::header::ACCESS_CONTROL_ALLOW_HEADERS,
 			"Content-Type, Authorization",
 		)
+		.header(hyper::header::CONTENT_TYPE, "application/json")
```
mably: Looks like you can even do something like:

```
.header(hyper::header::ContentType::json())
```
Implies UTF-8 encoding though.

Source: https://hyper.rs/hyper/0.8.0/hyper/header/struct.ContentType.html

rentenmark: I was looking for this and couldn't find it, thanks!

wallet/src/controller.rs
rentenmark: Yes we noticed some content coming back that was just text, that must be from here. I'll wrap it up and see what you think

rentenmark: This looks like more of a systemic issue. It looks like the API returns many (all?) errors as text without JSON wrapping. I can tackle this but it would take a bit more time. Perhaps we assume if the status code is OK then the content type must be JSON?

rentenmark:
This is a huge change that would break a ton of clients. Fixing v1 api is probably not worthwhile. I understand v2 is underway, we are looking forward to it. So, I am basically throwing in the towel on this.

For now, it would still be useful, I think, to make sure all successful API responses return content type json.

What do you guys think?

wallet/src/controller.rs
antiochp: That's probably correct yes. I'm not sure we promise _anything_ if the response is not an OK (and that's not a good thing).

rentenmark:
closed in favor of #2545

garyyu:
@rentenmark 
Indeed it's a systematic issue that almost all error cases of Grin APIs return errors as text without JSON wrapping, and without `Content-Type` header.

And it make sense to take care of this on API V2, since the change could make the clients incompatible. We'd better to have an open issue there to remind us this task.

And thanks for catching this:-)

PR #2543, opened 2019-02-08
Allow to peers behind NAT to get up to preferred_max connections

hashmap:
If peer has only outbound connections it's most likely behind NAT and we should not stop it from getting more outbound connections.

With that change a node with a public IP eventually still has 4 outbounds peers, but node behind the NAT jumps higher than 8 from time to time.

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -186,7 +186,7 @@ fn monitor_peers(
 	// maintenance step first, clean up p2p server peers
 	peers.clean_peers(config.peer_max_count() as usize);
 
-	if peers.peer_count() > peers.peer_outbound_count() && peers.healthy_peers_mix() {
+	if peers.healthy_peers_mix() {
 		return;
 	}
 
@@ -315,7 +315,7 @@ fn listen_for_addrs(
 	let addrs: Vec<SocketAddr> = rx.try_iter().collect();
 
 	// If we have a healthy number of outbound peers then we are done here.
-	if peers.healthy_peers_mix() {
+	if peers.peer_count() > peers.peer_outbound_count() && peers.healthy_peers_mix() {
 		return;
 	}
 
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -186,7 +186,7 @@ fn monitor_peers(
 	// maintenance step first, clean up p2p server peers
 	peers.clean_peers(config.peer_max_count() as usize);
 
-	if peers.healthy_peers_mix() {
+	if peers.peer_count() > peers.peer_outbound_count() && peers.healthy_peers_mix() {
 		return;
 	}
 
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -518,12 +518,8 @@ impl Peers {
 
 	/// We have enough peers, both total connected and outbound connected
 	pub fn healthy_peers_mix(&self) -> bool {
-		let peer_count = self.peer_count();
-		let outbound_count = self.peer_outbound_count();
-
-		peer_count > outbound_count
-			&& peer_count >= self.config.peer_min_preferred_count()
-			&& outbound_count >= self.config.peer_min_preferred_count() / 2
+		self.enough_peers()
+			&& self.peer_outbound_count() >= self.config.peer_min_preferred_count() / 2
 	}
 
 	/// Removes those peers that seem to have expired
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -518,8 +518,12 @@ impl Peers {
 
 	/// We have enough peers, both total connected and outbound connected
 	pub fn healthy_peers_mix(&self) -> bool {
-		self.enough_peers()
-			&& self.peer_outbound_count() >= self.config.peer_min_preferred_count() / 2
+		let peer_count = self.peer_count();
+		let outbound_count = self.peer_outbound_count();
+
+		peer_count > outbound_count
+			&& peer_count >= self.config.peer_min_preferred_count()
+			&& outbound_count >= self.config.peer_min_preferred_count() / 2
 	}
 
 	/// Removes those peers that seem to have expired


ignopeverell:
Doesn't that mean that NAT'd peers are always going to be polling others for more peers?

hashmap:
We still have `max_preferred_peers` limit. As I understand an additional check was added to limit number of outbound peers, it's why stuck with 4 outbounds peers by default no matter what. In my previous pr I increased this limit to `min_preferred_peers`, with this fix a node with a public IP will stop getting more outbound peers (and eventually will have 4), while natted node will nicely oscillate between max and min pref peers. Also it's more semantically correct to not call a case when we have only outbound peers `healthy_mix`

ignopeverell:
Why the DNM change? Also I agree this behaves more correctly than previously, but it does look like the current code, with this merged, will keep nagging other peers for more?

hashmap:
I had a glitch on my node, but I see now that it was unrelated to this change. The current code is too modest when runs behind NAT, it doesn't get more than min_peers limit just because we encourage it to get more inbound peers, which is against its nature.

ignopeverell:
We may be slightly talking past each other. I tried this PR and it does the job. But even though I have 17 peers now (more than my preferred count), my node keeps asking other peers for their peer list:

```
20190213 04:53:39.252 WARN grin_servers::grin::seed - monitor_peers: 0.0.0.0:3414 ask 145.14.158.174:3414 for more peers
...
20190213 04:53:59.281 WARN grin_servers::grin::seed - monitor_peers: 0.0.0.0:3414 ask 145.14.158.174:3414 for more peers
...
20190213 04:54:19.315 WARN grin_servers::grin::seed - monitor_peers: 0.0.0.0:3414 ask 145.14.158.174:3414 for more peers
...
```

And that for every single peer I'm connected to. This is due to `monitor_peers` never being short-circuited now on NAT'd peers.

hashmap:
@ignopeverell you are right. I introduced a wrong abstraction in my previous PR by extracting a duplicate code into a function. However in some cases code duplication is just coincidence and may be better than a wrong abstraction.
Anyway, that additional condition is applicable to only one place when we call this function, so I will put it there, not in the function.

hashmap:
Don't see it anymore:
```
tail -f ~/.grin/main/grin-server.log | grep monitor_peers
20190214 14:56:35.747 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 8 connected (8 most_work). all 11219 = 443 healthy + 0 banned + 10776 defunct
20190214 14:56:35.747 DEBUG grin_servers::grin::seed - monitor_peers: no preferred peers
20190214 14:56:56.087 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 7 connected (7 most_work). all 11219 = 444 healthy + 0 banned + 10775 defunct
20190214 14:56:56.087 DEBUG grin_servers::grin::seed - monitor_peers: no preferred peers
20190214 14:57:16.262 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 10 connected (6 most_work). all 11220 = 425 healthy + 0 banned + 10795 defunct
20190214 14:57:16.262 DEBUG grin_servers::grin::seed - monitor_peers: no preferred peers
20190214 14:57:36.463 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 10 connected (9 most_work). all 11220 = 426 healthy + 0 banned + 10794 defunct
20190214 14:57:36.463 DEBUG grin_servers::grin::seed - monitor_peers: no preferred peers
20190214 14:57:56.573 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 6 connected (6 most_work). all 11220 = 427 healthy + 0 banned + 10793 defunct
20190214 14:57:56.573 DEBUG grin_servers::grin::seed - monitor_peers: no preferred peers
20190214 14:58:16.918 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 7 connected (7 most_work). all 11220 = 404 healthy + 0 banned + 10816 defunct
20190214 14:58:16.918 DEBUG grin_servers::grin::seed - monitor_peers: no preferred peers
20190214 14:58:37.588 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 10 connected (10 most_work). all 11243 = 408 healthy + 0 banned + 10835 defunct
```

servers/src/grin/seed.rs
```diff
@@ -186,7 +186,7 @@ fn monitor_peers(
 	// maintenance step first, clean up p2p server peers
 	peers.clean_peers(config.peer_max_count() as usize);
 
-	if peers.healthy_peers_mix() {
+	if peers.peer_count() > peers.peer_outbound_count() && peers.healthy_peers_mix() {
```
ignopeverell: Should that be `>=`?

hashmap: Don’t think so, but not sure anymore:) the logic is - exit if we have at least 1 inbound peer and we have enough outbounds

ignopeverell: But for NAT'd peers there'll never be any inbound? So they could have 100 oubounds and this will still never return true?

hashmap: Oh, wrong place to add this check, thanks for spotting it. It's supposed to be where we connect

servers/src/grin/seed.rs
```diff
@@ -315,7 +315,7 @@ fn listen_for_addrs(
 	let addrs: Vec<SocketAddr> = rx.try_iter().collect();
 
 	// If we have a healthy number of outbound peers then we are done here.
-	if peers.healthy_peers_mix() {
+	if peers.peer_count() > peers.peer_outbound_count() && peers.healthy_peers_mix() {
```
antiochp: So we're basically saying we do not need to try initiating more outbound connections if - 
* _at least_ one inbound connection, and
* healthy_peers_mix

And `healthy_peers_mix` is - 

* peer_count >= min_preferred, and
* outbound_count >= min_preferred / 2

And that first check (at least one inbound connection) is basically to determine if we are behind a NAT or not?

I think I get the logic but it feel confusing.

Is it worth introducing a `peer.peer_inbound_count()` fn to make this more a little more explicit?
And then branch clearly on that first check - "do this if we think we are a public node, otherwise do this other thing"?

hashmap:
@antiochp 
> And that first check (at least one inbound connection) is basically to determine if we are behind a NAT or not?

Yes, you can also read it that way - if we have only outbounds peers there is no need to check how healthy the mix is, because there is no mix.

> Is it worth introducing a `peer.peer_inbound_count()` fn to make this more a little more explicit?

That's what I did in the beginning, but such check is pretty expensive

PR #2545, opened 2019-02-08
 add content disposition for OK responses 

rentenmark:
following on to discussions from #2541 this is a smaller PR to fix up the OK responses only

### DIFF
--- a/wallet/src/controller.rs
+++ b/wallet/src/controller.rs
@@ -692,15 +692,12 @@ fn response<T: Into<Body>>(status: StatusCode, text: T) -> Response<Body> {
 			"access-control-allow-headers",
 			"Content-Type, Authorization",
 		);
-	
+
 	if status == StatusCode::OK {
-		builder = builder
-			.header(hyper::header::CONTENT_TYPE, "application/json");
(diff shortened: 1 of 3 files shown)


ignopeverell:
While we're cleaning this, shouldn't those return `204`s instead?

kargakis:
> While we're cleaning this, shouldn't those return `204`s instead?

You may break more people than it's worth by changing this now.

wallet/src/controller.rs
```diff
@@ -677,20 +677,27 @@ fn create_ok_response(json: &str) -> Response<Body> {
...
-		)
-		.body(text.into())
-		.unwrap()
+		);
+
+	if status == StatusCode::OK {
+		builder = builder.header(hyper::header::CONTENT_TYPE, "application/json");
+	}
+
+	builder.body(text.into()).unwrap()
```
garyyu: Perhaps add one line of comment here to remind the caller of this `response` function, the `text` here is required to be a JSON format.

Or even we can change the variable name `text` here as for example `json_formated_text`?

ignopeverell:
@kargakis `<html mode:pedantic="false" xmlns:mode="http://www.w3.org/1999/xhtml/pedantism">Alright then</html> `

PR #2546, opened 2019-02-08
fix: wallet coin selection respects max_block_weight

### LINKED ISSUE #2510: Wallet coin selection can build a tx that exceeds max_block_weight
This was not really an issue until we allowed `max_block_weight` to be configured per `chain_type`.

`UserTesting` and `AutomatedTesting` now define an artificially low `max_block_weight` to make testing blocks and txs at the block weight limit more convenient.

The coin selection logic (and overall tx building logic) should take `max_block_weight` into consideration to ensure txs are below the max permissible weight.

Not a huge issue right now but it does mean the default coin selection strategy can build invalid txs on `UserTesting` (workaround is to specify `-s smallest`).

i1skn:
Deprecate "soft" max_outputs limit and introduce "hard" max_outputs
limit based on max_block_weight. Should fix https://github.com/mimblewimble/grin/issues/2510


ignopeverell:
Looks like this introduces some bona fide test compilation failures.

i1skn:
@ignopeverell my fault, fixed!

wallet/src/libwallet/internal/selection.rs
```diff
@@ -253,7 +262,7 @@ where
 		amount,
 		current_height,
 		minimum_confirmations,
-		max_outputs,
+		calculate_max_inputs_in_block(change_outputs),
```
ignopeverell: Apologies if this is a misunderstanding, I'm not too familiar with this code. But shouldn't you calculate the max number of inputs for `change_outputs + 1` to account for the receiver's output?

i1skn: This is correct, but first we need understand to we need change. First, we calculate assuming that there will be no change (this gives us also the lowest fee). We check do we have `fee+amount` on our balance or not and only then we figure out do we need a change output or we can transfer without it.    
If we need one - we set:
https://github.com/mimblewimble/grin/blob/6a6a423f3ff8339ac1d52acec1f08d49622e2be9/wallet/src/libwallet/internal/selection.rs#L301
and then select coins using `num_outputs`
https://github.com/mimblewimble/grin/blob/6a6a423f3ff8339ac1d52acec1f08d49622e2be9/wallet/src/libwallet/internal/selection.rs#L322-L330 

At least that's how I get it, probably @yeastplume can verify does this make sense or not.

yeastplume: Yes, correct, that's what happens.

yeastplume:
LGTM, but merge conflict introduced, sorry.

I think I'm okay with removing the concept of a manually specifiable `max_outputs` in the name of reducing the exposed complexity.

PR #2547, opened 2019-02-08
Generate API secret in wallet --here mode

ignopeverell:
Tried to find the right spot to insert this. The config handling has stratified a bit and may need some cleanup at some point as well, but this just tries to fix a specific problem.


yeastplume:
Yep, think that's fine

PR #2549, opened 2019-02-09
Refactor: hashed trait to disambiguate use cases

JeremyRubin:
This PR aims to disambiguate the cryptographic hash used in consensus critical applications from the rust implementation-defined std::hash trait.

Problematically, there are couple user defined functions called 'hash', in addition to the standard one, which might be accidentally called.

To remedy this, I changed the cryptographic hash trait fn to be called `crypto_hash`.

Where we were previously worried about accidentally getting the hash of a container struct rather than the inner item, rather than shadow it with the correct result, I shadowed it with an deprecated function that calls `unimplemented!` and returns `!`. The three instances of this are Block, HeaderEntry, Tip. Thus, using the result of a call should trigger unreachable expression warnings/errors, and should cause linter errors for deprecated function use, and if we somehow did manage to call the wrong version, tests should cause panics if the wrong method is somehow used,

If the crypto_hash of such a container is actually needed, UFCS can resolve it explicitly.

This PR is a bit noisy and large, and careful review is required to ensure it is consistent. I think the noise is a large improvement in readability long term.

### DIFF
--- a/api/src/handlers/blocks_api.rs
+++ b/api/src/handlers/blocks_api.rs
@@ -105,7 +105,7 @@ impl BlockHandler {
 	fn parse_input(&self, input: String) -> Result<Hash, Error> {
 		if let Ok(height) = input.parse() {
 			match w(&self.chain).get_header_by_height(height) {
-				Ok(header) => return Ok(header.hash()),
+				Ok(header) => return Ok(header.crypto_hash()),
 				Err(_) => return Err(ErrorKind::NotFound)?,
 			}
 		}
--- a/api/src/handlers/chain_api.rs
+++ b/api/src/handlers/chain_api.rs
@@ -139,7 +139,7 @@ impl OutputHandler {
 		// TODO - possible to compact away blocks we care about
 		// in the period between accepting the block and refreshing the wallet
 		let block = w(&self.chain)
-			.get_block(&header.hash())
+			.get_block(&header.crypto_hash())
 			.map_err(|_| ErrorKind::NotFound)?;
 		let outputs = block
 			.outputs()
--- a/api/src/handlers/pool_api.rs
+++ b/api/src/handlers/pool_api.rs
@@ -91,7 +91,7 @@ impl PoolPushHandler {
 					};
 					info!(
 						"Pushing transaction {} to pool (inputs: {}, outputs: {}, kernels: {})",
-						tx.hash(),
+						tx.crypto_hash(),
 						tx.inputs().len(),
 						tx.outputs().len(),
 						tx.kernels().len(),
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -291,7 +291,7 @@ impl OutputPrintable {
 			commit: output.commit,
 			spent,
 			proof,
-			proof_hash: util::to_hex(output.proof.hash().to_vec()),
+			proof_hash: util::to_hex(output.proof.crypto_hash().to_vec()),
 			block_height,
 			merkle_proof,
 			mmr_index: output_pos,
@@ -489,7 +489,7 @@ pub struct BlockHeaderInfo {
 impl BlockHeaderInfo {
 	pub fn from_header(header: &core::BlockHeader) -> BlockHeaderInfo {
 		BlockHeaderInfo {
-			hash: util::to_hex(header.hash().to_vec()),
+			hash: util::to_hex(header.crypto_hash().to_vec()),
 			height: header.height,
 			previous: util::to_hex(header.prev_hash.to_vec()),
 		}
@@ -533,7 +533,7 @@ pub struct BlockHeaderPrintable {
 impl BlockHeaderPrintable {
 	pub fn from_header(header: &core::BlockHeader) -> BlockHeaderPrintable {
 		BlockHeaderPrintable {
-			hash: util::to_hex(header.hash().to_vec()),
+			hash: util::to_hex(header.crypto_hash().to_vec()),
 			version: header.version,
 			height: header.height,
 			previous: util::to_hex(header.prev_hash.to_vec()),
@@ -621,7 +621,7 @@ impl CompactBlockPrintable {
 		cb: &core::CompactBlock,
 		chain: Arc<chain::Chain>,
 	) -> CompactBlockPrintable {
-		let block = chain.get_block(&cb.hash()).unwrap();
+		let block = chain.get_block(&cb.crypto_hash()).unwrap();
 		let out_full = cb
 			.out_full()
 			.iter()
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -89,8 +89,8 @@ impl OrphanBlockPool {
 			let height_hashes = height_idx
 				.entry(orphan.block.header.height)
 				.or_insert(vec![]);
-			height_hashes.push(orphan.block.hash());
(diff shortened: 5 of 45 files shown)


JeremyRubin:
Thanks for the review!

I agree that it's noisy.

The reason for adding the header_hash method is that the struct innards for HeaderEntry aren't pub, so I didn't want to expose new fields -- but it should be safe to do -- and I wanted to keep it symmetric with Block. For Tip I didn't as we sometimes directly use last_block_h anyways. I'd be happy to change to direct access rather than helper method throughout the PR.

In your example this is something the user already *has* to know. If they need the block hash (and not the header hash), they need to use UFCS to get it. The difference between the patch and the current codebase is that if a user messes up and doesn't use UFCS, they should get all sorts of warnings/compiler errors.

I think the case of 'got a warning or error' is better than "didn't know what they were doing". I agree that it is troubling though.

In order to improve the troubling nature moreso, we can fix it with a little bit of (inlineable) indirection and type noise as follows:

If we introduce a new trait ConsensusHashable, we can have that trait have a method called consensus_hash which resolves to 'the thing you probably want'. Then from an API perspective, we call consensus_hash everywhere instead and consensus_hash resolves whether we want to get the hash of a field or of the whole struct.

As a second step, we can also make ConsensusHashable able to return typed hashes, which can further protect us from accidentally comparing, storing, etc hashes of different types or to raw hashes. This adds a *lot* of type noise and diff size (changing all function signatures/structs returning typed hashes), but is a good longer term approach.

This approach seems a bit heavy handed on the 80/20 cost benefit analysis, but if you want something that feels more complete this would be how to handle it I think (It's possible to just introduce ConsensusHashable without typed hashes).

ignopeverell:
I don't have a better solution to propose either but I'm still of the opinion that:

1. Hashing for the purpose of storing in a map (Rust hash) and hashing as a way to summarize the data (generally our hash) should be made more explicitly different.
2. But in agreement with @antiochp that this PR tends to make it a little more complex or noisy than it should.

antiochp:
Do we have _any_ cases where we want to `hash` something, say to put it in a map, where do do _not_ want to use the "crypto" hash?

I think the complexity is more around things like - 

* `block.hash` is actually wrapping `block.header.hash`
* `header_entry.hash` is actually wrapping `header.hash`

i.e. we have various wrapper structs and alternative representations of core structs, where the underlying hash does not want to change.

For example these should all be the same (for convenience and to eliminate any surprises) - 

```
# hash is sometimes - 
#     a field
#     a function
#     a function wrapping a function on a nested struct
block.hash == compact_block.hash == header.hash == header_entry.hash
```

I think we can probably solve this with a couple of common patterns that are maybe more robustly enforced than we do currently.

JeremyRubin:
Ok I think I can motivate an example of why I am grossed out by the current pattern and think it is difficult to review.

```rust

trait Hashable {
    fn hash(&self);
}
struct A {
    
}

impl Hashable for A {
    fn hash(&self) {
        println!("Hashable::hash")
    }
}

impl A {
    fn hash(&self) {
        println!("A::hash")
    }
}
fn call<T : Hashable>(t:&T) {
    t.hash();
}

fn main() {
    let a = A{};
    a.hash();
    call(&a)
}
```

Outputs

```
A::hash
Hashable::hash
```

There is no correct, general way to rewrite call with the given traits and inherents. Adding a consensus_hash trait does solve this problem though.

```rust
trait Hashable {
    fn hash(&self);
}
struct A {
    
}

impl Hashable for A {
    fn hash(&self) {
        println!("Hashable::hash")
    }
}
trait ConsensusHashable {
    fn consensus_hash(&self);
}

impl A {
    fn hash(&self) {
        println!("A::hash")
    }
}
impl ConsensusHashable for A {
    fn consensus_hash(&self) {
        // Implicit resolution goes to inherent
        self.hash();
        // or UFCS inherent
        A::hash(self);
        // or UFCS hashable trait
        Hashable::hash(self)
    }
}
fn call<T : ConsensusHashable>(t:&T) {
    t.consensus_hash();
}

fn main() {
    let a = A{};
    a.consensus_hash();
    call(&a)
}
```

JeremyRubin:
Btw here is an example of what typed hashes might look like (can figure out a cleaner version, but this compiles). Here there is more burden (we can get close to having default impl by having a new trait, `ConsensusHashIsDefault` which must be manually impl'd).

What's nice about doing something like this is that while it doesn't guarantee someone uses the correct hash in all cases, if that hash is to be used in a validation interface or relayed somewhere, the function signatures should usually be relatively clear what data we need. 

```rust
use std::marker::PhantomData;
struct Hash<T>(PhantomData<T>);

trait Hashable {
    type ConsensusHashType;
    fn hash(&self) -> Hash<Self::ConsensusHashType> where Self :Sized;
}

struct Header;
struct Block {}

impl Hashable for Block {
    type ConsensusHashType = Header;
    fn hash(&self) -> Hash<Self::ConsensusHashType> {
        println!("Hashable::hash::<Block>");
        Hash(PhantomData)
    }
}
impl Hashable for Header {
    type ConsensusHashType = Self;
    fn hash(&self) -> Hash<Self::ConsensusHashType> {
        println!("Hashable::hash::<BlockHeader>");
                Hash(PhantomData)

    }
}
impl Block {
    fn hash(&self) -> Hash<Self> {
        println!("A::hash");
        Hash(PhantomData)
    }
}
fn call<T : Hashable>(t:&T) {
    let _x : Hash<<T as Hashable>::ConsensusHashType> = t.hash();
}


fn check_hash(h: &Hash<Header>) -> bool {
    true
}

fn main() {
    let a = Block{};
    let h : Hash<Block> = a.hash();
    let h_ : Hash<Header> = Hashable::hash(&a);
    check_hash(&h);
    check_hash(&h_);
    call(&a)
}
```

```
  --> src/main.rs:46:16
   |
46 |     check_hash(&h);
   |                ^^ expected struct `Header`, found struct `Block`
   |
   = note: expected type `&Hash<Header>`
              found type `&Hash<Block>`
```

antiochp:
Taking a step back and thinking about this from a different angle - 

The problem would appear to be caused by the fact we have a "default" implementation of the `Hashed` trait that is automatically implemented by anything `Writeable` -

```
impl<W: ser::Writeable> Hashed for W {
	fn hash(&self) -> Hash {
		ser::Writeable::write(self, &mut hasher).unwrap();
		let mut ret = [0; 32];
		hasher.finalize(&mut ret);
		Hash(ret)
	}
...
```

Which works fine for most cases, but occasionally we don't want to use this default implementation.

So what would happen if we simply got rid of this default impl? And just explicitly implemented it for the structs we actually care about?

There are not many - 
* Block
* BlockHeader
* CompactBlock
* ShortId
* Input
* Output
* TxKernel
* some tuples of various things

This way we don't need the complexity around typed hashes. We just have explicit hash() impls on the things that need them.

Does that make sense?

JeremyRubin:
That's reasonable to me, so long as we don't mind the bloat of the 10 odd manual impls.

I'd still also like to rename `fn hash()`, but I concur that this solves the MRO issue.

ignopeverell:
> so long as we don't mind the bloat of the 10 odd manual impls

Macro?

JeremyRubin:
closed in favor of https://github.com/mimblewimble/grin/pull/2573

PR #2553, opened 2019-02-09
add repost method into wallet owner api

garyyu:
add a `repost` API, which is identical to the wallet command `grin wallet repost -i x`.

[Update]:
Without this `repost` API, there's no way to repost a pending `sending` transaction in case of any unexpected failure of last `post_tx` call (unless you can cache the finalized tx data and call `post_tx` again, but with this new `repost` API there's no need to cache that finalized tx data).

But I will not merge it before the review/s, welcome anyone interested in wallet APIs to take a review on this.


yeastplume:
:+1: Looks good to me.

PR #2563, opened 2019-02-12
Update contributing guidelines

yeastplume:
Will give a couple of days for reviews or suggestions before merging


lehnberg:
Does it make sense to also include specific instructions to update any documentation that is affected by the PR in question?

JeremyRubin:
I think that you could make pull_request_template.md with sections for:
```
```
which might be a bit friendlier for first time contributors.

PR #2565, opened 2019-02-12
[docs] Add more details to wallet TLS setup instruction

kalambet:
While setting up TLS for the wallet found out that instruction is missing a couple of details.


hashmap:
Thanks for your contribution!

PR #2568, opened 2019-02-13
refactor: reserve capacity in aggregate

JeremyRubin:
Don't mean to inundate review with low hanging fruit, but I've just been fixing/cleaning stuff as I see it reading through the codebase.

In transaction aggregation, we add a bunch of stuff to unreserved vecs. This can be slow-ish as when we realloc we end up having to move stuff and wasteful because  we may also over allocate (e.g., memory doubling). It really depends on the implementation.

We care about making this faster because it's in the path of relay and in the path of pools creating blocks.

Allocating ahead of time is simple enough.


hashmap:
Thanks for you contribution!

PR #2570, opened 2019-02-13
Use failure for Pool and Committed erros

hashmap:
It doesn't play nice with failure-based error management, we have to
throw away an exisiting error and create a new one (failure-based)


hashmap:
@antiochp yes, also we introduced it in many crates. I'm with you, Rust error management could be improved.

PR #2572, opened 2019-02-13
API handlers refactoring

hashmap:
This PR introduces QueryParams to abstarct low-level parameter parsing in many handlers.
Also adds some API tests.


hashmap:
Ready for review. Pool API tests are still missing, but they worth a separate PR

PR #2569, opened 2019-02-13
clean-up the txhashset zip file and temp folder

garyyu:
To obsolete https://github.com/mimblewimble/grin/pull/2354

Compare to #2354, this PR:
1. don't need track the list of the zip files
2. don't need to worry about tracking filenames and handling shutdown/restart etc.

PR #2354, opened 2019-01-12
[DNM] clean-up the txhashset zip file and temp folder

garyyu:
My `floo/chain_data` folder becomes more and more bigger, 4.6G for 6 days:
```
$ du -s chain_data
4601608	chain_data

$ ls -l chain_data | grep txhashset
drwxrwxr-x 5 garyyu garyyu     4096 Dec 30 03:02 txhashset
-rw-r--r-- 1 garyyu garyyu 39838686 Jan  7 23:57 txhashset_snapshot_102503.zip
-rw-r--r-- 1 garyyu garyyu 45923180 Jan  9 18:23 txhashset_snapshot_14035.zip
...
-rw-r--r-- 1 garyyu garyyu 41937198 Jan  8 14:22 txhashset_snapshot_988669.zip
drwxr-xr-x 5 garyyu garyyu     4096 Jan  7 23:57 txhashset_zip_102503
drwxr-xr-x 5 garyyu garyyu     4096 Jan  9 18:23 txhashset_zip_14035
...
drwxr-xr-x 5 garyyu garyyu     4096 Jan  8 14:22 txhashset_zip_988669
```
There are a lot of temporary files there! Let's clean-up these garbage.

PR #2575, opened 2019-02-14
clean-up the txhashset zip file and temp folder

garyyu:
(cherry-pick from #2569)To obsolete https://github.com/mimblewimble/grin/pull/2354

Compare to #2354, this PR:
1. don't need track the list of the zip files
2. don't need to worry about tracking filenames and handling shutdown/restart etc.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -690,7 +690,7 @@ impl Chain {
 		}
 
 		// prepares the zip and return the corresponding Read
-		let txhashset_reader = txhashset::zip_read(self.db_root.clone(), &header, None)?;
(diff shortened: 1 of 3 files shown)


antiochp:
Does this need to be `1.1.0`? Feels simple enough to merge into master right now and target `1.0.2` (if this exists)?

ignopeverell:
@antiochp this isn't entirely technically motivated but I was thinking this improvement would be a good carrot to get more `1.1.0` upgrades.

antiochp:
:+1: That makes sense. Do we have a rough timeline for when 1.1.0 might be released?

ignopeverell:
Not really but I was hoping the Windows support would be in there. That seems to be getting close.

PR #2593, opened 2019-02-18
Remove wallet code [1.1.0]

yeastplume:
which is now in https://github.com/mimblewimble/grin-wallet

Notes:

* A couple of small simple structs (CbData and BlockFees) are just cut and paste from the wallet into mine_block.rs where needed. They're just for serialization purposes so a don't think there's a need to try and refactor anything over this.
* It might worth removing the `server` command from grin and simplifying command line args now, but we can leave that for another PR.
* tests in the `server` crate that use both the server and wallet are moved into a separate `integration` crate for now. It made more sense dependency-wise to put this crate into the grin-wallet project for now, but we're going to need a better strategy around these tests.


yeastplume:
Ready for review (if anyone wants to). Will give it a couple of days then merge.

PR #2599, opened 2019-02-19
[WIP] refactor: Arc-ify Transactions

JeremyRubin:
Before I invest further energy here, wanted to collect feedback. It doesn't pass build (yet) because there is some work to do to make tests pass.

In this PR, we make the pool APIs return/use Arcs instead of cloning transactions.

This should be fairly resource efficient because we no longer need to copy transactions and vectors of transactions so gratuitously, which saves us on memory and on runtime, but we also lose data locality and Arcs have some overhead so this isn't for free.


JeremyRubin:
I updated the tests, so they pass.

antiochp:
Current approach definitely does too much `cloning()` but this (at a glance) feels like a lot of added complexity in the code.

Feels like putting the `Arc<>` at the individual tx level is maybe the wrong place to do this?

JeremyRubin:
I don't think it adds that much complexity -- it's mostly just changing the type signatures and wrapping the txn in an Arc or cloning it where appropriate -- the first two commits are the ones to focus on. Updating the tests took more time than the patch itself.

w.r.t. granularity, there are a few places that are viable options:

1. Arc the whole `Vec<PoolEntry>`
This is an interesting idea, especially if we treat PoolEntry as a `Deref<Target=Transaction>`. Then we can just return the entries Vec as an `Vec<impl Deref>` and use clone on modify behavior of Arc. This feels more complex to me, and we may end up copying anyways, but we can do it in the less hot code path.
2. Arc the `PoolEntry`
This might work out, but in general we don't expose the PoolEntry type outside of the pool, we return transactions so this approach does not make sense to me.
3. Arc each `Vec<Inputs>`, `Vec<Outputs>`, etc per transaction
This seems a bit stupid, because it's the same as the approach given in terms of benefit with 3x more smart pointers. 
4. The above and convert type to a `Vec<Arc<Vec<Inputs>>>`
This seems a little bit gross, but has the benefit that coming out of aggregation we can just grad the Arcs to each Vec of inputs to aggregate, and implement a custom iterator to make it traverse it as if it were normal.
5. Arc each Input/Output
This seems silly
6. Add a non-pub TransactionInner struct which holds all the state for a transaction in an Arc and provide getter/setter access.
This has the benefit of preserving our type signatures in more places, at the expense of a bit of weirdness in the Transaction struct

ignopeverell:
I don't think this is too bad, but I may have a slight preference for your first proposition (`Arc<Vec<PoolEntry>>`). I think overall this should be a pool problem, with minimal impact on core. Stepping back maybe it'd be good to flesh out the lifecycle of a transaction instance and see if we can express it more explicitly.

1. A transaction is instantiated when we get it from the network. No way around this.
2. The pool then takes ownership of it. For its own internal handling it can aggregate and deaggregate. Ideally this would require minimal cloning.
3. For the purpose of mining, transactions can be made into a block. We can see this in 3 ways:
   1. Transactions move from the pool to stratum, to be re-added if the block isn't successfully mined. This would minimize cloning but introduces some coupling here, with stratum having the resubmit the aggregate.
   2. Transaction can be cloned into a block.
   3. Transaction references can be passed and arranged into a block. It seems this would require specific types.
4. When a valid compact block header arrives, transactions can be moved into that block to form a full block.
5. Receiving a full block instantiates all of its transactions.
6 Transactions are destroyed once a fully valid block has been accepted.

Did I forget anything? I think I'm the least worried about 3, a little cloning for mining doesn't seem like a huge issue. So the main focus could be on 2 (and arguably deaggregation isn't really a concern for core) and 4.

garyyu:
There's a feature request there for ["Switch from ready to draft in pull requests"](https://github.community/t5/How-to-use-Git-and-GitHub/Feature-Request-Switch-from-ready-to-draft-in-pull-requests/m-p/19107#M5695)   Let's go to 👍  it to support this feature :-)

quentinlesceller:
@JeremyRubin Any update on this? This seems pretty good and reduce the massive amount of cloning we do in the pool.

JeremyRubin:
Didn't receive a strong enough indication that this approach was desired so haven't worked on it since.

antiochp:
Closing for now.
We should revisit post 3.0.0 and go back over comments here, thinking through tx lifecycle -https://github.com/mimblewimble/grin/pull/2599#issuecomment-467158430

PR #2601, opened 2019-02-19
add a new configure item for log_max_files

garyyu:
Currently the log rotation files limitation is hardcoded as 32 log files. That means at maximum we can only keep 512M logs by default (512M here refers to the original file size before gzip), the only way is to change / enlarge `log_max_size` configuration.

This PR add a new config `log_max_files`, default value is still `32`, to replace that hardcoded `count`.

### DIFF
--- a/util/src/types.rs
+++ b/util/src/types.rs
@@ -46,6 +46,8 @@ pub struct LoggingConfig {
 	pub log_file_append: bool,
 	/// Size of the log in bytes to rotate over (optional)
 	pub log_max_size: Option<u64>,
+	/// Number of the log files to rotate over (optional)
+	pub log_max_files: Option<u32>,
 	/// Whether the tui is running (optional)
 	pub tui_running: Option<bool>,
 }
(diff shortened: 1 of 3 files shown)


util/src/types.rs
```diff
@@ -60,6 +62,7 @@ impl Default for LoggingConfig {
 			log_file_path: String::from("grin.log"),
 			log_file_append: true,
 			log_max_size: Some(1024 * 1024 * 16), // 16 megabytes default
+			log_max_files: Some(32),              // 32 log files to rotate over by default
```
hashmap: We use `32` in 2 places, would you consider introducing a constant?

ignopeverell:
@garyyu saw the above?

garyyu:
Ah, I saw it but then forgot to change, will modify it now :-)

util/src/types.rs
garyyu: Done, thanks :-)

PR #2603, opened 2019-02-19
[WIP] problem identified with clean_output_index callback

antiochp:
We do not allow duplicate outputs in the UTXO set.

But we _do_ allow duplicates in the full TXO set.

The `compact` pruning logic does not correctly account for this when removing entries from the `output_pos` index and assumes they are unique across the entire TXO set.

I _think_ we can fix this by simply having the `clean_output_index` callback check for `is_unspent()` based on the actual entry in `output_pos` index not just blindly removing entries from the index.

----

Edit: Turns out the above _is_ a problem but not one we are actually encountering because we are hitting another issue prior to this - 

The `save_prune` logic is attempting to call `delete_output_pos` with an empty byte slice. So we never actually reconstruct the `output_pos` keys correctly (they are indexed by commitment).

We never actually delete _any_ entries from the `output_pos` index.

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -292,6 +292,27 @@ impl TxHashSet {
 		let rewind_rm_pos = input_pos_to_rewind(&horizon_header, &head_header, &batch)?;
 
 		{
+			// Callback for every output commitment being removed from the data file.
+			//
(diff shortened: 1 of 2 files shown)


antiochp:
Closing this - tracking underlying issue(s) in #2606.

PR #2609, opened 2019-02-20
Simplify (and fix) output_pos cleanup during chain compaction

### LINKED ISSUE #2606: The output_pos index not being cleaned up correctly during compaction
See #2603 

We read raw bytes from the data file during the pruning/compaction process and use these bytes to clean the `output_pos` index up.

1) For each pruned chunk of bytes we attempt to call `delete_output_pos` but these bytes do not simply represent a commitment, they represent an output_identifier.

2) The chunk of bytes is actually empty...

3) It is not actually safe to do this as we can have duplicate outputs in the TXO set if one is spent and one is unspent. And in this scenario it is not safe to clean the `output_pos` index up based purely on spent outputs.

antiochp:
Our existing approach to cleaning up the `output_pos` index was flawed for a variety of reasons.
See #2603 and #2606 for details.

Resolves #2606.

tl;dr lack of type safety from passing around a slice of bytes caused us to never remove old entries from the `output_pos` index. Additionally we could not safely delete them based on the callbacks during pruning as we allow duplicates across full TXO set (both spent and unspent).

This PR implements a simplified approach to maintaining the `output_pos` index - during compaction we simply clear out the existing index and rebuild it from scratch based on current UTXO set.

* Removed the callback in check_compact (passing a raw byte slice around)
* Expose `(key, value)` in our `SerIterator` db iterator.
* Use a single db `batch` for the entire chain `compact()` process.
* Rename `compact_blocks_db -> remove_historical_blocks` (`compact_block` used elsewhere)
* Introduces `leaf_pos_iter()` on the PMMR backend 
    * we can now conveniently iterate over unspent outputs in the output PMMR
* Reworked and simplified `output_pos` maintenance during chain compaction - 
    * clear the index out (using the iterator on the db), then
    * rebuild the index (using the iterator on the PMMR backend)
* Reworked `verify_rangeproofs()` to use the new leaf_pos iterator
* Reworked `verify_kernel_signatures()` for consistency

----

Compaction is pretty fast overall (~1.5s on mainnet at height 51,000).
Rebuilding the full `output_pos` index is a fraction of this -

```
20190220 15:21:10.202 DEBUG grin_chain::txhashset::txhashset - txhashset: starting compaction...
20190220 15:21:10.543 DEBUG grin_chain::txhashset::txhashset - txhashset: check_compact output mmr backend...
20190220 15:21:10.943 DEBUG grin_chain::txhashset::txhashset - txhashset: check_compact rangeproof mmr backend...
20190220 15:21:11.468 DEBUG grin_chain::txhashset::txhashset - txhashset: ... compaction finished
20190220 15:21:11.556 DEBUG grin_chain::txhashset::txhashset - txhashset: rebuild_index: 36920 UTXOs, took 0s
```

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -205,19 +205,25 @@ impl TxHashSet {
 			.get_last_n_insertions(distance)
 	}
 
-	/// Get the header at the specified height based on the current state of the txhashset.
-	/// Derives the MMR pos from the height (insertion index) and retrieves the header hash.
-	/// Looks the header up in the db by hash.
-	pub fn get_header_by_height(&self, height: u64) -> Result<BlockHeader, Error> {
+	/// Get the header hash at the specified height based on the current state of the txhashset.
+	pub fn get_header_hash_by_height(&self, height: u64) -> Result<Hash, Error> {
 		let pos = pmmr::insertion_to_pmmr_index(height + 1);
 		let header_pmmr =
 			ReadonlyPMMR::at(&self.header_pmmr_h.backend, self.header_pmmr_h.last_pos);
 		if let Some(entry) = header_pmmr.get_data(pos) {
-			let header = self.commit_index.get_block_header(&entry.hash())?;
-			Ok(header)
+			Ok(entry.hash())
 		} else {
-			Err(ErrorKind::Other(format!("get header by height")).into())
+			Err(ErrorKind::Other(format!("get header hash by height")).into())
 		}
+	}
+
+	/// Get the header at the specified height based on the current state of the txhashset.
+	/// Derives the MMR pos from the height (insertion index) and retrieves the header hash.
+	/// Looks the header up in the db by hash.
+	pub fn get_header_by_height(&self, height: u64) -> Result<BlockHeader, Error> {
+		let hash = self.get_header_hash_by_height(height)?;
+		let header = self.commit_index.get_block_header(&hash)?;
+		Ok(header)
 	}
 
 	/// returns outputs from the given insertion (leaf) index up to the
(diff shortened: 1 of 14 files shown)


ignopeverell:
:+1: on the approach. Will need a little more time for a proper review however.

chain/src/txhashset/txhashset.rs
```diff
@@ -1323,13 +1315,23 @@ impl<'a> Extension<'a> {
 		let total_kernels = pmmr::n_leaves(self.kernel_pmmr.unpruned_size());
 		for n in 1..self.kernel_pmmr.unpruned_size() + 1 {
```
JeremyRubin: Why not use the new iter here

chain/src/txhashset/txhashset.rs
```diff
@@ -1351,30 +1353,32 @@ impl<'a> Extension<'a> {
...
-						commits.clear();
-						proofs.clear();
-						debug!(
-							"txhashset: verify_rangeproofs: verified {} rangeproofs",
-							proof_count,
-						);
-					}
-				}
+		for pos in self.output_pmmr.leaf_pos_iter() {
+			let out = self
```
JeremyRubin: Personally think matching over both as a tuple and explicitly returning is a bit clearer here

store/src/leaf_set.rs
```diff
@@ -199,4 +199,9 @@ impl LeafSet {
 	pub fn is_empty(&self) -> bool {
 		self.len() == 0
 	}
+
+	/// Iterator over positionns in the leaf_set (all leaf positions).
+	pub fn iter<'a>(&'a self) -> Box<Iterator<Item = u64> + 'a> {
```
JeremyRubin: You can probably avoid the box with an impl Iterator?

chain/src/txhashset/txhashset.rs
antiochp: Good question! The kernel MMR is non-prunable so we do not actually maintain a leaf_set (all leaves are always in the leaf_set). And we have not (yet) implemented the new iterator for this case.

chain/src/txhashset/txhashset.rs
antiochp: :+1: Agree.

store/src/leaf_set.rs
antiochp: There was a problem somewhere when I attempted to do this but I did not look into it too closely (just added a `Box<>`...). Let me take another look and see if I can work out what was going on.

antiochp: Got it working. Still need the `Box<>` for `leaf_pos_iter()` on the Backend trait but `impl Iterator` works here.

ignopeverell:
Just needs conflict resolution before merging.

antiochp:
Gahh so close. Will fix the conflicts shortly.

PR #2615, opened 2019-02-21
Switch from LRU cache to CuckooCache

JeremyRubin:
This PR adds the cuckoocache data structure to Grin. Cuckoocache has some unique advantages over a typical LRU cache. See https://github.com/bitcoin/bitcoin/pull/8895 for more discussion and https://github.com/JeremyRubin/cuckoocache https://docs.rs/cuckoocache/0.1.0/cuckoocache/.

The primary benefit to Grin codebase is that Cuckoocache is threadsafe, meaning it can be read and erased from without requiring mutability. This enables multiple concurrent threads to use it for validation at once, through the RWLock::read. 

I've added the Cuckoocache as an independent crate, like the LRU cache is. The LRU cache has a dependency on https://github.com/contain-rs/linked-hash-map, which is in "Maitnence mode". It has an extensive test suite, which I ported from the implementation used in bitcoin (I authored it -- hence re-licensing under my name), and no dependencies.

A couple points:

1) in the current patch, we never call contains with discardable=true. This means entries will only be deleted via the generation maturity (which is closer to LRU behavior). When might we validate and be OK erasing the previously validated results?
2) Technically, because of hashing, there is no need to have two separate caches, we could just have a single `verificationcache<SomeEnum<Type1, Type2>>` or `VerificationCache<Hash>`.
3) I haven't done extensive performance testing of this versus the original code
4) I didn't refactor the name from LRUCache to something like CuckooCache as it seemed easier to do it in a non functional patch later.


antiochp:
Purely on the name alone :+1: (cc @tromp)

Let me spend some time over the next couple of days to review but I see no reason why we shouldn't replace the existing `LruCache` with this.

JeremyRubin:
Great, thanks!

No need to review them at the same time, but I will add some more commits to further demonstrate benefit (e.g., moving the RWLock inside the VerifierCache and splitting the lock per cache ). We can drop the rwlock because we no longer have `&mut` methods on VerifierCache.

core/src/core/verifier_cache.rs
```diff
@@ -15,52 +15,90 @@
...
+impl<E: DefaultHashable> SaltedHasher<E> for VerifierHasher {
+	fn new() -> Self {
+		Self {
+			salt: Hash::from_vec(&random::<[u8; 32]>()[..]),
+		}
+	}
+	fn hashes(&self, e: &E) -> [u32; 8] {
+		let mut hs = [0u32; 8];
+
+		for (v, h) in (self.salt, e)
```
antiochp: What's going on with this funky hasher? Can't we just use our own `hash` impl?

JeremyRubin: We do use our own hash implementation -- `(self.salt, e).hash()`. The concern is that the hashes need to be salted with a private salt. Otherwise I can easily DoS the network by finding a hash cycle in the cache. The rest of the funkyness is just to cast from [u8;32] -> [u32;8]. I don't know of any real way to do that? Maybe we could do...

```rust
let mut x = [0u32; 8];
{
  let y : &mut [u8; 32] = unsafe{std::mem::transmute(&mut x)}
  y = (self.salt, e).hash();
}
x
```

but then we're a smidgen unsafe.

antiochp: Ah - I missed the actual `hash()` call. :+1:

I get what you're saying - this way everyone hashes differently.
Let me think on this a bit. Makes we wonder if we need to consider doing this elsewhere.

antiochp: Actually - can we use the `siphash` stuff in `id.rs`? This would give us a way of getting short hashes, salted with a private value locally.

antiochp: And related - if the DoS is an issue, do we need to be doing something similar to this for the db/index itself as we store various things based on their hashes in the db?

tromp: agree with antiochp that using a keyed hash function like our existing siphash24 is cleaner than this salted construct.

JeremyRubin: In this case we need 256 bits of hash, so we'd have to siphash with a lot of different keys.

I suppose we could get away with using fewer bits, but we need at least enough to support 8x the number of bits needed to address our table.

Most likely we should salt in the DB too.

JeremyRubin: In this case we need 256 bits of hash, so we'd have to siphash with a lot of different keys.

I suppose we could get away with using fewer bits, but we need at least enough to support 8x the number of bits needed to address our table.

Most likely we should salt in the DB too.

tromp: in what circumstance would 128 bits of siphash key not suffice?

note that our Cuckoo Cycle proof of work uses a variant of siphash24 with a full 256-bit key,
so we could also consider using that instead of the standard siphash24...

JeremyRubin: The way that cuckoocache works we use 8 insertion indexes. Our method of generating them is to get a 256 bit hash. And split it up into 8 and then use lemire's hashing trick to project each of 8 into the range of the table. Because we want to support up to a 2**32 element table, we need to have 32 bits per index. 

We also store only the hashed value, not the value itself, so it's important that the hash is collision resistant.

JeremyRubin: The way that cuckoocache works we use 8 insertion indexes. Our method of generating them is to get a 256 bit hash. And split it up into 8 and then use lemire's hashing trick to project each of 8 into the range of the table. Because we want to support up to a 2**32 element table, we need to have 32 bits per index. 

We also store only the hashed value, not the value itself, so it's important that the hash is collision resistant.

JeremyRubin: The way that cuckoocache works we use 8 insertion indexes. Our method of generating them is to get a 256 bit hash. And split it up into 8 and then use lemire's hashing trick to project each of 8 into the range of the table. Because we want to support up to a 2**32 element table, we need to have 32 bits per index. 

We also store only the hashed value, not the value itself, so it's important that the hash is collision resistant.

JeremyRubin: The way that cuckoocache works we use 8 insertion indexes. Our method of generating them is to get a 256 bit hash. And split it up into 8 and then use lemire's hashing trick to project each of 8 into the range of the table. Because we want to support up to a 2**32 element table, we need to have 32 bits per index. 

We also store only the hashed value, not the value itself, so it's important that the hash is collision resistant.

antiochp: The blake2b crate we use may also allow us to get 256 bits in a single digest?

JeremyRubin: We do already, I'm not computing the hash multiple times, just once?

tromp: considering the difficulty of getting more than 64 approved bits of siphash output, and the need for 8x32 bits for cuckoocache, I think using blake2b as a keyed hash function may be best?!

JeremyRubin: I'm not using siphash, I'm using blake2b as a salted hash. I'm fine using keyed instead of salted, we just don't have support for it in the code currently. The hash method uses the Hashed trait. At least I'm fairly certain I am -- is there a reason you think otherwise?

@antiochp can I add one point to the 'we should rename the hash method name' tally?

tromp: Checking the blake2b documentation at https://blake2.net/blake2.pdf, I see that it supports both keyed hashing, and a salt. The difference is that the salt is xored into the IV (initialization vector), while the key is processed as the first message block. The former seems preferable as long as one is fine with 16 bytes salts. So Jeremy was ahead of me all along:-) Sorry for the unnecessary siphash discussion.

I would just prefer the salt to be passed into the blake2b constructor as part of the blake2b parameter block, rather than being passed around as the first element of a rust pair to be hashed.

JeremyRubin: Fair enough -- it just requires some additional code to support using keyed hashing, which we don't currently have (but the blake2 crate does).

I agree it's cleaner to pass it into the parameter block.

JeremyRubin:
I'd like to make sure there's been sufficient review on the cuckoocache crate itself before merge, but I'll un-WIP it now.

JeremyRubin:
As noted it can be done with a std::mem:: transmute, but that requires unsafe. 

I can't think of another way to do it that is cleaner. You generally can not go from a u8 array to a u32 because of alignment. You can do vice versa though. So any solution will require having an explicit u32 -> [u8;4] cast. 

If you care about the endianness, the prior version I had pushed up used the native endian version, but it was a bit more verbose to write out.

I think we maybe can and perhaps should just trust the compiler to emit something good here -- maybe we can check to see what it puts out is bad so we aren't premature optimizing

tromp:
We don't care about endianness, and we'll just go with the extra copy loop if Rust can't safely cast the u32 array into a byte array.

antiochp:
@JeremyRubin Is this good to go?

ignopeverell:
I think @JeremyRubin would like a review of his `cuckoocache` crate. Which is on my list but haven't gotten to it yet.

JeremyRubin:
It's here: https://github.com/JeremyRubin/cuckoocache

I'd be OK with merging if we're far from release and can make an thorough review on a pre-release checklist.

antiochp:
Closing this for now - we can pick it up again post 3.0.0 HF if desired.

Issue #2584, opened 2019-02-15
Caching Hash Computations


JeremyRubin:
see: https://github.com/mimblewimble/grin/pull/2567#issuecomment-464091200

Opening this issue to discuss caching computation of hashes.

For example, when we sort a list of O(N) elements by hash, we end up computing O(N Log N) cryptographic hashes of our data.

This is inefficient, so we would like to do something better -- this issue is a space to discuss potential solutions. Here are some suggestions to kick off:

1) For critical structs where we are hashing often, compute the hash on construction (assuming immutable) and store it in the struct (downside: memory usage. upside: cache efficiency, easy to code). Network serialization/de-serialization ignores this field and computes based on incoming data.
2) Store in the struct an Option<Box<Hash>>, only compute when needed. May be None forever. upside: don't use what you don't need -- only storing 1 pointer (8 bytes) v.s. 32 bytes, downside is when we do need it, having to store 40 bytes and losing locality
3) when doing std::eq, if we expect a heavily biased negative result, doing a cmp with:
`siphash(x) == siphash(y) && x.hash() == y.hash()`
the first clause can cheaply prove to us that x != y, the second clause proves that x == y.
Thus we save on hashing where it's obviously not needed. 
4) A wrapper type which holds hashed data -- upside: if we don't need it, don't use it. downside: would require reallocating to move items into a new vector, e.g., before sorting. Makes APIs messier
5) Careful algorithms writing -- e.g., see https://github.com/mimblewimble/grin/pull/2567. Upside: It is written such that the caching only takes up two Hashes worth of space, which is better than . Downside: it only works for particular algorithms, which are more difficult to write and review as a result.
5) stupid simple cache: a siphash driven LRU cache would let you look up elements for much cheaper with a fixed memory space. E.g., we allocate 1MB for hashes of inputs, and store (input, input.hash()) at siphash(input) % 1MB  (perhaps using something like my cuckoocache if more performance needed). Upside: bounded memory, easily tunable. Downside: sharing space per hashable type requires a big enum or just having a cache table per hash cached type, ownership is annoying (RWLock can help)
6) Permutation sorting: E.g., if we have a Vec<Input>, we also have a Vec<Hash>, and we write a custom algorithm to sort Vec<Input> by Vec<Hash> as the key lookup, and also sort Vec<Hash> at the same time. Easy to drop the Vec<Hash> when no longer needed, or pass on when needed later.

antiochp:
[...]

PR #2617, opened 2019-02-22
[WIP] Cache Hash Computations

JeremyRubin:
This PR demonstrates how we can add a hash caching functionality to existing classes (or not!) easily.

In response to https://github.com/mimblewimble/grin/issues/2584

The trait-fu needed to work this code is pretty ridiculous, especially since Rust doesn't support negative traits (e.g., `impl<T: ~DefaultTrait>`) or default trait functions without experimental flags, but it compiles somehow ;) It does feel a little fragile, if we wanted to add _more_ hash functionality in the future, it... should be doable, but I wouldn't want to be that future maintainer. That said, I can't see wanting more hash functionality later, so maybe this is good 'nuf.

The biggest issue with this approach is that functions like outputs_mut which modify the outputs become hard to work with -- I just defaulted to having these ones destroy the cache permanently as they don't seem to be used outside of testing or wallet code. Validation should be fine.

It would also be pretty easy to feature flag/runtime flag enable or disable the caching behavior on a per-struct basis, which is cool.

I'd also consider not using the Option and just storing a ZERO_HASH and checking for that, since it should be an error.

Definitely a WIP, don't merge till there's consensus on this approach and on which structs should implement the caching hasher.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -14,7 +14,7 @@
 
 //! Transactions
 
-use crate::core::hash::{DefaultHashable, Hashed};
+use crate::core::hash::{DefaultHashCache, DefaultHashable, GetHashCache, Hash, Hashed};
 use crate::core::verifier_cache::VerifierCache;
 use crate::core::{committed, Committed};
 use crate::keychain::{self, BlindingFactor};
@@ -50,6 +50,7 @@ enum_from_primitive! {
 }
 
 impl DefaultHashable for KernelFeatures {}
+impl DefaultHashCache for KernelFeatures {}
 
 impl Writeable for KernelFeatures {
 	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
@@ -172,6 +173,7 @@ pub struct TxKernel {
 }
 
 impl DefaultHashable for TxKernel {}
+impl DefaultHashCache for TxKernel {}
 hashable_ord!(TxKernel);
 
 impl ::std::hash::Hash for TxKernel {
@@ -760,9 +762,15 @@ pub struct Transaction {
 	pub offset: BlindingFactor,
 	/// The transaction body - inputs/outputs/kernels
 	body: TransactionBody,
+	hash_cache: Option<Hash>,
 }
 
 impl DefaultHashable for Transaction {}
+impl GetHashCache for Transaction {
+	fn get_cached(&self) -> Option<Hash> {
+		self.hash_cache
+	}
+}
 
 /// PartialEq
 impl PartialEq for Transaction {
@@ -793,7 +801,12 @@ impl Readable for Transaction {
 	fn read(reader: &mut dyn Reader) -> Result<Transaction, ser::Error> {
 		let offset = BlindingFactor::read(reader)?;
 		let body = TransactionBody::read(reader)?;
-		let tx = Transaction { offset, body };
+		let mut tx = Transaction {
+			offset,
+			body,
+			hash_cache: None,
+		};
+		tx.hash_cache = Some(tx.hash());
 
 		// Now "lightweight" validation of the tx.
 		// Treat any validation issues as data corruption.
@@ -831,7 +844,13 @@ impl Transaction {
 		Transaction {
 			offset: BlindingFactor::zero(),
 			body: Default::default(),
+			hash_cache: None,
 		}
+	}
+	fn recompute_hash(mut self) -> Transaction {
+		self.hash_cache = None;
+		self.hash_cache = Some(self.hash());
+		self
 	}
 
 	/// Creates a new transaction initialized with
@@ -843,13 +862,19 @@ impl Transaction {
 		let body =
 			TransactionBody::init(inputs, outputs, kernels, false).expect("sorting, not verifying");
 
-		Transaction { offset, body }
+		let mut tx = Transaction {
+			offset,
+			body,
+			hash_cache: None,
+		};
+		tx.hash_cache = Some(tx.hash());
+		tx
 	}
 
 	/// Creates a new transaction using this transaction as a template
 	/// and with the specified offset.
 	pub fn with_offset(self, offset: BlindingFactor) -> Transaction {
-		Transaction { offset, ..self }
+		Transaction { offset, ..self }.recompute_hash()
 	}
 
 	/// Builds a new transaction with the provided inputs added. Existing
@@ -860,6 +885,7 @@ impl Transaction {
 			body: self.body.with_input(input),
 			..self
 		}
+		.recompute_hash()
 	}
 
 	/// Builds a new transaction with the provided output added. Existing
@@ -870,6 +896,7 @@ impl Transaction {
 			body: self.body.with_output(output),
 			..self
 		}
+		.recompute_hash()
 	}
 
 	/// Builds a new transaction with the provided output added. Existing
@@ -880,6 +907,7 @@ impl Transaction {
 			body: self.body.with_kernel(kernel),
 			..self
 		}
+		.recompute_hash()
 	}
 
 	/// Get inputs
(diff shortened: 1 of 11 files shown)


core/src/core/transaction.rs
```diff
@@ -831,8 +844,14 @@ impl Transaction {
 		Transaction {
 			offset: BlindingFactor::zero(),
 			body: Default::default(),
+			hash_cache: None,
 		}
 	}
+	fn recompute_hash(mut self) -> Transaction {
```
antiochp: This is going to be biggest concern here I think - making sure we have covered all cases of mutating existing data and _guaranteeing_ we are not using a stale hash somewhere.

antiochp:
Maybe worth doing some profiling/benching (not sure what with) and see what hotspots we currently have with hashes.

core/src/core/transaction.rs
JeremyRubin: Yep!

The best way to do this is going to be to make the types non pub (only gettable via getter which provides immutable ref), or where mut is needed, to make it take a callback, e.g.

```rust
transaction.outputs_mut(|outputs| {
   outputs.push(x)
});

impl Transaction {
  fn outputs_mut<F, R>(&mut self, f: F)-> where F: Fn(&mut Vec<Output>) -> R {
    let r = f(&mut self.outputs);
    self.recompute_hashes();
    r
  } 
}
```

JeremyRubin:
So it's actually a bit more than that -- we need to specifically look for things that we are hashing more than once -- e.g., if we spend a lot of time hashing blockheaders but we only hash them one time it won't help us, unless doing the hashing earlier is an advantage (e.g., not needing to hash a transaction during some validation step if it's sitting around in some pool)

Maybe adding a static `HashMap<TypeName, HashMap<Hash, u64>>` and outputting every hash with more than 1 in the bucket 100k hashes would let us get an idea for which hashes we'd save on.

DavidBurkett:
Grin++ caches hashes on certain objects, but honestly, it doesn't really provide much value. And I only did so because Grin++ almost exclusively uses immutable objects, so it's safe and easy to implement. Hashing isn't one of those things that take very long to do, so it's not worth much added complexity, especially if it introduces the possibility of stale hashes. There are far better ways to improve performance - mostly around p2p message processing, mutex locking, DB lookups, and bulletproof, signature, & kernel sum verifications. Just my $. 02

JeremyRubin:
I ran a node for about 10 hours with the instrumentation on hash set to count time for entries which are hashed more than once. Note the type names may appear more than once through differing paths.
```
 TIME_TOTAL    TYPE NAME
 22.877902906s    grin_core::core::block::BlockHeader
 10.828195722s    (u64, &grin_core::core::block::BlockHeader)
 9.694138593s    core::block::BlockHeader
 5.817870788s    (u64, &(grin_core::core::hash::Hash, grin_core::core::hash::Hash))
 263.927663ms    core::transaction::Output
 217.76041ms    core::transaction::Input
 95.719932ms    core::transaction::TxKernel
 94.28371ms    grin_core::core::transaction::Transaction
 83.973957ms    core::core::BlockHeader
 69.683209ms    secp256k1zkp::pedersen::RangeProof
 60.857147ms    grin_core::core::transaction::TxKernel
 31.019969ms    core::transaction::Transaction
 23.258116ms    grin_core::core::compact_block::CompactBlock
 20.496898ms    core::core::CompactBlock
 18.114322ms    pow::types::Proof
 10.804556ms    core::core::TxKernel
 9.145361ms    (u64, &grin_core::core::transaction::Input)
 9.020528ms    &core::transaction::TxKernel
 7.865762ms    core::id::ShortId
 463.305µs    (core::transaction::KernelFeatures, u64, u64)
 371.066µs    core::transaction::KernelFeatures
 361.89µs    (&grin_core::core::hash::Hash, u64)
 245.515µs    (&core::hash::Hash, u64)
 0ns    core::core::Transaction
 0ns    (u64, &secp256k1zkp::pedersen::RangeProof)
 0ns    (u64, &grin_core::core::transaction::TxKernel)
 0ns    (u64, &grin_core::core::transaction::Output)
```

So it may be worth it to implement caching for the blockheader. This doesn't rule out hashing for the other entries, as caching may move the time added by them out of band of the hot part of the code, but that will be a bit harder to instrument. This also doesn't rule out other things being cached because I don't know how closely last night's load mirrors a worst case load or load under future network conditions.


@DavidBurkett while it isn't a lot of time saved by caching, it does happen in a pretty latency sensitive part of the code (verification/propagation) which does make a difference for mining clients. Lastly, it's also something that we *know* gets linearly worse per marginal item, whereas something like a PMMR lookup is sublinear per marginal item, so decreasing hashing time very well be more important to growing the system. Also I am working on the performance of other parts of the code as well, so I don't think it's mutually exclusive.

JeremyRubin:
Instrumentation code below:

```rust
use std::collections::HashMap;
 lazy_static! {
     static ref TMAP: std::sync::Mutex<HashMap<String, (std::time::Duration, HashMap<Hash, (usize, std::time::Duration)>)>> =
         std::sync::Mutex::new(HashMap::new());
     static ref COUNT: std::sync::atomic::AtomicUsize = std::sync::atomic::AtomicUsize::new(0);
 }
 impl<D: DefaultHashable + GetHashCache> Hashed for D {
         fn hash(&self) -> Hash {
                 let t1 = std::time::Instant::now();
                 /*if let Some(h) = self.get_cached() {
                         return h;
                 }*/
                 Writeable::write(self, &mut hasher).unwrap();
                 let mut ret = [0; 32];
                 hasher.finalize(&mut ret);
                 let t2 = t1.elapsed();
                 let mut tmap = TMAP
                         .lock()
                         .unwrap();
                         let ent = tmap.entry(unsafe { std::intrinsics::type_name::<Self>() }.to_string())
                         .or_insert((std::time::Duration::new(0,0), HashMap::new())).1
                         .entry(Hash(ret))
                         .or_insert((0, std::time::Duration::new(0, 0)));
                         ent.0 += 1;
                         ent.1 += t2;
                 if COUNT.fetch_add(1, std::sync::atomic::Ordering::Relaxed) % 100000 == 0 {
                     let mut v : Vec<_> = tmap.iter_mut().map(|(typename, (total, hashes))| {
                         let time_extra : std::time::Duration = hashes.values().filter(|&(c, _)| *c > 1).map(|&(_, d)| d).sum();
                         //*total += time_extra;
                         (time_extra, typename)
                     }).collect();
                     v.sort();
                     v.reverse();
                     let mut s = String::from("Hashing Data:\n\nTIME_TOTAL    TYPE NAME");
                     for (time_extra, typename) in v {
                         s += &format!("\n{:?}    {}", time_extra, typename);
                     }
                     debug!("{}", s);
                 }
                 Hash(ret)
         }
 }
```

DavidBurkett:
@JeremyRubin Awesome job doing the legwork for all of this! I agree, the _when_ is way more important than the _how often_. If we naively pretend Grin runs in one thread, then it looks like we're spending ~.1% of the time doing unnecessarily BlockHeader hashing. That's certainly worth looking at.

I'm not able to dig into the code right this second, but do we ever mutate BlockHeader? I have nothing against caching hashes for the header, but stale hashes are a huge concern of mine, and I'd hesitate to add much complexity in order to avoid them.

Since you're looking at performance of everything, any idea why each peer I connect to increases my CPU usage by 1-2%? Seems outrageous for threads that should be idle 90% of the time.

JeremyRubin:
So if you look at the code for blockheader, when we are hashing, we already have an immutability/caching issue -- we *only* hash the proof of work! If other fields change, then we're already in an invalid state. This also means that the hashing can be made non default (I'll PR this separately) to defer to the PoW's proof.

antiochp:
going to close this PR for now - if there is interest in picking up the `BlockHeader` specific changes then :+1: as it does look the vast majority of the cost of caching is focused there _and_ as mentioned we do non-default hashing (on the PoW) for headers.
The benefit there is the PoW is not going to change on an existing block, so these hashes should be relatively robust to cache without risking introducing stale cached values.

PR #2620, opened 2019-02-23
[WIP] Add status call to wallet foreign endpoint

cayblood:
---
name: Add status call to wallet foreign endpoint
about: Bitforex (and maybe other exchanges) requires an HTTPS address to transfer grin coins to. I was going to use an ELB on AWS to do the SSL termination for me, but ELB needs a way of doing a health check on any endpoint it sends requests to. This PR adds a status check to the wallet foreign endpoint. I've updated the API documentation. I also tried to write a unit test for the endpoint, but the test framework you're using for the wallet foreign endpoint doesn't seem to simulate actual HTTP requests, and this test is so trivial that nothing more than checking for a 200 response is necessary. Perhaps someone could suggest a testing approach that would be compatible with your other test scaffolding. Thanks in advance.
title: ''
labels: ''
assignees: ''

### DIFF
--- a/wallet/src/controller.rs
+++ b/wallet/src/controller.rs
@@ -679,7 +679,21 @@ where
 		))
 	}
 
-	fn handle_request(&self, req: Request<Body>) -> WalletResponseFuture {
+	fn handle_get_request(&self, req: Request<Body>) -> WalletResponseFuture {
+		match req
+			.uri()
+			.path()
+			.trim_right_matches("/")
+			.rsplit("/")
+			.next()
+			.unwrap()
+		{
+			"status" => Box::new(ok(response(StatusCode::OK, ""))),
+			_ => Box::new(ok(response(StatusCode::BAD_REQUEST, ""))),
+		}
+	}
+
+	fn handle_post_request(&self, req: Request<Body>) -> WalletResponseFuture {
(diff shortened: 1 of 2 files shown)


wallet/src/controller.rs
```diff
@@ -679,7 +679,21 @@ where
 		))
 	}
 
-	fn handle_request(&self, req: Request<Body>) -> WalletResponseFuture {
+	fn handle_get_request(&self, req: Request<Body>) -> WalletResponseFuture {
+		match req
```
hashmap: We added some helpers to `api` crate, perhaps you may want to use it,  eg this macro does the same thing but safer, without calling `unwrap` https://github.com/mimblewimble/grin/blob/master/api/src/web.rs#L143 an example: https://github.com/mimblewimble/grin/blob/master/api/src/handlers/blocks_api.rs#L70

yeastplume:
This looks good as an enhancement, but please note https://github.com/mimblewimble/grin/pull/2550 and https://github.com/mimblewimble/grin-wallet

ignopeverell:
So should this be moved to the `grin-wallet` repository?

quentinlesceller:
Indeed @cayblood please open this pull request on the [grin-wallet](https://github.com/mimblewimble/grin-wallet) repository. Thanks!

PR #2626, opened 2019-02-25
[1.1.0] Travis automated windows builds

yeastplume:
Adds windows as a travis build target for tagging only. 

Annoyingly, there appears to be an issue with travis windows builds whereby if an encrypted env variable is defined the windows build just hangs. Right now the build relies on someone's API key referenced as `$GITHUB_TOKEN`, which causes the build to hang if included. This is only being used to generate a changelog on the linux build.

This will probably be fixed at some stage, but in the meantime, windows builds will have to be done separately using the following highly annoying process:

* Perform tag/build as usual
* Remove $GITHUB_TOKEN from the travis CI project settings
* comment windows release target back into the build matrix, comment other targets out, commit changes
* Perform another tag (possibly appending `-win` to it), and let it build/release
* Re-insert $GITHUB_TOKEN into project in readiness for next build

Hopefully travis ci addresses this before release.

We'll probably need to regenerate the github api key at some stage and ensure it's updated in the travis ci project settings and in .travis.yml.  As a reminder, the command to create an encrypted version to directly embed in travis.yml (under api_key) is:

`travis encrypt [token] deploy.api_key`

Also note this process has only been tested in grin-wallet to create pre-releases: https://github.com/mimblewimble/grin-wallet/releases

### DIFF
--- a/.auto-release.sh
+++ b/.auto-release.sh
@@ -6,7 +6,7 @@ export CHANGELOG_GITHUB_TOKEN="$token"
 
 tagname=`git describe --tags --exact-match 2>/dev/null || git symbolic-ref -q --short HEAD`
 
-echo 'make a tarball for the release binary...\n'
+echo 'package the release binary...\n'
 
 if [[ $TRAVIS_OS_NAME == 'osx' ]]; then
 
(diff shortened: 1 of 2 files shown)


hsk81:
@yeastplume You may want to create e.g. a `grin-bot` GitHub account, for which you would then create a Travis CI key (using travis CLI). The `grin-bot` would be a collaborator for the `grin` repository, with *only* a `public_repository` scope.

This way, your own account (with your other repositories) would still be protected in case of a leak of the token, and also limiting any potential damage the `grin-bot` could do. The `grin-bot` could do a *draft* release only, which you could then check, and then publish as a (pre-)release.

Please see also the *warning* at Travis CI w.r.t. GitHub Releases: https://docs.travis-ci.com/user/deployment/releases/#authenticating-with-an-oauth-token:

> **Warning:** the `public_repo` and `repo` scopes for GitHub oauth tokens grant write access to all of a user’s (public) repositories. For security, it’s ideal for `api_key` to have write access limited to only repositories where Travis deploys to GitHub releases. The suggested workaround is to create a machine user — a dummy GitHub account that is granted write access on a per repository basis.

PR #2628, opened 2019-02-25
Dandelion++ Rewrite

### LINKED ISSUE #2176: Dandelion++ Algorithm Changes

antiochp:
Resolves #2176.

Replaces #2344. Minimizing unrelated changes and keeping focused on Dandelion++.

This is a minimal implementation of Dandelion++ to replace our existing minimal implementation of the original Dandelion.

* Introduce `DandelionEpoch` to track current epoch stem/fluff and current outbound relay peer
* Simplify existing Dandelion monitor
* Get rid of `PoolEntryState`
* Each epoch lasts for (by default) 10 mins.
    * Configurable via `epoch_secs`
* If node is in "stem" epoch then we send stem txs on to outbound relay peer
* If nodes is in "fluff" epoch then we hold stem txs (allowing opportunity to aggregate)
* Dandelion monitor runs every 10s, if current epoch is "fluff" -
    * If epoch expired, aggregate and broadcast all stem txs
    * If any stem txs older than 30s (`aggregation_secs`), aggregate and fluff all stem txs
    * Otherwise wait, allowing stem txs opportunity to aggregate
* Handle expired (3 mins, `embargo_secs`) by broadcasting individual txs

### DIFF
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -157,18 +157,12 @@ impl TransactionPool {
 			tx,
 		};
 
+		// If not stem then we are fluff.
 		// If this is a stem tx then attempt to stem.
 		// Any problems fallback to fluff later.
-		// If not stem then we are fluff.
-		let fluff = if stem {
-			// Attempt to add to stempool, notifying adapter to relay to outbound peer.
-			// Fall back to fluffing (via txpool) if stempool interaction fails.
-			self.add_to_stempool(entry.clone(), header)
-				.and_then(|_| self.adapter.stem_tx_accepted(&entry.tx))
-				.is_err()
-		} else {
-			true
-		};
+		let fluff = !stem || self.add_to_stempool(entry.clone(), header)
+			.and_then(|_| self.adapter.stem_tx_accepted(&entry.tx))
+			.is_err();
 
 		// Fluff (broadcast) the tx if tagged as fluff or if
 		// stemming failed for any reason.
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -23,9 +23,7 @@ use self::core::core::verifier_cache::VerifierCache;
 use self::core::core::{transaction, Block, BlockHeader, Transaction, Weighting};
 use self::util::RwLock;
 use crate::pool::Pool;
-use crate::types::{
-	BlockChain, PoolAdapter, PoolConfig, PoolEntry, PoolEntryState, PoolError, TxSource,
-};
+use crate::types::{BlockChain, PoolAdapter, PoolConfig, PoolEntry, PoolError, TxSource};
 use chrono::prelude::*;
 use grin_core as core;
 use grin_util as util;
@@ -79,10 +77,8 @@ impl TransactionPool {
 	fn add_to_stempool(&mut self, entry: PoolEntry, header: &BlockHeader) -> Result<(), PoolError> {
 		// Add tx to stempool (passing in all txs from txpool to validate against).
 		self.stempool
-			.add_to_pool(entry, self.txpool.all_transactions(), header)?;
-
-		// Note: we do not notify the adapter here,
-		// we let the dandelion monitor handle this.
+			.add_to_pool(entry.clone(), self.txpool.all_transactions(), header)?;
+		self.adapter.stem_tx_accepted(&entry.tx);
 		Ok(())
 	}
 
@@ -159,28 +155,18 @@ impl TransactionPool {
 		self.blockchain.verify_coinbase_maturity(&tx)?;
 
 		let entry = PoolEntry {
-			state: PoolEntryState::Fresh,
 			src,
 			tx_at: Utc::now(),
 			tx,
 		};
 
-		// If we are in "stem" mode then check if this is a new tx or if we have seen it before.
-		// If new tx - add it to our stempool.
-		// If we have seen any of the kernels before then fallback to fluff,
-		// adding directly to txpool.
-		if stem
-			&& self
-				.stempool
-				.find_matching_transactions(entry.tx.kernels())
-				.is_empty()
-		{
-			self.add_to_stempool(entry, header)?;
-			return Ok(());
+		if stem {
+			self.add_to_stempool(entry.clone(), header)?;
+		} else {
+			self.add_to_txpool(entry.clone(), header)?;
+			self.add_to_reorg_cache(entry);
 		}
 
-		self.add_to_txpool(entry.clone(), header)?;
-		self.add_to_reorg_cache(entry);
 		Ok(())
 	}
 
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -159,14 +159,13 @@ impl TransactionPool {
 
 		// If not stem then we are fluff.
 		// If this is a stem tx then attempt to stem.
-		// Any problems fallback to fluff later.
-		let fluff = !stem || self.add_to_stempool(entry.clone(), header)
-			.and_then(|_| self.adapter.stem_tx_accepted(&entry.tx))
-			.is_err();
-
-		// Fluff (broadcast) the tx if tagged as fluff or if
-		// stemming failed for any reason.
-		if fluff {
+		// Any problems during stem, fallback to fluff.
+		if !stem
+			|| self
+				.add_to_stempool(entry.clone(), header)
+				.and_then(|_| self.adapter.stem_tx_accepted(&entry.tx))
+				.is_err()
+		{
 			self.add_to_txpool(entry.clone(), header)?;
 			self.add_to_reorg_cache(entry.clone());
 			self.adapter.tx_accepted(&entry.tx);
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -74,10 +74,10 @@ impl TransactionPool {
 		self.blockchain.chain_head()
 	}
 
+	// Add tx to stempool (passing in all txs from txpool to validate against).
 	fn add_to_stempool(&mut self, entry: PoolEntry, header: &BlockHeader) -> Result<(), PoolError> {
-		// Add tx to stempool (passing in all txs from txpool to validate against).
 		self.stempool
-			.add_to_pool(entry.clone(), self.txpool.all_transactions(), header)?;
+			.add_to_pool(entry, self.txpool.all_transactions(), header)?;
 		Ok(())
 	}
 
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -161,11 +161,11 @@ impl TransactionPool {
 		// Any problems fallback to fluff later.
 		// If not stem then we are fluff.
 		let fluff = if stem {
-			// First attempt to add to stempool and return an error if this fails.
-			self.add_to_stempool(entry.clone(), header)?;
-			// Then attempt to relay it to the next stem peer,
-			// falling back to fluffing locally if this fails.
-			self.adapter.stem_tx_accepted(&entry.tx).is_err()
+			// Attempt to add to stempool, notifying adapter to relay to outbound peer.
+			// Fall back to fluffing (via txpool) if stempool interaction fails.
+			self.add_to_stempool(entry.clone(), header)
+				.and_then(|_| self.adapter.stem_tx_accepted(&entry.tx))
+				.is_err()
 		} else {
 			true
 		};
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -161,7 +161,10 @@ impl TransactionPool {
 		// Any problems fallback to fluff later.
 		// If not stem then we are fluff.
 		let fluff = if stem {
+			// First attempt to add to stempool and return an error if this fails.
 			self.add_to_stempool(entry.clone(), header)?;
+			// Then attempt to relay it to the next stem peer,
+			// falling back to fluffing locally if this fails.
 			self.adapter.stem_tx_accepted(&entry.tx).is_err()
 		} else {
 			true
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -78,7 +78,6 @@ impl TransactionPool {
 		// Add tx to stempool (passing in all txs from txpool to validate against).
 		self.stempool
 			.add_to_pool(entry.clone(), self.txpool.all_transactions(), header)?;
-		self.adapter.stem_tx_accepted(&entry.tx);
 		Ok(())
 	}
 
@@ -120,8 +119,6 @@ impl TransactionPool {
 			let txpool_tx = self.txpool.all_transactions_aggregate()?;
 			self.stempool.reconcile(txpool_tx, header)?;
 		}
-
-		self.adapter.tx_accepted(&entry.tx);
 		Ok(())
 	}
 
@@ -160,11 +157,22 @@ impl TransactionPool {
 			tx,
 		};
 
-		if stem {
+		// If this is a stem tx then attempt to stem.
+		// Any problems fallback to fluff later.
+		// If not stem then we are fluff.
+		let fluff = if stem {
 			self.add_to_stempool(entry.clone(), header)?;
+			self.adapter.stem_tx_accepted(&entry.tx).is_err()
 		} else {
+			true
+		};
+
+		// Fluff (broadcast) the tx if tagged as fluff or if
+		// stemming failed for any reason.
+		if fluff {
 			self.add_to_txpool(entry.clone(), header)?;
-			self.add_to_reorg_cache(entry);
+			self.add_to_reorg_cache(entry.clone());
+			self.adapter.tx_accepted(&entry.tx);
 		}
 
 		Ok(())
--- a/pool/src/types.rs
+++ b/pool/src/types.rs
@@ -27,15 +27,14 @@ use failure::Fail;
 use grin_core as core;
 use grin_keychain as keychain;
 
-/// TODO - This now maps to the "epoch length"?
-/// Dandelion relay timer
-const DANDELION_RELAY_SECS: u64 = 600;
+/// Dandelion "epoch" length.
+const DANDELION_EPOCH_SECS: u64 = 600;
 
-/// Dandelion embargo timer
+/// Dandelion embargo timer.
 const DANDELION_EMBARGO_SECS: u64 = 180;
 
-/// Dandelion patience timer
-const DANDELION_PATIENCE_SECS: u64 = 10;
+/// Dandelion aggregation timer.
+const DANDELION_AGGREGATION_SECS: u64 = 30;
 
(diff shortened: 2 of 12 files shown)


antiochp:
Default Dandelion config - 

```
#########################################
### DANDELION CONFIGURATION           ###
#########################################
[server.dandelion_config]

#dandelion epoch duration
epoch_secs = 600

#fluff and broadcast after embargo expires if tx not seen on network
embargo_secs = 180

#dandelion aggregation period in secs
aggregation_secs = 30

#dandelion stem probability (stem 90% of the time, fluff 10% of the time)
stem_probability = 90
```

antiochp:
This is from a node running on mainnet - 

```
20190312 08:40:19.360 DEBUG grin_p2p::protocol - handle_payload: received stem tx: msg_len: 1638
20190312 08:40:19.396 DEBUG grin_pool::pool - add_to_pool [stempool]: d346119cd8b5 (p2p) [in/out/kern: 1/2/1] pool: 1 (at block 000000a50b26)
20190312 08:40:19.396 INFO grin_servers::common::adapters - Fluff epoch. Aggregating stem tx(s). Will fluff via Dandelion monitor.
20190312 08:40:24.161 DEBUG grin_pool::pool - add_to_pool [stempool]: a3498982130f (p2p) [in/out/kern: 6/2/1] pool: 0 (at block 000000a50b26)
20190312 08:40:24.177 DEBUG grin_pool::pool - add_to_pool [stempool]: d346119cd8b5 (p2p) [in/out/kern: 1/2/1] pool: 1 (at block 000000a50b26)
20190312 08:40:26.166 DEBUG grin_servers::grin::dandelion_monitor - dand_mon: Found 2 txs in local stempool to fluff
```

This was an "in the wild" tx aggregation via Dandelion++ which I don't think we have seen very often with the existing Dandelion impl.

2 txs arrived within 30s of each other on this node which was in a "fluff" epoch. 
Dandelion monitor ran, aggregated them and "fluffed" the resulting tx, broadcasting out to the network.

servers/src/grin/dandelion_monitor.rs
```diff
@@ -44,211 +46,153 @@ pub fn monitor_transactions(
...
-
-		let res = tx_pool.adapter.stem_tx_accepted(&agg_tx);
-		if res.is_err() {
-			debug!("dand_mon: Unable to propagate stem tx. No relay, fluffing instead.");
+	let cutoff_secs = dandelion_config
+		.aggregation_secs
+		.expect("aggregation secs config missing");
+	let cutoff_entries = select_txs_cutoff(&tx_pool.stempool, cutoff_secs);
+
+	if adapter.is_expired() {
```
ignopeverell: Did you mean to keep those conditions around the comments? Seem that code doc shouldn't rely on actual code :-)

servers/src/grin/dandelion_monitor.rs
```diff
@@ -44,211 +46,153 @@ pub fn monitor_transactions(
...
-				identifier: "?.?.?.?".to_string(),
-			};
+	let header = tx_pool.chain_head()?;
 
-			tx_pool.add_to_pool(src, agg_tx, false, &header)?;
-		}
-	}
+	let fluffable_txs = {
+		let txpool_tx = tx_pool.txpool.all_transactions_aggregate()?;
+		let txs: Vec<_> = all_entries.iter().map(|x| x.tx.clone()).collect();
```
ignopeverell: Seems that this clone isn't required, given that `validate_raw_txs` internally clones as well?

servers/src/common/types.rs
```diff
@@ -421,3 +424,94 @@ impl chain::TxHashsetWriteStatus for SyncState {
...
+		self.start_time = Some(Utc::now().timestamp());
+		self.relay_peer = peers.outgoing_connected_peers().first().cloned();
+
+		// If stem_probability == 90 then we stem 90% of the time.
+		let mut rng = rand::thread_rng();
+		let stem_probability = self
+			.config
+			.stem_probability
+			.expect("stem_probability config missing");
+		self.is_stem = rng.gen_range(0, 101) < stem_probability;
```
ignopeverell: Off by one? If I set a probability of 100, the condition should always be true, and in this case won't if the `rng` comes up with 100.

pool/src/transaction_pool.rs
```diff
@@ -159,28 +152,32 @@ impl TransactionPool {
...
-				.stempool
-				.find_matching_transactions(entry.tx.kernels())
-				.is_empty()
-		{
-			self.add_to_stempool(entry, header)?;
-			return Ok(());
+		// If this is a stem tx then attempt to stem.
+		// Any problems fallback to fluff later.
+		// If not stem then we are fluff.
+		let fluff = if stem {
```
ignopeverell: Maybe clearer as `!stem || self.add_to_stempool ...`

servers/src/grin/dandelion_monitor.rs
antiochp: Hmm - tried to make it clear what was going on here. Now I'm looking at fresh I see what you are saying. I'll rewrite this and clean it up.

servers/src/grin/dandelion_monitor.rs
antiochp: Yes. One of those situations where I fight with various combinations of syntax before _finally_ realizing the solution is to use `into_iter()`.

servers/src/common/types.rs
antiochp: 🤦‍♂️ Yes. Should be `<=` here.

antiochp:
```
rng.gen_range(0, 100) < stem_probability

# if stem probability is 0% then false at 0 and false at 99
# if stem probability is 100% then true at 0 and true at 99
```

antiochp: Not sure why I was having so much trouble getting my head around that...

pool/src/types.rs
```diff
@@ -27,62 +27,61 @@ use failure::Fail;
...
 const DANDELION_EMBARGO_SECS: u64 = 180;
 
-/// Dandelion patience timer
-const DANDELION_PATIENCE_SECS: u64 = 10;
+/// Dandelion aggregation timer.
+const DANDELION_AGGREGATION_SECS: u64 = 30;
 
 /// Dandelion stem probability (stem 90% of the time, fluff 10%).
-const DANDELION_STEM_PROBABILITY: usize = 90;
+const DANDELION_STEM_PROBABILITY: u64 = 90;
```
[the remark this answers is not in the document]
antiochp: :+1:
I kept it as a u64 for consistency with the other config values, but it doesn't really make sense to do this.
Given your suggestion I think we should limit this to a `u8`.

servers/src/common/types.rs
```diff
@@ -421,3 +424,94 @@ impl chain::TxHashsetWriteStatus for SyncState {
...
+		self.start_time = Some(Utc::now().timestamp());
+		self.relay_peer = peers.outgoing_connected_peers().first().cloned();
+
+		// If stem_probability == 90 then we stem 90% of the time.
+		let mut rng = rand::thread_rng();
+		let stem_probability = self
+			.config
+			.stem_probability
+			.expect("stem_probability config missing");
+		self.is_stem = rng.gen_range(0, 100) < stem_probability;
```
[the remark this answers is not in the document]
antiochp: Min random value is 0.
Max random value is 99.

Min config value is 0%.
Max config value is 100%.

If config is 0 we want all random values to result in false `0 < 0` etc.
If config is 100 we want all random values to result in true `99 < 100` etc.

If config is 50 we want 0-49 to be false and 50-99 to be true (i.e. true 50% of the time).

So does `rng.gen_range(0, 100) < stem_probability` not give us what we want here?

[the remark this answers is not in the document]
quentinlesceller: Indeed just figured it out. This confused me way too much. ^^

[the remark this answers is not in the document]
antiochp: Same here - spent way too long today staring at that line of code... Thanks @ignopeverell for spotting the issue!

lehnberg:
@antiochp is the node able to set its own fluff/stem status? Or was that just something for testing here, where a node was set to “100% fluff mode”? Could nodes decide on this individually in the wild? 

If so, is that a problem? Could it be used to attack the network somehow? By deploying lots of "100% fluff, never broadcast" nodes that “drains” all the transactions? I know we use embargo timers, but would that then lead to the originating nodes sending out the transactions at expiry of embargo?

antiochp:
@lehnberg nodes can choose any value between 0% and 100% in terms of stem probability.

What you are suggesting is not limited to Dandelion in any way. I could run a modified node that simply did not broadcast any txs (or blocks for that matter). There's nothing we could do to prevent that.

servers/src/grin/dandelion_monitor.rs
```diff
@@ -44,211 +46,143 @@ pub fn monitor_transactions(
...
 	let mut tx_pool = tx_pool.write();
 
-	let header = tx_pool.chain_head()?;
-
-	let stem_txs = tx_pool
-		.stempool
-		.get_transactions_in_state(PoolEntryState::ToStem);
-
-	if stem_txs.is_empty() {
+	let all_entries = select_txs_cutoff(&tx_pool.stempool, 0);
```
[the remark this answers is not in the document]
antiochp: I was trying to keep it consistent between the two txpool queries - but there's really no point doing that as you say. Will revise.

lehnberg:
Understood. In such an event and at the embargo timer expiration, would it lead to the originator broadcasting their transaction? Or could it plausibly also be other nodes doing that?

Given that the originator will have the shortest time left on their embargo timer, it would be them doing that, right?

antiochp:
There's some randomness in terms of the timer length - so even though the originator starts their timer first its still reasonably random who would actually broadcast.
With Dandelion++ we're not introducing a delay at each stem hop, so all the timers are running pretty close together. With the original Dandelion we were adding a delay at each hop and the originator was far more likely to be the one to broadcast in that scenario as their timer was more likely to expire first.

pool/src/types.rs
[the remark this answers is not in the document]
antiochp: We now use `u8` for the probability and `u16` for the various dandelion timers.

servers/src/common/types.rs
```diff
@@ -421,3 +424,94 @@ impl chain::TxHashsetWriteStatus for SyncState {
...
+	/// It is expired if start_time is older than the configured epoch_secs.
+	pub fn is_expired(&self) -> bool {
+		let expired = if let Some(start_time) = self.start_time {
+			let epoch_secs = self.config.epoch_secs.expect("epoch_secs config missing") as i64;
+			Utc::now().timestamp().saturating_sub(start_time) > epoch_secs
+		} else {
+			true
+		};
+		expired
+	}
```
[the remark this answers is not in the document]
antiochp: Ahh - this was leftover (and unnecessarily convoluted) from adding some debug logging in here earlier. Fixed.

antiochp:
Note: 40 lines of code removed! :rocket:

antiochp:
ok merging this into 1.1.0 🤞

kargakis:
@antiochp does https://github.com/mimblewimble/grin/blob/master/doc/dandelion/dandelion.md also need to get updated?

quentinlesceller:
Indeed @kargakis this doc reflects the previous version of Dandelion.

antiochp:
Yes it does and should. I totally forgot about that. I'll open an issue to track this.

antiochp:
Note my `Help Wanted` tag on that issue 😄

lehnberg:
Thinking about this some more (belatedly), I'm becoming less convinced of the privacy benefits of aggregation before fluffing.

With the probability of a node being in stem mode set to 90% and fluff mode set to 10%, it implies that  only 1/10th of the nodes on the network will be fluff nodes, i.e. "sinks" where transactions become aggregated. 

If this is the case, deploying malicious surveillance nodes with fluff mode = 100%  will have a relatively high weight in the pool of total fluff nodes. These can then start collecting data on stem transactions. Even if the nodes end up aggregating and broadcasting the transactions correctly (thus disguising any issue on the network), they will be able to deduce the original unaggregated transactions (as they will be receiving them in piecemeal one by one).

Wouldn't _some_ aggregation by honest nodes be better than nothing, i.e. not even attempting to aggregate? Maybe. But I think there's a problem to have a system that users believe achieves a certain degree of privacy, if this can easily be circumvented.

In the dandelion++ paper, the node's choice between stemming or fluffing is pseudorandom, implying a 50% coin flip between the two choices. This at least would make it much harder to deploy enough surveillance nodes to capture transactions. But it also would mean less chance of aggregation. 

Perhaps the entire objective of aggregation should be removed from our thinking around Dandelion?
Not sure if any of this makes any sense. But I'm hoping someone could convince me why aggregation pre-fluffing makes a big difference from a privacy perspective (or from any perspective at all for that matter).

antiochp:
> If this is the case, deploying malicious surveillance nodes with fluff mode = 100% will have a relatively high weight in the pool of total fluff nodes. These can then start collecting data on stem transactions. Even if the nodes end up aggregating and broadcasting the transactions correctly (thus disguising any issue on the network), they will be able to deduce the original unaggregated transactions (as they will be receiving them in piecemeal one by one).

If a malicious actor controlled 10% of the nodes on the network then they would have 50% of the "sinks" and would therefor see roughly 50% of _all_ stem txs. Is that your thinking?

How likely is it that a malicious actor would be able to do this?

Kind of feels like if we are in this situation there is not a lot anybody can do in terms of preventing them observing txs and constructing a reasonably complete tx graph.

antiochp:
> In the dandelion++ paper, the node's choice between stemming or fluffing is pseudorandom, implying a 50% coin flip between the two choices.

Is that true? I thought it was still 90/10.

lehnberg:
> If a malicious actor controlled 10% of the nodes on the network then they would have 50% of the "sinks" and would therefor see roughly 50% of all stem txs. Is that your thinking?

Yes, how many nodes are there on the network today? ~100? This means that 10 nodes will be fluffing nodes at any given time. If a malicious actor deploys 10 always-fluffing nodes as well, it means they see 50% of all transactions.

> How likely is it that a malicious actor would be able to do this?

It's 5x as likely compared to if the probability was a fair coin flip and we had 50 honest fluff nodes in my example above. 

> Is that true? I thought it was still 90/10.

Page 29:9 / 9 in the Dandelion++ pdf (emphasis mine):
> (3) Transaction Forwarding (relay): Each time a node receives a stem-phase transaction from another node, it either relays the transaction or diffuses it. **The choice to diffuse transactions is pseudo- random, and is computed from a hash of the node’s own identity and epoch number.** 

I don't think it's ever been 90/10 in the papers, this is our own construction, presumably put in place to encourage aggregation pre-fluffing. I struggle to understand what the benefits from this are. It's not a bad thing, but it's not clear to me why we should strive for it in our design objectives. If it happens, then great. But let's not siphon off all transactions to a few nodes with the sole intention for this to happen. It may have unintended consequences, and it's not clear what the gains are.

lehnberg:
Please disregard the statement of 90/10, I dug deeper in the papers, and now see that `q` (path length parameter) is recommended to be 0.1 or 0.2 (i.e. 10% or 20% likelihood of fluffing). I'm confused but nevertheless give up. Don't really feel any wiser.

lehnberg:
@antiochp in the Grin implementation, does a peer always stem their own transaction? this is what's instructed in the paper, i.e. to only ever fluff stem transactions received from other peers, but never their own.

antiochp:
> in the Grin implementation, does a peer always stem their own transaction?

Good question.
I actually thought we did yes. It was part of the original design and what I had intended to implement. But - I think we actually forgot about this (and I remember now it being kind of tricky because we lose the "is this our tx?" before we get to the point of deciding to stem/fluff.

The downside here is we may end up aggregating our own txs if our node is in a fluff epoch (and not a stem epoch).

I'm not actually sure what kind of downside we see with this implementation. 

_Some_ txs will end up appearing on the network from the nodes where they actually originated - but we won't know which txs these are, so I'm not entirely convinced we actually lose any privacy here?

lehnberg:
> But - I think we actually forgot about this (and I remember now it being kind of tricky because we lose the "is this our tx?" before we get to the point of deciding to stem/fluff.

Why is it important to keep the "is this our tx?" knowledge? In my [naive interpretation](https://github.com/mimblewimble/grin/blob/a25cf01b03a9e66dcd23930c5369e60d3b4cad62/doc/dandelion/dandelion.md#current-implementation) of how D++ was being implemented in Grin, I was presuming that nodes add any tx, whether received or originating from themselves to their own stempool and set an embargo timer on them.  If you stem it, and don't see it within the embargo, you will fluff it either way.

> I'm not actually sure what kind of downside we see with this implementation.

I think they added this requirement for a reason. Let's say we have a network of 100 honest nodes that are being observed. 10 of those will be fluff nodes, and 90 will be stem nodes. Let's also assume that each fluff node fluffs 5 transactions each epoch (which is a fairly high estimate I would imagine).

If a fluffing node fluffs its own transaction, an attacker will know that "every 10% of the time, the anonymity set of a node's own transaction will be 1 in 5". If they manage to figure out which one that was, then they would also reduce the anonymity set of the other transactions being fluffed to 1 in 4. This might sound like little additional information to give up, but if you monitor a network over time and build up a lot of data, this will have an impact on your model's prediction accuracy.

On the other hand, if a node always stems their own transaction regardless of whether it's in fluff or stem mode, the attacker will know that "every 100% of the time, the anonymity set of a node's own transaction will be 1 in 90".  Which would make it much harder to pinpoint the originating source, as you're hiding amongst all the other stem nodes.

Issue #2488, opened 2019-01-29
TxHashSet download speed not accurate


DavidBurkett:
My node was trying to sync the state with a peer that was uploading at 600 kB/s and got to 98% and then the peer stopped uploading (haven't investigated that yet). What I noticed though is that rather than download speed updating to 0 kB/s, it must be using the average over the entire timeframe, so the download speed slowly peters out (trust me, this is the technical term). Still, 5-10 minutes after the peer stopped uploading data, my node still says it's downloading at 200 kB/s.

Calculating this correctly could be a bit tricky, but even if we go with a less sophisticated approach, like updating every 5-10 seconds, it'd still be a whole lot better than what we do today.

c410-f3r:
cc #2633 

I think this issue can be closed now.

DavidBurkett:
I just experienced this exact same issue 2 days ago. I think it still needs some more work

yeastplume:
Just going to close this for sake of housekeeping, if you still think it needs work by all means reopen

PR #2633, opened 2019-02-27
Display the current download rate rather than the average when syncing the chain

Agreene:
Fix for https://github.com/mimblewimble/grin/issues/2488
Currently when syncing the chain, your displayed download rate is calculated as (Total downloaded size/ Total Time), aka the average download rate.  This can be extremely misleading if your download speed suddenly changes (like when you disconnect).

This PR adds to each TxHashsetDownload  the size and timestamp values of the previous TxHashsetDownload, which allows us to calculate the delta between them and get the current rate.

You can test by starting up a sync, letting it run for a minute, then disconnecting from the network; previous behavior was that the rate would slowly decrease, now it will drop to 0 within a few seconds.

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -327,12 +327,9 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 	) -> bool {
 		match self.sync_state.status() {
 			SyncStatus::TxHashsetDownload {
-				start_time: _,
-				prev_update_time: _,
 				update_time: old_update_time,
-				prev_downloaded_size: _,
 				downloaded_size: old_downloaded_size,
-				total_size: _,
+				..
 			} => {
 				self.sync_state
 					.update_txhashset_download(SyncStatus::TxHashsetDownload {
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -326,10 +326,20 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 		total_size: u64,
 	) -> bool {
 		match self.sync_state.status() {
-			SyncStatus::TxHashsetDownload { .. } => {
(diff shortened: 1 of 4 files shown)


servers/src/common/adapters.rs
```diff
@@ -326,10 +326,20 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 		total_size: u64,
 	) -> bool {
 		match self.sync_state.status() {
-			SyncStatus::TxHashsetDownload { .. } => {
+			SyncStatus::TxHashsetDownload {
```
ignopeverell: Doesn't a `, ..` at the end work here for fields you don't care about? Instead of specifying them explicitly.

src/bin/tui/status.rs
```diff
@@ -137,14 +140,14 @@ impl TUIStatusListener for TUIStatusView {
...
-						let start = start_time.timestamp_nanos();
+						let start = prev_update_time.timestamp_nanos();
 						let fin = Utc::now().timestamp_nanos();
 						let dur_ms = (fin - start) as f64 * NANO_TO_MILLIS;
 
 						format!("Downloading {}(MB) chain state for state sync: {}% at {:.1?}(kB/s), step 2/4",
 						total_size / 1_000_000,
 						percent,
-						if dur_ms > 1.0f64 { downloaded_size as f64 / dur_ms as f64 } else { 0f64 },
+						if dur_ms > 1.0f64 { (downloaded_size - prev_downloaded_size) as f64 / dur_ms as f64 } else { 0f64 },
```
ignopeverell: :+1:

servers/src/common/adapters.rs
Agreene: 👍 Updated

PR #2636, opened 2019-02-27
[1.1.0] Disable gen_gen crate

yeastplume:
gen_gen crate no longer builds due to wallet being extracted. Doesn't affect the build but it affects grin-wallet development when trying to reference the 1.1.0 branch directly (as it tries to build all Cargo.toml files it finds). This just renames gen_gen's Cargo.toml to _Cargo.toml and adds a note saying what's happened.


antiochp:
My limited understanding is `gen_gen` is _only_ needed for (re)building the genesis block. Is that correct?
So this is effectively archived for posterity like this?

yeastplume:
Yes, correct. (so this also gives any shitcoin forkers another annoyance).

PR #2641, opened 2019-02-28
Fix TxHashSet file filter for Windows.

DavidBurkett:
sanitized_name() converts `/` to `\\` on windows machines, so it does not match the regex filter.
This fix looks for both `/` and `\\`.

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -1465,7 +1465,7 @@ fn expected_file(path: &Path) -> bool {
 	lazy_static! {
 		static ref RE: Regex = Regex::new(
 			format!(
-				r#"^({}|{}|{})(/pmmr_(hash|data|leaf|prun)\.bin(\.\w*)?)?$"#,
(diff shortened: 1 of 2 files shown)


chain/src/txhashset/txhashset.rs
```diff
@@ -1465,7 +1465,7 @@ fn expected_file(path: &Path) -> bool {
 	lazy_static! {
 		static ref RE: Regex = Regex::new(
 			format!(
-				r#"^({}|{}|{})(/pmmr_(hash|data|leaf|prun)\.bin(\.\w*)?)?$"#,
+				r#"^({}|{}|{})((/|\\)pmmr_(hash|data|leaf|prun)\.bin(\.\w*)?)?$"#,
```
hashmap: Please add a test case, you can take it from my pr

yeastplume:
Great, thanks both of you for looking into this. Will merge, cherry pick to 1.1.0 then build/test.

@DavidBurkett Were you able to confirm that the txhashset zips and sends when requested from windows?

PR #2642, opened 2019-02-28
Specify grin or nanogrins in API docs where applicable

rentenmark:
## Problem

API docs are not specifying if amounts are in nanogrin or grin and it should be made explicit wherever possible.

## Solution

Add some docs related to nanogrin / grin where it tripped me up. This is a draft pull request so I am happy to flesh out more but I'm not entirely sure where else this information should be added or what the right information is.

### DIFF
--- a/doc/api/wallet_owner_api.md
+++ b/doc/api/wallet_owner_api.md
@@ -108,7 +108,7 @@ Attempt to update and retrieve outputs.
 * **Success Response:**
 
   * **Code:** 200
-  * **Content:** Array of
+  * **Content:** All listed amounts in nanogrin. Array of
 
     | Field                             | Type     | Description                             |
     |:----------------------------------|:---------|:----------------------------------------|
--- a/doc/api/wallet_owner_api.md
+++ b/doc/api/wallet_owner_api.md
@@ -341,7 +341,7 @@ Send a transaction either directly by http or file (then display the slate)
 
     | Field                         | Type     | Description                          |
     |:------------------------------|:---------|:-------------------------------------|
-    | amount                        | number   | Amount to send                       |
+    | amount                        | number   | Amount to send (in nanogrin)         |
     | minimum_confirmations         | number   | Minimum confirmations                |
     | method                        | string   | Payment method                       |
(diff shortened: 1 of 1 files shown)


doc/api/wallet_owner_api.md
```diff
@@ -341,7 +341,7 @@ Send a transaction either directly by http or file (then display the slate)
 
     | Field                         | Type     | Description                          |
     |:------------------------------|:---------|:-------------------------------------|
-    | amount                        | number   | Amount to send                       |
+    | amount                        | number   | Amount to send (in nanogrin)         |
```
antiochp: :+1: good idea...

hashmap:
If we accept grins somewhere in API it must be a mistake, one should never use float numbers to represent money

antiochp:
I'm guessing this is simply a documentation exercise and that is should be nanogrins everywhere.

yeastplume:
The command line wallet uses grins for obvious convenience sake, but that's the only place and they're converted immediately to nanogrins before any APIs are called

rentenmark:
> one should never use float numbers to represent money

This is being spoken with authority but is not authoritative. In the old world of buy side and sell side finance, amounts are represented frequenty _with absolute precision_ and that is absolutely fine. Often these numbers are represented in strings to make it overtly explicit that these are not IEEE 754 _floats_ they are _absolute precision_ numbers. 

I know this because I was there. 

That being said, API documentation in the old world is also often obsessive about documenting units. I believe we should be equally obsessive about documenting units. Here is [one excellent reason to document your units in your API](http://www.cnn.com/TECH/space/9909/30/mars.metric.02/) if you haven't seen this incident, it's a very interesting one. 

All told, I am very happy to work in nanogrins _everywhere_ and very glad that it's consistent.

I will go through and update docs to reflect nanogrins wherever applicable

ignopeverell:
Is this still a draft? Clarification looks good to me and AFAICT the code does the right thing.

rentenmark:
I will take this out of draft, I wanted to go through and find all the spots where this type of comment was prudent. I added one more comment to the docs

PR #2644, opened 2019-03-01
return the finalized slate from the post_tx endpoint

rentenmark:

The API for post_tx should return the finalized transaction slate for posterity.

This change may break some clients and would probably be best to wait for V2 but I wanted to put it here as a discussion point.  

No unit tests for API at the moment

Only touching outer skin, really just integration points.

WIP 

* Updates any documentation that's affected by the PR

https://github.com/mimblewimble/grin/blob/master/doc/api/wallet_owner_api.md#post-post-tx

### DIFF
--- a/wallet/src/controller.rs
+++ b/wallet/src/controller.rs
@@ -445,7 +445,7 @@ where
 		&self,
 		req: Request<Body>,
 		api: APIOwner<T, C, K>,
-	) -> Box<dyn Future<Item = (), Error = Error> + Send> {
+	) -> Box<dyn Future<Item = Slate, Error = Error> + Send> {
 		let params = match req.uri().query() {
 			Some(query_string) => form_urlencoded::parse(query_string.as_bytes())
 				.into_owned()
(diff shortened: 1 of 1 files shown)


garyyu:
`post_tx` return slate?!  but slate isn't the input of `post_tx`?

rentenmark:
Yes, if I understand things correctly the `api.post_tx` function modifies the slate transforming it from `responseTx` state to `finalTx` state. This returns the slate in its "final form"

I could absolutely be completely wrong about this. That's just me reading into the method signature.

jaspervdm:
Posting a transaction to the node doesn't change the slate in any way

garyyu:
@rentenmark 
Another option is to use `repost-tx` API, which don't need the caller to provide the input slate:
https://github.com/mimblewimble/grin/blob/master/doc/api/wallet_owner_api.md#post-repost-tx

And let's close this PR :-) but pls feel free to find us on support channel to continue the discussion about this.

PR #2651, opened 2019-03-05
Small QoL improvements for wallet developers

jaspervdm:
Some small changes to make the lives of wallet developers (including me) a bit easier. If preferred, I could create separate PRs for each but since they are pretty small changes I thought I'd bundle them up.

Changes:
- Add `pub` to the keychain in the builder `Context`. This is useful since I have a bunch of custom build functions that need to use the keychain
- Move `create_nonce` into the `Keychain` trait, this makes it easier to provide custom nonce derivation schemes. Edit: as yeastplume correctly points out below, this should not be used by wallet developers to change the nonce for regular transactions
- ~~Flag to turn off the fee check in slate signing/finalization. This is useful if the intermediate tx has a too low fee but the final tx has a high enough fee (due to added kernels). Simply adding the kernels to the slate tx won't work since there is an [assertion](https://github.com/mimblewimble/grin/blob/master/wallet/src/libwallet/slate.rs#L519) that there is only a single kernel in the tx~~ Will be re-done in a separate PR on grin-wallet
- Add `to_bytes` function to `WalletSeed`, easier than having to call `from_hex(seed.to_hex())`


yeastplume:
I have some concerns about facilitating a wide array of nonce creation schemes as it will affect expected behavior of restoring from seed. But refactoring it into a trait for now I think is fine.

Slate format changes need to go into a new slate version in the grin-wallet repo (slate versioning isn't contained in master,). Can I suggest aiming this PR at 1.1.0 and splitting it between here and the wallet?

jaspervdm:
Thanks for reviewing!

> I have some concerns about facilitating a wide array of nonce creation schemes as it will affect expected behavior of restoring from seed. But refactoring it into a trait for now I think is fine.

I agree, we should be careful with that. However for example in a multisignature wallet, restoring from a single seed is not expected to work in the same way. For this specific instance (and possibly a few others) it would be much easier to have the function inside the trait, rather than having to specify custom `build` functions.

> Slate format changes need to go into a new slate version in the grin-wallet repo (slate versioning isn't contained in master,). Can I suggest aiming this PR at 1.1.0 and splitting it between here and the wallet?

This PR doesn't actually change the (de)serialisation format, it is merely an internal flag to skip the fee check on finalisation. This is very useful for me in a specific application, but it might not be the best way to implement it. I could try and refactor it a bit to separate out functionality, so in my specific situation I can only call the functions that I actually need (which is basically `finalize` without a fee check). Would this be preferred?

yeastplume:
Agreed first point. On second, I think your suggestion is preferable to having fields in the slate that aren't serialized, mostly for readability's sake. Definitely needs to be done in `grin-wallet` though (`grin\wallet` files will all be removed when 1.1.0 is merged)

jaspervdm:
I removed the changes I made in the Slate, will refactor it and re-open on `grin-wallet`

PR #2660, opened 2019-03-07
https support for webhooks

mcdallas:
Implements the todos from #2598

- Adds https support
- Adds configurable `nthreads` and `timeout`

Usage:

Add these values to your `grin-server.toml`

```
[server.webhook_config]
nthreads = 4
timeout = 10
block_accepted_url = "http://127.0.0.1:8080/acceptedblock"
tx_received_url = "http://127.0.0.1:8080/tx"
header_received_url = "http://127.0.0.1:8080/header"
block_received_url = "http://127.0.0.1:8080/block"
```

### DIFF
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -248,6 +248,20 @@ pub struct WebHooksConfig {
 	pub block_received_url: Option<String>,
 	/// url to POST block data when a new block is accepted by our node (might be a reorg or a fork)
 	pub block_accepted_url: Option<String>,
+	/// number of worker threads in the tokio runtime
+	#[serde(default = "default_nthreads")]
+	pub nthreads: u16,
+	/// timeout in seconds for the http request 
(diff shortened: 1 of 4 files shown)


servers/src/common/types.rs
```diff
@@ -248,6 +248,20 @@ pub struct WebHooksConfig {
...
 	pub block_accepted_url: Option<String>,
+	/// number of worker threads in the tokio runtime
+	#[serde(default = "default_nthreads")]
+	pub nthreads: u16,
+	/// timeout in seconds for the http request
+	#[serde(default = "default_timeout")]
+	pub timeout: u16,
+}
+
+fn default_timeout() -> u16 {
```
hashmap: why do we need it as a function?

mcdallas: That's how serde works, you can't write `#[serde(default = "10")]` it needs a function name. https://serde.rs/attr-default.html

PR #2662, opened 2019-03-08
fix the build warning of deprecated trim_right_matches

garyyu:
In Ubuntu 18.04, the building has several warnings like this:
```
warning: use of deprecated item 'core::str::<impl str>::trim_right_matches': superseded by `trim_end_matches`
   --> wallet/src/controller.rs:303:6
    |
303 |                 .trim_right_matches("/")
    |                  ^^^^^^^^^^^^^^^^^^
    |
    = note: #[warn(deprecated)] on by default

```
this PR is a minor fix for this warning.

### DIFF
--- a/api/src/handlers/peers_api.rs
+++ b/api/src/handlers/peers_api.rs
@@ -79,7 +79,7 @@ impl Handler for PeerHandler {
 		}
 	}
 	fn post(&self, req: Request<Body>) -> ResponseFuture {
-		let mut path_elems = req.uri().path().trim_right_matches('/').rsplit('/');
+		let mut path_elems = req.uri().path().trim_end_matches('/').rsplit('/');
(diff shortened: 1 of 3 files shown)


hashmap:
I‘d suggest to start using QueryParams and friends 
https://github.com/mimblewimble/grin/blob/master/api/src/web.rs#L143

garyyu:
@hashmap 

Just find there're two ` path_elems.next()` there! I guess that's why the original author of this piece of code don't use the macro `right_path_element!`:
```
		let mut path_elems = req.uri().path().trim_end_matches('/').rsplit('/');
		let command = match path_elems.next() {
			None => return response(StatusCode::BAD_REQUEST, "invalid url"),
			Some(c) => c,
		};
		let addr = match path_elems.next() {
```
Any further improvement idea?

PR #2672, opened 2019-03-11
feat(ci): reintroduce code coverage

mrekucci:
The build result is determined as soon as all the required jobs
finish, while the "coverage" job may continue and may even fail.


ignopeverell:
Given that we changed CI to Azure pipelines, this will need to be changed as well.

PR #2677, opened 2019-03-14
Add flag to override run_tui setting

pharazen:
Add an additional flag to override TUI settings, needed for headless GUI wallets such as Kingfish and Supergrin.

### DIFF
--- a/src/bin/grin.yml
+++ b/src/bin/grin.yml
@@ -16,6 +16,11 @@ subcommands:
   - server:
       about: Control the Grin server
       args:
+        - tui_enabled:
+            help: Run with or without TUI, bypassing config
+            short: t
+            long: tui_enabled
+            takes_value: true
         - config_file:
             help: Path to a grin-server.toml configuration file
             short: c
--- a/src/bin/cmd/server.rs
+++ b/src/bin/cmd/server.rs
@@ -104,6 +104,10 @@ pub fn server_command(
 			server_config.p2p_config.port = port.parse().unwrap();
 		}
 
+		if let Some(tui_enabled) = a.value_of("tui_enabled") {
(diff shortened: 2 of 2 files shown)


rheza:
Please add it guys, Thanks

pharazen:
The reason to add the command is that GUI wallets (along with other future developments) want to run the node as a sub-process, but can't do this if the TUI is enabled. It's simple enough to edit the toml file, but then if the user also uses the TUI, they will be surprised to find it disabled. Similarly, a user may want to have their TUI disabled by default, but use a script, shortcut, etc to run in TUI mode. It is most flexible to offer the true/false flag and solves the sub-process issue.

src/bin/grin.yml
```diff
@@ -16,6 +16,11 @@ subcommands:
   - server:
       about: Control the Grin server
       args:
+        - tui_enabled:
+            help: Run with or without TUI, bypassing config
+            short: t
+            long: tui_enabled
+            takes_value: true
```
hashmap: Would you consider to not take value? For me`grin -t true` looks awkward comparing to `grin -t`, also boolean value anyway modeled by presence/absence of the argument

ignopeverell: Yup, agreed here.

kargakis:
> but then if the user also uses the TUI, they will be surprised to find it disabled

Not sure I understand this argument. Why would I be surprised to find the TUI disabled in case the GUI process edits the grin server config but not be surprised in case the flag is used? Is this more of a documentation issue for the config file option?

> Similarly, a user may want to have their TUI disabled by default, but use a script, shortcut, etc to run in TUI mode.

Why `sed` is not enough?

ignopeverell:
Closing for lack of progress. Feel free to reopen if you're ready to continue on this.

PR #2679, opened 2019-03-14
Reduce number of unwraps in chain crate

hashmap:


servers/src/grin/sync/syncer.rs
```diff
@@ -209,9 +209,15 @@ impl SyncRunner {
 			}
 		} else {
 			// sum the last 5 difficulties to give us the threshold
-			let threshold = self
-				.chain
-				.difficulty_iter()
+			let diff_iter = match self.chain.difficulty_iter() {
```
antiochp: Note we instantiated the `difficulty_iter` inline prior to this change. So it immediately went out of scope.
The `difficulty_iter` maintains a lock on lmdb db and we need to be _very_ careful with how we use it and in what scope.

Its been the source of a lot of pain and debugging...

That said I'm pretty sure this is safe to do here as we are scoped to the `else { ... }` block.
But it is something we should confirm.

We may want to rework this slightly to ensure the scope of the `difficulty_iter` is more explicit.

hashmap: @antiochp I reduced the scope

PR #2685, opened 2019-03-18
use sandbox folder for txhashset validation on state sync

garyyu:
An improvement on the txhashset validation procedure: use a sandbox folder for new downloaded txhashset files and the validation is based on sandbox folder firstly. The local txhashset will be overwritten ONLY when validation is ok.

The sandbox folder will be cleaned immediately no matter what's the validation result.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -180,7 +180,8 @@ impl Chain {
 			let store = Arc::new(store::ChainStore::new(db_env)?);
 
 			// open the txhashset, creating a new one if necessary
-			let mut txhashset = txhashset::TxHashSet::open(db_root.clone(), store.clone(), None)?;
+			let mut txhashset =
+				txhashset::TxHashSet::open(db_root.clone(), store.clone(), None, false)?;
 
 			setup_head(&genesis, &store, &mut txhashset)?;
 			Chain::log_heads(&store)?;
@@ -842,6 +843,11 @@ impl Chain {
 		}
 	}
 
+	/// Clean the temporary sandbox folder
+	pub fn clean_txhashset_sandbox(&self) {
+		txhashset::clean_txhashset_sandbox(self.db_root.clone());
+	}
+
 	/// Writes a reading view on a txhashset state that's been provided to us.
 	/// If we're willing to accept that new state, the data stream will be
 	/// read as a zip file, unzipped and the resulting state files should be
@@ -863,17 +869,21 @@ impl Chain {
 
 		let header = self.get_block_header(&h)?;
 
-		{
-			let mut txhashset_ref = self.txhashset.write();
-			// Drop file handles in underlying txhashset
-			txhashset_ref.release_backend_files();
-		}
+		// Write hashset to sandbox
+		txhashset::clean_txhashset_sandbox(self.db_root.clone());
+		txhashset::zip_write(
+			self.db_root.clone(),
+			txhashset_data.try_clone()?,
+			&header,
+			true,
+		)?;
 
-		// Rewrite hashset
-		txhashset::zip_write(self.db_root.clone(), txhashset_data, &header)?;
-
-		let mut txhashset =
-			txhashset::TxHashSet::open(self.db_root.clone(), self.store.clone(), Some(&header))?;
+		let mut txhashset = txhashset::TxHashSet::open(
+			self.db_root.clone(),
+			self.store.clone(),
+			Some(&header),
+			true,
+		)?;
 
 		// The txhashset.zip contains the output, rangeproof and kernel MMRs.
 		// We must rebuild the header MMR ourselves based on the headers in our db.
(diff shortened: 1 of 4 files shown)


chain/src/chain.rs
```diff
@@ -882,6 +892,26 @@ impl Chain {
...
 
+		// sandbox validation ok, go to overwrite txhashset and clean the sandbox
+		{
+			// Before overwriting, drop file handles in underlying txhashset
+			{
+				let mut txhashset_ref = self.txhashset.write();
+				txhashset_ref.release_backend_files();
+			}
+
+			txhashset::zip_write(self.db_root.clone(), txhashset_data, &header, false)?;
```
antiochp: 1. We extract the zip file out into the sandbox.
2. Then validate the contents of the sandbox.
3. Then if valid we (re)extract the zip file out into the real location.

(3) Feels wrong to me. I thinkwe should take the _validated_ contents of the sandbox and copy them across to the txhashset dir. We should not (re)extract the zipfile.

The 2nd extract step just increases the possibility of us doing it _slightly_ differently the 2nd time, particularly with us passing in a different flag (true/false) the 2nd time.

Proposal - 

1. Extraction into sandbox
2. Validate sandbox
3. Copy across validated files from sandbox into txhashset dir

chain/src/chain.rs
```diff
@@ -863,17 +869,21 @@ impl Chain {
...
-
-		// Rewrite hashset
-		txhashset::zip_write(self.db_root.clone(), txhashset_data, &header)?;
-
-		let mut txhashset =
-			txhashset::TxHashSet::open(self.db_root.clone(), self.store.clone(), Some(&header))?;
+		// Write hashset to sandbox
+		txhashset::clean_txhashset_sandbox(self.db_root.clone());
+		txhashset::zip_write(
+			self.db_root.clone(),
```
antiochp: Can we more explicitly separate the `sandbox` from the actual `txhashset` dir?
I don't think these should share the same `db_root` - its going to be easy for someone to unintentionally change this code in the future without realizing they are now writing to the _real_ txhashset dir.

chain/src/txhashset/txhashset.rs
```diff
@@ -41,6 +41,7 @@ use std::time::{Instant, SystemTime, UNIX_EPOCH};
 
 const HEADERHASHSET_SUBDIR: &'static str = "header";
 const TXHASHSET_SUBDIR: &'static str = "txhashset";
+const TXHASHSET_SANDBOX_SUBDIR: &'static str = "txhashset_sandbox";
```
antiochp: See earlier comment above.

Proposal: The "sandbox" should be under $TMPDIR and not under in the regular `db_root`.
The _only_ files saved to disk under `db_root` should be validated.

chain/src/chain.rs
garyyu: Accept.
Change from (re)extract to folder move (rename).

chain/src/chain.rs
garyyu: accept.
changed to `env::temp()` folder for sandbox.

chain/src/txhashset/txhashset.rs
garyyu: done with `env::temp` folder.

chain/src/txhashset/txhashset.rs
```diff
@@ -115,12 +114,12 @@ impl TxHashSet {
 		root_dir: String,
 		commit_index: Arc<ChainStore>,
 		header: Option<&BlockHeader>,
-		sandbox: bool,
+		sandbox_dir: Option<String>,
```
antiochp: We can just pass _either_ `root_dir` _or_ `sandbox_dir` as the first param and we don't need to handle the precedence rule here.

garyyu: Thanks @antiochp for the review.

You have `header/header_head` and `header/sync_head/` reading on this `open()` function,  but `header/` is at same level as `txhashset/` so surely it doesn't exist in sandbox folder.

That's why I keep this first `root_dir` param. 

Could you explain more about why we need put `header\` also into the `sandbox` folder?

```
			header_pmmr_h: PMMRHandle::new(
				&root_dir,
				HEADERHASHSET_SUBDIR,
				HEADER_HEAD_SUBDIR,
				false,
				None,
			)?,
			sync_pmmr_h: PMMRHandle::new(
				&root_dir,
				HEADERHASHSET_SUBDIR,
				SYNC_HEAD_SUBDIR,
				false,
				None,
			)?,
```

garyyu: Ah, perhaps that `rebuild` did the job :-)
```
		self.rebuild_header_mmr(&Tip::from_header(&header), &mut txhashset)?;
```

garyyu: but then what's the best to handle this `header/` folder in `sandbox`?
1. also move to the db root
2. call a `rebuild_header_mmr()` again after moving the `txhashset/` folder

Which one do you prefer?  Is the `rebuild_header_mmr()` a fast or slow processing?

garyyu: Tried with `moving the header/ folder`, but see a lot of `Corrupted storage` log which could be caused by not-atomic processing on store during `txhashset_write` processing?

Switch to `rebuild_header_mmr()` solution now.
And regarding the speed, the `rebuilding` spent 10 seconds on my slow mac air for current main chain height. not bad.

antiochp: Ahh ok.

```
chain_data/txhashset <- txhashset.zip unzips to here
chain_data/header <- local data only, not included in the zip file
```

The original intent was to keep the `header` dir out the `txhashset` dir to make it more explicit what originated from `txhashset.zip` and what was purely built locally.

Maybe we consider making the sandbox dir a couple of levels deep?

So we can have something like - 

```
chain_data/txhashset
chain_data/header
```

```
$sandbox_dir/txhashset
$sandbox_dir/header
```

Would this work?

I don't see we would not be able to copy across both the `txhashset` dir _and_ the `header` dir on successful validation.
A txhashset will have file references to files under both these dirs so we'd need to be careful with ordering.
i.e. copy them across then reopen the newly updated txhashset.

garyyu: @antiochp I don't know why, I tried the `header` dir moving on https://github.com/mimblewimble/grin/pull/2685/commits/2a6182b1a6d401fbde72457d1a8327c48b88c319, but there were a lot of `Corrupted storage` logs. So I switch to `rebuild_header_mmr` solution.

PR #2686, opened 2019-03-18
catch the panic to avoid peer thread quit early

garyyu:
For example, if a peer deserve a ban, we will lose the chance if the thread panic early.

Since the `ban` only happen after a success return from a processing function, for example on message `TxHashSetArchive`.

p2p/src/peers.rs:
```
	fn txhashset_write(&self, h: Hash, txhashset_data: File, peer_addr: PeerAddr) -> bool {
		if !self.adapter.txhashset_write(h, txhashset_data, peer_addr) {
			...
			self.ban_peer(peer_addr, ReasonForBan::BadTxHashSet);
			false
		} else {
			true
		}
	}
```

There could be more cases for such kind of panic on message processing functions, please let me know if you find a better way to give a universal panic catching on those processing functions.

Or at least one catch for one message (if needed), like this PR.

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -35,6 +35,7 @@ use grin_store;
 use grin_store::pmmr::{PMMRBackend, PMMR_FILES};
 use std::collections::HashSet;
 use std::fs::{self, File};
+use std::panic;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::time::{Instant, SystemTime, UNIX_EPOCH};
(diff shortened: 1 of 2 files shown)


chain/src/txhashset/txhashset.rs
```diff
@@ -1456,8 +1457,31 @@ pub fn zip_write(
...
 	fs::create_dir_all(txhashset_path.clone())?;
-	zip::decompress(txhashset_data, &txhashset_path, expected_file)
-		.map_err(|ze| ErrorKind::Other(ze.to_string()))?;
+
+	// catch the panic to avoid the thread quit
+	{
+		panic::set_hook(Box::new(|_info| {
+			// do nothing
+		}));
+		let result = panic::catch_unwind(|| {
```
antiochp: `zip::decompress` is our code (in the `util` crate).
Would it make sense to push this `catch` wrapper logic down into the util crate and make this a more general purpose solution?

I'm not sure I really like having this low-level `catch_unwind` in our `txhashset` code.

garyyu: Accept.
Move catch to the `util` crate.

PR #2698, opened 2019-03-21
[1.1.0] (JSON only) Serialization of TxKernel u64s as strings

yeastplume:
As part of wallet V2 API, I'm ensuring that all u64 fields returned by the API are serialized into JSON as strings as opposed to literals, as it's not guaranteed that consumers will know what to do with a u64 literal (javascript, for instance). TxKernel is embedded in the JSON slate, so the custom serializer needs to be added to it.

Note that the API will still be able to accept literals in u64 fields on deserialization, it will just produce them as strings by default on serialization.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -159,9 +159,11 @@ pub struct TxKernel {
 	/// Options for a kernel's structure or use
 	pub features: KernelFeatures,
 	/// Fee originally included in the transaction this proof is for.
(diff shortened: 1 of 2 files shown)


DavidBurkett:
Why not just require it to always be a string? Since this is a 'V2' API, there's no requirement of backward compatibility.

yeastplume:
It's just as easy to accept both and might remove an annoyance for clients that can deal with u64 literals

PR #2705, opened 2019-03-25
Add some better logging for get_outputs_by_id failure states

rentenmark:
We are having some issues with calls to `get_output_by_ids` failing when the wallet is attempting a refresh as well as when creating a new partial tx. This leads to a stale wallet balance and inability to send funds. Our current workaround is to restore from seed words. This leads to a working state for about 12 hours at which point we need to do another restore.


api/src/handlers/chain_api.rs
```diff
@@ -106,9 +106,10 @@ impl OutputHandler {
 
 		let mut outputs: Vec<Output> = vec![];
 		for x in commitments {
-			if let Ok(output) = self.get_output(&x) {
-				outputs.push(output);
-			}
+			match self.get_output(&x) {
+				Ok(output) => outputs.push(output),
+				Err(e) => error!("{}", e),
```
hashmap: would you consider adding some context to the error message (here and below)?

rentenmark: 👍

rentenmark:
@hashmap thanks for your review, I updated based on your comment

PR #2706, opened 2019-03-25
Evict transaction from txpool

### LINKED ISSUE #1417: Evict transactions from the pool by order of fee/weight

quentinlesceller:
Replace #1898 and fix #1417.
Not tested yet. 
When trying to add a transaction to the `tx_pool`, if the latter is full attempt to remove a transaction from it.
Will remove the most recent, low fee weight ratio transaction that has no children transactions.


core/src/core/transaction.rs
```diff
@@ -957,6 +957,11 @@ impl Transaction {
 		self.body.body_weight()
 	}
 
+	/// Calculate fee weight ratio
+	pub fn fee_weight_ratio(&self) -> u64 {
```
ignopeverell: Maybe a little more doc explaining the unit?

pool/src/pool.rs
```diff
@@ -427,6 +427,47 @@ impl Pool {
 		tx_buckets
 	}
 
+	/// Get a list of transaction that are potentially evictable, i.e. only transaction with no children
+	pub fn get_evictable_transactions(&self) -> Vec<Transaction> {
```
ignopeverell: Why not rely on `bucket_transactions`? Doesn't seem a lot more costly and would avoid all that duplication.

pool/src/transaction_pool.rs
```diff
@@ -184,6 +196,50 @@ impl TransactionPool {
...
+				Ordering::Greater
+			} else if fw_a == fw_b && entry_a.tx_at < entry_b.tx_at {
+				Ordering::Greater
+			} else {
+				Ordering::Less
+			}
+		});
+
+		// Compute a list of parent transaction
+		let evictable_transactions = self.txpool.get_evictable_transactions();
```
ignopeverell: Maybe relying on whole buckets would allow you to evict more than childless transactions?

ignopeverell:
Ping @quentinlesceller

antiochp:
Agree with @ignopeverell here - the underlying txpool code has been rewritten (and rewritten again) and I _think_ `bucket_transactions()` is almost exactly what you need here, each bucket being a single tx or potentially a small set of dependent txs, in each case a good candidate for eviction after we sort the buckets as desired.

quentinlesceller:
@ignopeverell @antiochp Okay I've been thinking about this and the best idea I have in mind is the following:

1. Whenever we receive a stem transaction, we check if adding it would overflow the current pool capacity.
1. Add the transaction to our transaction pool.
1. If overflow then `bucket_transactions` and remove the last one based on ordering  fee/weight ratio and date.

A. Nice things with this is that you can actually bump your bucket in the transaction pool.
B. Weird thing is that simply adding another transaction to a bucket can result in removing the whole bucket (if it is the last one). 
An additional check could be to check whether the fee/weight ratio of the new transaction is actually above the last bucket fee/weight ratio regardless of the fact that this new transaction belongs to an existing bucket (sorry for the long sentence 😅 ). But that would defeat A in some cases.
What do you think?

antiochp:
Thinking about what you're proposing - let me get back to you (please ping me if I forget).

antiochp:
> A. Nice thing with this is that you can actually bump your bucket in the transaction pool.

This is conceptually similar to CPFP in Bitcoin I believe. If you create a new 0-conf tx that will be aggregated to a tx in an existing bucket then you can increase the aggregate fee and the bucket tx will confirm faster (assuming a healthy fee market).

> B. Weird thing is that simply adding another transaction to a bucket can result in removing the whole bucket (if it is the last one).

Yeah this does not feel ideal to me. Seems like I could pay you, then basically prevent that tx ever getting confirmed by spending 0-conf with low fee txs dependent on it. I don't feel like I have a good intuitive understanding of if this is actually a problem though in reality.

> An additional verification could be to check whether the fee/weight ratio of the new transaction is above the last bucket fee/weight ratio regardless of the fact that this new transaction belongs to an existing bucket (sorry for the long sentence 😅 ). But that would defeat A. in some cases.
What do you think?

That may actually be the simplest solution to this. Basically the fee/weight ratio would always need to increase (or at least never decrease) when aggregating new txs into an existing bucket.

I'm not clear on how "But that would defeat A." would ever occur like this though? For (A) you would always need to _increase_ the fee/weight ratio to actually get the tx to confirm anyway?

antiochp:
Maybe the implementation of adding to a bucket requires fee/weight to not decrease?
If it would decrease we do not add to existing bucket, but create a new (single tx) bucket.
Then the ordering of these buckets by fee/weight would basically do the right thing when deciding to evict a bucket?

quentinlesceller:
Closing for https://github.com/mimblewimble/grin/pull/2797.

PR #2707, opened 2019-03-25
Reduce number of unwarps in servers crate

hashmap:
It doesn't include stratum server which is sufficiently changed in 1.1
branch and adapters, which is big enough for a separate PR.

### DIFF
--- a/pool/src/types.rs
+++ b/pool/src/types.rs
@@ -59,6 +59,24 @@ pub struct DandelionConfig {
 	pub stem_probability: Option<usize>,
 }
 
+impl DandelionConfig {
+	pub fn relay_secs(&self) -> u64 {
+		self.relay_secs.unwrap_or(DANDELION_RELAY_SECS)
+	}
+
+	pub fn embargo_secs(&self) -> u64 {
+		self.embargo_secs.unwrap_or(DANDELION_EMBARGO_SECS)
+	}
+
+	pub fn patience_secs(&self) -> u64 {
+		self.patience_secs.unwrap_or(DANDELION_PATIENCE_SECS)
+	}
+
+	pub fn stem_probability(&self) -> usize {
+		self.stem_probability.unwrap_or(DANDELION_STEM_PROBABILITY)
+	}
+}
+
 impl Default for DandelionConfig {
 	fn default() -> DandelionConfig {
 		DandelionConfig {
--- a/pool/src/types.rs
+++ b/pool/src/types.rs
@@ -44,37 +44,19 @@ const DANDELION_STEM_PROBABILITY: usize = 90;
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct DandelionConfig {
 	/// Choose new Dandelion relay peer every n secs.
-	#[serde = "default_dandelion_relay_secs"]
-	pub relay_secs: Option<u64>,
+	#[serde(default = "default_dandelion_relay_secs")]
+	pub relay_secs: u64,
 	/// Dandelion embargo, fluff and broadcast tx if not seen on network before
 	/// embargo expires.
-	#[serde = "default_dandelion_embargo_secs"]
-	pub embargo_secs: Option<u64>,
+	#[serde(default = "default_dandelion_embargo_secs")]
+	pub embargo_secs: u64,
 	/// Dandelion patience timer, fluff/stem processing runs every n secs.
 	/// Tx aggregation happens on stem txs received within this window.
-	#[serde = "default_dandelion_patience_secs"]
-	pub patience_secs: Option<u64>,
+	#[serde(default = "default_dandelion_patience_secs")]
+	pub patience_secs: u64,
 	/// Dandelion stem probability (stem 90% of the time, fluff 10% etc.)
-	#[serde = "default_dandelion_stem_probability"]
-	pub stem_probability: Option<usize>,
-}
-
-impl DandelionConfig {
-	pub fn relay_secs(&self) -> u64 {
-		self.relay_secs.unwrap_or(DANDELION_RELAY_SECS)
-	}
-
-	pub fn embargo_secs(&self) -> u64 {
-		self.embargo_secs.unwrap_or(DANDELION_EMBARGO_SECS)
-	}
-
-	pub fn patience_secs(&self) -> u64 {
-		self.patience_secs.unwrap_or(DANDELION_PATIENCE_SECS)
-	}
-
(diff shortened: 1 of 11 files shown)


JeremyRubin:
General concept ACK, good cleanup, looks correct.

I think it might make sense to get rid of the option type entirely in Dandelion config -- can these ever be None now?

I don't love the unwrap_retry macro, but it works and I don't have a better suggestion. One addressable concern is that it's possible to enter an infinite loop (v.s. just crashing). May also be better to do an exp backing on the timeout up to 1 sec, as 1 sec is kinda slow. Then again, I don't think we're expecting these to fail (ever) given they used to be unwraps...

nit: in the future, would be easier for me to review if you do one commit per module being refactored

hashmap:
@JeremyRubin thanks for the review!

>  think it might make sense to get rid of the option type entirely in Dandelion config 

My understanding is that it may be needed, because we deserialize toml file into this structure, some of fields may be missing, perhaps we should make fields private and treat it as internal representation

> One addressable concern is that it's possible to enter an infinite loop (v.s. just crashing).

That's what I was trying to prevent. This situation (eg `chain.head()`) shouldn't happen but we experienced it when all LMDB readers where taken by a large number of peers (we fixed it, shouldn't happen until the next time). Exponential backoff is a good suggestion but may be an overkill for this case.

>  in the future, would be easier for me to review if you do one commit per module being refactored

fair point, need to find a balance between size of a change and compactness (in topological sense). Some changes like introducing of `Result` bubble up.

JeremyRubin:
Hm. The deserialize argument is interesting, for which serde allows you to have a default value should the field be missing. 


Good point on the pausing if the failure is recoverable. I'm just worried about the case where failing is not recoverable and restarting is better, e.g., if we're running as a daemon and can restart (and make some progress) or if it's safest to just shut a malfunctioning node off completely. Given the tradeoffs in cryptocurrency, I'd rather shut my node down completely until manual intervention if it's not just a temporary resource limit induced race condition. 


Re the nit, certainly not several PRs for this, but just commits. Im not sure if we adhere to 'every commit compiles' but I think it can be nice for reviewers to have the interface change separate from the refactorings that can be more quickly scanned over. I always prefer more small commits as it is easier to bisect on if an issue comes up and can always be squashed on merge anyways. But this is off topic for here -- can chat about it out of band.

pool/src/types.rs
```diff
@@ -44,37 +44,19 @@ const DANDELION_STEM_PROBABILITY: usize = 90;
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct DandelionConfig {
 	/// Choose new Dandelion relay peer every n secs.
-	#[serde = "default_dandelion_relay_secs"]
-	pub relay_secs: Option<u64>,
+	#[serde(default = "default_dandelion_relay_secs")]
```
antiochp: This is going to cause all kinds of conflicts with 1.1.0 branch.
It might be worth not touching the config here and just doing it on the 1.1.0 branch.
We fixed the "default" serde config like you do here, but we did not go the additional step and get rid of the options (which I agree is :+1:).

hashmap: @antiochp ok, I've reverted this commit

antiochp: I suspect `#[serde = "foo"]` vs. `#[serde(default = "foo")]` has bitten us and continues to bite us all over our config parsing code.

servers/src/grin/sync/state_sync.rs
```diff
@@ -166,9 +166,25 @@ impl StateSync {
...
+					p2p::Error::Internal
+				})?;
 			for _ in 0..threshold {
-				txhashset_head = self.chain.get_previous_header(&txhashset_head).unwrap();
+				txhashset_head = self
+					.chain
+					.get_previous_header(&txhashset_head)
+					.map_err(|e| {
+						error!(
+							"chain error dirung getting a previous block header {}: {:?}",
```
antiochp: Unrelated - typo.

servers/src/grin/sync/syncer.rs
```diff
@@ -147,13 +161,13 @@ impl SyncRunner {
 			}
 
 			// if syncing is needed
-			let head = self.chain.head().unwrap();
+			let head = unwrap_or_retry!(self.chain.head());
```
antiochp: The syncer already loops and implicitly retries on next loop.
So it feels like we should just break out of this loop early if we fail any of these calls that _could_ potentially fail?

I'm not convinced we need to handle these individual failures at this level of granularity like this.

antiochp: Maybe one option would be to just kill the sync thread on any kind of (retry-able) failure and kick off a new one to replace it?

hashmap: What do you mean by granularity? I think we should handle error cases using `unwarap` as the last resort, which is quite granular.

Unwrap kills a thread silently, unless we expect it. Another option is to return `Result` from syncer and handle on the upper level, but to me it seems too complicated

hashmap:
@antiochp  what would be  advantage of relaunching of `sync_loop`? I mean it's easy to return `Result` and replace `continue` with `return Err()` but what would we get from it? My main goal is to not be in a situation when we have a node without sync thread, I'm fine with both tactical decisions in this particular case.

servers/src/grin/sync/syncer.rs
antiochp: Ahh I think I was misunderstanding what `unwrap_or_retry!()` was doing. I thought it was retrying the thing being wrapped in the macro. It's actually `continue` on the outer loop.

> Another option is to return Result from syncer and handle on the upper level, but to me it seems too complicated

I think this might be easier to read though.
We spawn a simple loop that handles errors (by logging errors and sleeping).
And the loop calls `sync_once()` or something similar returning `Result`.

ignopeverell:
@antiochp okay with the PR after @hashmap response? It'd also be easy going forward to lose information about the original error if we reduce the granularity.

ignopeverell:
Btw thanks for the review as well @JeremyRubin!

Make the server warping again. :sparkle:

servers/src/grin/sync/syncer.rs
kargakis: > Ahh I think I was misunderstanding what unwrap_or_retry!() was doing. I thought it was retrying the thing being wrapped in the macro. It's actually continue on the outer loop.

Yeah, this wasn't obvious to me either - either follow @antiochp's suggestion or add a comment in the continue statement in the macro that it's meant to continue the sync loop.

PR #2714, opened 2019-03-27
[PENDING APPROVAL] put phase outs of C32 and beyond on hold

tromp:
Limit Cuckatoo phaseouts to C31. This is a hard forking change that only takes effect when C32 would originally phase out in early 2021.


yeastplume:
@tromp Should we get this in for 2.0.0?

tromp:
At least a decision on this should be taken before freezing 2.0.0 as it requires an 18 month advance notice on Cuckatoo changes and includes such a change taking effect in Jan 2021.

I would very much like to see it approved. Can we decide in the next Gov meeting?

Issue #2709, opened 2019-03-25
Clarify /tmp use


kargakis:
Is it necessary to use /tmp for downloading txhashsets?

See https://www.grin-forum.org/t/chain-building-stuck-at-58/4553 for more info

antiochp:
Related - https://github.com/mimblewimble/grin/pull/2685

There is no specific reason we went with `$TMPDIR` beyond using a standard location. But as mentioned in the forum we have some fairly specific disk space requirements so want to ensure we use a disk with sufficient space.

It sounds like we may want to consider using a grin specific `tmp` dir under `~/home/.grin/main` or something similar?

I do think we need to segregate the `sandbox_dir` and/or `tmpdir` from our mainnet/floonet `chain_data` or `chain_data/txhashset` dirs.

kargakis:
At a minimum we should document this somewhere prominently.

garyyu:
Interesting!
Agree to use a grin specific `tmp` dir under `~/.grin/main/tmp` for mainnet and `~/.grin/floo/tmp` for floonet.
Also a specific `sandbox` dir under `~/.grin/main/tmp/txhashset`.

It deserve a dedicated PR to fix all these `folder`  name / location issues. And let #2685 purely only for the sandbox fixing.

kargakis:
Fixed by https://github.com/mimblewimble/grin/pull/2772

PR #2715, opened 2019-03-28
use a grin specific tmp dir instead of a system tmp dir

garyyu:
Fix for https://github.com/mimblewimble/grin/issues/2709

And improvement according to some recommendations:
- Randomize temporary file names
- Only write to files in directories owned and writable by the node user
- Using the ​`create_new`​ option from `​std::fs::OpenOptions`​ instead of `create`

TODO:
- [X] After https://github.com/mimblewimble/grin/pull/2685 merged, the `sandbox` dir also need switch to grin specific tmp dir.


p2p/src/protocol.rs
```diff
@@ -288,10 +288,24 @@ impl MessageHandler for Protocol {
 				self.adapter
 					.txhashset_download_update(download_start_time, 0, sm_arch.bytes);
 
-				let mut tmp = env::temp_dir();
-				tmp.push("txhashset.zip");
+				let mut tmp = self.adapter.get_tmp_dir();
```
ignopeverell: Perhaps now that logic would fit better in the chain, given that it already holds the preparation of the directory location?

servers/src/common/adapters.rs
```diff
@@ -369,6 +370,19 @@ impl p2p::ChainAdapter for NetToChainAdapter {
 			true
 		}
 	}
+
+	/// Specific tmp dir.
+	/// Normally it's ~/.grin/main/tmp for mainnet
+	/// or ~/.grin/floo/tmp for floonet
+	fn get_tmp_dir(&self) -> PathBuf {
```
ignopeverell: Why not delegate to the chain implementation here? Am I missing something?

garyyu: accept

p2p/src/protocol.rs
garyyu: What do you mean?  moving `save_txhashset_to_file()` function into `chain`?

yeastplume:
Something else needs to be verified here, when trying to test 1.1.0 I came across an issue on this line:

https://github.com/mimblewimble/grin/blob/325e32821dad60c0509ba938e7b0f5001aaa161c/chain/src/txhashset/txhashset.rs#L1466

The error I'm getting is:

`20190402 08:09:51.600 ERROR grin_chain::txhashset::txhashset - hashset_replace fail on txhashset. err: Invalid cross-device link (os error 18)`

What's actually happening is that the tmp dir is a seperate tmpfs mount, so rename won't work across mounted devices. This fix definitely needs to go into the 1.1.0 release.

p2p/src/protocol.rs
garyyu: @ignopeverell  remind ^ above question.

ignopeverell: I meant that whole code section, with getting the tmp dir, checking if it exists, creating it if not, appending file name, etc.

garyyu: done. moved the logic to chain

garyyu:
@ignopeverell  ready for another review, thanks

ignopeverell:
Given where we're at, how about targeting this for 1.1.0?

garyyu:
ok, how to move an existing PR to another branch?

garyyu:
Replaced by #2753 for branch 1.1.0

Issue #2542, opened 2019-02-08
Simplify error management


hashmap:
Currently we have one or multiple error enum or struct per crate. Major part of such errors are wrappers on top of error variants from other crates. As result we have a huge graph of dependencies between errors which may be a good playground for a cuckoo solver.  Here is an example:

```
$ rg "keychain::Error"
wallet/src/error.rs
44:     Keychain(keychain::Error),
180:impl From<keychain::Error> for Error {
181:    fn from(error: keychain::Error) -> Error {

pool/src/types.rs
189:    Keychain(keychain::Error),
223:impl From<keychain::Error> for PoolError {
224:    fn from(e: keychain::Error) -> PoolError {

chain/src/error.rs
76:     Keychain(keychain::Error),
218:impl From<keychain::Error> for Error {
219:    fn from(error: keychain::Error) -> Error {

core/src/core/committed.rs
28:     Keychain(keychain::Error),
43:impl From<keychain::Error> for Error {
44:     fn from(e: keychain::Error) -> Error {

core/src/core/transaction.rs
75:     Keychain(keychain::Error),
138:impl From<keychain::Error> for Error {
139:    fn from(e: keychain::Error) -> Error {

core/src/core/block.rs
61:     Keychain(keychain::Error),
99:impl From<keychain::Error> for Error {
100:    fn from(e: keychain::Error) -> Error {

wallet/src/libwallet/error.rs
61:     Keychain(keychain::Error),
254:impl From<keychain::Error> for Error {
255:    fn from(error: keychain::Error) -> Error {

core/src/libtx/error.rs
37:     Keychain(keychain::Error),
94:impl From<keychain::Error> for Error {
95:     fn from(error: keychain::Error) -> Error {
```
We have a lot of redundant code in each crate which converts error from other crates. At the same time the entire error management is pretty monolithic, single call can use errors from multiple crates.

Another issue that sometimes we cut backtrace on error conversion (`From` invocation).

I have a bit controversial proposal to extract errors into a separate crate, flatten (e.g to have single keychain error) and use it from all crates. We still have ability to add  a context specific for a particular call. It's just to start discussion, I'm sure we can find a better solution than we have now.

hashmap:
I checked some multi-crate projects, found some common patterns, here are my suggestions:

[...]

PR #2722, opened 2019-04-01
Return Result from methods of ChainAdapter

hashmap:
Most of the methods return nothing or bool which is used to decide if a
sender of a message should be banned or not. However underlying chain
implementation may fail so we need a way to reflect this fact in API.

Also it allows to reduce number of unwraps and makes the code more robust.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -853,6 +853,7 @@ dependencies = [
  "bytes 0.4.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "chrono 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "enum_primitive 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "grin_chain 1.0.2",
  "grin_core 1.0.2",
  "grin_pool 1.0.2",
  "grin_store 1.0.2",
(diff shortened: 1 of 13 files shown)


JeremyRubin:
Concept Ack.

I'm not in love with introducing a dependency from p2p to chain module for two reasons:

1) what if these methods should return a different error type in the future
2) it introduces a new dependency on a bunch of other stuff (including the keychain module)

 Is there an alternative approach -- can we move chain's error type to it's own module or maybe provide a module for cross-module errors (containing a enum with a variant for each module, whereas the `From<MyModuleError> for AllModulesError` wraps the serialized error in the appropriate variant?

hashmap:
@JeremyRubin  I had the same feelings and tried different approaches, but haven't find anything more reasonable than just use `chain::Error`. My main rationale that the code must be candid.

We deal with `ChainAdapter`, so we should accept the fact that we have a dependency on `chain` crate. The methods should return `chain::Error` for the same reason. Actually I changed the trait first and then worked with implementations, so it was a design decision.

I think we should tackle error management in a more holistic way, I created an issue to tack it a while ago https://github.com/mimblewimble/grin/issues/2542 and I had the same idea of error crate to rule them all. To be fair I don't think it's a right approach anymore,  but not sure.

garyyu:
@hashmap 

> I think we should tackle error management in a more holistic way, I created an issue to tack it a while ago #2542 and I had the same idea of error crate to rule them all.
👍 

> To be fair I don't think it's a right approach anymore, but not sure.

what's this `it`? ^   
#2542 or this PR?

hashmap:
@garyyu the idea to have a designated error crate I proposed as an option in #2542

PR #2724, opened 2019-04-02
Add first_seen field to LivePeerInfo

hashmap:
Totally nice to have, I personally found it useful. Also could be used
to support FIFO peer list.

Also some refactoring


hashmap:
@antiochp correct

PR #2731, opened 2019-04-04
Fix misuse of NoSync state

hashmap:
NoSync is basically Running state, but for some (historical?) reason we use it as a substitute of Initial state. The most annoying issue is chain reset every 5-10 minutes or so. This is a really expensive operations, not sure how safe it is.

### DIFF
--- a/servers/src/common/types.rs
+++ b/servers/src/common/types.rs
@@ -316,7 +316,7 @@ pub struct SyncState {
 }
 
 impl SyncState {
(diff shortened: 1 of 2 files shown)


antiochp:
> The most annoying issue is chain reset every 5-10 minutes or so.

Can you explain this more? I don't think all nodes see this (mine don't).

hashmap:
I misunderstood the problem, opened an issue to track it https://github.com/mimblewimble/grin/issues/2732

PR #2736, opened 2019-04-08
We bumped version to 1.0.3 earlier but missed the Cargo.lock file

antiochp:
Looks like we missed updating `Cargo.lock` when we bumped version numbers from `1.0.2` -> `1.0.3`.


antiochp:
Going to merge this into `master`.
Any build right now is updating `Cargo.lock` and causing diffs/conflicts locally.

Issue #1491, opened 2018-09-09
Duplicate commit for potential coinbase detected. Trying next derivation.


bladedoyle:
I am seeing this "duplicate commit" error frequently.  In some cases it will loop a few times trying new derivations and then continue, but other times it gets stuck in this retry loop forever and the node needs to be restarted.

https://ipfs.io/ipfs/QmZDXDCJzd5cBmFifuKMQKZYdJ5gVsnH3Vd8HFj4zTCTHa
https://ipfs.io/ipfs/QmcByH7AT9RJVhNVjUqoHJMaEFtjmC4SoVeMhdtzXgoTQo

Its possible that this issue is contributing to high numbers of "late" shares:  
https://github.com/mimblewimble/grin/issues/1216

yeastplume:
A lot of work has gone into this recently, particularly in the case of multiple wallets using the same seed. Please reopen if you see similar on mainnet

PR #2737, opened 2019-04-09
Use next available key on duplicate coinbase commitment

bmhjyu:
If a coinbase commitment hits a duplication and there is no transactions to mine, it is possible that coinbase cannot be built for quite a long time. This change tries to build coinbase with an empty key identifier immediately in this case. The listening wallet will then use the next available key to build a coinbase commitment, which is different from the previous one and is unlikely to hit the duplication.

### DIFF
--- a/servers/src/mining/mine_block.rs
+++ b/servers/src/mining/mine_block.rs
@@ -51,12 +51,15 @@ pub fn get_block(
 		wallet_listener_url.clone(),
 	);
 	while let Err(e) = result {
+		let mut new_key_id = key_id.to_owned();
 		match e {
 			self::Error::Chain(c) => match c.kind() {
 				chain::ErrorKind::DuplicateCommitment(_) => {
 					debug!(
 						"Duplicate commit for potential coinbase detected. Trying next derivation."
 					);
+					// use the next available key to generate a different coinbase commitment
+					new_key_id = None;
 				}
 				_ => {
 					error!("Chain Error: {}", c);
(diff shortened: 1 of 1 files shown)


garyyu:
Thanks for your 1st PR 👍 
I'm afraid I don't understand here `it is possible that coinbase cannot be built for quite a long time.`, do you see the real case?  and why `quite a long time`? I see there just a 100ms sleep there.

Could you please share more info?

garyyu:
perhaps https://github.com/mimblewimble/grin/issues/1491 related.

servers/src/mining/mine_block.rs
```diff
@@ -51,12 +51,15 @@ pub fn get_block(
...
 	while let Err(e) = result {
+		let mut new_key_id = key_id.to_owned();
 		match e {
 			self::Error::Chain(c) => match c.kind() {
 				chain::ErrorKind::DuplicateCommitment(_) => {
 					debug!(
 						"Duplicate commit for potential coinbase detected. Trying next derivation."
 					);
+					// use the next available key to generate a different coinbase commitment
+					new_key_id = None;
```
garyyu: 👍

bmhjyu:
> Thanks for your 1st PR 
> I'm afraid I don't understand here `it is possible that coinbase cannot be built for quite a long time.`, do you see the real case? and why `quite a long time`? I see there just a 100ms sleep there.
> 
> Could you please share more info?

We have met this in production some time (stratum server stuck for hours). The commitment seems only depend on the amount and the key itself. If we hit the duplication and there is no transactions, the amount will always be just the coinbase reward itself. If we do not change the key, the commitment does not change until there is a transaction (amount changes as there are transaction fees).

PR #2739, opened 2019-04-09
correct a mistake in intro

dumingxiao:
k1 is the private key. We should sign a transaction with the private key.


garyyu:
Normally, when we say `sign sth with P`, if `P` here is a public key, we know what it means.
Does this indeed cause some people confusing?

dumingxiao:
> Normally, when we say `sign sth with P`, if `P` here is a public key, we know what it means.
> Does this indeed cause some people confusing?
Sorry, did I misunderstand the description? Here you want to use a public key to sign as a new private key?

dumingxiao:
> Normally, when we say `sign sth with P`, if `P` here is a public key, we know what it means.
> Does this indeed cause some people confusing?

Sorry, did I misunderstand the description? Here you want to use a public key to sign as a new private key?

jaspervdm:
To verify the signature you only need `x*G` so I think its valid to see the signature as one for `x*G`. The phrases "sign with `x*G`" and "sign with `x`" usually mean the same thing and are used interchangably.

Issue #2691, opened 2019-03-19
Investigate alternatives to TravisCI 


lehnberg:
As discussed in the last dev meeting. 

## Motivation
* See https://github.com/mimblewimble/grin/issues/2667
* Current windows build process are highly manual where you have to temporarily remove the github token from the project.
* [TravisCI was acquired by Ibera](https://techcrunch.com/2019/01/23/idera-acquires-travis-ci/) end of Jan 2019.
* Ibera sent a wave of [termination letters to Travis CI employees](
https://www.reddit.com/r/devops/comments/at3oyq/it_looks_like_ibera_is_gutting_travis_ci_just_a/) end of Feb 2019. There is no mention of shutting down Travis CI, or shutting it down for Open Source projects, however the loss of experienced employees is unlikely to help stabilise the platform, or get bugs fixed. 

## Candidate discussions
* https://github.com/ligurio/awesome-ci
* https://internals.rust-lang.org/t/which-ci-platform-should-rust-use/9322/29

### Gitlab
We have access to https://gitlab.com/groups/mimblewimble

It seems it can be integrated with Github:
https://about.gitlab.com/solutions/github/

... And that this is free for open source projects:
> If you have a public, open source project on GitHub you can now take advantage of free CI/CD on GitLab.com. As part of our commitment to open source, we offer all public projects our highest tier features (Gold) for free. While other CI/CD vendors limit you to running a handful of concurrent jobs, GitLab.com gives open source projects hundreds of concurrent jobs with 50,000 free CI pipeline minutes.

### Microsoft Azure Pipelines
https://news.ycombinator.com/item?id=18983586

### [DroneCI](https://drone.io)

[Free cloud hosting](https://cloud.drone.io) for open source projects

### [Codefresh?](https://codefresh.io/)

---

## General questions

1. What's a sensible first step here?
1. How have other orgs done it? 
1. What are the workflows in practice?

sesam:
> the owner of https://gitlab.com/groups/mimblewimble/-/group_members,
>
Who can I donate the project to?
Google knows the steps, once we have a user.

It's currently set up to do
cargo test --all

It triggers on merge to github master.


And it has been a bit wobbly, so might need more help. Or if the wobblyness
os gone after splitting grin into more parts and taking out the curl:ed
wallet angular site, maybe just enable and all is good.

lehnberg:
[...]

PR #2744, opened 2019-04-10
Set up CI with Azure Pipelines

quentinlesceller:
Replace https://github.com/mimblewimble/grin/pull/2742 and fix https://github.com/mimblewimble/grin/issues/2691. 

While GitLab is a great candidate for CI, it does not support macOS and Windows builds with the SaaS solution.

This PR adds Azure Pipelines as Grin CI.

Pros:
- Windows, Linux and macOS builds.
- Details on each job directly on Github.
- [Unlimited CI/CD minutes](https://azure.microsoft.com/en-us/blog/announcing-azure-pipelines-with-unlimited-ci-cd-minutes-for-open-source/).

Cons:
- No caching yet (but it is in the [pipeline](https://devblogs.microsoft.com/devops/adding-caching-to-azure-pipelines/)).
- Microsoft.
- Documentation scattered and not always super clear.

To do:
Complete release process with Azure Pipelines.

Example of build process here:
- Azure side: https://dev.azure.com/qlesceller/qlesceller/_build/results?buildId=49
- Github side: https://github.com/quentinlesceller/grin/pull/3

If you don't like the facts that there is 12 tasks running on GitHub I can always let it display as one task.

If this is accepted next steps would be to:
- @ignopeverell Remove GitLab integration and add Azure Pipelines on Grin repo.
- Add complete release process (which is doable https://nbsoftsolutions.com/blog/azure-pipelines-for-rust-projects).


quentinlesceller:
I guess you need to create an Azure account. I'll let you know when to merge I'm trying to do the GitHub release process and once that is done we'll be able to merge.

ignopeverell:
Where are we at with Azure CI? This PR needs some minor conflict fix.

quentinlesceller:
@ignopeverell as you can see I needed some cleanup https://github.com/mimblewimble/grin/pull/2752. In order to debug Azure Pipeline CI. You can follow my progress [on GitHub](https://github.com/quentinlesceller/grin/pull/3) and on [Azure](https://dev.azure.com/qlesceller/qlesceller/_build/results?buildId=96&view=logs). Windows tests are still no passing but I'm making very good progress.
Fixing the merge conflict btw.

store/tests/pmmr.rs
```diff
@@ -776,7 +776,7 @@ fn cleanup_rewind_files_test() {
 		&data_dir,
 		expected,
 		prefix_to_delete,
-		seconds_to_delete_after,
+		seconds_to_delete_after + 1,
```
quentinlesceller: For those wondering why I had to do this:
This test create file such as they are removed if older than ```seconds_to_delete_after```. 
However this test is very fast so basically it does that.
1. Create file that are `seconds_to_delete_after` old.
2. Do some operation
3. Remove file that are older than `seconds_to_delete_after`.

However, on Windows CI (as you can see here https://dev.azure.com/qlesceller/qlesceller/_build/results?buildId=105&view=logs&jobId=75e8fe06-bfd6-5564-cb5c-dff4fd81606c&taskId=c33ab3a6-027f-5d3f-2c09-9b84aecf111c&lineStart=1608&lineEnd=1608&colStart=1&colEnd=9) this duration is in seconds thus the deletion never happens. 
```
let duration_since_accessed = now.duration_since(accessed)?;
  if duration_since_accessed <= cleanup_duration {
    return Ok(0); // these files are still too new
}
// Deletion happens here
```

antiochp: 😑

quentinlesceller:
Okay this is done here is an example of what a release looks like: https://github.com/quentinlesceller/grin/releases. 
The changeling is not as nice though (but it still can be generated manually if needed).
@ignopeverell I'll need the name of the connection and I'll update the yml file.
Once this is done we can merge and proceed to a test beta release @yeastplume. 
IMO we should keep Travis file around in case we want to switch back (but just disable the check).

PR #2748, opened 2019-04-11
[DNM] Simplify logic for writing msgs onto the wire

antiochp:
Looking into sending kernel data (see #2743) and went down the rabbit hole of exactly how we end up serializing "attachments" when sending a msg response to a peer.

I think we can clean the `write` logic up pretty significantly, reducing the amount of intermediate writes and allocations - 

* get rid of our custom `write_all()` implementation
* use `buf_writer` to wrap the tcp connection
* use `buf_reader` to wrap reading the txhashset attachment
* track bytes written internally in `BinWriter`
* use `io::copy()` to handle robust write from reader to writer

TODO - 
- [ ] Take this a step further and write the msg body straight to tcp connection via `ser::serialize()` (need to be construct the header as we do this as we care about body size in the header)
- [ ] We needed to touch stratum code here - confirm this still works
- [ ] More testing

### DIFF
--- a/core/src/ser.rs
+++ b/core/src/ser.rs
@@ -279,9 +279,10 @@ pub fn deserialize<T: Readable>(source: &mut dyn Read) -> Result<T, Error> {
 }
 
 /// Serializes a Writeable into any std::io::Write implementation.
-pub fn serialize<W: Writeable>(sink: &mut dyn Write, thing: &W) -> Result<(), Error> {
-	let mut writer = BinWriter { sink };
-	thing.write(&mut writer)
+pub fn serialize<W: Writeable>(sink: &mut dyn Write, thing: &W) -> Result<u64, Error> {
+	let mut writer = BinWriter::new(sink);
+	thing.write(&mut writer)?;
+	Ok(writer.bytes_written as u64)
 }
 
 /// Utility function to serialize a writeable directly in memory using a
@@ -559,13 +560,17 @@ impl<T: Hashed> VerifySortedAndUnique<T> for Vec<T> {
 /// Utility wrapper for an underlying byte Writer. Defines higher level methods
 /// to write numbers, byte vectors, hashes, etc.
 pub struct BinWriter<'a> {
+	bytes_written: usize,
 	sink: &'a mut dyn Write,
 }
 
 impl<'a> BinWriter<'a> {
 	/// Wraps a standard Write in a new BinWriter
-	pub fn new(write: &'a mut dyn Write) -> BinWriter<'a> {
-		BinWriter { sink: write }
+	pub fn new(sink: &'a mut dyn Write) -> BinWriter<'a> {
+		BinWriter {
+			bytes_written: 0,
+			sink,
+		}
 	}
 }
 
@@ -576,6 +581,7 @@ impl<'a> Writer for BinWriter<'a> {
 
 	fn write_fixed_bytes<T: AsFixedBytes>(&mut self, fixed: &T) -> Result<(), Error> {
 		let bs = fixed.as_ref();
+		self.bytes_written = self.bytes_written.saturating_add(bs.len());
 		self.sink.write_all(bs)?;
 		Ok(())
 	}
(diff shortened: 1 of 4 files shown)


hashmap:
Can we really remove our `write_all`? We need to deal with `WouldBlock` error because we use non-blocking IO.

antiochp:
> Can we really remove our write_all? We need to deal with WouldBlock error because we use non-blocking IO.

You are right - this approach won't work as it stands. 

I'm kind of confused though because I tested this in various scenarios for most of yesterday and could not get a WouldBlock situation to occur (and so convinced myself we could do this safely).

I think the source of my confusion (after doing some more reading) can be summed up as - 
* We call `set_nonblocking(true)` to put both the listener and the stream in "non-blocking" mode.
* But we do not actually do any async polling to leverage this. 
    * We "poll" in a loop with a `sleep()`
        * Worse than that we first loop to do a `read` and then loop _again_ to do a separate `write`.
        * Nothing is interleaved. 
        * i.e. We do not read and write simultaneously (for a given peer).

The existing code does something roughly like this -

```
stream.set_nonblocking(true);
thread.spawn {
	loop {
		// Check the read end via read loop.
		// Read single msg (or none if no msg to read).
		loop {
			// ..reading from stream and checking for WouldBlock and other errors 

			thread::sleep();
		}

		// Write the "response" if we should respond to the msg.
		if response {
			// Write single response.
			loop {
				// .. writing to stream and checking for WouldBlock and other errors	

				thread.sleep();
			}
		}

		// Check the write end via write loop.
		// Write single msg (or none if no msg to write).
		loop {
			// .. writing to stream and checking for WouldBlock and other errors

			thread.sleep();
		}

		// Check the close channel.

		thread::sleep();
	}
}
``` 

* We spawn a thread per connected peer.
* We start a "polling" loop that first reads _something_ if it can then writes _something_ if it can, then checks if the connection should be closed. Then it sleeps for a bit and repeats the loop.
* Reading consists of looping, reading what it can until it has read a full msg, sleeping in the internal loop.
* Writing consists of looping, writing until it has written the full msg, sleeping in the internal loop.

So we read, then write, then read, then write at the granularity of whole msgs.

tl;dr We do not really get _any_ benefit from using "non blocking" IO like this.

----

I know we have discussed this before and we know Rust IO is kind of janky, hence our `read_exact` and `write_all` "workarounds" to deal with `WouldBlock` handling.
But I think after this iteration of experimentation and investigation I _finally_ have a better understanding of exactly what has been bothering me with this approach.

hashmap:
> I'm kind of confused though because I tested this in various scenarios for most of yesterday and could not get a WouldBlock situation to occur (and so convinced myself we could do this safely).

It depends on IO load and peers behaviour I think

> So we read, then write, then read, then write at the granularity of whole msgs.
tl;dr We do not really get any benefit from using "non blocking" IO like this.

I totally understand you, it was my confusion too, why we use non-blocking io if we don't multiplex sockets etc. However if we have blocking IO we will block on a read operation until we read smth from the peer, imagine we need to send a header, we can't do it until we receive something (eg ping in 10 seconds). Alternative solution is to clone tcp stream and run 2-3 threads per peer (separate for read and write). I implemented it a while ago but found the result even more confusing

hashmap:
> So we read, then write, then read, then write at the granularity of whole msgs.

In fact we may read then may write, `may` is the key here, because we don't have request-response protocol, we basically send and receive messages in random order and this trick allows us to support this behaviour

antiochp:
> In fact we may read then may write, may is the key here, because we don't have request-response protocol, we basically send and receive messages in random order and this trick allows us to support this behaviour

Yes sorry, should have been clearer there. Each time we read/write either 0 or 1 msg.

My realization was that it is always "per msg" though. so even though we are using "non blocking IO" we never read part of a msg then write part of a msg. We always read the full msg (if there is one to read).

antiochp:
Closing - @hashmap convinced me this is a bad idea.
We _do_ want non blocking IO. But the way we do it today is pretty confusing (and likely room for some cleanup here).

Issue #2594, opened 2019-02-18
node cpu usage


nikito7:
PID USER      PR  NI    VIRT    RES    SHR S %CPU %MEM     TIME+ COMMAND
11061 user      20   0 10.760g  85924  62380 S 34.1  8.5   0:28.62 grin          

17:44:08 up 10:20,  1 user,  load average: 0.47, 0.55, 0.86

Constant high cpu usage, it's normal?

Default configs.

DavidBurkett:
My mac node sits at ~20% CPU. I assume one of the thread sleep values is unnecessarily low, or one of the containers is really inefficient, but I haven't had a chance to track it down. Grin++ sits at <0.1%, so there's certainly nothing core to the protocol that requires constant high CPU usage.

nikito7:
changed:

cat .grin/main/grin-server.toml | grep tui
run_tui = false

now <6% cpu usage

nikito7:
Sooo,

no tui, <8 peers = < 10% cpu
no tui, >50 peers = > 70% cpu

And Grin++, 50 peers = < 5% cpu

nikito7:
Bug or feature?

Grin never stop discovering new peers

There is a loop

nikito7:
windows preview grin

with tui, 50 peers: 10% cpu
no tui, 50 peers: <1% cpu

amazarache:
I have followed this thread, trying to achieve less CPU usage for grin node, but without luck :(
disabling tui doesn't lower the CPU usage in my case

processor usage stays between 30-50% load in both cases.

nikito7:
peer_max_count = 8
peer_min_preferred_count = 4

amazarache:
Whatever values I used for for peers, grin server loads processor to 40-50%

nikito7:
Its normal at begin, peers discovery issues

nikito7:
maybe fixed in
20190616 21:08:27.629 INFO grin - This is Grin version 1.1.1-beta.1 (git v1.1.0-3-gd03a81f), built for x86_64-unknown-linux-gnu by rustc 1.35.0 (3c235d560 2019-05-20).

nikito7:
Current status:

Vps 2Ghz
50% cpu load
275 peers

yeastplume:
Are there any actions to be taken here or can this be closed?

DavidBurkett:
This was solved long ago. Should be good to close.

PR #2754, opened 2019-04-15
Increase peer max count significantly

ignopeverell:
I think several of our peer-related issues stem from a max count of peers that's much too low. At this point a node can easily support 250 peers on a low-spec machine and so I don't see a lot of reasons to have this limit so low. Having it much higher should help a lot in connecting and keeping peers.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -228,7 +228,7 @@ fn comments() -> HashMap<String, String> {
 #ban_window = 10800
 
 #maximum number of peers
-#peer_max_count = 250
+#peer_max_count = 125
 
 #preferred minimum number of peers (we'll actively keep trying to add peers
 #until we get to at least this number
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -228,7 +228,7 @@ fn comments() -> HashMap<String, String> {
 #ban_window = 10800
 
 #maximum number of peers
-#peer_max_count = 25
+#peer_max_count = 250
 
 #preferred minimum number of peers (we'll actively keep trying to add peers
 #until we get to at least this number
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -48,7 +48,7 @@ pub const MAX_LOCATORS: u32 = 20;
 const BAN_WINDOW: i64 = 10800;
 
 /// The max peer count
-const PEER_MAX_COUNT: u32 = 250;
+const PEER_MAX_COUNT: u32 = 125;
 
 /// min preferred peer count
 const PEER_MIN_PREFERRED_COUNT: u32 = 8;
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -48,7 +48,7 @@ pub const MAX_LOCATORS: u32 = 20;
 const BAN_WINDOW: i64 = 10800;
 
 /// The max peer count
-const PEER_MAX_COUNT: u32 = 25;
+const PEER_MAX_COUNT: u32 = 250;
 
 /// min preferred peer count
 const PEER_MIN_PREFERRED_COUNT: u32 = 8;


DavidBurkett:
While I would like to see nodes accepting more peers, did you see this today: https://github.com/mimblewimble/grin/issues/2594#issuecomment-483218582 And @bladedoyle is in gitter right now talking about high cpu usage due to having too many peers.

ignopeverell:
These seem more related to too many txhashset downloads. While increasing peer counts would make that more likely, we should really just throttle those to reasonable levels. I'm actually suspecting someone is being obnoxious right now. In any case, I wouldn't want to block something that'd improve the network overall because we're being stupid in allowing any txhashset download.

hashmap:
I think we also should  increase `PEER_MIN_PREFERRED_COUNT`, it's important for nat'ed users. On the other hand a peer consumes 0.7% cpu in average (see a screenshot form medium ec2 instance) so I'm not sure about `we can easily support 250 peers on a low-spec machine`

Screenshot 2019-04-16 at 08 16 32

ignopeverell:
My seed is a 2 vCPUs, 2GB machine, but I agree it'd be good to do some profiling to cut some of that CPU activity. I'm not sure I understand the part about `PEER_MIN_PREFERRED_COUNT` though. Isn't the issue that non NAT'd nodes reject everything above 25 peers right now, refusing new peers that they could potentially accept easily?

hashmap:
@ignopeverell The main reason is that we propagate new headers (and the rest) to `min_count` peers. Also need to check the current code, we have `healthy_mix` check which makes (used to make?) nat'd nodes gravitate towards min_count.

antiochp:
If we do this we should do review our logic for the various bits of broadcast logic.
I suspect we broadcast to _all_ connected peers when broadcasting some things (txs?)
Others I'm pretty sure we limit (but we may limit as a function of our peer counts).

hashmap:
@antiochp you are right, we have it for headers only https://github.com/mimblewimble/grin/blob/master/p2p/src/peers.rs#L336  and we use max for tx and compact blocks https://github.com/mimblewimble/grin/blob/master/p2p/src/peers.rs#L373

config/src/comments.rs
```diff
@@ -228,7 +228,7 @@ fn comments() -> HashMap<String, String> {
 #ban_window = 10800
 
 #maximum number of peers
-#peer_max_count = 25
+#peer_max_count = 250
```
antiochp: Unrelated - but can we use `const` from code elsewhere to parameterize/templatize these values?
I haven't looked to closely into how painful this would be to do this but its a .rs file so should be possible?

ignopeverell:
Had a look before, but I think if we're connected to 250 we actually want to relay to 250. Alternatively, I could reduce to say 100 to start with until we have figured out CPU bottlenecks. Thoughts?

config/src/comments.rs
```diff
@@ -228,7 +228,7 @@ fn comments() -> HashMap<String, String> {
 #ban_window = 10800
 
 #maximum number of peers
-#peer_max_count = 25
+#peer_max_count = 125
```
kargakis: Shouldn't this option be uncommented? Otherwise, it's not really a default but more like a suggestion.

kargakis: Or is the default driven by PEER_MAX_COUNT below?

ignopeverell: The default is driven by `PEER_MAX_COUNT`. Ideally config comments that are generated should be driven by that, as @antiochp mentioned, but it's a little out of scope for this PR.

kargakis: Opened https://github.com/mimblewimble/grin/issues/2781

nikito7:
Just stats:

14:00:52 up 12 days, 13:42,  1 user,  load average: 1.93, 2.26, 1.82

        RX packets 91974348  bytes 16989604567 (15.8 GiB)
        TX packets 107252270  bytes 64276425320 (59.8 GiB)

20190626 13:59:15.499 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 196 connected (176 most_work). all 4467 = 199 healthy + 5 banned + 4263 defunct

PR #2756, opened 2019-04-16
[1.1.0 -> Master] Merge master into 1.1.0 -> Then create new master

yeastplume:
Update merge from master into 1.1.0. It's starting to get much harder to determine which changes should be included in the case of conflicts, so I think we're going to need to turn this branch into our master very soon and create a legacy 1.0.x branch from the current master (can discuss in the meeting later).

Everyone who's made recent changes to master should carefully examine the merge here to make sure their changes have been picked up properly.


quentinlesceller:
Tests also passed here. Looks like Travis is stuck.

PR #2757, opened 2019-04-16
cleanup how we deal with fixed size vs variable size data files

antiochp:
* introduce a `SizeInfo` enum with 2 variants - 
    * `SizeInfo::FixedSize`
    * `SizeInfo::VariableSize`

The `VariableSize` variant maintains a `size_file` internally.

This replaces the current impl where we pass around a couple of seemingly unrelated options which made it unnecessarily complicated.

Refactor only - no change to how any of this works (enum replaces two existing options).


antiochp:
Will hold off merging this. Guessing we merge this onto `master` once `1.1.0` is merged and released.

PR #2758, opened 2019-04-16
bucket_transactions is now fee_to_weight aware

antiochp:
Reworked `bucket_transactions` to make it `fee_to_weight` aware.
We also return the vec of underlying txs from `bucket_transactions`. We used to return the aggregated bucket txs.

`bucket_transactions` now performs the following logic - 
* Buckets consist of a vec of txs and track the aggregate fee_to_weight ~and  a "depth"~.
* ~Bucket depth is incremented if a bucket depends on an earlier bucket.~
    * buckets only ever have dependencies on buckets with higher fee_to_weight
* We aggregate (cut-through) dependent transactions within a bucket *unless* adding a tx
would reduce the aggregate fee_to_weight, in which case we start a new bucket ~(and increment the depth).~
* We then sort the buckets by ~depth (ascending) _and_~ fee_to_weight (descending) to preserve dependency ordering and maximize both cut-through and overall fees.

The vec of txs returned by `bucket_transactions` satisfies the following - 
* No tx is dependent on any subsequent tx in the vec
* Txs that can be cut-through are adjacent (as long as fee_to_weight is not adversely impacted)
* Cut-through takes max tx weight into account

This prevents low fee 0-conf txs from "piggy-backing" off of higher fee txs in the pool.
While enabling low fee 0-conf txs stuck in the pool to be spent via high fee 0-conf txs (CPFP style).

The pool logic will now attempt to maximize cut-through as long as overall fees are not adversely impacted.

A nice side-effect of all this is related to tx eviction (see #2706). The txs at the end of the vec of txs returned by `bucket_transactions` are good candidates for eviction.
They are likely to have low `fee_to_weight` _and_ can be safely evicted without affecting dependent txs and without impacting cut-through.

### DIFF
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -396,11 +396,9 @@ impl Pool {
 		}
 
 		// Now sort them by depth (ascending) to maintain dependency ordering.
-		// And then sort them (within each depth) by fee_to_weight descending.
+		// And then sort them (within each depth) by fee_to_weight (descending).
 		// Txs with no dependencies will be toward the start of the vec.
 		// Txs with a big chain of dependencies will be toward the end of the vec.
-		// Will satisfying the "depth" ordering we will then sort txs such that high
-		// fee_to_weight come first.
 		tx_buckets.sort_unstable_by_key(|x| (x.depth, -(x.fee_to_weight as i64)));
 
 		tx_buckets
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -26,6 +26,7 @@ use self::util::RwLock;
 use crate::types::{BlockChain, PoolEntry, PoolError};
 use grin_core as core;
 use grin_util as util;
+use std::cmp::Reverse;
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
 
@@ -302,12 +303,13 @@ impl Pool {
 		Ok(())
 	}
 
-	/// Buckets consist of a vec of txs and track the aggregated fee_to_weight and  a "depth".
-	/// The depth is incremented if a bucket necessarily depends on an earlier bucket.
+	/// Buckets consist of a vec of txs and track the aggregate fee_to_weight.
 	/// We aggregate (cut-through) dependent transactions within a bucket *unless* adding a tx
-	/// would reduce the aggregate fee_to_weight, in which case we start a new bucket (and increment the depth).
-	/// We can then sort the buckets by depth and fee_to_weight to preserve dependency ordering and maximize
-	/// both cut-through and overall fees.
+	/// would reduce the aggregate fee_to_weight, in which case we start a new bucket.
+	/// Note this new bucket will by definition have a lower fee_to_weight than the bucket
+	/// containing the tx it depends on.
+	/// Sorting the buckets by fee_to_weight will therefore preserve dependency ordering,
+	/// maximizing both cut-through and overall fees.
 	fn bucket_transactions(&self, weighting: Weighting) -> Vec<Transaction> {
 		let mut tx_buckets: Vec<Bucket> = Vec::new();
 		let mut output_commits = HashMap::new();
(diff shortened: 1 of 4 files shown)


antiochp:
I guess the complication here is we end up building buckets with dependencies between them (one bucket depends on a prior bucket).
Which is fine as long as we then sort them in a consistent way (by descending `fee_to_weight`) but this is not necessarily clear in the code and the dependency between buckets is implicit.

Maybe we're fine with this assumption that the buckets will be sorted consistently.

quentinlesceller:
The first idea bother me on the fact that making buckets possibly dependant of each other can be tricky for a lot of things.
For the second method, let's say you have this transaction pool with the following buckets (each bucket is Vec<Vec< Transaction >>):

```
Bucket A <<Tx1>>
Bucket B <<Tx2;Tx2,1>;<Tx2,2>> where Tx2,1 and Tx2,2 are dependant on Tx2 but not dependant on each other.
Bucket C <Tx3>
```
Let's say you must take 2 transactions to build a block.
You order the transactions by fee over weight and let's say `fee_weight(Tx3)>fee_weight(Tx2,2)`. So you end up taking Tx1 and <Tx2;Tx2,1>. So basically you flatten the buckets in a list of transactions and  again order them by fee over weight and take the first X txs. I think that could work.
> To evict txs we sort the buckets and evict later txs from cheaper buckets.

 This I think we need to the same as above (flatten the buckets  and take the last one) as example `fee_weight(Tx3)>fee_weight(Tx2,2)`.
Overall I don't directly see any flaws in this.


> Alternatively: Maybe the simplest solution is for bucket_transactions() to handle the sorting internally (then we can keep it all consistent).

What do you mean?

antiochp:
@quentinlesceller Take another look.

Reworked this so `bucket_transactions` returns a vec of the underlying txs (and not the aggregate bucket txs).
These are sorted such that dependencies flow right to left (no tx is dependent on txs later in the vec).

If you want to evict a single tx you can use the last one in the vec.
No other tx depends on it.
It has low fee_to_weight.
It is unlikely to participate in any cut-through.

pool/src/pool.rs
```diff
@@ -397,7 +394,20 @@ impl Pool {
 				}
 			}
 		}
+
+		// Now sort them by depth (ascending) to maintain dependency ordering.
+		// And then sort them (within each depth) by fee_to_weight descending.
+		// Txs with no dependencies will be toward the start of the vec.
+		// Txs with a big chain of dependencies will be toward the end of the vec.
+		// Will satisfying the "depth" ordering we will then sort txs such that high
```
quentinlesceller: Small typo `While satisfying the "depth" ordering we will then sort txs such that high` ?

pool/src/pool.rs
```diff
@@ -358,25 +353,27 @@ impl Pool {
...
 						} else {
-							// Aggregated tx is not valid so discard this new tx.
-							is_rejected = true;
+							// Otherwise put it in its own bucket at the end.
+							// Note we increment the depth here to track the dependency.
+							tx_buckets
+								.push(Bucket::new_with_depth(entry.tx.clone(), bucket.depth + 1));
 						}
 					} else {
 						// Aggregation failed so discard this new tx.
```
quentinlesceller: Out of curiosity out frequently this happens and how?

antiochp: The aggregation failure?

Basically the only way this would fail is if the aggregated tx ended up being too big to fit in a block. Say we attempted two huge transactions that in aggregate was larger than our block weight limit. This should happen only rarely.

The only other way this could happen is if the pool got itself into a bad state somehow with a double spend or something similar in there.
This should never happen in practice as we check for consistency at various stages of the tx pool lifecycle.

ignopeverell:
Sorry for being so late to the party. So here we'd prioritize a transaction with no dependent ("child") transaction over a potentially much higher paying one with one dependent that lowers its parent's fee/weight ratio?

antiochp:
> So here we'd prioritize a transaction with no dependent ("child") transaction over a potentially much higher paying one with one dependent that lowers its parent's fee/weight ratio?

Not quite no. We would bucket these up into 3 buckets.
1) tx A with no dependents
2) the high fee tx B 
3) tx C (lower fee), dependent on B

i.e. B and C would not be bucketed together because C would lower the overal fee/weight ratio. But they would both still be included in (separate) buckets.

We would then sort the buckets by fee/weight ratio (while maintaining the ordering constraint between B and C).

So in your example B then A then C.

antiochp:
Said another way - 

It is possible to bump a tx _up_ the priority list by sending a high-fee dependent tx.
It is not possible to bump a tx _down_ the priority list by sending a low-fee dependent tx.

ignopeverell:
Thanks a lot for the detailed explanation, all makes sense now. Very nice!

PR #2761, opened 2019-04-18
pass peer_info around rather than peer_addr (includes protocol version)

antiochp:
We discussed using `PROTOCOL_VERSION` to handle various changes to p2p msgs.
This PR changes the various adapter fns to pass around a ref to a `PeerInfo` rather than simply a `PeerAddr`.
The `PeerInfo` contains both the `PeerAddr` and the `version`.

This would allow the various handlers to handle msgs in a protocol version specific way, based on the peer version via the initial hand/shake msg.


antiochp:
cc @DavidBurkett

DavidBurkett:
Looks good to me!

PR #2762, opened 2019-04-18
make field in tx_body public

yeastplume:
Required for slate conversion logic in https://github.com/mimblewimble/grin-wallet/pull/85


yeastplume:
:+1: thanks!

PR #2764, opened 2019-04-19
relax our p2p rules around matching protocol versions

antiochp:
We currently only allow peer connections if the "protocol version" is an exact match between us and the peer.
We will start refusing peer connections in either direction if we attempt to increment any protocol version numbers on the network.
This PR removes this rule and permits peer connections for any combination of protocol version values.

If/when we need to introduce thresholds or incompatible version checking in the future we can, but for now we just accept arbitrary version numbers.

Also - expose _our_ version number via the TUI.

### DIFF
--- a/p2p/src/handshake.rs
+++ b/p2p/src/handshake.rs
@@ -86,12 +86,7 @@ impl Handshake {
 		// write and read the handshake response
 		write_message(conn, hand, Type::Hand)?;
 		let shake: Shake = read_message(conn, Type::Shake)?;
-		if shake.version != PROTOCOL_VERSION {
-			return Err(Error::ProtocolMismatch {
-				us: PROTOCOL_VERSION,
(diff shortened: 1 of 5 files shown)


antiochp:
cc @DavidBurkett 
We're going to need to be more flexible if anybody decides to bump their protocol version number.

And we can reintroduce any constraints on "legacy" or minimum versions as necessary later.

DavidBurkett:
So the way it was coded before, any increase in protocol version would've caused a chain split? That doesn't seem very useful 😁. Glad you noticed this.

antiochp:
Yeah it was just placeholder code but _inflexible_ placeholder code...

antiochp:
We should get this merged in for `1.1.0` to be sure a decent number of nodes post `1.1.0` can support the protocol version number getting bumped.

PR #2765, opened 2019-04-19
initial support for kernel data download

antiochp:
Extracted "kernel download" from #2743 without all the code attempting to rebuild the MMR on the receiving end.

We cannot (yet) bump the protocol version number without splitting the network (see #2764 for a fix).
So this is untouched in this PR.

If/when this is released we will have _some_ nodes on the network that are able to provide a "kernel data" download.
We won't do anything with this just yet, but we can verify we can download it and parse the kernels out of the data.
Your node _will_ get banned if you attempt to request a kernel data download from a peer that does not yet support this msg type. (So be careful using the `/kerneldownload` test api endpoint).

----

Proposal: Get this merged in prior to `1.1.0` being released. This way we can ideally have a healthy number of peers that support the kernel data download msg type.
Once `1.1.0` is released we can bump the protocol version if deemed necessary (#2764 adds support for flexible protocol versioning and peering).

If we can be reasonably confident most nodes support kernel data downloads then we can make progress of beginning to leverage this as part of fast sync (separate PR).


ignopeverell:
Tried to merge for you but think I messed something up...

PR #2773, opened 2019-04-22
Increasing thread sleeps for socket listener thread and per-connection threads

DavidBurkett:
This significantly decreases cpu usage, especially for nodes with >20 peers. There's very little reason for the thread sleeps to be 1 ms. The only (theoretical) advantage is during mining, but as long as miners have a sufficient number of peers (~10+), there should be no practical effect.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -244,7 +244,7 @@ fn poll<H>(
 	let _ = thread::Builder::new()
 		.name("peer".to_string())
 		.spawn(move || {
-			let sleep_time = time::Duration::from_millis(1);
+			let sleep_time = time::Duration::from_millis(5);
 			let mut retry_send = Err(());
 			loop {
 				// check the read end
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -71,7 +71,7 @@ impl Server {
 		let listener = TcpListener::bind(addr)?;
 		listener.set_nonblocking(true)?;
 
-		let sleep_time = Duration::from_millis(1);
+		let sleep_time = Duration::from_millis(5);
 		loop {
 			// Pause peer ingress connection request. Only for tests.
 			if self.stop_state.lock().is_paused() {


mcdallas:
Some results: I increased mine from 1ms to 10ms and cpu usage went from 90% to 30%

p2p/src/conn.rs
```diff
@@ -244,7 +244,7 @@ fn poll<H>(
 	let _ = thread::Builder::new()
 		.name("peer".to_string())
 		.spawn(move || {
-			let sleep_time = time::Duration::from_millis(1);
+			let sleep_time = time::Duration::from_millis(5);
```
garyyu: I saw @mcdallas reported 10ms test result, are we sure the `5ms` here has the same improvement?

garyyu:
Perhaps we should merge all p2p threads into one single thread, in the future. But for the moment, it's good to have this quick and simple improvement 👍

p2p/src/conn.rs
DavidBurkett: It won't be the same improvement, but it will be close. Setting the conn thread sleep too high could have negative consequences, especially for miners. We can continue to tweak it as necessary, but this seems a happy middle ground for now.

garyyu:
After 7 hours test, with this 5ms modification, before upgrading:
```
Total Peers:          112
3414 Total Connection:  116
3414 Total Established: 115

62.8%  CPU LOAD for Grin  (not stable, but a typical value)
```

After upgrading:
```
Total Peers:          164
3414 Total Connection:  164
3414 Total Established: 164

51.8% CPU LOAD for Grin (typical value)
```

Because the peers number different, we can't directly compare the improvement, but an estimated value 80% improvement here, anyway it deserves to have this PR, thanks @DavidBurkett

kargakis:
I am seeing a ~50% cpu usage reduction in GCP - nice one @DavidBurkett !
grin-cpu

PR #2778, opened 2019-04-25
Use blocking IO in P2P to avoid short sleeps

hashmap:
It creates constant CPU load. Instead we create 2 threads - one for
reading, one for writing, both block for 1 second then check stop flag,
if it is not set the process continue.
This approach produces 5-10 times less CPU load, the node without TUI
consumes around 2% CPU on my machine vs 10-30%.
Also some refactoring of connection tracking was done, partially to
simplify the code partially to defeat the borrow checker.


DavidBurkett:
1) Was the 10-30% with 5ms timeouts (https://github.com/mimblewimble/grin/pull/2773) or just with 1ms?
2) How does this affect shutdowns?

hashmap:
We merged some parts of this pr already, the last piece is #2855

PR #2784, opened 2019-04-29
Gracefully shutdown if SIGINT was sent in TUI mode

### LINKED ISSUE #2799: If TUI is enabled Ctrl-C doesn't start graceful shutdown
On all of my nodes except one (?) `Ctrl-C` kills TUI, but the process continue running. Second `Ctrl-C` kills the process but non of them start shutdown procedure. On one node process exits immediately without any attempt to shutdown (eg close connections).
It seems we don't handle `SIGINT` in TUI mode. I tried to add a handler in UI code but it's not invoked, I assume ncurses backend handles the signal.

hashmap:
Fixes #2799

Also 2 Arc's were replaced by one server's instance.
It is needed for p2p thread management in https://github.com/mimblewimble/grin/pull/2778, currently there is no point where we could store thread handles and join them because `thread::join`
consume the caller, which is impossible in case of Arc.

### DIFF
--- a/servers/src/grin/server.rs
+++ b/servers/src/grin/server.rs
@@ -75,12 +75,12 @@ impl Server {
 	/// to poll info about the server status
 	pub fn start<F>(config: ServerConfig, mut info_callback: F) -> Result<(), Error>
 	where
-		F: FnMut(Arc<Server>),
+		F: FnMut(Server) -> thread::JoinHandle<()>,
 	{
 		let mining_config = config.stratum_mining_config.clone();
 		let enable_test_miner = config.run_test_miner;
 		let test_miner_wallet_url = config.test_miner_wallet_url.clone();
-		let serv = Arc::new(Server::new(config)?);
+		let serv = Server::new(config)?;
 
 		if let Some(c) = mining_config {
 			let enable_stratum_server = c.enable_stratum_server;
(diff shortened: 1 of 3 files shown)


servers/src/grin/server.rs
```diff
@@ -101,13 +101,8 @@ impl Server {
 			}
 		}
 
-		info_callback(serv.clone());
-		loop {
```
antiochp: lol

src/bin/cmd/server.rs
```diff
@@ -45,21 +45,16 @@ fn start_server_tui(config: servers::ServerConfig) {
 	// everything it might need
 	if config.run_tui.unwrap_or(false) {
 		warn!("Starting GRIN in UI mode...");
-		servers::Server::start(config, |serv: Arc<servers::Server>| {
-			let running = Arc::new(AtomicBool::new(true));
-			let _ = thread::Builder::new()
-				.name("ui".to_string())
-				.spawn(move || {
```
antiochp: So we now run the TUI in the main thread and not in a separate ui thread?
Is there any downside to this?

hashmap: It's symmetrical with non-tui case. Also  to be fair I'm not sure, it seems that `cursive` creates a separate thread anyway, but I'm not an expert. What I see is that SIGINT is magically handled (in ncurses backend or most likely in C code, because I didn't find any traces of it in `cursive`) and it stops only UI, so we are able to call `server.stop()` in the main thread after that.

antiochp: :+1:

PR #2790, opened 2019-04-30
bump rand crate to 0.6 (was 0.5)

antiochp:
shuffle was deprecated, fix this up


antiochp:
@ignopeverell want to take a quick look and make sure I have not done something dumb here moving from -

```
thread_rng().shuffle(&mut res);
```

to

```
res.shuffle(&mut thread_rng());
```

PR #2791, opened 2019-05-01
Wrap MsgHeader in MsgHeaderWrapper for Known/Unknown msg type support

### LINKED ISSUE #2782: unknown msg type treated as CorruptedData
We are not very friendly with our peers when we encounter an unknown message type.
We treat any unexpected msg type (u8) as `CorruptedData` and immediately ban the peer.

https://github.com/mimblewimble/grin/blob/304ae444ab993cbc5a37ec7b12161451fa140ec9/p2p/src/msg.rs#L231-L246

We may want to consider being more accepting and simply drop the msg.

Although we may want to think this through carefully as we would not necessarily know how many bytes we want to ignore on that connection.

From gitter - 

> David Burkett @DavidBurkett 16:23
It just seems like every decision made around the p2p network actively discourages any protocol upgrades
Which doesn't align well with a coin that advertises itself as cutting-edge and "beta"

antiochp:
Introduce a `MsgHeaderWrapper` enum with `Known` and `Unknown` variants.

* The `Known` variant wraps a `MsgHeader` and we process it as before.
* The `Unknown` variant wraps a `msg_len` and we call `read_discard(msg_len)` to discard the unwanted bytes from the stream.

If we do encounter an unknown msg type then we do not know how to deserialize it safely.
But we do want to read the bytes (based on `msg_len`) so we can read any subsequent msg correctly off the stream.

Resolves #2782.


p2p/src/msg.rs
```diff
@@ -228,19 +251,50 @@ impl Writeable for MsgHeader {
...
+				Ok(MsgHeaderWrapper::Known(MsgHeader {
+					magic: m,
+					msg_type,
+					msg_len,
+				}))
+			}
+			None => {
+				// Unknown msg type, but we still want to limit how big the msg is.
+				// Default to max_block_size (4x as above for space to change things).
+				let max_len = max_block_size() * 4;
```
hashmap: why `max_block_size`?

p2p/src/msg.rs
```diff
@@ -228,19 +251,50 @@ impl Writeable for MsgHeader {
...
+				if msg_len > max_len {
```
hashmap: Mostly unrelated, because we do the same above, but does it mean if we get a too large message we (this peer thread) will stuck forever checking the size?

p2p/src/protocol.rs
```diff
@@ -353,9 +353,16 @@ impl MessageHandler for Protocol {
 
 				Ok(None)
 			}
-
-			_ => {
-				debug!("unknown message type {:?}", msg.header.msg_type);
+			Type::Error => {
```
hashmap: You can combine checks `Type::Error | Type::Hand | Type::Shake`

p2p/src/msg.rs
```diff
@@ -158,13 +154,28 @@ pub fn read_body<T: Readable>(h: &MsgHeader, stream: &mut dyn Read) -> Result<T,
 	ser::deserialize(&mut &body[..]).map_err(From::from)
 }
 
+/// Read (an unknown) message from the provided stream and discard it.
+pub fn read_discard(msg_len: u64, stream: &mut dyn Read) -> Result<(), Error> {
+	let mut buffer = vec![0u8; msg_len as usize];
```
hashmap: I wish we could skip allocation here, but I'm not sure we can do it with async io

p2p/src/msg.rs
antiochp: Ahh good question - `max_block_size()` is the largest msg size we support in `max_msg_size()`.

We should probably pull this out into something more meaningful like `default_max_msg_size()`
and document this better.

p2p/src/protocol.rs
antiochp: :+1: I had it broken out explicitly like this when making these changes so it was easier to see all msg types in there. But I'll merge these like you suggest.

p2p/src/msg.rs
antiochp: Yeah. We can always revisit this and optimize it later.
There might be something we can do down in read_exact so we're not allocating the whole thing like this.

p2p/src/msg.rs
antiochp: I need to check but I'm guessing what we _should_ be doing in this scenario is banning the peer.
We should _never_ receive a msg that exceeds the size limits from an honest peer.

I'm not sure we are doing this today though, but I had assumed we were.

antiochp: I assume that what happens here is - 

* we attempt to deserialize the msg header
  * 11 bytes (fixed size) read from stream
  * deserialization error due to `msg_len` exceeding `max_len`
* `try_break!` loops again
* we then attempt to deserialize the next msg header
  * 11 bytes (fixed size) read from stream
  * the magic bytes will not match as these 11 bytes are not actually a msg header
* `try_break!` is now effectively looping looking for a valid 11 byte chunk starting with magic bytes

I was actually thinking about something similar to this yesterday related to connection dropping unexpectedly. I was wondering if the sender and the receiver could get "out of alignment" as we identify msg based on the first 2 bytes (the magic number).

It looks like this could actually happen here as well.
If the next bytes are not the magic bytes then we are going to read 11 bytes at a time and if we are misaligned by even one byte we will not identify the next msg header correctly.

A couple of alternative approaches - 

1. If we don't read a full msg we drop the connection. We cannot rely on the next bytes read being the next pair of magic bytes.
2. Alternatively - consider adding some logic to read (and discard) bytes until we encounter the next pair of magic bytes? I'm not sure if this is an approach often used in practice?

antiochp: I'm going to open a separate issue to track this.

antiochp: Issue #2793.

PR #2794, opened 2019-05-02
cleanup check_connections

### LINKED ISSUE #2793: msg deserialization error causes peer thread to loop continuously
We fail to deserialize a p2p msg if the `msg_len` exceeds a defined `max_len` (per msg type).

If this happens we have read the 11 bytes of the msg header off the stream but we have not read the msg body itself.

We require every msg to begin with a pair of "magic bytes" that indicate the start of a msg header.

* we attempt to deserialize the msg header
    * 11 bytes (fixed size) read from stream
    * deserialization error due to msg_len exceeding max_len
* try_break! loops again
* we then attempt to deserialize the next msg header
    * 11 bytes (fixed size) read from stream
    * the magic bytes will not match as these 11 bytes are not actually a msg header
* try_break! is now effectively looping looking for a valid 11 byte chunk starting with magic bytes

If the next bytes are not the magic bytes then we are going to read chunks of 11 bytes at a time and if we are misaligned by even one byte we will not identify the next msg header correctly.

In fact this is not limited to the `msg_len` check. Any error occurring during deserialization of the msg header or the msg itself will cause this.

Is my understanding here correct?
* Exceeding `max_len` is treated as a serialization error
* We do not drop the connection on serialization errors
* Our msg polling logic will loop again on a serialization error
* Once we are out of alignment" we are effectively stuck reading garbage off the stream

A couple of alternative approaches -

1. If we don't read a full msg successfully we need to drop the connection. We cannot rely on the next bytes read being the next pair of magic bytes.
2. Alternatively - consider adding some logic to read (and discard) bytes until we encounter the next pair of magic bytes? I'm not sure if this is an approach often used in practice?

Related #2791.

antiochp:
See https://github.com/mimblewimble/grin/issues/2793#issuecomment-488618006 for more context.

[description tbd]

Resolves #2793.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -286,19 +286,22 @@ fn poll<H>(
 					}
 				}
 
-				// check the close channel
-				if let Ok(_) = close_rx.try_recv() {
-					debug!(
-						"Connection close with {} initiated by us",
-						conn.peer_addr()
-							.map(|a| a.to_string())
-							.unwrap_or("?".to_owned())
-					);
+				// check the close channel, reading all the close msgs if any.
+				let close_msgs: Vec<_> = close_rx.try_iter().collect();
+				if !close_msgs.is_empty() {
 					break;
 				}
 
 				thread::sleep(sleep_time);
 			}
+
+			debug!(
+				"Shutting down connection with {}.",
+				conn.peer_addr()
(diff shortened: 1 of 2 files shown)


antiochp:
Replaced with #2796.

PR #2796, opened 2019-05-02
remove the error_channel and simplify how we close peer connections

antiochp:
Replaces #2794 with a simpler solution.

Existing behavior -

* `try_break!` on certain errors will send an error event to the `error_channel`
* `peers.check_connection` tries to read off the `error_channel`
    * on receiving an error event it will then put a close event on the `close_channel`
* the peer connection poll loop then tries to read off the `close_channel`
    * on receiving a close event it closes the connection

So we have - 

```
error -> error_channel -> close_channel -> close
```

In addition `try_break!` _also_ breaks out of the connection poll loop and immediately closes the connection. 
So none of the above actually affects anything. We have already broken out of the poll loop and we will never read from the peer specific `close_channel`.

This PR simplifies all of the above by replacing the "put an error event on the error_channel" with a simple "stop (close connection) and remove peer".

There is no need for the `error_channel` as we can close peer connection directly without this intermediate layer of indirection.

There are 3 places where we actually use these connections -
1. When we receive a msg from a peer
1. When we broadcast a msg out to peer(s)
1. When we ping all our peers (similar but slightly different to above)

For (1) we want to close the connection if `try_break!` encounters a serious enough peer related error.
We already handle this by breaking out of the poll loop (error_channel never actually used).
For (2) and (3) we now simply `close()` and remove the peer from our list of peers. 

Previously we were relying on the complex `is_connected()` logic to identify a closed or error condition peer connection (and for it to internally clean up the peer and connection).

We now simply attempt to send the msg and on error close the connection (for both ping and broadcast paths).

Related - #2801 removes option wrapper on the peer connection which will make `peer.is_connected()` here even simpler as we no longer need to handle the case where a peer has no initialized connection.

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -42,11 +42,7 @@ const MAX_PEER_MSG_PER_MIN: u64 = 500;
 ///   For example: 'Disconnected' state here could still be 'Healthy' and could reconnect in next loop.
 enum State {
 	Connected,
-	Disconnected,
 	Banned,
-	// Banned from Peers side, by ban_peer().
-	//   This could happen when error in block (or compact block) received, header(s) received,
-	//   or txhashset received.
 }
 
 pub struct Peer {
@@ -173,9 +169,12 @@ impl Peer {
 		false
 	}
 
-	/// Whether this peer is still connected.
+	/// Whether this peer is currently connected.
 	pub fn is_connected(&self) -> bool {
-		self.check_connection()
+		if self.connection.is_none() {
+			return false;
+		}
+		State::Connected == *self.state.read()
 	}
 
 	/// Whether this peer has been banned.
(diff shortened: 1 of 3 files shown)


p2p/src/peer.rs
```diff
@@ -173,9 +169,12 @@ impl Peer {
 		false
 	}
 
-	/// Whether this peer is still connected.
+	/// Whether this peer is currently connected.
 	pub fn is_connected(&self) -> bool {
-		self.check_connection()
+		if self.connection.is_none() {
```
hashmap: do we set `connection` to None somewhere?

p2p/src/peers.rs
```diff
@@ -256,7 +257,14 @@ impl Peers {
...
 				Ok(true) => count += 1,
 				Ok(false) => (),
-				Err(e) => debug!("Error sending {} to peer: {:?}", obj_name, e),
+				Err(e) => {
+					debug!(
+						"Error sending {:?} to peer {:?}: {:?}",
+						obj_name, &p.info.addr, e
+					);
+					p.stop();
+					self.peers.write().remove(&p.info.addr);
```
hashmap: Do I understand correctly that we replace the logic of marking peers in `check_connection` with this error handling, assuming if an error happened and we closed the connection on peer's thread exit we will get an error here and remove this peer?

p2p/src/peer.rs
antiochp: I believe there is a small period of time during the initialization of a new peer connection where we instantiate the peer instance but we do not yet have a live connection.
I don't think we ever set a connection back to None once a peer is up and running.

antiochp: Related - #2801 (get rid of the option wrapping the peer connection).
The cleanup in #2801 is probably worth doing regardless of what we decide here.

p2p/src/peers.rs
antiochp: Updated description for this PR which should explain what the intent is here.

PR #2797, opened 2019-05-02
Evict transaction from transaction pool

quentinlesceller:
Replace https://github.com/mimblewimble/grin/pull/2706 and fix https://github.com/mimblewimble/grin/issues/1417.

Thanks to https://github.com/mimblewimble/grin/pull/2758 we can drastically simplify the eviction process when the transaction pool is full:

1. Whenever we receive a stem transaction, we check if adding it would overflow the current pool capacity.
2. Add the transaction to the transaction pool.
3. If overflow then `bucket_transactions` and remove the last one (No other tx depends on it, it has low fee_to_weight and is unlikely to participate in any cut-through).

### DIFF
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -138,7 +138,13 @@ impl TransactionPool {
 		}
 
 		// Do we have the capacity to accept this transaction?
-		self.is_acceptable(&tx, stem)?;
+		let acceptability = self.is_acceptable(&tx, stem);
+		let mut evict = false;
+		if !stem && acceptability.as_ref().err() == Some(&PoolError::OverCapacity) {
+			evict = true;
+		} else if acceptability.is_err() {
+			return acceptability;
+		}
 
 		// Make sure the transaction is valid before anything else.
 		// Validate tx accounting for max tx weight.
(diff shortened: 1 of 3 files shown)


pool/src/transaction_pool.rs
```diff
@@ -171,9 +177,37 @@ impl TransactionPool {
...
+	// Remove the last transaction from the flattened bucket transactions.
+	// No other tx depends on it, it has low fee_to_weight and is unlikely to participate in any cut-through.
+	pub fn evict_from_txpool(&mut self) -> Result<(), PoolError> {
+		// Get bucket transactions
+		let bucket_transactions = self.txpool.bucket_transactions(Weighting::NoLimit);
+
+		// Get last transaction and remove it
+		match bucket_transactions.last() {
+			Some(evictable_transaction) => {
+				// Remove transaction
```
antiochp: We should be able use `entries.remove()` here I think.

pool/src/transaction_pool.rs
```diff
@@ -171,9 +177,37 @@ impl TransactionPool {
...
+				self.txpool.entries = self
+					.txpool
+					.entries
+					.iter()
+					.filter(|x| x.tx != *evictable_transaction)
+					.map(|x| x.clone())
+					.collect::<Vec<_>>();
+				Ok(())
+			}
+			None => Err(PoolError::OverCapacity),
```
antiochp: This does not make sense - if we ever call `evict_from_pool` then we should be reasonably confident we have at least one tx in the pool.
But if entries was to ever be empty we should probably just do nothing and return (and not raise an error).

pool/src/transaction_pool.rs
quentinlesceller: I'm not sure this is doable since we are comparing a `PoolEntry` with a `Transaction`.
If both were `PoolEntry` a `remove_item(...)` would be possible.

pool/src/transaction_pool.rs
quentinlesceller: Indeed fixed.

pool/src/transaction_pool.rs
antiochp: Ah good point.

PR #2801, opened 2019-05-03
Simplify peer connection handling

antiochp:
The peer connection used to be wrapped in an option.
This complicated the initialization of a peer (with active connection) due to the period between creating the peer and assigning the connection to it.
The workaround was the `connection!` macro to handle the case where the connection was `None`.

This PR simplifies this. `Peer::accept()` and `Peer::connect()` now return a peer with active connection already assigned.

We have less special cases to handle like this.

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -61,6 +61,19 @@ impl fmt::Debug for Peer {
 }
 
 impl Peer {
+	fn new(info: PeerInfo, conn: TcpStream, adapter: Arc<dyn NetAdapter>) -> Peer {
+		let state = Arc::new(RwLock::new(State::Connected));
+		let tracking_adapter = TrackingAdapter::new(adapter);
+		let handler = Protocol::new(Arc::new(tracking_adapter.clone()), info.clone());
+		let connection = Mutex::new(conn::listen(conn, handler));
+		Peer {
+			info,
+			state,
+			tracking_adapter,
+			connection,
+		}
+	}
+
 	pub fn accept(
 		mut conn: TcpStream,
 		capab: Capabilities,
@@ -71,20 +84,7 @@ impl Peer {
 		debug!("accept: handshaking from {:?}", conn.peer_addr());
 		let info = hs.accept(capab, total_difficulty, &mut conn);
 		match info {
-			Ok(info) => {
-				let state = Arc::new(RwLock::new(State::Connected));
-				let tracking_adapter = TrackingAdapter::new(adapter);
-				let handler = Protocol::new(Arc::new(tracking_adapter.clone()), info.clone());
-
-				let peer = Peer {
-					info,
-					state,
-					tracking_adapter,
-					connection: Mutex::new(conn::listen(conn, handler)),
-				};
-
-				Ok(peer)
-			}
+			Ok(info) => Ok(Peer::new(info, conn, adapter)),
 			Err(e) => {
 				debug!(
 					"accept: handshaking from {:?} failed with error: {:?}",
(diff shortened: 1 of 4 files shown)


p2p/src/peer.rs
```diff
@@ -53,17 +53,8 @@ pub struct Peer {
...
-}
-
-macro_rules! connection {
-	($holder:expr) => {
-		match $holder.connection.as_ref() {
-			Some(conn) => conn.lock(),
-			None => return Err(Error::ConnectionClose),
-			}
-	};
+	tracking_adapter: Arc<TrackingAdapter>,
```
hashmap: Do we really need `Arc` here and below? Would be nice to get rid of this indirection (we didn't have it before)

p2p/src/peer.rs
```diff
@@ -250,13 +235,14 @@ impl Peer {
 			total_difficulty,
 			height,
 		};
-		connection!(self).send(ping_msg, msg::Type::Ping)
+		self.connection.lock().send(ping_msg, msg::Type::Ping)
```
hashmap: I wonder should we introduce `send` method on Peer? This construction is verbose and reveal a lot of internal structure of the object, which is against law of Demeter, which is not a law of course, but a nice principle to follow https://hackernoon.com/object-oriented-tricks-2-law-of-demeter-4ecc9becad85

antiochp: :+1: doing this now.

p2p/src/peer.rs
antiochp: I added this for consistency with peer state above but I agree its not necessary.
Let me take another pass over this.

p2p/src/peer.rs
```diff
@@ -73,27 +65,30 @@ impl fmt::Debug for Peer {
...
 		hs: &Handshake,
 		adapter: Arc<dyn NetAdapter>,
 	) -> Result<Peer, Error> {
 		debug!("accept: handshaking from {:?}", conn.peer_addr());
-		let info = hs.accept(capab, total_difficulty, conn);
+		let info = hs.accept(capab, total_difficulty, &mut conn);
 		match info {
-			Ok(peer_info) => Ok(Peer::new(peer_info, adapter)),
+			Ok(info) => {
+				let state = Arc::new(RwLock::new(State::Connected));
```
hashmap: This chunk looks like a Peer constructor (same code here and in `connect`), can we extract it into `Peer::new` function or smth  like that? We need to pass `conn` and `adapter`, all details would be hidden there

p2p/src/peer.rs
```diff
@@ -197,66 +197,58 @@ impl Peer {
...
 		};
-		connection!(self).send(ping_msg, msg::Type::Ping)
+		self.send(ping_msg, msg::Type::Ping)
 	}
 
 	/// Send the ban reason before banning
 	pub fn send_ban_reason(&self, ban_reason: ReasonForBan) -> Result<(), Error> {
 		let ban_reason_msg = BanReason { ban_reason };
-		connection!(self)
+		self.connection
```
hashmap: `self.send()`?

p2p/src/peer.rs
```diff
@@ -311,7 +303,9 @@ impl Peer {
 	pub fn send_tx_kernel_hash(&self, h: Hash) -> Result<bool, Error> {
 		if !self.tracking_adapter.has_recv(h) {
 			debug!("Send tx kernel hash {} to {}", h, self.info.addr);
-			connection!(self).send(h, msg::Type::TransactionKernel)?;
+			self.connection
```
hashmap: `self.send()`?

p2p/src/peer.rs
```diff
@@ -356,38 +350,40 @@ impl Peer {
...
 	pub fn send_stem_transaction(&self, tx: &core::Transaction) -> Result<(), Error> {
 		debug!("Send (stem) tx {} to {}", tx.hash(), self.info.addr);
-		connection!(self).send(tx, msg::Type::StemTransaction)
+		self.send(tx, msg::Type::StemTransaction)
 	}
 
 	/// Sends a request for block headers from the provided block locator
 	pub fn send_header_request(&self, locator: Vec<Hash>) -> Result<(), Error> {
-		connection!(self).send(&Locator { hashes: locator }, msg::Type::GetHeaders)
+		self.connection
```
hashmap: `self.send()`?

p2p/src/peer.rs
antiochp: Bad search and replace and missed these multi-line instances...

PR #2809, opened 2019-05-07
Header version cleanup

antiochp:
* Introduce `HeaderVersion(u16)` for type safety around header versioning
* Cleanup pow ser/deser (header version is unused)

Header version is converted to the underlying u16 during json serialization for the api.
Header version is serialized/deserialized as the inner u16 vie read/write.
Default header version is `1` (currently the _only_ version in use).
pow ser/deser was passing an unused version around - cleaned this up so we can now implement `Readable/Writeable` traits on `ProofOfWork` struct.

This is (should be) purely a refactor - no change to behavior or functionality.

I went down this rabbit hole because I was investigating "protocol version" specific serialization/deserialization and discovered we are already using "versioning" for block headers (even if not for the same reason).

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -22,7 +22,7 @@ use std::fmt;
 use std::iter::FromIterator;
 use std::sync::Arc;
 
-use crate::consensus::{reward, REWARD};
+use crate::consensus::{self, reward, REWARD};
 use crate::core::committed::{self, Committed};
 use crate::core::compact_block::{CompactBlock, CompactBlockBody};
 use crate::core::hash::{DefaultHashable, Hash, Hashed, ZERO_HASH};
@@ -298,6 +298,14 @@ impl Readable for BlockHeader {
 			|| timestamp < MIN_DATE.and_hms(0, 0, 0).timestamp()
 		{
 			return Err(ser::Error::CorruptedData);
+		}
+
+		// Check the block version before proceeding any further.
+		// We want to do this here because blocks can be pretty large
+		// and we want to halt processing as early as possible.
+		// If we receive an invalid block version then the peer is not on our hard-fork.
+		if !consensus::valid_header_version(height, version) {
+			return Err(ser::Error::InvalidBlockVersion);
 		}
 
 		Ok(BlockHeader {
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -178,6 +178,13 @@ impl Default for HeaderVersion {
 	}
 }
 
+impl HeaderVersion {
+	/// Constructor taking the provided version.
+	pub fn new(version: u16) -> HeaderVersion {
+		HeaderVersion(version)
+	}
+}
+
 impl From<HeaderVersion> for u16 {
 	fn from(v: HeaderVersion) -> u16 {
 		v.0
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -168,11 +168,40 @@ impl Hashed for HeaderEntry {
 	}
 }
 
+/// Some type safety around header versioning.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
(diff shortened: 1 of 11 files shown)


core/src/core/block.rs
```diff
@@ -168,11 +168,47 @@ impl Hashed for HeaderEntry {
 	}
 }
 
+/// Some type safety around header versioning.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+pub struct HeaderVersion(pub u16);
+
+impl Default for HeaderVersion {
```
DavidBurkett: I'm not sure it makes sense to implement Default for HeaderVersion. In 2 months, all new blocks will have to be version 2, and having a default at that point seems like a good way to introduce logical errors.

core/src/pow/types.rs
```diff
@@ -239,9 +239,19 @@ impl Default for ProofOfWork {
 	}
 }
 
-impl ProofOfWork {
-	/// Read implementation, can't define as trait impl as we need a version
-	pub fn read(_ver: u16, reader: &mut dyn Reader) -> Result<ProofOfWork, ser::Error> {
```
DavidBurkett: I don't think you can do this. You're going to be adding that version back for our scheduled hard fork, aren't you?

antiochp: I don't think the check is going to happen here during deserialization. 

We have `consensus::valid_header_version()` and this is happening orthogonal to the ser/deser process (called via `pipe::validate_header()`).

I _think_ we originally had this passed in here when we first introduced the AR/AT dual PoW impl or when we opened up the C31+ flexibility.

That said - this change was one of the things I wanted to discuss via this PR.

core/src/core/block.rs
antiochp: Default post-HF should be version 2 at that point. At least that's the intent here.
We may want to make it more explicit though and not use "default".

DavidBurkett: Let's assume for a second that it will be version 2 post hard fork. How do you manage that? Do we want default to look up the current height to know if we're >= block 250000? I would assume no. My vote is to remove default.

core/src/pow/types.rs
DavidBurkett: I don't know much about the format of our PoW, but I assumed the protocol bump would change how we have to serialize/deserialize ProofOfWork. If that's not true, then I agree with your change.

antiochp: I'd leaning toward simplifying and standardizing how we `read` (to use our existing traits) for now. If and when we discover this needs to be adjusted we can revisit this decision.

antiochp:
Note: I added a header version validation step to the header deserialization stage.
We're going to want to halt header (and block) processing as early as possible if we know the header is an incorrect version.

core/src/pow/types.rs
ignopeverell: Ideally the PoW change wouldn't change its format, but just details about graph construction or cycle finding. Not sure we'll be able to keep it that way every time but still a good goal to start with.

core/src/core/block.rs
ignopeverell: Was slightly agreeing with @DavidBurkett but then realized versions are rarely built by themselves, they're built with their headers. And then we have a bunch of code (mostly tests) that just want a header without caring much.

antiochp: The way I was thinking about this was in terms of two ways we need to create headers - 

* new/current headers - we want the default to be version 2 post hark fork
  * for tests (as @ignopeverell points out) it is reasonable to default to version 2 here
* historical headers (say during fast sync) - we want to create non-default headers taking height into consideration
  * any tests touching pre and post hard fork behavior will want to use non-default versions

But to be honest I have not thought too deeply about this - 
  * we only have a single version today
  * its not unreasonable to have a default now, until we decide otherwise
  * its easy to rework this as necessary in the future

PR #2811, opened 2019-05-07
Protocol version type safety

antiochp:
* Introduce a `ProtocolVersion(u32)` struct for safely representing protocol versions

Related #2809.

This gives us a nice struct on which to build some protocol_version support while preventing accidental use of arbitrary `u32` values.

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -362,7 +362,7 @@ pub struct Shake {
 impl Writeable for Shake {
 	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
 		self.version.write(writer)?;
-		ser_multiwrite!(writer, [write_u32, self.capabilities.bits()]);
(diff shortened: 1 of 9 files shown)


p2p/src/msg.rs
```diff
@@ -325,11 +361,8 @@ pub struct Shake {
 
 impl Writeable for Shake {
 	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
-		ser_multiwrite!(
-			writer,
-			[write_u32, self.version],
-			[write_u32, self.capabilities.bits()]
-		);
+		self.version.write(writer)?;
+		ser_multiwrite!(writer, [write_u32, self.capabilities.bits()]);
```
DavidBurkett: You can get rid of the ser_multiwrite! here, and just use writer.write_u32(self.capabilities.bits())

antiochp: :+1:

Issue #2740, opened 2019-04-10
Bittorrent support for txhashset zip download


ignopeverell:
I've had this in mind for quite a while and was actually the main reason to opt for a simple package file. The idea is simply to have peers download the file over bittorrent to speed it up, instead of chunking using some file-specific formats (i.e. kernels, then outputs, etc). The risk brought by a dependency against a bittorrent client or library (https://github.com/Luminarys/synapse or https://github.com/GGist/bip-rs come to mind) can be limited by simply running the download in a different process.

Note that this would also require some sort of regular snapshotting of the state. Right now a client asks for the txhashset at a specific block and if another one came a couple blocks after, another archive would have to be produced. But there is no reason for it to work this way, archives could just be produced on demand and reused over 500 or 1000 blocks.

antiochp:
Related - #2743 may be redundant now 😞...

I think there is still a case to be made for splitting the txhashset into two separate downloadable payloads - 
* UTXO set (outputs and rangeproofs)
* Kernels 

The reason I'm arguing this is because of how different these two bits of data behave - specifically one is prunable and compactable (outputs + rangeproofs) vs. the kernels which are not.
Kernel download can be simplified significantly (see #2743) due to the unprunable nature of the data - we do not need the hash file, just the data file.

ignopeverell:
I don't disagree and I think #2743 is still good to have. Torrenting 2 files (or more) vs one doesn't make much difference.

antiochp:
:+1: Just wanted to make sure #2743 wasn't going down the wrong path. I think a couple of torrents would actually make a lot of sense - the kernel torrent being simply the `pmmr_data.bin` file at some predefined set of "snapshot" set of block heights (every 1,000 blocks say).

I'm not that familiar with bittorrent but I'm assuming if I torrent a file (say kernels up to height 100,000) and you independently build the "same" file then they would both generate the same file hash and be seen as duplicates?

ignopeverell:
Yes, that's correct. Bittorrent also has an envelop format so you can download the files all at once or individually.

cadmuspeverell:
@ignopeverell I fail to understand how bittorrent gains us anything.  How would we maintain consensus on the metadata in what would be our equivalent of a torrent file?  Torrents just give us a structure committed to by cryptographic hashes.  We already have such a structure: Merkle Mountain Ranges.

[...]

Issue #2806, opened 2019-05-06
txhashset snapshot zips and dirs still using up a lot of disk space


antiochp:
We clean up "old" fast sync snapshot dirs and zip files but I suspect we need to be more aggresive with these and only keep the last n around.
Ideally we can get rid of them shortly after using them. Maybe our node can determine this based on how we serve the txhashset request to the syncing peer.

```
-rw-r--r--  1 root root    0 Feb 20 16:33 grin.lock
drwxr-xr-x  4 root root 4.0K Jan 15 15:38 header
drwxr-xr-x  2 root root 4.0K Jan 15 15:38 lmdb
drwxr-xr-x  2 root root 4.0K Feb 25 09:17 peer
drwxr-xr-x  5 root root 4.0K May  5 12:03 txhashset
-rw-r--r--  1 root root 242M May  5 16:34 txhashset_snapshot_100376.zip 
-rw-r--r--  1 root root 241M May  5 14:39 txhashset_snapshot_123868.zip 
-rw-r--r--  1 root root 241M May  5 15:13 txhashset_snapshot_15110.zip
-rw-r--r--  1 root root 241M May  5 15:03 txhashset_snapshot_329292.zip
-rw-r--r--  1 root root 242M May  5 18:57 txhashset_snapshot_454730.zip
-rw-r--r--  1 root root 243M May  6 01:33 txhashset_snapshot_454978.zip
-rw-r--r--  1 root root 243M May  5 22:22 txhashset_snapshot_475856.zip
-rw-r--r--  1 root root 241M May  5 13:03 txhashset_snapshot_527162.zip 
-rw-r--r--  1 root root 241M May  5 14:52 txhashset_snapshot_694734.zip
-rw-r--r--  1 root root 243M May  6 00:44 txhashset_snapshot_69812.zip
-rw-r--r--  1 root root 244M May  6 05:22 txhashset_snapshot_833786.zip
-rw-r--r--  1 root root 242M May  5 21:29 txhashset_snapshot_85703.zip
-rw-r--r--  1 root root 243M May  6 02:49 txhashset_snapshot_879868.zip
-rw-r--r--  1 root root 242M May  5 18:05 txhashset_snapshot_896996.zip
drwxr-xr-x  5 root root 4.0K May  5 16:34 txhashset_zip_100376 
drwxr-xr-x  5 root root 4.0K May  5 14:38 txhashset_zip_123868 
[...]

PR #2813, opened 2019-05-08
Generate txhashset archives on 720 block intervals.

cadmuspeverell:
This is the first of several PRs to resolve https://github.com/mimblewimble/grin/issues/2740. With this PR, txhashset archives will only be generated for blocks 250, 500, 750, etc. This inadvertently solves https://github.com/mimblewimble/grin/issues/2806 as well.

Marked as WIP to encourage review and discussion while I write tests.


garyyu:
Finally, all 3 brothers are here 😃 👍

antiochp:
Hello brother.

cadmuspeverell:
This is ready for review whenever you get a chance @antiochp

p2p/src/protocol.rs
```diff
@@ -251,15 +251,17 @@ impl MessageHandler for Protocol {
 					sm_req.hash, sm_req.height
 				);
 
-				let txhashset = self.adapter.txhashset_read(sm_req.hash);
```
antiochp: We need to think this through a bit. Our peer sends us a height and hash (the `sm_req`) but we are now effectively ignoring this and just sending the most recent snapshot.

I wonder if we need to take the height send by our peer and quantize this to the nearest previous snapshot (i.e. the closest `720` snapshot to that height).

Alternatively maybe we should be ignoring the values passed in by our peer - all we care about is serving the most recent snapshot?

cadmuspeverell: I thought about it for a while, but I don't think we would want to provide them any other txhashset. I considered verifying that the requested header is within an expected range, but it seemed unlikely enough to not justify any additional complexity.

antiochp:
I need to think through the horizon behavior here a bit more but looks good so far.

cadmuspeverell:
> I need to think through the horizon behavior here a bit more but looks good so far.

Sure, take your time.  This does effectively change the horizon temporarily for newly synced nodes.  Once most nodes have this change, it might make sense to just make txhashset_archive_header the new cut_through_horizon.  Perhaps as part of the first hard fork?

antiochp:
Currently we make some guarantees/assumptions about what blocks are available from peers.

We maintain 7 days of full block history based on `global::cut_through_horizon()`.
But on initial sync we ask for a txhashset based on 48 hours with `global::state_sync_threshold()`.

So every running node has at least 48 hours of full blocks (after successful sync).
And every running node has up to 7 days of full blocks based on compaction/pruning.

One of the edge cases we need to account for here is when we ask for full blocks during fast sync from another node that has recently sync'd. i.e. Other nodes (possible a majority of nodes on the network) are only guaranteed to have blocks from 48 hours ago. Currently this is _exactly_ 48 hours ago.

So there are a couple of conflicting requirements here pulling in competing directions -

* If we ask for a txhashset older than 48 hours then we risk not being able to reliably sync full blocks from that point onwards.
* If we ask for a txhashset younger than 48 hours then _we_ can (selfishly) sync but we are unable to provide 48 hours worth of blocks to other peers who may also by sync'ing.

So currently we are kind of stuck doing the same thing as all other peers and setting our sync_threshold to 48 hours.

If we were to deploy a node with different behaviour by for example using more of a "step function" in terms of 740 block periods then we risk being a bad citizen with respect to these conflicting requirements. We either have too few blocks locally (temporarily) or we are asking for too many blocks and cannot reliably retrieve them (if our peers are also syncing).

cadmuspeverell:
> We  either have too few blocks locally (temporarily) or we are asking for  too many blocks and cannot reliably retrieve them (if our peers are also syncing).

The way it's coded in this PR, it would be the latter.  This was coded that way intentionally, for the reasons you listed above.  Although initial sync might be delayed very slightly if it requests blocks from a peer that just synced, it shouldn't be a large problem because the significant majority of peers would have more than 48 hours worth of blocks.

garyyu:
@cadmuspeverell  thanks for your work and how are you doing?

I would like to take a look and test this PR and could you please solve the conflict?  thanks.

antiochp:
Q) If we only produce a new txhashset every 12 hours - do we now reuse one already generated if we are asked for it more than once in a 12 hour period?

cadmuspeverell:
> Q) If we only produce a new txhashset every 12 hours - do we now reuse one already generated if we are asked for it more than once in a 12 hour period?
> 
> I don't remember if we implemented de-duping behavior before (we had a smaller window before).

Yep!  That logic was already there - and seemed to work well when I was doing my testing.  I'll resolve that darned conflict now and then maybe @garyyu can try it out to make sure I didn't miss anything?  :smile_cat:

garyyu:
After my 2nd reading everything is good on my view, and thanks @cadmuspeverell for this 1st PR 👍 and a long review :-)

antiochp:
Looks like this PR was merged into `master` after the release of `1.1.0`.
Can we revert this on `master` and retarget it for the `2.0.0` branch (its tagged as `2.x.x` milestone). 

If we do need to push `1.1.x` for any reason we will do this from `master` and we only want to include critical fixes.

I'm not 100% sure if this is the approach we will always take but we have some tight time constraints between `1.1.0` and `2.0.0` and we decided to keep `master` clean for a `1.1.1` release if necessary.

garyyu:
> its tagged as 2.x.x milestone

but it was tagged as 1.1.1 milestone before, you can see above history.
And I don't see any risk for this PR in 1.1.1, and we didn't yet plan a release date about 1.1.1, did we?

yeastplume:
@garyyu There is no 1.1.1 milestone anymore, in recent meetings we've collectively decided that the next release will be 2.0.0, and that there will only be a 1.1.1 if there is a critical issue uncovered that warrants it. Also, given the time constraints we should only be pushing PRs that are hard-fork related to the branch, and labeling any other fixes or changes as post 2.0.0 (i've created a 2.x.x milestone for that)

Sorry this wasn't communicated properly as there have been one or two extra meeting schedules over the past days, but we really feel it's important that everyone stick to this plan for the next few weeks.

PR #2814, opened 2019-05-08
[WIP] simplify (maybe) is_bad_data peer banning

antiochp:
Experimental change to `header_received` to rework the peer banning logic - 

* adapter returns `Result<(), chain::Error>` and not `Result<bool, chain::Error>`
* adapter no longer responsible for intercepting the "bad data" error and translating it into an `Ok(false)`
* we ban the peer based on `error.is_bad_data()` 

I'm making the argument here that `adapters.rs` should not be responsible for identifying "bad data" errors and translating them into `Ok(false)`. 
This way `adapters.rs` simply passes any `chain::Error` through to the p2p layer. 
The p2p layer is responsible for deciding to ban the peer based on the underlying `chain::Error` which it already needs to deal with.

### DIFF
--- a/p2p/src/peer.rs
+++ b/p2p/src/peer.rs
@@ -500,7 +500,7 @@ impl ChainAdapter for TrackingAdapter {
 		&self,
 		bh: core::BlockHeader,
 		peer_info: &PeerInfo,
-	) -> Result<bool, chain::Error> {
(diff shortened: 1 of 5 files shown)


antiochp:
Closing this for now - it was simply a spike to try out a refactoring and was incomplete.
Will revisit at some future point.

Issue #2585, opened 2019-02-16
db/store full (forked) blocks cleanup


antiochp:
We compact/prune the blocks db periodically to remove historical full blocks from the db older than 1 week (unless running in archive mode).

But - we do this based on the current chain by traversing back from the head of the chain.

So if we see any short lived forks or reorgs we will never actually clean these forked blocks up and they will remain in the db indefinitely.

i.e. if we have the following blocks in the chain -

`B1 <- B2 <- B3` etc.

We will never remove any forked blocks, say `B2'` for example.

pml10628:
1ede817398a84d4c8533f423444681cf

192-sean:
Where would I start if I were to tackle this.

antiochp:
Right now we traverse back through historical blocks and then go to the db for each one to be removed. But I _think_ a better approach may be start at the db and simply iterate over the blocks in the db, removing any that are candidates for removal.

This might be a good introduction to the lower level db access stuff, so I'd start looking around our lmdb code to see how we might use an iterator over full blocks. We do something similar in the peers db (via an iterator).

daogangtang:
I would like to process this.

daogangtang:
Hi, @antiochp  I have the following questions:

If we clean old blocks by iterating db directly, how we check which block is at a short-lived fork? or even a fork? 

And, should we remove all folk blocks from db?

And I proposed a request on #2807 , does it walk the right way?

Thank you.

hayeah:
My understanding is that instead of using `batch.get_previous_header(&current)` to iterate through the blocks, you should access the underlying store of the Batch type (which is a LMDB instance)

```rust
/// An atomic batch in which all changes can be committed all at once or
/// discarded on error.
pub struct Batch<'a> {
	db: store::Batch<'a>,
}
```

Create a LMDB iterator to access ALL the blocks, regardless of whether the block is in the chain or not.

Then we need to have an efficient way to check whether a particular block exists on the chain (if not, remove it).

To check whether a block is on chain, one naive way is to iterate through the chain (using `get_previous_header`), and save block hashes in a map. The CUT_THROUGH_HORIZON is ~10k blocks, so we'll have to keep 10k entries.

antiochp:
> Then we need to have an efficient way to check whether a particular block exists on the chain (if not, remove it).

```
chain.is_on_current_chain()
```

This checks the header against the header at that height based on our current chain - if they match then we're good (header at height `h` is the one we think it is).

[...]

PR #2815, opened 2019-05-09
fix: try to fix issue #2585 by adding block cleanup from db directly.

daogangtang:
fix: try to fix issue #2585 by adding block cleanup from db directly.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1029,10 +1029,10 @@ impl Chain {
 		let tail_hash = txhashset.get_header_hash_by_height(head.height - horizon)?;
 		let tail = batch.get_block_header(&tail_hash)?;
 
-		// Remove old blocks and short lived fork blocks
+		// Remove old blocks (including short lived fork blocks) which height < tail.height
 		// here b is a block
 		for (_, b) in batch.blocks_iter()? {
-			if self.is_on_current_chain(&b.header).is_err() || b.header.height < tail.height {
+			if b.header.height < tail.height {
 				let _ = batch.delete_block(&b.hash());
 				count += 1;
 			}
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1029,14 +1029,14 @@ impl Chain {
 		let tail_hash = txhashset.get_header_hash_by_height(head.height - horizon)?;
 		let tail = batch.get_block_header(&tail_hash)?;
 
-		// Remove old blocks and short live fork blocks
-		// here b is a lived block
-		let _ = batch.iter_lived_blocks()?.map(|(_, b)| {
+		// Remove old blocks and short lived fork blocks
+		// here b is a block
+		for (_, b) in batch.blocks_iter()? {
 			if self.is_on_current_chain(&b.header).is_err() || b.header.height < tail.height {
 				let _ = batch.delete_block(&b.hash());
 				count += 1;
 			}
-		});
+		}
 
 		batch.save_body_tail(&Tip::from_header(&tail))?;
 
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -34,7 +34,7 @@ use crate::types::{
 use crate::util::secp::pedersen::{Commitment, RangeProof};
 use crate::util::{Mutex, RwLock, StopState};
 use grin_store::Error::NotFoundErr;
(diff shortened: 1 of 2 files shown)


chain/src/chain.rs
```diff
@@ -1065,6 +1065,49 @@ impl Chain {
 			count, tail.height
 		);
 
+		// Remove short live fork blocks
+		// Get all block headers in db
+		let all_block_headers = self.store.all_block_headers()?;
```
cadmuspeverell: You'll want to iterate over all blocks, not block headers. We never compact/delete the block headers, so using your approach would end up iterating over 100s of thousands of headers. Blocks are compacted frequently, which means we never have more than a few weeks worth of them. Just iterate over each block in the store, check `is_on_current_chain` and check block.header.height >= tail.height. If either of those is false, call delete_block.

daogangtang: Hi, updated with new algorithm u advised :)

chain/src/store.rs
```diff
@@ -378,6 +378,13 @@ impl<'a> Batch<'a> {
 			db: self.db.child()?,
 		})
 	}
+
+	/// An iterator to all lived block now
+	pub fn iter_lived_blocks(&self) -> Result<SerIterator<Block>, Error> {
```
antiochp: Maybe rename this to `blocks_iter()` for consistency with `difficulty_iter()` (and others)?

daogangtang: Ok.

chain/src/chain.rs
```diff
@@ -1026,38 +1026,18 @@ impl Chain {
...
-			}
-			if current.height <= 1 {
-				break;
-			}
-			match batch.get_previous_header(&current) {
-				Ok(h) => current = h,
-				Err(NotFoundErr(_)) => break,
-				Err(e) => return Err(From::from(e)),
+		// Remove old blocks and short live fork blocks
+		// here b is a lived block
```
antiochp: I'm not sure what you mean by "lived" block - maybe reword these comments?

daogangtang: Ok.

chain/src/chain.rs
```diff
@@ -1026,38 +1026,18 @@ impl Chain {
...
+		let _ = batch.iter_lived_blocks()?.map(|(_, b)| {
```
antiochp: This can be simplified to - 

```
for (_, b) in batch.iter_lived_blocks()? {
    ...
}
```

daogangtang: Ok.

antiochp:
Looks like lmdb is locking up internally with this change - our "batch" is very susceptible to deadlocking with multiple accesses.
I'm looking to see if there are any obvious culprits here.

daogangtang:
> Looks like lmdb is locking up internally with this change - our "batch" is very susceptible to deadlocking with multiple accesses.
> I'm looking to see if there are any obvious culprits here.

Wow, this is my first attempt to contribute to grin, so lack experiences on it. :(

daogangtang:
Or should I use store_instance.iter() rather than batch.iter() in exported iterator in chain/src/store.rs?

antiochp:
Ha - It's the call to `is_on_current_chain()`. 
Turns out we cannot actually call this inside the iterator (as it tries to get a lock on the db but the batch already has a lock on it).
I'll take another pass later today and see if we can rework this to use the existing batch.

daogangtang:
In my opinion, maybe we don't need batch lock in the procedure of iterating over blocks in db, this is not necessary for the whole block cleanup, only necessary when delete one block. So, we can add an iterator to ChainStore implementation, rather than Batch.

antiochp:
I think we actually got the logic wrong here.

Now that we are iterating over _all_ blocks in the db, all we really care about is cleaning up blocks older than the threshold.

So we do not need to actually check for `is_on_current_chain()` - if its older than `tail.height` we simply delete it, whether on current chain or not.

And this should resolve the deadlocking issue as well I think.

daogangtang:
The height of a fork block MUST be smaller than tail.height? Surely?

antiochp:
> The height of a fork block MUST be smaller than tail.height? Surely?

Yes - we want to delete block older (i.e. lower height) than tail.height

PR #2819, opened 2019-05-09
Use read_u16 etc (byteorder) in StreamingReader

antiochp:
Our `StreamingReader` impl differs from the `BinReader` impl as we are not using byteorder to ensure consistent handling of multiple byte values.

BinReader - 

https://github.com/mimblewimble/grin/blob/998824ecf3e0d8a53c26b93060cc8bba4f51a65d/core/src/ser.rs#L321-L323

StreamingReader - 

https://github.com/mimblewimble/grin/blob/998824ecf3e0d8a53c26b93060cc8bba4f51a65d/core/src/ser.rs#L399-L402

This PR makes `StreamingReader` use the various `byteorder` fns and makes `BigEndian` explicit in these various read fns.

### DIFF
--- a/core/src/ser.rs
+++ b/core/src/ser.rs
@@ -392,33 +392,22 @@ impl<'a> StreamingReader<'a> {
 
 impl<'a> Reader for StreamingReader<'a> {
 	fn read_u8(&mut self) -> Result<u8, Error> {
-		let buf = self.read_fixed_bytes(1)?;
-		deserialize(&mut &buf[..])
+		self.stream.read_u8().map_err(map_io_err)
 	}
-
 	fn read_u16(&mut self) -> Result<u16, Error> {
-		let buf = self.read_fixed_bytes(2)?;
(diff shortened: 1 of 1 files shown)


antiochp:
See #2820. I think I remember now why I did not use the `byteorder` fns here - its related to `read_exact` and what we need to do to ensure we read the correct number of bytes off a non-blocking stream.

So I _think_ we may need to revisit how we do this in both `StreamingReader` and `BinReader`.
I suspect the `BinReader` impl actually puts us at risk of partial reads etc.

ignopeverell:
I think you need to rebase master on your PR for CI to work.

antiochp:
Closing. See #2820.
This appears to be the wrong approach - we need to use a combination of our custom `read_exact` with explicit `BigEndian` byte ordering, for both `BinReader` and `StreamingReader`.

Will open a new PR later.

PR #2821, opened 2019-05-10
make sure we use BigEndian when reading numbers via StreamingReader

### LINKED ISSUE #2820: byteorder crate does not play nicely with our "non-blocking" IO
Related #2819.

We jump through a lot of hoops with `read_exact` in our `util` crate to make sure we can safely and robustly read multiple bytes via non-blocking tcp streams.

But when we call the various low-level `byteorder` derived fns we skip this and go straight to the `read_exact` impl in `std::io::Read` -

https://docs.rs/byteorder/1.3.1/src/byteorder/io.rs.html#110-113

https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact

It seems like it is entirely possible for a u64 or other multi-byte value here to cause our non-blocking IO to fail on a partial read.
We can be in a situation where we read say 5 bytes of the 8 bytes of a u64 and return early with effectively corrupted data.

antiochp:
Resolves #2820.

This PR fixes a narrow part of #2820 where we are not using `BigEndian` explicitly in `StreamingReader`.

This has an additional benefit where we get rid of a couple of levels of indirection. We used to deserialize numbers (`u64` etc.) in a `StreamingReader` by instantiating a new `BinReader` within the `deserialize` call for each number being deserialized.

### DIFF
--- a/core/src/ser.rs
+++ b/core/src/ser.rs
@@ -393,32 +393,32 @@ impl<'a> StreamingReader<'a> {
 impl<'a> Reader for StreamingReader<'a> {
 	fn read_u8(&mut self) -> Result<u8, Error> {
 		let buf = self.read_fixed_bytes(1)?;
-		deserialize(&mut &buf[..])
+		Ok(buf[0])
 	}
 
 	fn read_u16(&mut self) -> Result<u16, Error> {
 		let buf = self.read_fixed_bytes(2)?;
-		deserialize(&mut &buf[..])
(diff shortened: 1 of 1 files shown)


hashmap:
What if we just switch to blocking io?

antiochp:
> What if we just switch to blocking io?

That would be one option that would presumably simplify this. I'm not 100% convinced blocking io is the right solution here.

Issue #2808, opened 2019-05-07
Documentation Bug


notorious-clay:
**Documentation Bug**
In the Documentation **Introduction to MimbleWimble and Grin** at the part Ownership after the last example it says: _"But because 85*G is a valid public key on the elliptic curve G..."_

It should be on the elliptic curve C because G is the generator point.

kargakis:
https://github.com/mimblewimble/grin/pull/2823 was just merged, this issue can be closed.

PR #2823, opened 2019-05-14
Fix the curve name in one section of intro

hashmap:


GandalfThePink:
I think that G is just the generating point and not the curve itself. G can be multiplied, i.e. k * G, but C instead is an equation that specified the curve. I think the old version was fine.

hashmap:
@GandalfThePink it seems we both agree that `G` is not the curve. What I'm trying to fix is exactly that statement: `on the elliptic curve _G_,`

GandalfThePink:
Yes, ... I must be colorblind. 

I then fully agree with your modification.

kargakis:
Fixes https://github.com/mimblewimble/grin/issues/2808

PR #2834, opened 2019-05-21
Check for known block when processing "header first"

antiochp:
When processing a header in "header first" propagation we currently perform a quick check to see if we have just processed this header.
This does not cover the case where a peer sends (intentionally or otherwise) an old header.

We should do the same check here that we already do when processing _full_ blocks - 
* Is the block for this header on the head of the current chain?
* Is the block for this header in our set of orphan blocks?
* Is the block for this header in our database?

If this is a header for a "known" block then we should halt processing of the header.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -75,10 +75,10 @@ fn process_header_for_block(
 
 // Check if we already know about this block for various reasons
 // from cheapest to most expensive (delay hitting the db until last).
-fn check_known(block: &Block, ctx: &mut BlockContext<'_>) -> Result<(), Error> {
-	check_known_head(&block.header, ctx)?;
-	check_known_orphans(&block.header, ctx)?;
-	check_known_store(&block.header, ctx)?;
+fn check_known(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(), Error> {
+	check_known_head(header, ctx)?;
+	check_known_orphans(header, ctx)?;
+	check_known_store(header, ctx)?;
 	Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


chain/src/pipe.rs
```diff
@@ -260,19 +260,10 @@ pub fn process_block_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) ->
...
-	Ok(())
-}
+	// Check if this header is already "known" from processing a previous block.
+	check_known(header, ctx)?;
 
-/// Quick in-memory check to fast-reject any block header we've already handled
-/// recently. Keeps duplicates from the network in check.
-/// ctx here is specific to the header_head (tip of the header chain)
-fn check_header_known(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(), Error> {
-	let header_head = ctx.batch.header_head()?;
```
garyyu: It said `ctx here is specific to the header_head (tip of the header chain)`.

but in `check_know()`, the `check_known_head(header, ctx)` is using `ctx.batch.head()`, not `header_head`.

antiochp: This code is deleted.

garyyu: I mean your new code has very different behavior from the old `check_header_known` which use header_head.  now you're using `ctx.batch.head()`.

antiochp: We used to check for the header itself was known based on the header chain.
The only time we would know about headers and not yet now about the full blocks is during initial sync.
So it's safe to check if we know about the header based on the full chain - we're processing "header first" in order to process full blocks.

PR #2836, opened 2019-05-22
[TUI] Replace S and R with corresponding ASCII arrows in bandwidth column

eupn:
This PR just replaces `S` (sent) and `R` (received) letters in TUI's `Peers and Sync`'s `Bandwidth` column with the corresponding ASCII arrows:

- Arrow up (↑) for *Sent*
- Arrow down (↓) for *Received*

Screenshot:

image

This makes bandwidth stats column more intuitive and easier to read. 
There shouldn't be issues with different terminals since they're all supporting ASCII.


antiochp:
Nice!

Issue #2833, opened 2019-05-17
Include reorg depth in the log (and ideally as part of the webhook payload)


kargakis:
When I see the following log I would like to know how deep the reorg is. Depending on the depth, an alert may be sent (I would like to get alerted when a reorg is hours deep but probably don't care about minutes).

```
May 14 14:35:16 my-node grin[32733]: 20190514 14:35:16.943 WARN grin_servers::common::hooks - block_accepted (REORG!): 00002b652b46 at 170558 (diff: 512514316382489)
May 14 19:27:08 my-node grin[32733]: 20190514 19:27:08.526 WARN grin_servers::common::hooks - block_accepted (REORG!): 0000002c9b08 at 170846 (diff: 513254790279165)
May 15 19:13:16 my-node grin[32733]: 20190515 19:13:16.613 WARN grin_servers::common::hooks - block_accepted (REORG!): 000006d06a7b at 172267 (diff: 516892312421434)
May 15 20:08:19 my-node grin[32733]: 20190515 20:08:19.692 WARN grin_servers::common::hooks - block_accepted (REORG!): 0000351edafc at 172319 (diff: 517025510854039)
May 17 11:21:22 my-node grin[32733]: 20190517 11:21:22.708 WARN grin_servers::common::hooks - block_accepted (REORG!): 0000067e21fd at 174618 (diff: 521965930623788)
```

kargakis:
Fixed by #2839

PR #2839, opened 2019-05-23
Report reorg depth in block status

eupn:
This PR implements reporting of chain reorganization depth in `BlockStatus::Reorg` enum member.

- Adds a `u64` field to `BlockStatus::Reorg` that expresses the depth of the chain reorg
- Addresses #2833 by including depth of the reorg in log and webhook payload
- Adds integration test that creates chain reorganization and checks that reorg is reported correctly

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -259,7 +259,7 @@ impl Chain {
 			if head.prev_block_h == prev_head.last_block_h {
 				is_next_block = true;
 			} else {
-				reorg_depth = Some(prev_head.height.saturating_sub(head.height));
+				reorg_depth = Some(prev_head.height.saturating_sub(head.height) + 1);
 			}
 		}
 
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -254,15 +254,18 @@ impl Chain {
 		let is_more_work = head.is_some();
 
 		let mut is_next_block = false;
+		let mut reorg_depth = None;
 		if let Some(head) = head {
 			if head.prev_block_h == prev_head.last_block_h {
 				is_next_block = true;
+			} else {
+				reorg_depth =  Some(prev_head.height.saturating_sub(head.height));
 			}
 		}
 
 		match (is_more_work, is_next_block) {
 			(true, true) => BlockStatus::Next,
-			(true, false) => BlockStatus::Reorg,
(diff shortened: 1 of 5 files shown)


chain/src/chain.rs
```diff
@@ -254,15 +254,18 @@ impl Chain {
...
 				is_next_block = true;
+			} else {
+				reorg_depth = Some(prev_head.height.saturating_sub(head.height) + 1);
 			}
 		}
 
 		match (is_more_work, is_next_block) {
 			(true, true) => BlockStatus::Next,
-			(true, false) => BlockStatus::Reorg,
+			(true, false) => BlockStatus::Reorg(reorg_depth.unwrap_or(0)),
```
antiochp: Nice. Thank you for adding this to the enum and not passing around another variable...

servers/src/common/hooks.rs
```diff
@@ -262,9 +263,9 @@ impl WebHook {
 impl ChainEvents for WebHook {
 	fn on_block_accepted(&self, block: &core::Block, status: &BlockStatus) {
 		let status = match status {
-			BlockStatus::Reorg => "reorg",
-			BlockStatus::Fork => "fork",
-			BlockStatus::Next => "head",
+			BlockStatus::Reorg(depth) => format!("reorg:{}", depth),
```
antiochp: If we are building json here then might make sense to add a new (optional) field for the reorg depth.
It would save clients from having to parse this out of the string on the other side.

antiochp:
```
# optional depth field, only used for reorg status
{
  "hash": "abc",
  "status: "reorg",
  "depth": 1,
  "data": { 
    ... 
  }
}

# maybe status is optionally a dict and not just a simple value (harder to parse though)
{
  "hash": "abc",
  "status: {
    "reorg": 1,
  },
  "data": { 
    ... 
  }
}
```

eupn: @antiochp I've made option no. 1 that adds optional `depth` field on the top level.

antiochp: :+1: Looks good to me. 

We should get some feedback from someone who uses this endpoint.

eupn: @kargakis 👀

kargakis: Looks good to me, thanks for working on this @eupn !

PR #2843, opened 2019-05-25
get `/v1/blocks` without merkle_proof

invis87:
add flag 'no_merkle_proof' to `/v1/blocks` api to get blocks without checking merkle proof

### DIFF
--- a/api/src/handlers/blocks_api.rs
+++ b/api/src/handlers/blocks_api.rs
@@ -79,15 +79,18 @@ impl Handler for HeaderHandler {
 ///
 /// Optionally return results as "compact blocks" by passing "?compact" query
 /// param GET /v1/blocks/<hash>?compact
+///
+/// Optionally turn off the Merkle proof extraction by passing "?no_merkle_proof" query
+/// param GET /v1/blocks/<hash>?no_merkle_proof
 pub struct BlockHandler {
 	pub chain: Weak<chain::Chain>,
 }
 
 impl BlockHandler {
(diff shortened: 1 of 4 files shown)


hashmap:
Thanks for the contribution! LGTM, would you mind to add a test to https://github.com/mimblewimble/grin-wallet/blob/master/integration/tests/simulnet.rs I know it seems strange, but we moved integration tests to the wallet repo.

invis87:
@hashmap Yep, ofc, I didn't find tests on `grin` project thats why my PR without tests. But now I know where are they hide :)

hashmap:
@invis87 yeah, this is confusing and less safe, I think we should bring node-specific tests back

invis87:
@hashmap sorry for stupid question, but how can I run integration tests in `grin_wallet`:
1. run it
2. run on top of my changes on `grin`

I found that integration tests are disabled in `grin_wallet` :(

But of course I can add tests to `"http://{}:{}/v1/blocks/{}?no_merkle_proof"` in the same manner as it is done for `"http://{}:{}/v1/blocks/{}?compact"`. If I do so - should it be different PR or... How to do it right :) ?

hashmap:
There is no such thing as a stupid question. We discussed this situation during this week’s dev meeting and decided to move node-related tests back to grin repo. I’d recommend to create an issue to implement this test, we can do it after moving the tests. I’ll test your code manually meanwhile, thanks!

invis87:
@hashmap can you close https://github.com/mimblewimble/grin/issues/1477 please?

Issue #2218, opened 2018-12-23
Security checklist


ignopeverell:
Some feedback from the libsecp audit and our use in grin, all simple things we could add for a little more defense in depth:

* Zeroing of sensitive data through Drop (password, mnemonic)
* Check zeroing of private keys in libsecp
* More randomized tests to check invalid range proofs don't validate
* Same for aggsigs

brumball:
randomness sources if even necessary.
E.g. https://github.com/mimblewimble/rust-secp256k1-zkp/blob/master/src/aggsig.rs#L36
uses `thread_rng` which, AFAIK, is backed by `EntropyRng` instead of `OsRng`

ignopeverell:
Randomness was already checked, with rand 0.5+ it's fine.

yeastplume:
Closing, out of date

PR #2847, opened 2019-05-25
Fill BlindingFactor with zeros on Drop

eupn:
This is a first (and probably naïve) step towards security of sensitive data such as `BlindingFactor`.

In response to #2218. Cleaning of `secp`'s `SecretKey` and wallet's mnemonic should be done in corresponding modules.

This PR:
* Adds [zeroize](https://crates.io/crates/zeroize) crate as a dependency for [keychain](https://github.com/mimblewimble/grin/tree/master/keychain)
* Adds `Zeroize` derive for `BlindingFactor` in a way that underlying byte array is filled with zeroes
* Removes `Copy` derive from `BlindingFactor` since it is impossible to implement both `Drop` and `Copy` at the same time. All implicit copying were replaced with more explicit `.clone()`
* Adds unit test for `BlindingFactor` zeroing

Some references:
https://github.com/dalek-cryptography/curve25519-dalek/issues/11
https://github.com/rust-lang/rfcs/issues/2533
https://www.youtube.com/watch?v=cQ9wTyYCdNU

### DIFF
--- a/keychain/src/types.rs
+++ b/keychain/src/types.rs
@@ -231,10 +231,10 @@ impl fmt::Display for Identifier {
 #[derive(Default, Clone, PartialEq, Serialize, Deserialize, Zeroize)]
 pub struct BlindingFactor([u8; SECRET_KEY_SIZE]);
 
-// Intentionally empty `Default` implementation to prevent leakage.
+// Dummy `Debug` implementation that prevents secret leakage.
 impl fmt::Debug for BlindingFactor {
 	fn fmt(&self, _f: &mut ::std::fmt::Formatter<'_>) -> fmt::Result {
-		Ok(())
+		write!(f, "BlindingFactor(<secret key hidden>)")
 	}
 }
 
--- a/keychain/src/types.rs
+++ b/keychain/src/types.rs
@@ -489,8 +489,38 @@ mod test {
 	use rand::thread_rng;
 
 	use crate::types::{BlindingFactor, ExtKeychainPath, Identifier};
+	use crate::util::secp::constants::SECRET_KEY_SIZE;
 	use crate::util::secp::key::{SecretKey, ZERO_KEY};
 	use crate::util::secp::Secp256k1;
+	use std::slice::from_raw_parts;
+
+	// This tests cleaning of BlindingFactor (e.g. secret key) on Drop.
+	// To make this test fail, just remove `Drop` implementation for `BlindingFactor`.
+	#[test]
+	fn blinding_factor_clear_on_drop() {
+		// Create buffer for blinding factor filled with non-zero bytes.
+		let bf_bytes = [0xAA; SECRET_KEY_SIZE];
+		let ptr = {
+			// Fill blinding factor with some "sensitive" data
+			let bf = BlindingFactor::from_slice(&bf_bytes[..]);
+			bf.0.as_ptr()
+
+			// -- here BlindingFactor should be zeroed since it is `Drop`ed
+		};
+
+		// Unsafely get data from where BlindingFactor was in memory. Should be all zeros.
+		let bf_bytes = unsafe { from_raw_parts(ptr, SECRET_KEY_SIZE) };
+
+		// There should be all zeroes.
+		let mut all_zeros = true;
+		for b in bf_bytes {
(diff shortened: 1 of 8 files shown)


hashmap:
We use `zeroize` crate in some parts of grin, would you consider using it?

keychain/src/types.rs
```diff
@@ -227,12 +228,13 @@ impl fmt::Display for Identifier {
 	}
 }
 
-#[derive(Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[derive(Default, Clone, PartialEq, Serialize, Deserialize, Zeroize)]
 pub struct BlindingFactor([u8; SECRET_KEY_SIZE]);
 
+// Intentionally empty `Default` implementation to prevent leakage.
```
garyyu: This line of comment confuse me a bit, is it talking about `Default` derive?  or `Debug`? because next line is Debug impl.

eupn: Good catch! Obviously a typo. I meant to make blank Debug impl, not Default.

antiochp: Minor point but this got me thinking - can we do something other thank blank here?
We could print out `********` or `<REDACTED> `something, that way we can actually check the logs and know if we are doing we probably shouldn't. If its just blank we'll just never see it.

Also less likely that someone will come in, ignore the comment and "fix" the impl.

eupn: Seems reasonable!

keychain/src/types.rs
```diff
@@ -227,12 +228,13 @@ impl fmt::Display for Identifier {
...
 
-#[derive(Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[derive(Default, Clone, PartialEq, Serialize, Deserialize, Zeroize)]
 pub struct BlindingFactor([u8; SECRET_KEY_SIZE]);
 
+// Dummy `Debug` implementation that prevents secret leakage.
 impl fmt::Debug for BlindingFactor {
-	fn fmt(&self, f: &mut ::std::fmt::Formatter<'_>) -> fmt::Result {
-		write!(f, "{}", self.to_hex())
+	fn fmt(&self, _f: &mut ::std::fmt::Formatter<'_>) -> fmt::Result {
```
garyyu: why `_f`?
should be `f`

elichai:
FYI,
Every time you pass this struct by value it's get copied in memory without calling the Drop trait.

The better way to implement it is using the Pin struct or just putting it in a smart pointer(i.e Box) so that only the pointer will be copied and not the data itself

hashmap:
@elichai could you elaborate on why drop is not called when a cloned value gets out of its scope?

elichai:
@hashmap because semantically when you pass something by value it's not dropped. In reality the compiler will prefer to copy the value when passing by value (especially if it's on the stack)

Example https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=edc392a036c0d880224f0a94d6336f69

garyyu:
@elichai  👍 
and that's exact what I'm worrying, as comments in https://github.com/mimblewimble/grin/pull/2851#issuecomment-497566933.  thanks for a nice playground example to show that.

> The better way to implement it is using the Pin struct or just putting it in a smart pointer(i.e Box) so that only the pointer will be copied and not the data itself

Could you please also demo this in your example?
And another method to solve this is always use the explicit `b = a.clone()` instead of `b = a`.

elichai:
@garyyu With Box: https://play.rust-lang.org/?gist=26f5b9609348f4b6e287bf808ab0161a

The problem is that even though it works it's not **guaranteed**, Wrapping the Box in Pin guarantees it (e.g. https://play.rust-lang.org/?gist=3a3fabc5c44b3017b64dee1d22169d45 )

cadmuspeverell:
Maybe we could also use mlock to prevent paging to disk?

elichai:
I think that paging is a whole different conversation that deserves its own issue

garyyu:
@elichai 

> The problem is that even though it works it's not guaranteed, Wrapping the Box in Pin guarantees it 

👍  Thanks, now I understand. Would you mind to submit PRs to fix this?

PR #2848, opened 2019-05-26
Support new Bulletproof rewind scheme

jaspervdm:
Changes to the keychain and builders that are needed for https://github.com/mimblewimble/grin-wallet/issues/105. 

I introduce a new `ProofBuild` trait, that is used in building and rewinding a proof. At the moment there are two of them, `LegacyProofBuilder` and `ProofBuilder`. Wallets can supply the transaction build functions with any struct that implements `ProofBuild`. `grin-wallet` v2.0.0 will only use `ProofBuilder` for creating new outputs, but use `LegacyProofBuilder` to rewind proofs before the hard fork height.

I also did some refactoring around switch commitments. With the new rewind scheme there is a version byte specifically for the switch commitment, which means we have the possibility to turn switch commitments off or use a different type for specific outputs, without requiring additional grinding.
For now all outputs that are created will use the `Regular` type. In the future we could have full support on the wallet level for this, but this will have to be done in another PR.

This PR also has a `ViewKey` struct, that can be used to scan the chain. Currently it cannot detect switch commitment outputs because it requires 2 extra functions to be added to libsecp. See [here](https://github.com/mimblewimble/grin/pull/2848/files#diff-a6d7fb0a9c7c4f62cd0ba642f817f037R164) for more information

### DIFF
--- a/core/src/libtx/proof.rs
+++ b/core/src/libtx/proof.rs
@@ -18,7 +18,7 @@ use crate::blake2;
 use crate::keychain::{Identifier, Keychain, SwitchCommitmentType};
 use crate::libtx::error::{Error, ErrorKind};
 use crate::util::secp::key::SecretKey;
-use crate::util::secp::pedersen::{Commitment, ProofInfo, ProofMessage, RangeProof};
+use crate::util::secp::pedersen::{Commitment, ProofMessage, RangeProof};
 use crate::util::secp::{self, Secp256k1};
 
 /// Create a bulletproof
@@ -80,20 +80,26 @@ where
 	K: Keychain,
 	B: ProofBuild,
 {
-	let nonce = b.rewind_nonce(&commit)
+	let nonce = b
+		.rewind_nonce(&commit)
 		.map_err(|e| ErrorKind::RangeProof(e.to_string()))?;
 	let info = k
 		.secp()
-		.rewind_bullet_proof(commit, nonce, extra_data, proof)
-		.map_err(|e| ErrorKind::RangeProof(e.to_string()))?;
+		.rewind_bullet_proof(commit, nonce, extra_data, proof);
+	if info.is_err() {
+		return Ok(None);
+	}
+	let info = info.unwrap();
 
 	let amount = info.value;
-	let check = b.check_output(&commit, amount, info.message)
+	let check = b
+		.check_output(&commit, amount, info.message)
 		.map_err(|e| ErrorKind::RangeProof(e.to_string()))?;
 
 	Ok(check.map(|(id, switch)| (amount, id, switch)))
 }
 
+/// Used for building proofs and checking if the output belongs to the wallet
 pub trait ProofBuild {
 	/// Create a BP nonce that will allow to rewind the derivation path and flags
 	fn rewind_nonce(&self, commit: &Commitment) -> Result<SecretKey, Error>;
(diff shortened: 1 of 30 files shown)


keychain/src/types.rs
```diff
@@ -468,7 +468,19 @@ pub trait Keychain: Sync + Send + Clone {
 	fn derive_key(&self, amount: u64, id: &Identifier) -> Result<SecretKey, Error>;
 	fn commit(&self, amount: u64, id: &Identifier) -> Result<Commitment, Error>;
 	fn blind_sum(&self, blind_sum: &BlindSum) -> Result<BlindingFactor, Error>;
-	fn create_nonce(&self, commit: &Commitment) -> Result<SecretKey, Error>;
+
+	/// Create a BP nonce that will allow to rewind the derivation path and flags
+	fn create_rewind_nonce(&self, commit: &Commitment, legacy: bool) -> Result<SecretKey, Error>;
```
cadmuspeverell: When we redesign code we're really just replacing today's legacy code with tomorrow's.  Let's make it easier on our future selves!  Consider substituting a version byte for the legacy bool.

keychain/src/keychain.rs
```diff
@@ -33,6 +33,45 @@ pub struct ExtKeychain {
...
+impl ExtKeychain {
+	fn derive_key_internal(&self, amount: u64, id: &Identifier, switch: bool) -> Result<SecretKey, Error> {
+		let mut h = self.hasher.clone();
+		let p = id.to_path();
+		let mut ext_key = self.master;
+		for i in 0..p.depth {
+			ext_key = ext_key.ckd_priv(&self.secp, &mut h, p.path[i as usize])?;
+		}
+
+		match switch {
```
cadmuspeverell: I realize you are just moving this code but why would we not always use switch commitments?  If wallets are creating outputs without using switch commits they risk the possibility of losing funds if Grin is ever forced to require revealing the pre-switch commit thanks to QC or smth.

jaspervdm:
@cadmuspeverell thanks for checking the code, however I did some big refactoring on it. I decided that the keychain should be a general key derivation tool, it shouldn't even be aware of the fact of "legacy" proofs. This is why I created a `ProofBuild` trait, see edited opening post. Happy to hear your thoughts on it!

Regarding switch commitments: I agree in general switch commitments should be used. However in some very specific scenario's wallet implementators might not want to use them. For this reason and having the possibility to move to a different scheme in the future I added a `SwitchCommitmentType` enum. The proofs decode and encode them, but for now on the wallet level it is assumed that they are always of the `SwitchCommitmentType::Regular` type.

yeastplume:
Still a ways to go, but I've had a look over it and absolutely happy with the approach, I likely would have done something similar. Totally in agreement with not loading up the keychain with too much more functionality at this stage (given we've been talking a bit about where it should ultimately go)

core/src/libtx/proof.rs
```diff
@@ -55,35 +70,652 @@ pub fn verify(
...
+	}
+}
+
+/// The legacy proof builder, used before the first hard fork
+pub struct LegacyProofBuilder<'a, K>
+where
+	K: Keychain,
+{
+	keychain: &'a K,
+	root_hash: Vec<u8>,
```
DavidBurkett: Stealing this is as effective as stealing the wallet seed. Make sure you erase the memory when you're done with it. "Drop" in rust, I believe.

PR #2851, opened 2019-05-27
Fix keychain to work with non-Copy secp::SecretKey

eupn:
This PR allows keychain to work with `secp::SecretKey` which made non-`Copy`able in https://github.com/mimblewimble/rust-secp256k1-zkp/pull/51

This fix should be merged as soon as https://github.com/mimblewimble/rust-secp256k1-zkp/pull/51 is.

*Update*: https://github.com/mimblewimble/rust-secp256k1-zkp/pull/51 is merged, so this PR can be merged as well.

### DIFF
--- a/keychain/src/extkey_bip32.rs
+++ b/keychain/src/extkey_bip32.rs
@@ -149,7 +149,7 @@ impl BIP32Hasher for BIP32GrinHasher {
 }
 
 /// Extended private key
-#[derive(Copy, Clone, PartialEq, Eq, Debug)]
+#[derive(Clone, PartialEq, Eq, Debug)]
 pub struct ExtendedPrivKey {
 	/// The network this key is to be used on
 	pub network: [u8; 4],
@@ -399,7 +399,7 @@ impl ExtendedPrivKey {
 	where
 		H: BIP32Hasher,
 	{
-		let mut sk: ExtendedPrivKey = *self;
+		let mut sk: ExtendedPrivKey = self.clone();
 		for cnum in cnums {
 			sk = sk.ckd_priv(secp, hasher, *cnum)?;
 		}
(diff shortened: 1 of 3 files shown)


antiochp:
Why `Clone` and not `Copy`? What's the benefit?

eupn:
@antiochp 
> Why `Clone` and not `Copy`? What's the benefit?

It is impossible to implement both `Copy` and `Drop`. Therefore, if one wants to use a custom implementation of `Drop` (which zeroes underlying data, for example), he should remove `Copy` derive. Since `Copy` trait is no longer implemented, the only option left is to use `.clone` to satisfy borrow-checker. Hope this makes sense.

hashmap:
https://doc.rust-lang.org/1.30.0/book/2018-edition/ch04-01-what-is-ownership.html

If a type has the Copy trait, an older variable is still usable after assignment. Rust won’t let us annotate a type with the Copy trait if the type, or any of its parts, has implemented the Drop trait. If the type needs something special to happen when the value goes out of scope and we add the Copy annotation to that type, we’ll get a compile time error.

garyyu:
Similar comment as in https://github.com/mimblewimble/grin-wallet/pull/121#issuecomment-497533632

garyyu:
Related:  https://github.com/mimblewimble/grin/pull/2847#issuecomment-499268857

yeastplume:
the PR in rust-secp256k1 may have been merged, but a new version hasn't been released. @jaspervdm I assume this will need to go in for 2.0.0 when we do a new release of rust-secp for all of the other 2.0.0 changes?

jaspervdm:
If it has been merged in rust-secp already then yes we need this for v2.0.0

yeastplume:
Fixed via #2848

PR #2854, opened 2019-05-27
Remove TODO about transaction eviction

eupn:
Just removes `TODO` that is misleading since TX eviction is actually already happening at the [top level](https://github.com/mimblewimble/grin/blob/4b88a9b5bd951f1a755c7d4281f0785e985d13c1/pool/src/transaction_pool.rs#L181).


quentinlesceller:
Indeed thanks!

PR #2855, opened 2019-05-27
[2.x.x] Use blocking IO in P2P to reduce CPU load

hashmap:
Instead we create 2 threads - one for reading, one for writing, both block for 1 second then check stop flag, if it is not set the process continue.
The current master consumes 20-25% CPU on my machine with ~ 20 peers, this version 9-12% with the same number of peers (with TUI enabled).

Also we can use all higher level reader/writer methods like read_string, read_exact etc


p2p/src/conn.rs
```diff
@@ -98,12 +103,7 @@ impl<'a> Message<'a> {
 		while written < len {
 			let read_len = cmp::min(8000, len - written);
 			let mut buf = vec![0u8; read_len];
-			read_exact(
-				&mut self.stream,
-				&mut buf[..],
-				time::Duration::from_secs(10),
-				true,
-			)?;
+			self.stream.read_exact(&mut buf[..])?;
```
garyyu: Is this now controlled by `IO_TIMEOUT` (i.e. 1 second) ?  if the buf can't be filled in 1 second, `read_exact` will return a TIMEOUT error. So, now the minimum speed is 8kByte per second, instead of old 800 Byte/s. 

Is my understanding correct?

p2p/src/msg.rs
```diff
@@ -160,14 +152,14 @@ pub fn read_item<T: Readable>(stream: &mut dyn Read) -> Result<(T, u64), Error>
 /// until we have a result (or timeout).
 pub fn read_body<T: Readable>(h: &MsgHeader, stream: &mut dyn Read) -> Result<T, Error> {
 	let mut body = vec![0u8; h.msg_len as usize];
-	read_exact(stream, &mut body, time::Duration::from_secs(20), true)?;
+	stream.read_exact(&mut body)?;
```
garyyu: Similar question here, what's the biggest possible length for a `body`?  perhaps a Block is the biggest? then it's maximum possible value 1.5MB for example.

The timeout is 1 second and minimum speed is 1.5MB/s ?

garyyu:
I like the performance improvement 👍

p2p/src/conn.rs
hashmap: Not exactly, `read` (syscall) timeout occurs when we we can't read a single byte in 1 second. In our `read_exact` we didn't have a timeout for a single read, but for the entire operation (filling a buffer). In Rust `read_exact` we don't have any high-level timeouts but can control the timeout for a single read. 
So it's quite opposite, the previous impl was more demanding for connection speed, a new one will work for any connection faster than 1 byte/sec. A drawback that a peer's thread can be blocked by `read_exact` for a while on a slow connection, that what we can experience during shutdown.

p2p/src/msg.rs
hashmap: See my previous comment, the minimum speed is still 1 byte/sec

p2p/src/conn.rs
garyyu: > read (syscall) timeout occurs when we we can't read a single byte in 1 second.

In the case of a peer's sending thread doesn't get the CPU resources for 1 second, does this `read_exact` will return a timeout error with part of buff filled?

garyyu: @hashmap   remind above question in case you missed it :-)

hashmap: Ah, thanks! Well, this is timeout for a single read operation which is basically syscall, I'm not an expert in preemptive scheduling but I think it's safe to assume that this situation is very unlikely.

antiochp:
I updated this PR to merge onto `milestone/2.x.x`.
There are a few conflicts around the protocol versioning changes but nothing too serious.

hashmap:
Synced from the scratch, no issues

hashmap:
CPU consumption without TUI
this PR (1.7%):
Screenshot 2019-07-12 at 21 10 12

master (22.9%):
Screenshot 2019-07-12 at 21 13 07
it seems to consume a bit more memory, but the difference is not that big, so it sounds like a good tradeoff

PR #2856, opened 2019-05-28
[2.x.x] Writeable protocol version aware

antiochp:
This is the other "half" of #2824.
We now pass a "protocol version" into serialization/deserialization based on the version provided by our peer, or our local db storage.
This allows us to read/write different protocol versions in a robust way - making version negotiation across the p2p layer possible along with local db upgrades and backward compatibility.

This is not actively in use yet - everything remains on `protocol version 1` but this will allow nodes to support (and upgrade safely) to the proposed kernel serialization in #2859.


antiochp:
Going to merge this. See #2824 for original context and discussion.

PR #2859, opened 2019-05-28
fee and lock_height maintained in kernel features variants

antiochp:
This PR introduces better type safety around the various kernel variants. 
The existing binary serialization/deserialization format is preserved.

Kernel feature variants - 
* Plain
    * fee only
* Coinbase
    * no fee 
    * no lock_height
* HeightLocked
    * fee
    * lock_height

This is an incremental pre-requisite for supporting "variable size" kernels.

This PR does not affect binary serialization/deserialization of the different kernel variants.
Each variant will still serialize both a fee and a lock_height, defaulting to 0, regardless of variant.

Existing API json serialization is also preserved via mapping logic in `TxKernelPrintable`.

The actual serialization changes will be tackled in a subsequent PR.

There is a corresponding PR for grin-wallet https://github.com/mimblewimble/grin-wallet/pull/129 that makes sure "VersionedSlate" json serialization is not affected by these changes. Changes to grin-wallet in that PR are mainly for compatibility with the code changes here.

----

Each kernel feature will continue to include - 
* 1 byte for the features identifier
* 8 bytes for the fee (0 is not applicable)
* 8 bytes for the lock_height (0 if not applicable)
* 32 bytes for the excess
* 64 bytes for the signature

For example a Plain kernel will be as follows (note 0 for lock_height) - 
```
00 | 00 00 00 00 01 f7 8a 40 | 00 00 00 00 00 00 00 00 | 08 b1 ... 22 d8 | 33 11 ... b9 69
```

----

I'm going to open an RFC for the actual protocol version support for multiple serialization strategies (removing the unnecessary 8 bytes where fee not applicable etc.) This will affect the network and will need some careful consideration around the release process.

### DIFF
--- a/core/src/core/transaction.rs
+++ b/core/src/core/transaction.rs
@@ -35,34 +35,102 @@ use std::cmp::{max, min};
 use std::sync::Arc;
 use std::{error, fmt};
 
-// Enum of various supported kernel "features".
-enum_from_primitive! {
-	/// Various flavors of tx kernel.
-	#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
-	#[repr(u8)]
-	pub enum KernelFeatures {
-		/// Plain kernel (the default for Grin txs).
-		Plain = 0,
-		/// A coinbase kernel.
-		Coinbase = 1,
-		/// A kernel with an expicit lock height.
-		HeightLocked = 2,
+/// Various flavors of tx kernel.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+pub enum KernelFeatures {
+	/// Plain kernel (the default for Grin txs).
+	Plain { fee: u64 },
+	/// A coinbase kernel.
+	Coinbase,
+	/// A kernel with an expicit lock height.
+	HeightLocked { fee: u64, lock_height: u64 },
+}
+
+impl KernelFeatures {
+	const PLAIN_U8: u8 = 0;
+	const COINBASE_U8: u8 = 1;
+	const HEIGHT_LOCKED_U8: u8 = 2;
+
+	pub fn as_u8(&self) -> u8 {
+		match self {
+			KernelFeatures::Plain { .. } => KernelFeatures::PLAIN_U8,
+			KernelFeatures::Coinbase => KernelFeatures::COINBASE_U8,
+			KernelFeatures::HeightLocked { .. } => KernelFeatures::HEIGHT_LOCKED_U8,
+		}
+	}
+
+	/// We cannot simply hash the features as v1 includes "optional" fields.
+	///
+	/// i.e. A coinbase kernel has a fee of 0 in v1 (but is omitted when hashing the msg).
+	///
+	/// msg = hash(features)                       for coinbase kernels
+	///       hash(features || fee)                for plain kernels
+	///       hash(features || fee || lock_height) for height locked kernels
+	pub fn kernel_sig_msg(&self) -> Result<secp::Message, Error> {
+		let x = self.as_u8();
+		let hash = match self {
+			KernelFeatures::Plain { fee } => (x, fee).hash(),
+			KernelFeatures::Coinbase => (x).hash(),
+			KernelFeatures::HeightLocked { fee, lock_height } => (x, fee, lock_height).hash(),
+		};
+
+		let msg = secp::Message::from_slice(&hash.as_bytes())?;
+		Ok(msg)
 	}
 }
 
-impl DefaultHashable for KernelFeatures {}
-
 impl Writeable for KernelFeatures {
 	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
-		writer.write_u8(*self as u8)?;
+		// Writing kernel features is a little complex right now due to various
+		// optional fields that are not optional in v1 serialization.
+		// TODO - Improve how we serialize kernel features.
+		// TODO - Consider a v1 features struct that does the right thing here.
+
+		// This is effectively v1 serialization (always include both fee and lock_height).
+		match self {
+			KernelFeatures::Plain { fee } => {
+				writer.write_u8(self.as_u8())?;
+				writer.write_u64(*fee)?;
+				writer.write_u64(0)?;
+			}
+			KernelFeatures::Coinbase => {
+				writer.write_u8(self.as_u8())?;
+				writer.write_u64(0)?;
+				writer.write_u64(0)?;
+			}
+			KernelFeatures::HeightLocked { fee, lock_height } => {
+				writer.write_u8(self.as_u8())?;
+				writer.write_u64(*fee)?;
+				writer.write_u64(*lock_height)?;
+			}
+		}
+
 		Ok(())
 	}
 }
 
 impl Readable for KernelFeatures {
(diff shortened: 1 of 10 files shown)


antiochp:
@tromp - You are likely interested in this. Moved `fee` and `lock_height` into the relevant kernel features enum variants. Reworked a bunch of stuff to work with this revised approach.

antiochp:
> with_fee applies to KernelFeatures::HeightLocked as well; currently panics...

We actually only do these in a particular order, we add a fee, then we add a lock_height.
I was torn here - do we complicate the code to support more flexibility or do we just do what needs to be done to get the current approach to work?

Ideally we'd rewrite more of this and rethink these builders a bit - it makes less sense to build a tx in stages like this when two stage both touch the underlying features (and potentially in incompatible ways).

antiochp:
> You replaced the fee field by a fee() method, but why not do the same for lock_height() ?

Felt like `fee` was a common enough thing to want to expose as a fn like this. We always sum _all_ the fees up for all kernels when calculating the fee on a tx.

Whereas lock_height is a more specific concept - it makes no sense for coinbase kernels for example. So we should expect the client to do the right thing and explicitly look for kernels that have lock heights.
I suspect (may be wrong though) that "relative lock heights" are going to behave differently enough that a `lock_height()` fn is not going to work there, at which point we start adding even more fns exposing feature specific stuff.
Fees are fees though - regardless of any future feature variants (at east that feels like a reasonable assumption).

tromp:
> we add a fee, then we add a lock_height.

Not just add. Currently you can start with plain, then call with_fee any number of times to repeatedly change fee, then call with_lock_height any number of times to first add lock_height and later repeatedly change lock_height.
Given this freedom of changes, it might be semantically simpler to allow changing the fee on a height locked tx as well?!

tromp:
> lock_height is a more specific concept - it makes no sense for coinbase kernels for example. 

Fees make no sense for coinbase kernels either; no more than lock height.

> Fees are fees though - regardless of any future feature variants

The same can be argued for (absolute) lock heights.
Of course relative_lock_height would be a different thing.

antiochp:
> Fees make no sense for coinbase kernels either; no more than lock height.

Yeah agreed - I'm going to look at making the summing feature aware and just summing those that actually have fees, to be explicit about it (and for consistency with lock_height).

antiochp:
@yeastplume I'm looking at the slate v2 stuff in grin-wallet and it _looks_ like we don't actually serialize a tx anywhere without using wallet specific representations.

For example - 
https://github.com/mimblewimble/grin-wallet/blob/b1738e33ed05d8237d0251d18070fc0d4a0101b8/libwallet/src/slate_versions/v2.rs#L167-L168

Do we ever serialize/deserialize the core tx or kernel anywhere in the wallet code?
Is the final api post/broadcast to the node the only instance where we do this?

yeastplume:
Correct, the wallet only uses the transaction struct within slates, it doesn’t de/serialize anywhere

antiochp:
> Shouldn't with_fee return an optional error instead of panicing?

We're kind of limited with this because of the way the builders are implemented.
We do some _interesting_ things with these as part of the tx building, chaining multiple tx builders together. And there is no easy way of handling error inside this.

```
	Box::new(
		move |_build, (tx, kern, sum)| -> (Transaction, TxKernel, BlindSum) {
			(tx, kern.with_fee(fee), sum)
		},
	)
```

The `panic!` was the easiest way of working around this. We make a chained `with_fee` simply invalid in certain situations, i.e. when we attempt to call it on a coinbase kernel, which is a logic error and should never happen.

antiochp:
> What stops us from making Readable and Writable impls of KernelFeatures that strip out unnecessary fee and lockheight?

This is pretty much the entire purpose of the proposed RFC here - https://github.com/mimblewimble/grin-rfcs/pull/21

We need to support backward compatibility for nodes on the network that expect those redundant bytes to be present.

antiochp:
> kernel_sig_msg could be refactored with
let tuple = match self { ... };
let msg = secp::Message::from_slice(&tuple.hash().as_bytes())?;
to avoid applying hash() in every match branch.

This unfortunately won't work because the tuples are not all the same arity and rust treats them as different types.

core/src/core/transaction.rs
```diff
@@ -272,20 +352,16 @@ impl TxKernel {
...
 	}
 
 	/// Verify the transaction proof validity. Entails handling the commitment
 	/// as a public key and checking the signature verifies with the fee as
 	/// message.
 	pub fn verify(&self) -> Result<(), Error> {
-		if self.is_coinbase() && self.fee != 0 || !self.is_height_locked() && self.lock_height != 0
-		{
-			return Err(Error::InvalidKernelFeatures);
-		}
```
garyyu: Does this mean now the TxKernel can have a `fee`>0 for coinbase? and `lock_height`=0 with KernelFeatures=HeightLocked?

because you removed this checking on `verify()`.

antiochp: No. A coinbase kernel has no fee at all, there is no field for it any more.
And anything other than a height locked kernel cannot have a lock height value.

garyyu: I mean if now we broadcast a new block with a coinbase kernel like the following (took part of data from your example):
```sh
01 | 00 00 00 00 01 f7 8a 40 | 00 00 00 00 00 00 00 00 | 08 b1 ... 22 d8 | 33 11 ... b9 69
```

Will it split the Grin network into 2 parts? the 1st part with all old version nodes which reject this block because it has a `fee>0` for coinbase kernel, and the 2nd part with all new version nodes which accept this block?

core/src/core/transaction.rs
```diff
@@ -35,34 +35,115 @@ use std::cmp::{max, min};
...
-		let features =
-			KernelFeatures::from_u8(reader.read_u8()?).ok_or(ser::Error::CorruptedData)?;
+		let features = match reader.read_u8()? {
+			KernelFeatures::PLAIN_U8 => {
+				let fee = reader.read_u64()?;
+				let _lock_height = reader.read_u64()?;
+				KernelFeatures::Plain { fee }
+			}
+			KernelFeatures::COINBASE_U8 => {
+				let _fee = reader.read_u64()?;
```
garyyu: Indeed, the new code here will just ignore the `fee` for coinbase kernel. (pls look https://github.com/mimblewimble/grin/pull/2859/files#r318871293)

tromp:
Good catch, Gary.
As long as there's coinbase fee data serialized anywhere, we cannot ignore it but must keep verifying that it's zero.

antiochp:
Let me see if I understand what you're saying - 

On the reading side -
* old nodes read the fee bytes on coinbase kernels and verify the fee is 0
* new nodes ignore these bytes entirely (they are read but discarded)

On the writing side - 
* old nodes write the fee out
* new nodes write 0 out in place of the fee

So if a misbehaving old node (or a malicious node) were to write out a non-zero fee on a coinbase kernel - 
* old nodes would reject this as invalid
* new nodes would simply ignore the fee _and treat the kernel as valid_

I hadn't thought through that scenario as neither old nor new nodes would do this under normal scenarios. It would have to be a malformed (and hence invalid) kernel.

Is this a correct understanding of this?

Good catch. Thanks @garyyu.
We should add validation when deserializing the new kernel features - i.e. not just ignore the fee bytes but also check they are all 0s.

(And similar for the lock_height).

PR #2862, opened 2019-05-29
Speedup shutdown

hashmap:
I made an suboptimal (aka stupid) decision to stop and wait for peers
one by one which makes shutdown very slow - O(n). This PR decouples sending
stop signal from waiting a thread to exit. On top of it in Peers we
first send stop signal to all peers and only after that start waiting
for them to exit. It gives us a constant time of shutdown in most of the
cases.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -181,18 +181,17 @@ impl StopHandle {
 		}
 	}
 
-	pub fn stop_and_wait(&mut self) {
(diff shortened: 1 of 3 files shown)


garyyu:
Thanks for finding this 👍 

It happened several times when I kill the process, but I didn't realize it's a new "bug" :-)  I thought sth was wrong in my system and just use `kill -9`.

PR #2866, opened 2019-05-30
PoW HardFork

tromp:
Grin's first mainnet hardfork!

From well before launch, Grin planned to hardfork every 6 month during the first 2 years.
This is pretty much a necessity for any chain with an ASIC-resistant PoW, as an immutable PoW
falls to ASICs sooner or later. Grin's GPU oriented PoW, Cuckaroo29, is no exception.

The PoW tweak was introduced in the Grin Forum post at https://www.grin-forum.org/t/mid-july-pow-hardfork-cuckaroo29-cuckarood29.

The crucial change in this PR is in function global::create_pow_context:

        match chain_type {
                // Mainnet has Cuckaroo(d)29 for AR and Cuckatoo30+ for AF
    +           ChainTypes::Mainnet if edge_bits == 29 && valid_header_version(height, HeaderVersion::new(2))
    +                                            => new_cuckarood_ctx(edge_bits, proof_size),
                ChainTypes::Mainnet if edge_bits == 29 => new_cuckaroo_ctx(edge_bits, proof_size),
                ChainTypes::Mainnet => new_cuckatoo_ctx(edge_bits, proof_size, max_sols),

                // Same for Floonet
    +             ChainTypes::Floonet if edge_bits == 29 && valid_header_version(height, HeaderVersion::new(2))
    +                                            => new_cuckarood_ctx(edge_bits, proof_size),
                ChainTypes::Floonet if edge_bits == 29 => new_cuckaroo_ctx(edge_bits, proof_size),
                ChainTypes::Floonet => new_cuckatoo_ctx(edge_bits, proof_size, max_sols),

in combination with a change in consensus::valid_header_version
that requires version 2 at heights [HARD_FORK_INTERVAL, 2 * HARD_FORK_INTERVAL) on mainnet and at heights [FLOONET_FIRST_HARD_FORK,  2 * HARD_FORK_INTERVAL) on flooonet,

with all other changes just putting up the necessary scaffolding to make this work.

If all goes well, then the first block with header version 2, at height WEEK_HEIGHT * 52/2 =
60*24*7*26 = 262080, will roll around on Jul 17, and will have to be solved by either good old Cuckatoo31+, or by the new Cuckarood29. We will be able to test this a few weeks prior on Floonet.

Any nodes that fail to upgrade will not be able to advance to height 262080, as the current consensus::valid_header_version will simply return false. So there cannot actually be any chain split.
One side will advance, and the other will remain stuck until they upgrade as well.

The major work left to be done is in grin-miner. We may initially only offer a CUDA miner there as the CPU mean miner is an exceedingly complex beast.


core/src/pow/cuckarood.rs
```diff
@@ -0,0 +1,191 @@
...
+			}
+			i = j ^ 1;
+			n += 1;
+			if i == 0 {
+				break;
+			}
+		}
+		if n == self.params.proof_size {
+			Ok(())
+		} else {
```
quentinlesceller: Unnecessary else?

tromp: necessary, unless you add a return before Ok....

quentinlesceller: Indeed.

core/src/global.rs
```diff
@@ -155,10 +156,14 @@ where
 	let chain_type = CHAIN_TYPE.read().clone();
 	match chain_type {
 		// Mainnet has Cuckaroo29 for AR and Cuckatoo30+ for AF
```
quentinlesceller: Could you update this comment by any chance?

core/src/global.rs
```diff
@@ -155,10 +156,14 @@ where
...
+		ChainTypes::Mainnet if edge_bits == 29 && valid_header_version(_height, HeaderVersion::new(2))
+						=> new_cuckarood_ctx(edge_bits, proof_size),
```
quentinlesceller: I think you can rename `_height` to `height` since we use it now. Also Cuckatoo31+ instead of Cuckatoo30+?

tromp: thanks for careful reading. fixed now.

quentinlesceller:
@tromp Can you fix the merge conflict?

tromp:
Fixed conflict.

tromp:
NOTE that while the PoW hardfork happens 32 days earlier on Floonet, at height 216000, it will still have version 1 headers. To make Floonet have version 2 headers at 216000 would require making consensus::valid_header_version dependent on CHAIN_TYPE...

core/src/consensus.rs
```diff
@@ -127,22 +127,42 @@ pub const MAX_BLOCK_WEIGHT: usize = 40_000;
 /// Fork every 6 months.
 pub const HARD_FORK_INTERVAL: u64 = YEAR_HEIGHT / 2;
 
+/// Floonet first hard fork height, set to happen around 2019-06-20
+pub const FLOONET_FIRST_HARD_FORK: u64 = 184_320;
```
quentinlesceller: For the record this will happen on 2019-06-19 22:56:46. Can we move it a bit around June 20?

tromp: Are you assuming an 60s block time average? 
Not sure we have the graph rate consistency for that.
I often see average block times well over 60s.
I'd say this is more likely to hit on June 20 than on June 19...

quentinlesceller: Yeah, assuming a constant 60s per block. Starting at block height 166_469 on 2019-06-07 13:25:46 UTC.

tromp:
ok; moved it forward by a half day

yeastplume:
It would seem these tests are hanging in the current commit:

```
test hard_forks ... ok
test test_secondary_pow_scale ... ok
test test_secondary_pow_ratio ... test test_secondary_pow_ratio has been running for over 60 seconds
test adjustment_scenarios ... test adjustment_scenarios has been running for over 60 seconds
test next_target_adjustment ... test next_target_adjustment has been running for over 60 seconds
```

yeastplume:
Odd... I ran CI again and it passed this time, but on a second local run I get a different set of tests hanging. Something definitely inconsistent here:

```
test test_secondary_pow_scale ... ok
test test_secondary_pow_ratio ... test test_secondary_pow_ratio has been running for over 60 seconds
test hard_forks ... test hard_forks has been running for over 60 seconds
test adjustment_scenarios ... test adjustment_scenarios has been running for over 60 seconds
test next_target_adjustment ... test next_target_adjustment has been running for over 60 seconds
```

Issue #2865, opened 2019-05-30
Downloading Blocks 110%


garyyu:
(On **Floonet**) Now, my local Grin node is showing "Longest chain @141012 vs Us @155467",  "Downloading Blocks 110%, ..."

Happened several times in these days, and trying to understand the root cause. With the latest master branch source.

<img width="576" alt="Screen Shot 2019-05-30 at 4 21 05 PM" src="https://user-images.githubusercontent.com/1852227/58619538-ac413b80-82f7-11e9-951c-c91089ab3e4f.png">

<img width="1248" alt="Screen Shot 2019-05-30 at 4 20 56 PM" src="https://user-images.githubusercontent.com/1852227/58619573-b95e2a80-82f7-11e9-8f97-795369f7ead9.png">

garyyu:
Check the peers, and find this `Longest Chain` peer:
```
{
"capabilities":{"bits":15},
"user_agent":"MW/Grin 1.0.3",
"version":1,
"addr":"54.148.113.184:13414",
"direction":"Inbound",
"total_difficulty":427388621917598,
"height":141012
}
```

It's quite strange why our banning mechanism doesn't work.  will look into the logs.

antiochp:
Something weird going on with your peers there.

Looking here - 
https://grinscan.net/block/155466

total_diff: `470,779,341,967,660`

But your peer list is showing significantly lower total_diff values.

And here - https://grinscan.net/block/141012

total_diff: `427,388,621,917,598`

This looks correct.

According to your local peer data that one _is_ your most work peer.

antiochp:
https://grinscan.net/block/141012

vs.

https://floonet.grinscan.net/block/155466

Somehow we have a `mainnet` node peering with your `floonet` node.

garyyu:
Regarding why this peer hasn't been banned, I think the root cause is here:
```
		// Ban this fraud peer which claims a higher work but can't send us the real headers
		if now > *stalling_ts + Duration::seconds(120)
			&& highest_height == peer.info.height()
		{
			self.peers
				.ban_peer(peer.info.addr, ReasonForBan::FraudHeight);
			...
```

`highest_height == peer.info.height()` <-- this only handle the case of a peer with higher height than us, but in this issue, the peer `54.148.113.184` has a lower height but with much more work.

garyyu:
And regarding the wrong peer (`mainnet` node peering with `floonet` node) , here is the log:
```
[...]

PR #2867, opened 2019-05-30
fix: fraud peer ban doesn't work in some cases

garyyu:
fix https://github.com/mimblewimble/grin/issues/2865

### DIFF
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -143,13 +143,10 @@ impl HeaderSync {
 				if let Some(ref stalling_ts) = self.stalling_ts {
 					if let Some(ref peer) = self.syncing_peer {
 						match self.sync_state.status() {
-							SyncStatus::HeaderSync {
-								current_height: _,
-								highest_height,
-							} => {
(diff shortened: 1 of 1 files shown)


servers/src/grin/sync/header_sync.rs
```diff
@@ -143,13 +143,10 @@ impl HeaderSync {
 				if let Some(ref stalling_ts) = self.stalling_ts {
 					if let Some(ref peer) = self.syncing_peer {
 						match self.sync_state.status() {
-							SyncStatus::HeaderSync {
-								current_height: _,
-								highest_height,
-							} => {
+							SyncStatus::HeaderSync { .. } | SyncStatus::BodySync { .. } => {
```
antiochp: Do we even need to check for HeaderSync or BodySync here now?
If we got this far then we know we are syncing headers?

garyyu: In case a HeaderSync state switched to StateSync and spent several minutes on txhashset downloading, and after that, switching back to HeaderSync.
So, perhaps not bad to check the state here.

antiochp: :+1:

PR #2874, opened 2019-06-05
Make Valid Header Version dependant on ChainType

quentinlesceller:
Linked to https://github.com/mimblewimble/grin/pull/2866

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -130,19 +130,56 @@ pub const HARD_FORK_INTERVAL: u64 = YEAR_HEIGHT / 2;
 /// Check whether the block version is valid at a given height, implements
 /// 6 months interval scheduled hard forks for the first 2 years.
 pub fn valid_header_version(height: u64, version: HeaderVersion) -> bool {
-	// uncomment below as we go from hard fork to hard fork
-	if height < HARD_FORK_INTERVAL {
-		version == HeaderVersion::default()
-	/* } else if height < 2 * HARD_FORK_INTERVAL {
-		version == 2
-	} else if height < 3 * HARD_FORK_INTERVAL {
-		version == 3
-	} else if height < 4 * HARD_FORK_INTERVAL {
-		version == 4
-	} else if height >= 5 * HARD_FORK_INTERVAL {
-		version > 4 */
-	} else {
-		false
+	let chain_type = global::CHAIN_TYPE.read().clone();
+	match chain_type {
+		global::ChainTypes::Mainnet => {
+			// uncomment below as we go from hard fork to hard fork
+			if height < HARD_FORK_INTERVAL {
+			version == HeaderVersion::default()
+			/* } else if height < 2 * HARD_FORK_INTERVAL {
+				version == 2
+			} else if height < 3 * HARD_FORK_INTERVAL {
+				version == 3
+			} else if height < 4 * HARD_FORK_INTERVAL {
+				version == 4
+			} else if height >= 5 * HARD_FORK_INTERVAL {
+				version > 4 */
+			} else {
+				false
+			}
+		},
+		global::ChainTypes::Floonet => {
+			// uncomment below as we go from hard fork to hard fork
+			if height < HARD_FORK_INTERVAL-28*DAY_HEIGHT {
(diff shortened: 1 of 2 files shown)


core/src/consensus.rs
```diff
@@ -130,19 +130,57 @@ pub const HARD_FORK_INTERVAL: u64 = YEAR_HEIGHT / 2;
...
+				version == HeaderVersion::new(4)
+			} else if height >= 5 * HARD_FORK_INTERVAL {
+				version > HeaderVersion::new(4) */
+			} else {
+				false
+			}
+		}
+		global::ChainTypes::Floonet => {
+			// uncomment below as we go from hard fork to hard fork
+			if height < HARD_FORK_INTERVAL - 28 * DAY_HEIGHT {
```
jaspervdm: Maybe its better to have a hardcoded height here and set the next one to that height plus `HARD_FORK_INTERVAL`?

core/src/consensus.rs
```diff
@@ -130,19 +130,57 @@ pub const HARD_FORK_INTERVAL: u64 = YEAR_HEIGHT / 2;
...
-		version == 3
-	} else if height < 4 * HARD_FORK_INTERVAL {
-		version == 4
-	} else if height >= 5 * HARD_FORK_INTERVAL {
-		version > 4 */
-	} else {
-		false
+	let chain_type = global::CHAIN_TYPE.read().clone();
+	match chain_type {
+		global::ChainTypes::Mainnet => {
```
jaspervdm: This is duplicate code right? It is the same as the 3rd option of the `match`. Why not remove this one and just have floonet and "everything else"

tromp: agreed with Jasper. if everything non-floonet is to behave the same then that should be reflected in a single branch of the code...

tromp:
This PR should be combined with #2866 which no longer needs to make a height adjustment for floonet in create_pow_context ...

core/src/consensus.rs
cadmuspeverell: Why would we want to use `HARD_FORK_INTERVAL` at all when the exact heights are static and known?  The fewer calculations performed in consensus code the safer and predictable it all feels.

jaspervdm: The exact height for the second hard fork on Floonet is not yet known. It is unlikely that blocks will keep coming in at a 1 minute interval for 6 months straight, so i see it as setting an upperbound now, then when we get close to the next hardfork we can adjust it to an explicit height.

core/src/consensus.rs
antiochp: :+1: On keeping `dev` and `mainnet` as similar as possible.

core/src/consensus.rs
quentinlesceller: 👍 @jaspervdm on setting an exact block height. Floonet is currently at bock 165,019 which means choosing `HARD_FORK_INTERVAL - 28 * DAY_HEIGHT` does not make any sense for a fork around June 20.

quentinlesceller: Added `FLOONET_FIRST_HARD_FORK` at height 185,200.

quentinlesceller:
Agree with you @tromp. How you'd like to proceed? I can close this PR and you can maybe put everything in #2866 or merge this (once approved) in 2.0 branch and fix conflicts on #2866. Let me know, either way is fine for me

antiochp:
In terms of this vs. #2866 I would lean toward merging these as separate PRs.

tromp:
> In terms of this vs. #2866 I would lean toward merging these as separate PRs.

Actually, I was planning to write something like #2866 myself if ppl agreed to add a CHAIN dependency to valid_header_version. These PR seemingly cannot be separate as #2866 necessitates a change in my create_pow_context. I will proceed to merge this into my PR. It will only have the distinction of a separate commit.

jaspervdm:
@tromp If we merge this PR in before yours, you could just pull in the changes and use them.

tromp:
I incorporated this PR into my https://github.com/mimblewimble/grin/pull/2866 where I felt it belonged.

quentinlesceller:
Closing for #2866.

PR #2875, opened 2019-06-06
commit cargo.lock

garyyu:
- commit cargo.lock
- add a log to check the possible difference between **requested** `TxHashSet` and **response** `TxHashSet`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -655,7 +655,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "grin"
-version = "1.1.0"
+version = "1.1.1-beta.1"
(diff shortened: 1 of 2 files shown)


antiochp:
I'm not 100% sure this makes sense. We do not have a `1.1.1` release and ideally never will.

Ideally `master` at this point is identical to `1.1.0` and nothing is merged in here unless the decision is made to fix a critical issue.

(We should probably have had at least one other person reviewing this before merging to `master`).

hashmap:
@antiochp agree with the last point. Cargo.lock update was a leftover from https://github.com/mimblewimble/grin/commit/dfb4d5afae35a3d0e7507556011a9645e0de2521 though

PR #2876, opened 2019-06-06
Always stem local txs if configured that way (unless explicitly fluffed)

antiochp:
We identify "our" txs as those with a tx src of `push-api`.

* If wallet explicitly pushes with `fluff` then we fluff.
* If node is configured as `always_stem_our_txs=true` then we will stem our txs, regardless of current epoch.
* Otherwise we stem/fluff based on current epoch.
* Stem txs (arriving via stem relay) will always stem/fluff based on current epoch.

Also cleaned up the dandelion_config to get rid of the unnecessary `Option<>` wrappers.

TODO - 

- [x] Consider making `TxSource` an enum as part of this for better type safety
    (right now we are identifying our txs based on something called the `debug_name` ...)
- [x] rename "local txs" to "our txs" (local does not make much sense in wider context)

----

Default generated config as follows - 

```
#########################################
### DANDELION CONFIGURATION           ###
#########################################
[server.dandelion_config]

#dandelion epoch duration
epoch_secs = 600

#fluff and broadcast after embargo expires if tx not seen on network
embargo_secs = 180

#dandelion aggregation period in secs
aggregation_secs = 30

#dandelion stem probability (stem 90% of the time, fluff 10% of the time)
stem_probability = 90

#always stem our (pushed via api) txs regardless of stem/fluff epoch (as per Dandelion++ paper)
always_stem_our_txs = true
```

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -211,6 +211,14 @@ fn comments() -> HashMap<String, String> {
 	);
 
 	retval.insert(
+		"always_stem_our_txs".to_string(),
+		"
+#always stem our (pushed via api) txs regardless of stem/fluff epoch (as per Dandelion++ paper)
+"
+		.to_string(),
+	);
+
+	retval.insert(
 		"[server.p2p_config]".to_string(),
 		"#test miner wallet URL (burns if this doesn't exist)
 #test_miner_wallet_url = \"http://127.0.0.1:3415\"
--- a/api/src/handlers/pool_api.rs
+++ b/api/src/handlers/pool_api.rs
@@ -82,10 +82,7 @@ impl PoolPushHandler {
 						.map_err(|e| ErrorKind::RequestError(format!("Bad request: {}", e)).into())
 				})
 				.and_then(move |tx: Transaction| {
-					let source = pool::TxSource {
-						debug_name: "push-api".to_string(),
-						identifier: "?.?.?.?".to_string(),
-					};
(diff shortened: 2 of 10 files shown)


antiochp:
cc @lehnberg 

I took another look at what would be required to do this, to follow the spec as described in the Dandelion++ paper.
It was actually pretty striaghtforward to implement this.

We _do_ want to be able to override this "always stem local" behavior to make Dandelion testing easier. Its really hard to test Dandelion behavior when "local" txs are handled differently.

lehnberg:
Thanks @antiochp this is great! I saw your question in the docs ticket but didn't have a chance to reply yet. Yes - we want to always stem our own transactions - otherwise txos in a fluff can be attributed to the fluffing node to a predictable degree.

config/src/comments.rs
```diff
@@ -210,6 +210,14 @@ fn comments() -> HashMap<String, String> {
 		.to_string(),
 	);
 
+	retval.insert(
+		"always_stem_local".to_string(),
+		"
+#always stem local txs regardless of stem/fluff epoch (as per Dandelion++ paper)
```
kargakis: "local txs" sounds a bit weird, is this term used in the dandelion paper? Maybe "always_stem_ours" as a config option and "#always stem our own txs..." as a comment?

antiochp: Yeah good idea - will update.

PR #2878, opened 2019-06-07
[2.x.x] Take age of txs into consideration when bucketing and sorting txs in txpool

antiochp:
Turns out we were not taking tx "age" into consideration when bucketing and sorting txs in the txpool.
Which means that at block capacity it is basically a lottery whether your tx will get included in the next block...

This PR fixes this.

We maintain txs in the pool in insertion order so we do not need to track "age" explicitly. 
Bucket creation follows tx insertion order so we just need to take bucket creation order into consideration when sorting the buckets.

----
Existing behavior - 
* aggregate txs in pool if bucket fee_to_weight is not reduced
* sort buckets by fee_to_weight descending
* aggregate buckets up to max tx size

Revised behavior -
* aggregate txs in pool if bucket fee_to_weight is not reduced
* sort buckets by fee_to_weight descending _and_ age (oldest first)
* aggregate buckets up to max tx size

----

Our txpool bucketing and sorting logic still picks txs that maximize the overall block fee. But now also prioritizes txs that have been in the txpool the longest.

### DIFF
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -355,7 +355,7 @@ impl Pool {
 					// This is the common case for non 0-conf txs in the txpool.
 					// We assume the tx is valid here as we validated it on the way into the txpool.
 					insert_pos = Some(tx_buckets.len());
-					tx_buckets.push(Bucket::new(entry.tx.clone()));
+					tx_buckets.push(Bucket::new(entry.tx.clone(), tx_buckets.len()));
 				}
 				Some(pos) => {
 					// We found a single parent tx, so aggregate in the bucket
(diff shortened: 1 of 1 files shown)


antiochp:
> One thing I do not understand is why an old but low fee transaction should be prioritised over a new high fee transaction.

It won't be. We will always prioritize by `fee_by_weight` first. 
But if we have two txs with the _same_ `fee_by_weight` then we should prioritize the oldest one first (and we don't currently do this).

quentinlesceller:
Okay in that case it makes perfect sense to add this. LGTM 👍

PR #2892, opened 2019-06-12
Modify mine_block to use wallet V2 API

yeastplume:
Changes the `create_coinbase` call to use the V2 wallet api's `build_coinbase` function

### DIFF
--- a/servers/src/mining/mine_block.rs
+++ b/servers/src/mining/mine_block.rs
@@ -18,10 +18,10 @@
 use crate::util::RwLock;
 use chrono::prelude::{DateTime, NaiveDateTime, Utc};
 use rand::{thread_rng, Rng};
+use serde_json::{json, Value};
 use std::sync::Arc;
 use std::thread;
 use std::time::Duration;
-use serde_json::{json, Value};
 
 use crate::api;
 use crate::chain;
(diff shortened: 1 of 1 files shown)


servers/src/mining/mine_block.rs
```diff
@@ -265,15 +266,48 @@ fn get_coinbase(
...
+	let res: String = api::client::send_request(req).map_err(|e| {
+		let report = format!(
+			"Failed to get coinbase from {}. Is the wallet listening? {}",
+			dest, e
+		);
+		error!("{}", report);
+		Error::WalletComm(report)
+	})?;
+
+	let res: Value = serde_json::from_str(&res).unwrap();
```
hashmap: Wouldn't be cleaner to create a type for the response (I assume it's Result-like) instead of dealing with raw json Value?

servers/src/mining/mine_block.rs
```diff
@@ -265,15 +266,48 @@ fn get_coinbase(
...
+	if res["error"] != json!(null) {
+		let report = format!(
+			"Failed to get coinbase from {}: Error: {}, Message: {}",
+			dest, res["error"]["code"], res["error"]["message"]
+		);
+		error!("{}", report);
+		return Err(Error::WalletComm(report));
 	}
+
+	let cb_data = res["result"]["Ok"].clone();
```
hashmap: for example it's hard to tell should `Ok` be in lower case or not (or does it matter)

servers/src/mining/mine_block.rs
yeastplume: It just seemed simpler for this one-off in the code to do this way, also had the notion of making it obvious what the json structure is underneath for anyone looking at the code. If it were a full wallet API client, I'd say definitely, yes.

yeastplume:
@hashmap can you live with this as is? Happy to come back to refine it later, but it's tested in this form and works and we need this in place for floonet testing very shortly.

kargakis:
@yeastplume @hashmap can one of you open an issue to track the follow-up task? Thanks

PR #2898, opened 2019-06-14
Introduce HTTP Client and proxy support

hashmap:
This PR adds support of HTTP proxy in Grin API client, which is useful
for some network setups and for Tor support. So now in addition to API
secret which we pass to functions like get/post we need to pass at
least one additional parameter. Also we instantiate HTTP clinet
connection objects each time we call such functions. To make
configuration management simpler and enable reuse of HTTP client a
struct GrinClient is introduced.

Not sure why diff looks so massive, basically all functions were moved into the struct implementation and `api_secret` parameter was moved to constructor.


kargakis:
[?w=1](https://github.com/mimblewimble/grin/pull/2898/files?w=1) is your friend when reviewing this PR.

servers/src/mining/mine_block.rs
```diff
@@ -259,7 +259,7 @@ fn get_coinbase(
 /// Will retry based on default "retry forever with backoff" behavior.
 fn create_coinbase(dest: &str, block_fees: &BlockFees) -> Result<CbData, Error> {
 	let url = format!("{}/v1/wallet/foreign/build_coinbase", dest);
-	match api::client::post(&url, None, &block_fees) {
+	match api::client::GrinClient::default().post(&url, &block_fees) {
```
kargakis: I know that you are not changing anything here but how does this call work if the wallet uses auth?

hashmap: I have the same question actually

api/src/client.rs
```diff
@@ -29,203 +30,220 @@ use tokio::runtime::Runtime;
...
-	send_request(req)?;
-	Ok(())
-}
+impl GrinClient {
+	pub fn new(api_secret: Option<String>, proxy_addr: Option<String>) -> GrinClient {
+		let https = hyper_rustls::HttpsConnector::new(1);
+		let client = Client::builder().build::<_, Body>(https);
+		GrinClient {
+			api_secret,
+			proxy_addr,
```
kargakis: Actually, one related question: this is not used currently. How are you meant to use it?

chisa0a: Similar question/comment, this may require some extra plumbing using [hyper-proxy](https://lib.rs/crates/hyper-proxy). Is this PR just meant to lay the initial ground work?

hashmap:
I wasn’t clear, yes, it’s the ground work. I plan to have 3 prs to simplify the review process:
- refactor the existing code, I added proxy url for illustration purposes, to support socks proxy we may need more fields
- add socks proxy support
- enable this functionality in grin and grin- wallet, eg add fields to config files etc
I could combine it to have less prs.

quentinlesceller:
What's the status of this PR @hashmap. If still relevant can you fix the merge conflicts?

quentinlesceller:
Going to close this PR as it is one year old and no activity ever since. @hashmap feel free to reopen and rebase on latest master if needed.

Issue #2889, opened 2019-06-11
store both mmr index and block height into database for output


garyyu:
Currently, both `/v1/txhashset/outputs` API and `/v1/chain/outputs/byids` API are heavily used in wallet, and both API are very slow.

The heaviest part of these 2 API are `get_header_for_output` which gets the block header from an output commit.

https://github.com/mimblewimble/grin/blob/a89a1024b6dd7cc5d3def6b41edae0e46b55658d/chain/src/chain.rs#L1225-L1257

which has to do a loop searching to find the right height, and each loop contains two `get_header_by_height` function call.

The loop count depends on which commit we're querying, but it's normal case to get Log2(height) loops.  For example:  if the final result height is `159538`, then we need `16` loops, and that means `32` calls to `get_header_by_height`.

And  that `get_header_by_height` is not a light-cost function call:
https://github.com/mimblewimble/grin/blob/a89a1024b6dd7cc5d3def6b41edae0e46b55658d/chain/src/txhashset/txhashset.rs#L229-L236

---

But we have a simple way to change that and get a much faster improvement on these two APIs.

The way is to have both mmr index and block height saving to database for output.

Currently, when apply a new block, we save the output position (MMR index) into the database:
```rust
	pub fn apply_block(&mut self, b: &Block) -> Result<(), Error> {
		self.apply_header(&b.header)?;

		for out in b.outputs() {
			let pos = self.apply_output(out)?;
			// Update the output_pos index for the new output.
			self.batch.save_output_pos(&out.commitment(), pos)?;   <<<< Here !
		}
		...
```
If we change this `pos` to `(pos, b.header.height)`, then all those pains will be gone :-)

(but we have a pain to upgrade the database for the compatibility in one new version which take this improvement)

And Thanks @DavidBurkett 's discussion on https://github.com/GrinPlusPlus/GrinPlusPlus/issues/63#issuecomment-500658757 which leads me to realize this performance problem.

antiochp:
:+1: Makes a lot of sense to me to store these in the db.

antiochp:
For backward compatibility and ease of migration we might want to consider adding an additional index and avoid needing to migrate the existing one - 

```
A) commitment -> output_pos
B) output_pos -> block_height
```

Index (A) has the benefit of minimal chance of key collision due to using output commitments.
As long as index (B) was used in conjunction with (A) and not in isolation then we should be fine.
[...]

PR #2903, opened 2019-06-19
store both mmr index and block height into database for output

garyyu:
PR for #2889.

**5~10** times speed improvement for `/v1/chain/outputs` and `/v1/txhashset/outputs` APIs, which will make the wallet `restore`, `check` and `refresh outputs` super fast comparing to current version.

---

For the database migration speed, I tested on current Floonet and spent `5` seconds for this migration on my laptop (`Intel(R) Core(TM) i5-8210Y CPU @ 1.60GHz`).
```
20190618 22:50:44.688 DEBUG grin_chain::chain - rebuild_height_for_pos: rebuilding 175910 output_pos's height...
20190618 22:50:49.731 DEBUG grin_chain::chain - rebuild_height_for_pos: done
``` 

---

The test examples of these APIs:

```Bash
$ time curl -0 -XGET http://127.0.0.1:13413/v1/chain/outputs/byids?id=081f8c2343739bf28de457d80cc87efe62255a0bcc21e13022eefa7e5979b94529,080322b5046fce3390884b9035839ca296cdf7cfefbb4e7847dc226b35c0b6665f,083da72de4f6f22abce3c9ca9d826167c717b10eb44cfe80da4f715615738a7b85,094edc18bdaf8d0b0848b1d3e4991cd0e13a9d5aa755bc757469b4269abe4f6d5d,080384cb6acd92c1c15d20e9cd57fc74de70214866546b6dc3c9ff67d8ffcf374c
[
  {
    "commit": "081f8c2343739bf28de457d80cc87efe62255a0bcc21e13022eefa7e5979b94529",
    "height": 160001,
    "mmr_index": 470092
  },
  {
    "commit": "080322b5046fce3390884b9035839ca296cdf7cfefbb4e7847dc226b35c0b6665f",
    "height": 160001,
    "mmr_index": 470093
  },
  {
    "commit": "083da72de4f6f22abce3c9ca9d826167c717b10eb44cfe80da4f715615738a7b85",
    "height": 160003,
    "mmr_index": 470096
  },
  {
    "commit": "094edc18bdaf8d0b0848b1d3e4991cd0e13a9d5aa755bc757469b4269abe4f6d5d",
    "height": 160003,
    "mmr_index": 470097
  },
  {
    "commit": "080384cb6acd92c1c15d20e9cd57fc74de70214866546b6dc3c9ff67d8ffcf374c",
    "height": 160005,
    "mmr_index": 470099
  }
]
real	0m0.027s
user	0m0.010s
sys	0m0.009s


$ time curl -0 -XGET http://127.0.0.1:13413/v1/txhashset/outputs?start_index=100000\&max=1000 2>/dev/null >/dev/null

real	0m0.080s
user	0m0.009s
sys	0m0.009s
```


chain/src/chain.rs
```diff
@@ -461,13 +461,9 @@ impl Chain {
 	/// spent. This querying is done in a way that is consistent with the
 	/// current chain state, specifically the current winning (valid, most
 	/// work) fork.
-	pub fn is_unspent(&self, output_ref: &OutputIdentifier) -> Result<Hash, Error> {
+	pub fn is_unspent(&self, output_ref: &OutputIdentifier) -> Result<(Hash, u64, u64), Error> {
```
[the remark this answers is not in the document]
garyyu: Improved with a type.

chain/src/store.rs
```diff
@@ -111,11 +112,42 @@ impl ChainStore {
...
+				"Output position for: {:?}",
+				commit
+			))),
+			Ok(Some((pos, _height))) => Ok(pos),
+			Err(e) => Err(e),
+		}
+	}
+
+	/// Get PMMR pos and block height for the given output commitment.
+	pub fn get_output_pos_height(&self, commit: &Commitment) -> Result<(u64, u64), Error> {
```
antiochp: Might be worth returning a `OutputMMRPosition` here to wrap the two u64 values.
Not store them in the db. We don't need to duplicate the hash in there, but create them as needed based on the u64 in the db.

store/src/lmdb.rs
```diff
@@ -396,7 +396,7 @@ where
 {
 	fn deser_if_prefix_match(&self, key: &[u8], value: &[u8]) -> Option<(Vec<u8>, T)> {
 		let plen = self.prefix.len();
-		if plen == 0 || key[0..plen] == self.prefix[..] {
+		if plen == 0 || (key.len() >= plen && key[0..plen] == self.prefix[..]) {
```
antiochp: What was this change needed for?

antiochp: Oh I get it now after rereading it. We just check the length for safety so it doesn't panic.

antiochp:
@garyyu Is this good to go? I can merge if so.

chain/src/store.rs
garyyu: If I understand it well, it means no need to wrap this two u64 values? 
Let me know if you want a new type for this two u64, then I will refactor it.

store/src/lmdb.rs
garyyu: Yes. Actually I saw a panic here on a test, don't know exact reason, but not bad to give a safe protection here.

garyyu:
@antiochp  ok if you don't need that new type for the (u64, u64). I have run a lot of tests on several nodes for this PR, no problem found.

antiochp:
Testing this locally.

garyyu:
> In `rebuild_index` we do the following -
> 
> ```
> 		for pos in self.output_pmmr.leaf_pos_iter() {
> 			if let Some(out) = self.output_pmmr.get_data(pos) {
> 				self.batch.save_output_pos(&out.commit, pos)?;
> 				count += 1;
> 			}
> 		}
> ```
> 
> i.e. We are still calling `save_output_pos()` and I think we should be calling the new `save_output_pos_height()` instead?

No, I keep it to avoid a big refactoring. Instead, after each time calling this `rebuild_index`, there will be another `rebuild_height_for_pos` calling to migrate pos index to pos+height index.

There're 2 locations for this `rebuild_index`:
- https://github.com/mimblewimble/grin/blob/b0858cd035fbfab89c268d8b172fe30b4e05ab63/chain/src/chain.rs#L984  for `txhashset_write` function.
- https://github.com/mimblewimble/grin/blob/b0858cd035fbfab89c268d8b172fe30b4e05ab63/chain/src/chain.rs#L1096 for `compact` function.


---
(BTW, I want another PR to give a complete refactoring on the current `txhashset_write` function, which currently has multiple `rebuild` split into multiple locking: `rebuild_header_mmr`, `rebuild_index`, `rebuild_header_mmr`, `rebuild_height_for_pos`. There could be a race condition here if a new block coming in and seize the locking among these `rebuild`, even in a very low possibility.)

PR #2910, opened 2019-06-21
[DNM] handle the HF invalid header scenario safely during chain init

antiochp:
<b>Edit: As @DavidBurkett mentions we are not necessarily rewinding the MMR safely here. We need to rewind all the spent output pos and we do not know these if we cannot read the (now invalid) headers and corresponding hashes. This is going to take some more thinking through...</b>

----

We observed the following when upgrading a node in place _after_ the HF block height had passed.
This node had a header at height 185040 with `version=1` (mined by a miner on pre-HF code).
This header was in the local db and became invalid (was originally valid) once the node had been upgraded.
Reading this header from the db at startup caused the node to fail to startup.

```
stack backtrace:
   0: failure::backtrace::internal::InternalBacktrace::new::h51d5229f6fcf6886 (0x560bf29d4962)
             at /home/mcordner/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/backtrace/internal.rs:44
   1: failure::backtrace::Backtrace::new::h8ccade931a77636b (0x560bf29d377d)
             at /home/mcordner/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/backtrace/mod.rs:111
   2: <failure::error::error_impl::ErrorImpl as core::convert::From<F>>::from::hbbda90237adbe1e8 (0x560bf23e9719)
             at /home/mcordner/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/error/error_impl.rs:19
   3: <failure::error::Error as core::convert::From<F>>::from::hd89c1f37810a28b6 (0x560bf23d1b92)
             at /home/mcordner/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/error/mod.rs:36
   4: <T as core::convert::Into<U>>::into::h5744f998b1cc987b (0x560bf22f0df2)
             at /rustc/3c235d5600393dfe6c36eeed34042efad8d4f26e/src/libcore/convert.rs:540
   5: failure::context::Context<D>::with_err::h43fc73673ba83dbd (0x560bf23baf1f)
             at /home/mcordner/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/context.rs:105
   6: failure::Fail::context::h1c4ac468da15b50d (0x560bf22f0edf)
             at /home/mcordner/.cargo/registry/src/github.com-1ecc6299db9ec823/failure-0.1.5/src/lib.rs:160
[...]

PR #2912, opened 2019-06-23
handle the HF invalid header scenario safely during chain init

antiochp:
Replaces #2910.

[tbd - description of why this necessary with some context around the floonet HF]

This PR - 
* txhashset extension and header_extension how take a `head` (chain `tip`) and not a `header`
* build the head MMR in parallel with the sync MMR (if we increase total work)
* refactor, pull out `setup_genesis_head`
* cleanup and simplify `update_head` and `update_header_head`
* make setup_head more robust in terms of rewinding if we cannot read header from db
* `input_pos_to_rewind` (to rewind "spent" UTXO pos) now takes a height and can operate over a set of input_pos without needing to read headers from the db
* reader header hashes from the header (and sync) MMR so we can determine _which_ headers we need to rewind without necessarily needing to read/deserialize them from the db

TODO - 
- [ ] clean up error! logging

----

This got too big and unwieldy again.
Related - #2918 (which should go a long way toward making this fix simpler to impl).

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -34,7 +34,6 @@ use crate::types::{
 };
 use crate::util::secp::pedersen::{Commitment, RangeProof};
 use crate::util::RwLock;
-use grin_store::Error::NotFoundErr;
 use std::collections::HashMap;
 use std::fs::{self, File};
 use std::io::Read;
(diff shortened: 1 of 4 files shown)


antiochp:
Closing - will replace with smaller PRs.

PR #2913, opened 2019-06-24
fix wording in comment

NaN-git:
This PR corrects a misleading comment that seems to mix up primary and secondary proof of work.

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -59,8 +59,8 @@ pub const YEAR_HEIGHT: u64 = 52 * WEEK_HEIGHT;
 /// Number of blocks before a coinbase matures and can be spent
 pub const COINBASE_MATURITY: u64 = DAY_HEIGHT;
 
-/// Ratio the secondary proof of work should take over the primary, as a
-/// function of block height (time). Starts at 90% losing a percent
+/// Target ratio of secondary proof of work to primary proof of work,
+/// as a function of block height (time). Starts at 90% losing a percent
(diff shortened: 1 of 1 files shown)


antiochp:
I'm not sure these changes make sense?

Old
> Ratio the secondary proof of work should take over the primary, as a function of block height (time). Starts at 90% losing a percent approximately every week. Represented as an integer between 0 and 100.

New
> Ratio the secondary proof of work should be taken over by the primary, as a function of block height (time). Starts at 90% losing a percent approximately every week. Represented as an integer between 0 and 100.

This part no longer makes sense - 
> Ratio the secondary proof of work should be taken over by the primary

NaN-git:
> I'm not sure these changes make sense?
> 
> Old
> 
> > Ratio the secondary proof of work should take over the primary, as a function of block height (time). Starts at 90% losing a percent approximately every week. Represented as an integer between 0 and 100.
> 
> New
> 
> > Ratio the secondary proof of work should be taken over by the primary, as a function of block height (time). Starts at 90% losing a percent approximately every week. Represented as an integer between 0 and 100.
> 
> This part no longer makes sense -
> 
> > Ratio the secondary proof of work should be taken over by the primary

Yeah, you're right. After reading your comment I understood the original comment. My understanding of the comment was that the secondary POW should take over the primary POW.

antiochp:
> Can it be changed to something like target ratio of secondary POW to primary POW?

Yes. I think this does make more sense. :+1:

NaN-git:
> Yes. I think this does make more sense. +1
Done

PR #2914, opened 2019-06-25
zeroize: Upgrade to v0.9

garyyu:
can't build because of this error:
```
error: failed to select a version for the requirement `zeroize = "^0.8"`
  candidate versions found which didn't match: 0.9.1, 0.9.0, 0.7.0, ...
```
`zeroize` 0.8 is now marked as "yanked" so we have to upgrade to 0.9.

---

(Note: this PR need wait the `grin_secp256k1zkp` version updated as `0.7.7` into crates.io:  https://crates.io/crates/grin_secp256k1zkp )

### DIFF
--- a/util/Cargo.toml
+++ b/util/Cargo.toml
@@ -22,11 +22,11 @@ log = "0.4"
 walkdir = "2"
 zip = { version = "0.4", default-features = false }
 parking_lot = {version = "0.6"}
-zeroize = "0.5.2"
+zeroize = "0.9"
 
 [dependencies.grin_secp256k1zkp]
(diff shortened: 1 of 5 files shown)


util/Cargo.toml
```diff
@@ -22,11 +22,11 @@ log = "0.4"
 walkdir = "2"
 zip = { version = "0.4", default-features = false }
 parking_lot = {version = "0.6"}
-zeroize = "0.5.2"
+zeroize = "0.9"
```
yeastplume: Just noting Drop is implemented by `ZeroingString` in  `util/types.rs` so should be okay here.

PR #2915, opened 2019-06-25
Branching guidelines into CONTRIBUTING.md

yeastplume:
Add description of current branching model into CONTRIBUTING.md (avoiding shame at upcoming meeting)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,6 +24,18 @@ If submitting a PR consisting of documentation changes only, please try to ensur
 
 The development team will be happy to help and guide you with any of these points and work with you getting your PR submitted for approval. Create a PR with [WIP] in the title and ask for specific assistance within the issue, or contact the dev team on any of the channels below.
 
+# Branching Model
+
+Grin's current branching model is as follows:
+
+`master` is used for development of the current point release. For instance, if the last released version of Grin is 1.2.0, master will contain changes intending to go into 1.2.1, 1.2.2 etc. Breaking and backwards-incompatible changes should not occur between point releases, and should not be merged into master.
+
+The repository will contain a branch from the next upcoming minor or major release. If `master` contains changes for the 1.2.x series and the next planned minor release is 1.3.0, there will be a branch called `milestones/1.3.0` into which all breaking changes intended for 1.3.0 should be merged.
+
+Periodially, (and particularly in a situation where minor/major releases are undergoing a long development cycle), changes from `master` should be commit into the current major/minor branch so as to avoid a large, messy merge before release.
+
+Just before a major or minor release, the development branch is merged back into `master`, tested, tagged and released. It then becomes the new master for point releases in the series. The previous major/minor branch is deleted, and a new branch is created with the version number of the next anticipated major/minor release.
+
 # Find Us
 
 When you are starting to contribute to grin, we really would appreciate if you come by the gitter chat channels.


CONTRIBUTING.md
```diff
@@ -24,6 +24,18 @@ If submitting a PR consisting of documentation changes only, please try to ensur
 
 The development team will be happy to help and guide you with any of these points and work with you getting your PR submitted for approval. Create a PR with [WIP] in the title and ask for specific assistance within the issue, or contact the dev team on any of the channels below.
 
+# Branching Model
+
+Grin's current branching model is as follows:
+
+`master` is used for development of the current point release. For instance, if the last released version of Grin is 1.2.0, master will contain changes intending to go into 1.2.1, 1.2.2 etc. Breaking and backwards-incompatible changes should not occur between point releases, and should not be merged into master.
+
```
lehnberg: Is it expected that `master` always should be safe to build and be  deployable on production?

CONTRIBUTING.md
```diff
@@ -24,6 +24,18 @@ If submitting a PR consisting of documentation changes only, please try to ensur
...
+The repository will contain a branch from the next upcoming minor or major release. If `master` contains changes for the 1.2.x series and the next planned minor release is 1.3.0, there will be a branch called `milestones/1.3.0` into which all breaking changes intended for 1.3.0 should be merged.
+
```
lehnberg: Does it make sense to outline a scenario where `master` is for 1.2.x, there's a (frozen) `milestones/2.0.0`, and how people are supposed to do work that is targeted for a post `2.0.0` world?

CONTRIBUTING.md
yeastplume: Good question. Deployable on production I'd say no, it's expected to be bleeding edge and only releases should be deemed 'safe for production'. It should always build, however.

CONTRIBUTING.md
yeastplume: I think the current situation is fairly unique where we have a long and frozen 2.0.0, and I wouldn't expect it to be the norm (at least I'd hope not), so not sure explicitly outlining this scenario is needed right now. Of course, this could change

lehnberg: Oki. What triggered my question, is that based on the current contents of the document, I can't deduce how this is meant to be handled. It also doesn't feel like an edge case to me, given that we have at least three more of these scenarios scheduled over the next 18 months.

CONTRIBUTING.md
kargakis: Based on older discussions, the basic premise for the current branching model was that `master` is meant to be deployable (think of the simple case: clone+build+run) and breaking changes are developed as part of the milestone branches. If `master` is only meant to be buildable then what's the point of the milestone branches? I assume a milestone branch is also buildable if you get something merged into it, ie. it passes CI.

CONTRIBUTING.md
kargakis: There is a lot of conflicts risk involved in working on top of two moving branches but we could add a script in the repo that brings a dev branch up to date with a post 2.0.0 world (`git checkout milestones/2.0.0`->`git checkout -b mybranch`->`git merge master`)

yeastplume: The current situation whereby we've essentially frozen all development other than work that's directly hard-fork related is a contingency measure. I'd hope we get our panning together a bit better for future HFs so as this situation won't be repeated. That's why I'm not keen on putting together guidelines that reflect the current situation.

Issue #2906, opened 2019-06-21
grin server can't stop


garyyu:
In `1.1.0` release, look like we still have the peers stop issue?  the grin process can't stop. here is the shutdown logs:

<details><summary>shutdown logs</summary>
<p>

```
20190621 10:10:10.485 WARN grin::cmd::server - Received SIGINT (Ctrl+C) or SIGTERM (kill).
20190621 10:10:10.485 DEBUG grin_servers::common::types - sync_state: sync_status: NoSync -> Shutdown
20190621 10:10:10.552 INFO grin_servers::grin::server - connect_and_monitor thread stopped
20190621 10:10:11.015 INFO grin_servers::grin::server - sync thread stopped
20190621 10:10:11.015 INFO grin_servers::grin::server - dandelion_monitor thread stopped
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(86.85.139.156:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(178.63.97.197:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(52.59.238.178:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(46.4.91.48:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(157.230.142.11:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(120.78.134.123:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(109.124.18.120:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(18.222.188.55:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(68.183.167.183:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(217.123.171.4:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(178.62.232.49:3414))
20190621 10:10:11.015 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(217.182.192.59:3414))
[...]

PR #2917, opened 2019-06-25
Don't wait for p2p-server thread

hashmap:
Currently p2p.stop() stops and wait for all peers to exit, that's
basically all we need. However we also run a TCP listener in this thread
which is blocked on `accept` most of the time. We do an attempt to stop
it but it would work only if we get an incoming connection during the
shutdown, which is a week guarantee.

This fix remove joining to p2p-server thread, it stops all peers and
makes an attempt to stop the listener.

Fixes [#2906]


garyyu:
Thanks, looks good. Let me test it for 24hours on my seed server, hope we can catch up the 2.0.0 release.

Issue #2929, opened 2019-06-28
txhashset archive save to file fail (connection closed)


antiochp:
Added some logging to the txhashset download so we can track progress via the log file (on the 2.x.x branch) - 

```
20190628 09:40:09.321 DEBUG grin_p2p::peer - Asking 67.207.70.122:3414 for txhashset archive at 231880 000018532638.
20190628 09:40:14.687 DEBUG grin_p2p::protocol - handle_payload: txhashset archive for 000018532638 at 231880. size=253592404
20190628 09:40:25.693 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 2976000/253592404
20190628 09:40:36.748 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 9072000/253592404
20190628 09:40:47.769 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 16944000/253592404
20190628 09:40:58.829 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 24384000/253592404
...
20190628 09:49:02.918 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 238656000/253592404
20190628 09:49:14.008 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 245472000/253592404
20190628 09:49:25.148 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 250128000/253592404
20190628 09:49:36.447 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 251232000/253592404
20190628 09:49:38.456 ERROR grin_p2p::protocol - handle_payload: txhashset archive save to file fail. err=Connection(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" })
```

I have seen the above scenario a few time now. We download what appears to be _almost_ all the bytes and then the connection is dropped.

Seems like too much of a coincidence that we get _so close_ to getting the whole file and then fail to save it with a "connection closed" (like 99% of the way there on multiple occasions).

I _suspect_ we actually have all the bytes and that the connection is getting closed on the sending side as soon as it has sent everything over (for whatever reason).
We must be able to do something smarter on the receiving side to handle this more robustly.

Related - #2639.

antiochp:
Here's another instance of it happening - 

```
...
20190628 10:11:09.139 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 261024000/275364910
[...]

PR #2930, opened 2019-06-28
Replace SyncChannel with Channel in peer

hashmap:
We use a buffered SyncChannel (buffer size is 10) to store messages to
be sent to a remote peer. In send impl we use try_send which return
error if channel is closed (peer is disconnected) or the buffer is full.
Such error leads to dropping the peer.

slow it may lead to channel sending error hence the peer's drop.

We could increase buffer, but at the same time we use SyncChannel as
regular channel (with unlimited buffer), so it may make more sense to switch to it. It's unlikely that a peer can buffer too many messages, so wasting memory on preallocating a big buffer is not justified.

Addresses #2929

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -201,7 +201,7 @@ impl StopHandle {
 
 pub struct ConnHandle {
 	/// Channel to allow sending data through the connection
-	pub send_channel: mpsc::SyncSender<Vec<u8>>,
+	pub send_channel: mpsc::Sender<Vec<u8>>,
 }
 
 impl ConnHandle {
@@ -211,7 +211,7 @@ impl ConnHandle {
 	{
 		let buf = write_to_buf(body, msg_type)?;
 		let buf_len = buf.len();
-		self.send_channel.try_send(buf)?;
+		self.send_channel.send(buf)?;
 		Ok(buf_len as u64)
 	}
 }
@@ -261,7 +261,7 @@ pub fn listen<H>(
 where
 	H: MessageHandler,
 {
-	let (send_tx, send_rx) = mpsc::sync_channel(SEND_CHANNEL_CAP);
+	let (send_tx, send_rx) = mpsc::channel();
 	let (close_tx, close_rx) = mpsc::channel();
 
 	stream
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -98,8 +98,8 @@ impl From<io::Error> for Error {
 		Error::Connection(e)
 	}
 }
-impl<T> From<mpsc::TrySendError<T>> for Error {
-	fn from(e: mpsc::TrySendError<T>) -> Error {
+impl<T> From<mpsc::SendError<T>> for Error {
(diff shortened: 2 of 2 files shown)


hashmap:
I agree it’s risky to ship it with 2.0, but we do see peers massacre, would be good to try to fix it in 2.0, because the entire network will update, next chance will be in 6 month.

antiochp:
Just saw this locally with these changes, so it doesn't fix the issue with receiving txhashset - 

```
20190628 16:55:33.815 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 241248000/251122166
20190628 16:55:44.889 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 247584000/251122166
20190628 16:55:48.949 ERROR grin_p2p::protocol - handle_payload: txhashset archive save to file fail. err=Connection(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" })
20190628 16:55:48.949 DEBUG grin_p2p::conn - try_break: exit the loop: Connection(Os { code: 54, kind: ConnectionReset, message: "Connection reset by peer" })
20190628 16:55:48.949 DEBUG grin_p2p::conn - Shutting down connection with ?
```

That said - I do think these changes are still worth considering.

hashmap:
I’ll measure how often we see this issue, then we can discuss if the risk is worth of it

hashmap:
@antiochp sure, you need to update your peers to fix it

antiochp:
Even if this is the underlying reason why we are closing the connection - we are still sending at least 99% of the txhashset file across. It still feels like too much of a coincidence that we get that far along and then close the connection.

Are we returning from sending the final set of bytes while these bytes are still on their way out and closing the connection before they have time to finish sending?

hashmap:
It’s not a coincidence, our peer thread’s loop reads from a socket, writes to it, checks the stop channel, repeats. So even if the peer is stopped the thread will send all bytes before checking the stop channel. Then it immediately closes the connection. I’m not a TCP stack expert, but I speculate the socket closure breaks the last read on the receiver side.

antiochp:
Related - I think we are not necessarily handling error conditions with our `send_rx.try_recv()` call correectly either...

https://doc.rust-lang.org/std/sync/mpsc/enum.TryRecvError.html

https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.try_recv

> Compared with recv, this function has two failure cases instead of one (one for disconnection, one for an empty buffer).

We simply ignore _both_ of these and continue. It would probably be better to identify the "disconnection" state and actually do something about it).
Maybe this is not a huge issue as whatever caused the channel to close is likely in the process of cleaning up the peer anyway, but it feels wrong to ignore the "disconnection" when we know it happened.

antiochp:
I think we should do this for `2.x.x` if the changes on `2.0.0` (increase the cap to 100) perform smoothly.

hashmap:
@antiochp agree, it was closed by github because the branch was deleted

PR #2931, opened 2019-06-28
Increase peer's send buffer

hashmap:
slow buffer capacity 10 may be not enough,  hence the peer's drop.

Safer attempt to address #2929 in 2.0.0, more info https://github.com/mimblewimble/grin/pull/2930


hashmap:
I tested the original version with some additional logging (https://github.com/cyclefortytwo/grin/tree/test-peer-drop) can confirm that this situation (drop a peer because of the full buffer) happens.

PR #2934, opened 2019-06-30
Increase peer's send buffer (#2931)

hashmap:
slow buffer capacity 10 may be not enough,  hence the peer's drop.

Safer attempt to address #2929 in 2.0.0


hashmap:
It was merged to 2.0.0, pushing to master

PR #2939, opened 2019-07-04
Add 4th contact to SECURITY.md

j01tz:
Following up with agreement on point 5.3 from July 02 governance meeting, adds my PGP key for additional security contact.


quentinlesceller:
👍

PR #2940, opened 2019-07-06
Add param to include rangeproof in block api

jaspervdm:
This PR adds an optional `include_proof` param in the block API to include the bulletproofs

### DIFF
--- a/api/src/handlers/blocks_api.rs
+++ b/api/src/handlers/blocks_api.rs
@@ -87,10 +87,15 @@ pub struct BlockHandler {
 }
 
 impl BlockHandler {
-	fn get_block(&self, h: &Hash, include_merkle_proof: bool) -> Result<BlockPrintable, Error> {
+	fn get_block(
+		&self,
+		h: &Hash,
+		include_proof: bool,
+		include_merkle_proof: bool,
+	) -> Result<BlockPrintable, Error> {
 		let chain = w(&self.chain)?;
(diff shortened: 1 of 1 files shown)


api/src/handlers/blocks_api.rs
```diff
@@ -144,16 +149,18 @@ impl Handler for BlockHandler {
 		if let Some(param) = req.uri().query() {
 			if param == "compact" {
 				result_to_response(self.get_compact_block(&h))
+			} else if param == "include_proof" {
+				result_to_response(self.get_block(&h, true, true))
 			} else if param == "no_merkle_proof" {
-				result_to_response(self.get_block(&h, false))
+				result_to_response(self.get_block(&h, false, false))
```
antiochp: These permutations are getting a bit complex with multiple combinations of true/false for the two proofs.
Not for this PR but we should probably think about passing a struct around - build it based on query params and pass it to `build_block()`.

jaspervdm: Yes, makes sense

jaspervdm:
Hm seems like changing the base branch to v2.x.x will add a lot of other commits. Will close this and re-open one on the proper branch

antiochp:
I don't really understand why sometimes changing the target branch works perfectly and other times it pulls in dozens of unrelated commits...

hashmap:
It seems we squashed a few commits on master branch, so both `2.0.0` and `2.x.x` look a bit alien for git now

PR #2950, opened 2019-07-15
test: Re-enabled too large block test. Block tests now use ChainTypes::AutomatedTesting

JosephGoulden:
The AutomatedTesting chain type has a much smaller max block weight so it's possible to test the size validation in a short amount of time (it takes about one second now).

After enabling AutomatedTesting in the too_large_block test the serialized_size tests started failing as this is a shared global state and the serialized block size is smaller in the AutomatedTesting mode, so I updated the sizes.


hashmap:
Excellent, thanks for the contribution!

PR #2919, opened 2019-06-25
[2.x.x] Generate txhashset archives on 720 block intervals. 

quentinlesceller:
Replace https://github.com/mimblewimble/grin/pull/2813. Cause this previous PR was merged on master branch.

PR #2951, opened 2019-07-16
[2.x.x] Generate txhashset archives on 720 block intervals.

quentinlesceller:
Fixes #2919 and ultimately replaces #2813.


antiochp:
Note to self: We need to introduce a shorter "test" interval for this for local testing in "user testing" mode. We can do this in a separate PR.

PR #2956, opened 2019-07-20
fix: move checks for automated test mode into global.rs

JosephGoulden:
Moved some CI checks from pipe.rs to global. I think it makes sense to do this configuration in one place and it looks like the majority of it is done there now. The only real change to the automated tests here is that the genesis block timestamp is 60 minutes in the past rather than the time now.


antiochp:
I think I created a conflict here touching some adjacent code...

JosephGoulden:
Thanks for review. I fixed the conflicts. Also had to change block time on your new chain_test_helper to 60 from 160, hope that's ok.

antiochp:
We just merged `2.x.x` back into `master`.
Updated base branch for this PR back to `master`.
Once all tests pass I'll merge :+1:.

antiochp:
Getting a windows specific test failure on CI.

quentinlesceller:
Seems like a random failure. Relaunched.

PR #2961, opened 2019-07-23
schnorr signature batch verification

garyyu:
To complete the schnorr signature batch verification feature.

The related PRs:
- in `secp256k1-zkp` repo as https://github.com/mimblewimble/secp256k1-zkp/pull/47
- in `rust-secp256k1-zkp` repo as https://github.com/mimblewimble/rust-secp256k1-zkp/pull/56.

This should bring us over 50% performance improvement on schnorr signature verification:
> if we batch with 1000 signatures, the average time is 40.5/78.5 = 51.6% compared to no batching.


chain/src/txhashset/txhashset.rs
```diff
@@ -1323,28 +1323,39 @@ impl<'a> Extension<'a> {
...
 					.get_data(n)
 					.ok_or::<Error>(ErrorKind::TxKernelNotFound.into())?;
 
-				kernel.verify()?;
+				tx_kernels.push(kernel.kernel);
 				kern_count += 1;
 
-				if kern_count % 20 == 0 {
+				// batch on every 1024 kernels
+				if kern_count & 0x0400 == 0 {
```
antiochp: These `& 0x0400` are _really_ hard to read at a glance. Can we use modulo `%` here for simplicity?
The counts are integers, so it makes sense to just use integer modulo (rather than bitwise and).

chain/src/txhashset/txhashset.rs
```diff
@@ -1323,28 +1323,39 @@ impl<'a> Extension<'a> {
 
 		let mut kern_count = 0;
 		let total_kernels = pmmr::n_leaves(self.kernel_pmmr.unpruned_size());
+		let mut tx_kernels: Vec<TxKernel> = Vec::with_capacity(1_024);
```
antiochp: I think we can simplify this by just using an iterator and `chunks`.
https://doc.rust-lang.org/std/primitive.slice.html#method.chunks

That way we don't need to hand code the chunking logic.

core/src/core/transaction.rs
```diff
@@ -306,6 +306,34 @@ impl TxKernel {
...
+		let len = tx_kernels.len();
+		let mut sigs: Vec<secp::Signature> = Vec::with_capacity(len);
+		let mut pubkeys: Vec<secp::key::PublicKey> = Vec::with_capacity(len);
+		let mut msgs: Vec<secp::Message> = Vec::with_capacity(len);
+
+		let secp = static_secp_instance();
+		let secp = secp.lock();
+
+		for tx_kernel in tx_kernels {
+			if tx_kernel.is_coinbase() && tx_kernel.fee != 0
```
antiochp: Can we split the fee and lock_height verification out from this batch signature verification?
There are some pending changes in #2859 that rework this and these don't need to be batched up.
We can just do 2 passes over the kernels -
* verify fee and lock_height
* then batch verify the signatures

chain/src/txhashset/txhashset.rs
garyyu: That's for efficiency, `& 0x0400` is much faster than ` % sth`. For readable, that's why the comments line is there.

core/src/core/transaction.rs
garyyu: Sure, I can split it out and will rework it.
BTW, when do you plan to complete #2859?

quentinlesceller:
Rebased to master.

chain/src/txhashset/txhashset.rs
antiochp: > That's for efficiency

This is almost certainly not the bottleneck here. I'd much prefer the code be easy to read (and self documenting).

garyyu: 😄 Okay, let me change this.

chain/src/txhashset/txhashset.rs
garyyu: sorry for late response, let me finish this `chunks` today.

core/src/core/transaction.rs
garyyu: will try the splitting today and request a new review on this.

chain/src/txhashset/txhashset.rs
garyyu: Oops,  the `chunks` is useful for splitting an existing big vector, but here I need create and fill a new vector, and I don't want to create a huge vector and later to split it with `chunks`. So I think it's better to keep it like this.

Pls let me know if thought wrongly.

garyyu:
@antiochp  Updated. Could you take a review again? thanks.

garyyu:
Oops, just saw the `fee` and `lock_height` checking has been removed from kernel verify function in #2859: 
- https://github.com/mimblewimble/grin/pull/2859/files#diff-404c77cf79ebda67cdbd58c1c99daa8cR364-L288

So, let me rework it again to remove the `fee_height_verify` function which I just split from origin verify function.

garyyu:
@antiochp  I have a related question on https://github.com/mimblewimble/grin/pull/2859/files#r318043781 . Could you please have a look?

antiochp:
Testing this locally.

core/src/core/transaction.rs
```diff
@@ -852,11 +875,7 @@ impl TransactionBody {
 		};
 
 		// Verify the unverified tx kernels.
-		// No ability to batch verify these right now
-		// so just do them individually.
-		for x in &kernels {
-			x.verify()?;
-		}
+		TxKernel::batch_sig_verify(&kernels)?;
```
antiochp: I'm guessing we are fine just batching them all up like this in the context of a block.
But is there a limit to how many we can batch?

antiochp:
This looks significantly faster than 50% speedup to me. More like 5x - 

```
20190829 18:13:56.905 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 5000 signatures
20190829 18:13:57.253 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 10000 signatures
20190829 18:13:57.599 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 15000 signatures
20190829 18:13:57.952 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 20000 signatures
20190829 18:13:58.301 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 25000 signatures
20190829 18:13:58.655 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 30000 signatures
20190829 18:13:59.003 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 35000 signatures
20190829 18:13:59.351 DEBUG grin_chain::txhashset::txhashset - txhashset: verify_kernel_signatures: verified 40000 signatures
```

DavidBurkett:
Shouldn't be more than 50%, unless we were doing something really inefficient before. Fingers crossed that we were 😂

core/src/core/transaction.rs
garyyu: The limit should be big enough. At least 100,000 batching is OK:
```sh
running 1 test
Verifying aggsig batch of 100000
100000 signatures verified in 6408.601ms
test aggsig::tests::test_aggsig_batch ... ok
```

PR #2975, opened 2019-07-24
Update intro.md

rsoltanzadeh:
Significant clarifications, especially in the paragraph where excess value is introduced and the section about kernel offsets.


doc/intro.md
```diff
@@ -291,66 +286,62 @@ concept: _cut-through_. With this addition, a MimbleWimble chain gains:
...
 
     (in1, in2) -> (out1), (kern1)
 
-We also know that everything remaining can be used to reconstruct the other valid transaction -
+We also know that everything remaining can be used to reconstruct the other valid transaction:
 
     (in3) -> (out2), (kern2)
 
-To mitigate this we include a _kernel offset_ with every transaction kernel. This is a blinding factor (private key) that needs to be added back to the kernel excess to verify the commitments sum to zero -
+To mitigate this we include a _kernel offset_ with every transaction kernel. This is a public key that needs to be added to the kernel excess to ensure the commitments sum to zero:
```
antiochp: The offset is a blinding factor here.
If the offset was a commitment it would be trivial to build a tx with any arbitrary offset and any corresponding outputs (ask me how I know).

antiochp:
Looks good - I need to have another read through but :+1:

doc/intro.md
rsoltanzadeh: It says 
>This is a blinding factor (private key) that needs to be added back to the kernel excess to verify the commitments sum to zero.

If the offset is a blinding factor (scalar), then surely you can't add it to a curve point (the kernel excess is `(r2-r1)*G`). At the time of writing I realize that it's probably meant to say *excess value* and not *kernel excess*, which would make the equation
>`sum(outputs) - sum(inputs) = sum(kernel_excess) + kernel_offset*G`

rather than the currently written

>`sum(outputs) - sum(inputs) = sum(kernel_excess) + kernel_offset`

If I understand this correctly, the kernel excess is actually not `r*G` where *r* is the sum of the input and output blinding factors, but rather `(r-offset)*G`, and the offset (not-so-private key) is published so that the miners can validate the individual transactions during block creation.

I'll edit shortly if there are no objections (i.e. if I'm not wrong still).

antiochp: > ```sum(outputs) - sum(inputs) = kernel_excess + kernel_offset*G```

Yes I think this is correct.

I think we were trying to balance accuracy with simplicity here in the intro. Once you start bringing both the excess and the offset into the equation it gets significantly more complex to explain clearly.

sesam:
Really nice that you're working on improving the docs. This is really important for grin, IMHO.

I see the changed document has different column wrap. If you have time to rebase to get rid of the conflict that is currently preventing merging, maybe you could also have your text editor redo the line-breaks to match the original more narrow? I'm thinking that if the changes were easier to see, then this update could be easier to get through the review process.

quentinlesceller:
Just fixed the merge conflict. Going to have a thorough review today.

doc/intro.md
```diff
@@ -126,8 +126,8 @@ Curve Cryptography. `r*G + v*H` is called a _Pedersen Commitment_.
 As an example, let's assume we want to build a transaction with two inputs and one
 output. We have (ignoring fees):
 
-* `vi1` and `vi2` as input values.
-* `vo3` as output value.
+* *vi1* and *vi2* as input values.
```
quentinlesceller: For consistency sake we should use either \*x\* or \`x\` for mathematical notation.

quentinlesceller: Same as \_x\_ which is equivalent to \*x\*

doc/intro.md
```diff
@@ -192,25 +192,15 @@ She picks 113 say, and what ends up on the blockchain is:
...
+and that the transacting parties collectively can produce its private key (85 in the above example). The simplest way to do so is to require a signature built with the excess value (85),
+which then ensures that:
 
-The simplest way to do so is to require a signature built with the excess value (85),
-which then validates that:
-
-* The transacting parties collectively know the private key (the excess value 85), and
-* The sum of the transaction outputs, minus the inputs, sum to a zero value
-  (because only a valid public key, matching the private key, will check against
-  the signature).
```
quentinlesceller: Seems like we are loosing a bit of explanation here? I'd be for repeating things again just to clarify as much as possible.

doc/intro.md
```diff
@@ -256,14 +245,20 @@ the number. We won't elaborate on the range proof, but you just need to know
...
+    Carol's UTXO:      113*G + 2*H
+    Attacker's output: (113 + 99)*G + 2*H
 
+which can be signed by the attacker because Carol's blinding factor cancels out in the equation `Y - Xi`:
+
+    Y - Xi = ((113 + 99)*G + 2*H) - (113*G + 2*H) =  99*G
+    
+This output (`(113 + 99)*G + 2*H`) requires that both the numbers 113 and 99 are known in order to be spent; the attacker
+would thus have successfully locked Carol's UTXO. The requirement for a range proof for the blinding factor prevents this
+because the attacker doesn't know the number 113 and thus neither (113 + 99). A more detailed description of range proofs is further detailed in the [range proof paper](https://eprint.iacr.org/2017/1066.pdf).
```
quentinlesceller: 👍

doc/intro.md
```diff
@@ -291,66 +286,66 @@ concept: _cut-through_. With this addition, a Mimblewimble chain gains:
 * Extremely good scalability, as the great majority of transaction data can be
   eliminated over time, without compromising security.
 * Further anonymity by mixing and removing transaction data.
-* And the ability for new nodes to sync up with the rest of the network very
-  efficiently.
```
quentinlesceller: Why remove it? It's a bit of a repetition of point 1 but can still clarify things here no?

doc/intro.md
rsoltanzadeh: I didn't just delete that part. I rephrased it on lines 202-203.

doc/intro.md
rsoltanzadeh: I removed it beacuse it was repetition of the first point, yes. If you want to keep it I would suggest merging it together with the first point rather than having it separate as it strikes me as dishonest.

For example: *"Extremely good scalability as the great majority of transaction data can be eliminated over time without compromising security, decreasing the burden of running a full node as well as streamlining the synchronization of new nodes."*

doc/intro.md
rsoltanzadeh: The convention used in the document seems to be that variable names are in italics while mathematical expressions are code-formatted. I simply followed that convention.

doc/intro.md
quentinlesceller: 👍

doc/intro.md
quentinlesceller: Okay 👍

quentinlesceller:
Okay LGTM.

PR #2976, opened 2019-07-25
Prune list robustness

antiochp:
We read the "prune list" from external data (via a txhashset download).
So we need to be a little careful with what we expect this to contain.

Specifically we make assumptions about this being 1-indexed and therefore will never contain a `0` value.

This PR adds some robustness around this - we still make the same assumption but we are not a little more defensive in how we deal with it.

* Use `saturating_sub(1)` in preference to `- 1` to avoid overflows.
* Filter out `0` from our 1-indexed prune list when initializing the prune list.
* Add some `assert!` to make logic errors more visible if/when we misuse the prune list

### DIFF
--- a/store/src/prune_list.rs
+++ b/store/src/prune_list.rs
@@ -65,7 +65,7 @@ impl PruneList {
 		}
 	}
 
-	// Instatiate a new empty prune list.
+	/// Instatiate a new empty prune list.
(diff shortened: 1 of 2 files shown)


antiochp:
We've tested this for a bit in various scenarios.
Going to merge this.

PR #2981, opened 2019-07-30
Improve validation when processing orphan blocks

j01tz:
The block acceptance pipeline can potentially be improved by adding a proof of work validation check before adding orphan blocks to the orphan block pool. 

Currently adding an orphan block with an invalid proof of work is allowed and can consume unnecessary resources and potentially flush legitimate orphans from the pool if MAX_ORPHAN_SIZE is reached.

This PR adds a check based on the one proposed by @quentinlesceller to better validate orphan blocks before they are added to the orphan block pool. 
* Adds `validate_pow_only()` function to cheaply validate orphans
* Adds `validate_pow_only()` check to `pipe::process_block()` inside the orphan conditional before `​Err​(ErrorKind::​Orphan​.into())` is returned

This is a WIP as it needs more testing, someone may find a more efficient/elegant way for this check and tests still need to be added.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -53,6 +53,23 @@ fn check_known(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(), E
 	Ok(())
 }
 
+// Validate only the proof of work in a block header.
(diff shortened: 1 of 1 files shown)


antiochp:
Going to do some local testing of this - will merge after that.

PR #2982, opened 2019-07-30
Updates to support XORed wallet keys/Tokens

yeastplume:
Mostly to support: https://github.com/mimblewimble/grin-wallet/pull/200

* add an `option_seckey_serde` de/serialiser (using a seckey internally to represent the mask)
* add a `mask_master_key` helper function to the keychain trait + ExtKeychain implementation

Leaving as WIP for the time being in case there are more changes related to this


yeastplume:
Thanks!

PR #2984, opened 2019-07-31
TxHashSet Download Improvement

quentinlesceller:
This PR introduces additional checks during TxHashSet download.


antiochp:
Doing some testing of this locally and seeing `Invalid Block Time` errors when attempting to sync. I don't think it is related - but it looks like "user testing" mode is maybe not working right now.
Working on getting it fixed before continuing testing of this.

antiochp:
Tested locally in user testing syncing from one node to another. :+1:

Issue #2726, opened 2019-04-03
Peer connection drop


hashmap:
It's not an issue per se, it was a design decision, however it may be worth to discuss one more time.

Grin node periodically kills the excess of connections https://github.com/mimblewimble/grin/blob/master/p2p/src/peers.rs#L463 which is understandable but annoying for counterparts. From my experience a node connects and almost immediately get dropped. There are couple ideas how to improve it:

- Send list of peers before dropping a peer https://github.com/mimblewimble/grin/pull/2725 (however some peers don't need it, is it ok to spam?)
- FIFO dropping -  https://github.com/mimblewimble/grin/pull/2724 to support it (could be gameable)

Do we have some other ideas?

JeremyRubin:
@ethanheilman is probably a good person to look at this.

DavidBurkett:
Yes, it'd be great to get an expert opinion on this. My thoughts are: 
* FIFO dropping could allow someone with access to a few dozen IPs to fill all of the node's connections.
* More practically, FIFO dropping is annoying because you don't have any long-term, reliable peers.

I put some thought into it, and the best approach I can come up with to solve our current peer health problems, without sacrificing network security, would be to:
   1) Have separate incoming and outgoing connection quotas.
   2) If a node has filled its quota of incoming connections, any new connection requests should be responded to with a successful handshake (optional), a message with a list of peers (crucial on seed nodes for new peers who are trying to bootstrap), and then immediately dropped. The only exception would be if the peer has a higher chain height/difficulty, we would want to allow them to connect to make sure we're not being eclipse and/or sybil attacked.
   3) To combat the remote possibility of you and your peers becoming isolated from the main network, occasionally drop a random incoming connection or two when your incoming quota is full. Do the same with your outgoing connection pool, and immediately try to connect to another peer (preferably from a different subnet and/or geographical region).

Obviously, this doesn't have to all be implemented at once, but I'd prefer to work toward a system similar to this. I'd love to hear everyone's thoughts on this.

garyyu:
@hashmap 

> From my experience a node connects and almost immediately get dropped. 

When `dropping` a connection because of excessing, does/could the remote peer try connecting again and again?
I suppose yes.
Perhaps we should duplicate the `excess` logic on the `connection` request part, instead of current accept / drop / accept / drop / ...

> Send list of peers before dropping a peer #2725 (however some peers don't need it, is it ok to spam?)

[...]

PR #2985, opened 2019-07-31
Improve checking for p2p connection limits

j01tz:
Connection limiting can be improved by adding a check against maximum allowed peer count immediately after a connection is accepted.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -274,12 +274,18 @@ fn comments() -> HashMap<String, String> {
 #how long a banned peer should stay banned
 #ban_window = 10800
 
-#maximum number of peers
-#peer_max_count = 125
+#maximum number of inbound peer connections
+#peer_max_inbound_count = 117
+
+#maximum number of outbound peer connections
+#peer_max_outbound_count = 8
 
 #preferred minimum number of peers (we'll actively keep trying to add peers
 #until we get to at least this number
 #peer_min_preferred_count = 8
+
+#amount of incoming connections temporarily allowed to exceed peer_max_inbound_count
+#peer_listener_buffer_count = 8
 
 # 15 = Bit flags for FULL_NODE
 #This structure needs to be changed internally, to make it more configurable
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -275,14 +275,14 @@ fn comments() -> HashMap<String, String> {
 #ban_window = 10800
 
 #maximum number of inbound peer connections
-#peer_max_inbound_count = 117
+#peer_max_inbound_count = 128
 
 #maximum number of outbound peer connections
 #peer_max_outbound_count = 8
 
-#preferred minimum number of peers (we'll actively keep trying to add peers
-#until we get to at least this number
-#peer_min_preferred_count = 8
+#preferred minimum number of outbound peers (we'll actively keep trying to add peers
+#until we get to at least this number)
+#peer_min_preferred_outbound_count = 8
 
 #amount of incoming connections temporarily allowed to exceed peer_max_inbound_count
 #peer_listener_buffer_count = 8
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -126,10 +126,19 @@ impl Peers {
 		res
 	}
 
+	/// Get vec of peers we currently have an outgoing connection with.
 	pub fn outgoing_connected_peers(&self) -> Vec<Arc<Peer>> {
 		self.connected_peers()
 			.into_iter()
 			.filter(|x| x.info.is_outbound())
+			.collect()
+	}
+
+	/// Get vec of peers we currently have an incoming connection with.
+	pub fn incoming_connected_peers(&self) -> Vec<Arc<Peer>> {
+		self.connected_peers()
+			.into_iter()
+			.filter(|x| x.info.is_inbound())
 			.collect()
 	}
 
@@ -153,6 +162,11 @@ impl Peers {
 	/// Number of outbound peers currently connected to.
 	pub fn peer_outbound_count(&self) -> u32 {
 		self.outgoing_connected_peers().len() as u32
+	}
+
+	/// Number of inbound peers currently connected to.
+	pub fn peer_inbound_count(&self) -> u32 {
+		self.incoming_connected_peers().len() as u32
 	}
 
 	// Return vec of connected peers that currently advertise more work
@@ -433,7 +447,7 @@ impl Peers {
 	/// Iterate over the peer list and prune all peers we have
 	/// lost connection to or have been deemed problematic.
 	/// Also avoid connected peer count getting too high.
-	pub fn clean_peers(&self, max_count: usize) {
+	pub fn clean_peers(&self, max_inbound_count: usize, max_outbound_count: usize) {
 		let mut rm = vec![];
 
 		// build a list of peers to be cleaned up
@@ -477,16 +491,27 @@ impl Peers {
 			}
 		}
 
-		// ensure we do not still have too many connected peers
-		let excess_count = (self.peer_count() as usize)
-			.saturating_sub(rm.len())
-			.saturating_sub(max_count);
-		if excess_count > 0 {
-			// map peers to addrs in a block to bound how long we keep the read lock for
+		// check here to make sure we don't have too many outgoing connections
+		let excess_outgoing_count =
+			(self.peer_outbound_count() as usize).saturating_sub(max_outbound_count);
+		if excess_outgoing_count > 0 {
 			let mut addrs = self
-				.connected_peers()
+				.outgoing_connected_peers()
 				.iter()
-				.take(excess_count)
+				.take(excess_outgoing_count)
+				.map(|x| x.info.addr.clone())
+				.collect::<Vec<_>>();
+			rm.append(&mut addrs);
+		}
+
+		// check here to make sure we don't have too many incoming connections
+		let excess_incoming_count =
+			(self.peer_inbound_count() as usize).saturating_sub(max_inbound_count);
+		if excess_incoming_count > 0 {
+			let mut addrs = self
+				.incoming_connected_peers()
+				.iter()
+				.take(excess_incoming_count)
 				.map(|x| x.info.addr.clone())
 				.collect::<Vec<_>>();
 			rm.append(&mut addrs);
@@ -525,7 +550,7 @@ impl Peers {
 	/// We have enough peers, both total connected and outbound connected
 	pub fn healthy_peers_mix(&self) -> bool {
 		self.enough_peers()
-			&& self.peer_outbound_count() >= self.config.peer_min_preferred_count() / 2
+			&& self.peer_outbound_count() >= self.config.peer_min_preferred_outbound_count()
 	}
 
 	/// Removes those peers that seem to have expired
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -545,8 +545,8 @@ impl Peers {
 		}
 	}
 
-	/// We have enough peers, both total connected and outbound connected
-	pub fn healthy_peers_mix(&self) -> bool {
+	/// We have enough outbound connected peers
+	pub fn enough_outbound_peers(&self) -> bool {
 		self.peer_outbound_count() >= self.config.peer_min_preferred_outbound_count()
 	}
 
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -338,7 +338,8 @@ impl Peers {
 	/// A peer implementation may drop the broadcast request
 	/// if it knows the remote peer already has the block.
 	pub fn broadcast_compact_block(&self, b: &core::CompactBlock) {
-		let num_peers = self.config.peer_max_count();
+		let num_peers =
+			self.config.peer_max_inbound_count() + self.config.peer_max_outbound_count();
 		let count = self.broadcast("compact block", num_peers, |p| p.send_compact_block(b));
 		debug!(
 			"broadcast_compact_block: {}, {} at {}, to {} peers, done.",
@@ -355,7 +356,7 @@ impl Peers {
 	/// A peer implementation may drop the broadcast request
 	/// if it knows the remote peer already has the header.
 	pub fn broadcast_header(&self, bh: &core::BlockHeader) {
-		let num_peers = self.config.peer_min_preferred_count();
+		let num_peers = self.config.peer_min_preferred_outbound_count();
 		let count = self.broadcast("header", num_peers, |p| p.send_header(bh));
 		debug!(
 			"broadcast_header: {}, {} at {}, to {} peers, done.",
@@ -372,7 +373,8 @@ impl Peers {
 	/// A peer implementation may drop the broadcast request
 	/// if it knows the remote peer already has the transaction.
 	pub fn broadcast_transaction(&self, tx: &core::Transaction) {
-		let num_peers = self.config.peer_max_count();
+		let num_peers =
+			self.config.peer_max_inbound_count() + self.config.peer_max_outbound_count();
 		let count = self.broadcast("transaction", num_peers, |p| p.send_transaction(tx));
 		debug!(
 			"broadcast_transaction: {} to {} peers, done.",
@@ -543,14 +545,9 @@ impl Peers {
 		}
 	}
 
-	pub fn enough_peers(&self) -> bool {
-		self.peer_count() >= self.config.peer_min_preferred_count()
-	}
-
 	/// We have enough peers, both total connected and outbound connected
 	pub fn healthy_peers_mix(&self) -> bool {
-		self.enough_peers()
-			&& self.peer_outbound_count() >= self.config.peer_min_preferred_outbound_count()
+		self.peer_outbound_count() >= self.config.peer_min_preferred_outbound_count()
 	}
 
 	/// Removes those peers that seem to have expired
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -87,6 +87,10 @@ impl Server {
 					let peer_addr = PeerAddr(peer_addr);
 
 					if self.check_undesirable(&stream) {
+						// Shutdown the incoming TCP connection if it is not desired
+						if let Err(e) = stream.shutdown(Shutdown::Both) {
+							debug!("Error shutting down conn: {:?}", e);
+						}
 						continue;
 					}
 					match self.handle_new_peer(stream) {
@@ -197,7 +201,8 @@ impl Server {
 	/// Checks whether there's any reason we don't want to accept a peer
 	/// connection. There can be a few of them:
 	/// 1. Accepting the peer connection would exceed the configured maximum
-	/// allowed peer count.
+	/// allowed peer count. Note that seed nodes may wish to allow more than
+	/// the maximum configured connections to help with network bootstrapping.
 	/// 2. The peer has been previously banned and the ban period hasn't
 	/// expired yet.
 	/// 3. We're already connected to a peer at the same IP. While there are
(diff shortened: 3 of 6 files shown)


p2p/src/serv.rs
```diff
@@ -195,15 +195,23 @@ impl Server {
...
+	/// 2. The peer has been previously banned and the ban period hasn't
 	/// expired yet.
-	/// 2. We're already connected to a peer at the same IP. While there are
+	/// 3. We're already connected to a peer at the same IP. While there are
 	/// many reasons multiple peers can legitimately share identical IP
 	/// addresses (NAT), network distribution is improved if they choose
 	/// different sets of peers themselves. In addition, it prevent potential
 	/// duplicate connections, malicious or not.
 	fn check_undesirable(&self, stream: &TcpStream) -> bool {
+		if self.peers.peer_count() >= self.config.peer_max_count() {
```
hashmap: Unrelated to this change, but we always close connection before  returning true, what if we extract this logic to  the accept loop, where we call just `continue` atm?

DavidBurkett:
There's been lots of discussion about this in the past. See: https://github.com/mimblewimble/grin/issues/2726
https://github.com/mimblewimble/grin/pull/2725

I still believe that anytime we're dropping a new peer immediately after accepting the connection, it's imperative that we send a peer list first.

j01tz:
> I still believe that anytime we're dropping a new peer immediately after accepting the connection, it's imperative that we send a peer list first.

My initial thinking is that we want to consume as few resources as possible until we know the new connection is desirable (not from a banned peer, not from a known peer and not if it would exceed our maximum configured connection limit).

Otherwise it would allow "undesirable" connections to consume more resources than necessary (like sending peer lists), even after a ban. I don't think it makes sense to send a peer list to a banned peer for example?

DavidBurkett:
That's fine, we can check if they're banned before sending a peer list. But if we're disconnecting simply because we have too many incoming connections, we could have a real bootstrapping problem if we don't provide alternative peers.

j01tz:
> But if we're disconnecting simply because we have too many incoming connections, we could have a real bootstrapping problem if we don't provide alternative peers.

Agreed it would potentially impact bootstrapping ability with seed nodes in current form. 

Do you see this remaining an issue if the seed nodes don't maintain the proposed rule of shutting down new incoming connections while connection limit is reached without sending a peer list? 

Are there cases where non-seed nodes with default connection limits following this rule would likely impact bootstrapping?

p2p/src/serv.rs
j01tz: I think I addressed what you were referring with https://github.com/mimblewimble/grin/pull/2985/commits/a0145a5b83313e4cb082fe3b86a02a22533598a9. Please let me know if I'm mistaken, thanks

DavidBurkett:
Seed nodes would probably be enough, but what constitutes a seed node? We don't currently have a config setting for that, so we would have to add one.

j01tz:
After some more thought and further consideration of https://github.com/mimblewimble/grin/issues/2726:

Goals:
- limit accepting incoming connections (async cleanup not sufficient)
- allow network to grow even if all peers are already at max peer limit (maintain some flexibility)
- improve clarity around distinction between **incoming** and **outgoing** connections

Changes:
- replace `PEER_MAX_COUNT` with `PEER_MAX_INBOUND_COUNT` and `PEER_MAX_OUTBOUND_COUNT` to clarify and make configurable **incoming** and **outgoing** connection limits
- improve check at network level for tcp listener to avoid accepting too many **incoming** connections
    - include a buffer `PEER_LISTENER_BUFFER_COUNT` to exceed this limit if necessary, but not by much (default value is 8)
    - make this configurable for use with seed nodes etc
- increase minimum desired **outgoing** connections from 4 to 8
    - reduces chance of getting partitioned from the network
    - can improve block/tx propagation
    - makes eclipse attacks more expensive
- limit **outgoing** connections to 8
    - prevents exhausting potential **incoming** connection slots
        - not all nodes allow **incoming** connections
        - want to make good use of available **incoming** connections for the network
        - don't need to take up slots with many **outgoing** connections that could be used for **incoming**
        - reduces resources on nodes that don't allow **incoming** connections 

I'm updating the PR to [WIP] as this probably deserves more thought and testing, especially around limiting **outgoing** connections to 8.

The changes above may be too much and can be rolled back in favor of simply having the check in the listener with a configurable 8 peer buffer.

antiochp:
I actually like what is being proposed here. Making a more explicit distinction between _inbound_ and _outbound_ connection handling makes a lot of sense.

My only comment is this may be overkill short-term. We may want to tackle this is two stages - 
* this PR plus the proposed "8 peer temporary buffer"
* then tackle the larger inbound vs outbound suggestion proposed here

config/src/comments.rs
```diff
@@ -274,13 +274,19 @@ fn comments() -> HashMap<String, String> {
 #how long a banned peer should stay banned
 #ban_window = 10800
 
-#maximum number of peers
-#peer_max_count = 125
+#maximum number of inbound peer connections
+#peer_max_inbound_count = 117
+
+#maximum number of outbound peer connections
+#peer_max_outbound_count = 8
```
hashmap: 👍

config/src/comments.rs
```diff
@@ -274,13 +274,19 @@ fn comments() -> HashMap<String, String> {
 #how long a banned peer should stay banned
 #ban_window = 10800
 
-#maximum number of peers
-#peer_max_count = 125
+#maximum number of inbound peer connections
+#peer_max_inbound_count = 117
```
antiochp: We should just leave this as 125 or maybe even 128 for a nice round number.
It was arbitrary to begin with, definitely don't need to tweak them to add up to 125 in total.

p2p/src/types.rs
```diff
@@ -48,12 +48,22 @@ pub const MAX_LOCATORS: u32 = 20;
...
-/// The max peer count
-const PEER_MAX_COUNT: u32 = 125;
+/// The max inbound peer count
+const PEER_MAX_INBOUND_COUNT: u32 = 117;
+
+/// The max outbound peer count
+const PEER_MAX_OUTBOUND_COUNT: u32 = 8;
 
 /// min preferred peer count
 const PEER_MIN_PREFERRED_COUNT: u32 = 8;
```
antiochp: Should this one be inbound/outbound specific as well?

p2p/src/types.rs
```diff
@@ -268,11 +287,24 @@ impl P2PConfig {
 		}
 	}
 
-	/// return peer_max_count
+	/// return total maximum peer count (incoming + outgoing)
```
antiochp: Do we still need this total max count anywhere?

j01tz:
Changes in https://github.com/mimblewimble/grin/pull/2985/commits/7bfb54ecb7aba169d70a81f483684b03e63b77b2:
- increase `peer_max_inbound_count` from `117` to `128`
- remove `peer_min_preferred_count` in favor of `peer_min_preferred_outbound_count`
- consolidate redundant `enough_peers()` with `healthy_peers_mix()`
- remove `peer_max_count`
- better distinguish **outbound** connections in `p2p` and `servers`

Potential unresolved question:
- Do we want to broadcast compact blocks and transactions only to our outgoing connected peers (limited to 8) or temporarily exceed that limit to broadcast to all currently connected peers?
    - previously to broadcast we iterate through `connected_peers()` up to the `max_peer_count()` to broadcast to as many peers as we are connected to up to the maximum limit
    - this was kept the same in spirit so far, as the connection directionality is still ignored in `Peers::broadcast()` and we still attempt to broadcast up to the theoretical maximum connection limit

config/src/comments.rs
j01tz: Increased the limit to 128 here.

p2p/src/types.rs
j01tz: Removed `PEER_MIN_PREFERRED_COUNT` in favor of `PEER_MIN_PREFERRED_OUTBOUND_COUNT`

p2p/src/types.rs
j01tz: It was still used when broadcasting compact blocks and transactions for `Peers` but should be able to be removed in favor of making incoming and outgoing connections more explicit. Please see the unresolved question posted here https://github.com/mimblewimble/grin/pull/2985#issuecomment-521445528

DavidBurkett:
We definitely don't want to just broadcast to outgoing peers. Non-public nodes would never receive new blocks or transactions.

j01tz:
> We definitely don't want to just broadcast to outgoing peers. Non-public nodes would never receive new blocks or transactions.

That's where I ended up. I wasn't certain if nodes would request those on their own from peers or not (even if so the propagation delay could be ugly). It sounds like we definitely want to broadcast to all connected peers.

DavidBurkett:
I wouldn't say definitely broadcast to all. Figuring out the optimal number is more a graph theory problem. @tromp could probably tell us how to figure that out. We'd probably have to take into account expected percentage of public nodes, average number of incoming and outgoing peers per public node, and average number of outgoing peers per private node.

j01tz:
Agreed in the long term that a more complex mechanism will get us a more desirable result. My goal for this PR for now is to add a basic limit to incoming p2p connections by distinguishing incoming and outgoing connections without changing _too_ much. I think as is, the broadcasting should behave the same as before, but the question in general could probably use more thought.

antiochp:
> I think as is, the broadcasting should behave the same as before, but the question in general could probably use more thought.

Yeah I was actually wondering (briefly) about just broadcasting to outbound but then realized the same thing - some nodes only have outbound connections and they would never receive anything.

We should not distinguish between inbound/outbound when broadcasting (at least in the short-term) and just broadcast to max peers.

antiochp:
:+1:

config/src/comments.rs
```diff
@@ -274,12 +274,18 @@ fn comments() -> HashMap<String, String> {
...
-#maximum number of peers
-#peer_max_count = 125
+#maximum number of inbound peer connections
+#peer_max_inbound_count = 128
 
-#preferred minimum number of peers (we'll actively keep trying to add peers
-#until we get to at least this number
-#peer_min_preferred_count = 8
+#maximum number of outbound peer connections
+#peer_max_outbound_count = 8
```
DavidBurkett: What's the point of this? If this is greater than peer_min_preferred_outbound_count, do we continue to connect to more outbound peers?

servers/src/grin/seed.rs
```diff
@@ -300,11 +303,14 @@ fn listen_for_addrs(
 		return;
```
DavidBurkett: I think we can remove the `peers.peer_count() > peers.peer_outbound_count()` check on line 302, since `healthy_peers_mix()` covers it.

p2p/src/peers.rs
```diff
@@ -518,14 +545,9 @@ impl Peers {
 		}
 	}
 
-	pub fn enough_peers(&self) -> bool {
-		self.peer_count() >= self.config.peer_min_preferred_count()
-	}
-
 	/// We have enough peers, both total connected and outbound connected
 	pub fn healthy_peers_mix(&self) -> bool {
```
DavidBurkett: I'd advise changing the name of this now, since "healthy_peers_mix" doesn't really represent what it's checking. Possible options are something like `meets_preferred_outbound`, or preferably reverse the logic and call it `needs_outbound_connections`. The logic for managing connections was already built piecemeal through a series of defect corrections, so it's probably best to try to keep it from getting anymore complicated/confusing.

DavidBurkett:
> Agreed in the long term that a more complex mechanism will get us a more desirable result. My goal for this PR for now is to add a basic limit to incoming p2p connections by distinguishing incoming and outgoing connections without changing _too_ much. I think as is, the broadcasting should behave the same as before, but the question in general could probably use more thought.

It wouldn't be a more complex mechanism, I don't think. It's a complex formula to figure out by hand what the optimal number is, but then it's just a matter of broadcasting to n random peers. I'm ok with pushing this off though, just know that our network will be a whole lot noisier than it needs to be.

DavidBurkett:
I spoke with @tromp about this, and he found: https://www.quora.com/How-many-peers-is-a-Bitcoin-node-connected-to-at-all-times
It seems bitcoin chooses 8 random peers when broadcasting. I recall igno at one point also tossing out 8 as the magic number.

antiochp:
I think I originally introduced the "broadcast to 8 of our peers regardless of total peer connections" logic and only for headers. We never did this for transactions as far as I'm aware.

Currently we do the following - 
* headers - broadcast to peer_min_preferred_count
* compact blocks (initial broadcast from miner) - broadcast to peer_max_count
* transactions - broadcast to peer_max_count


But I think there is a lot of misinformation and misunderstanding out there on the internet regarding this. And some of the confusion comes from the inbound/outbound terminology. 
The "bitcoin node has 8 peers" thing is really old and I don't think its been like that for a long time. My understanding is bitcoin nodes connect to many more peers than 8.
And the documentation out there around exactly which peers get relayed to is confusing and ambiguous.

This documentation implies nodes send to _all_ their peers for block relay - 

https://bitcoin.org/en/p2p-network-guide#term-unsolicited-block-push
https://bitcoin.org/en/p2p-network-guide#term-standard-block-relay

And this documentations suggests transactions are sent to _all_ peer as well - 

https://en.bitcoin.it/wiki/Network#Standard_relaying



I'm not entirely sure which what the correct interpretation of bitcoin behavior is, but I'm less convinced that some magic 8 peers is used anywhere (even if we did this for header first broadcast, which I think may be wrong behavior).

DavidBurkett:
@antiochp Ok, I went to the actual source code, since documentation is inconsistent. Here's what I found.

 For compact blocks, looks like bitcoin actually only relays to up to 3 peers, but the whole system is different than ours (btc nodes request to receive compact blocks from specific peers). https://github.com/bitcoin/bips/blob/master/bip-0152.mediawiki and https://github.com/TheBlueMatt/bitcoin/blob/48efec82f3a18364c019577495914fcebb425e35/src/main.cpp#L470

```
* For nodes which have sufficient inbound bandwidth, sending a sendcmpct message with the first integer set to 1 to up to 3 peers is RECOMMENDED. If possible, it is RECOMMENDED that those peers be selected based on their past performance in providing blocks quickly (eg the three peers which provided the highest number of the recent N blocks the quickest), allowing nodes to receive blocks which come from those peers in only 0.5*RTT.
* Nodes MUST NOT send such sendcmpct messages to more than three peers, as it encourages wasting outbound bandwidth across the network.
```

For transactions, looks like it announces the availability of txs to all peers (similar to our tx kernel hash message): https://github.com/bitcoin/bitcoin/blob/67be6d7a177cb14ca8a68b9c6361dfaae9e64f2b/src/net_processing.cpp#L1311

It looks like it's the same for headers as well, but I could be reading the code wrong.

One difference we should note is bitcoin keeps track of each peer's "Inventory", so it doesn't send transactions and headers to nodes that are known to already have them. So based on all of this, my recommendation is:

* Continue broadcasting headers and peers to all nodes for now.
* In the future, add the ability to keep track of inventory of each connected peer
* Limit how many peers we request a compact block from (in cases where a well-connected node receives dozens of header messages at once)

antiochp:
> One difference we should note is bitcoin keeps track of each peer's "Inventory", so it doesn't send transactions and headers to nodes that are known to already have them.

We do something similar at the peer level, tracking hashes of blocks and transactions that have passed through that particular peer connection. It is implemented in the `TrackingAdapter` and tracks the 30 most recent hashes (tx and block). I believe this is pretty much entirely undocumented...

DavidBurkett:
I had no idea. Ok, seems like we're good then.

config/src/comments.rs
j01tz: `peer_max_outbound_count` lets us put a configurable limit on how many outgoing connections with peers we can have and by extension all connections. We continue to try to connect to more outbound peers until `peer_min_preferred_outbound_count` is reached (a "healthy peers mix"). By default these values are the same (we try to maintain connections on all available outgoing slots)

servers/src/grin/seed.rs
j01tz: That makes sense. They don't quite check the same thing but I think we only want the `healthy_peers_mix()` check here as you said. I'm pretty sure we don't care to check here that the total connected peers exceeds the outbound connected peers. This would require (I think) that we have inbound connections, and I'm pretty sure we want to `return` here any time our outbound connections exceeds our configured preferred minimum outbound connections ("a healthy number of outbound peers").

Removed in https://github.com/mimblewimble/grin/pull/2985/commits/fb9b06ecf6f86a8424aa92f7a991b8754a8a0302 unless there is a good reason to keep this as is.

config/src/comments.rs
DavidBurkett: But let's say you have `peer_min_preferred_outbound_count` of 8, and `peer_max_outbound_count` of 16, how would you ever end up with more than 8 connections?

p2p/src/peers.rs
j01tz: The naming here wasn't immediately intuitive to me either. My understanding is that it is a result of needing to maintain a healthy mix of outbound connections for dandelion to allow each peer to select a random outbound relay to send txs along the stem. Without ensuring a "healthy peers mix" apparently some nodes were not having enough slots for outbound connections.

With that context it makes a little more sense to me- a healthy peers mix is one
that has at least the minimum number of preferred outbound connections. So while I agree it doesn't strictly represent what is checked (outbound connections >= configured min preferred outbound connections), I think for now it does an ok job of capturing what we are checking for (that we have a healthy peers mix of outgoing connections).

DavidBurkett: Yea, I agree that's where the name came from, but you've now just changed the functionality, so my claim is the name no longer makes sense. Before, it used to compare inbound and outbound count so it made sense, but it's no longer about having a healthy mix, and is now just about making sure we have a minimum number of outbound.

config/src/comments.rs
j01tz: We may not after future refactoring to better clean up distinguishing incoming and outgoing connections. For now it happens when we broadcast compact blocks for example. There are probably other cases as well where outgoing connections are opened for one reason or another, exceeding the `peer_max_outbound_count` at least until `clean_peers()` is run to clean up any potential excess connections. In this case if `peer_max_outbound_count` was set to 16, I think we would hang on to those connections instead of dropping them when cleaning up peers.

DavidBurkett: 😕 You're saying we open outbound connections when broadcasting compact blocks? Where in the code does that take place, and what for?

j01tz: There are a few places in the code that will create temporary outgoing connections (I haven't verified it in broadcasting compact blocks in practice, but I think that is one area). You can set some test limits and watch logs to see the outgoing connections be temporarily exceeded. The hope of starting this process of distinguishing incoming and outgoing connections is to identify those places where we don't have full control over incoming/outgoing connections.

p2p/src/peers.rs
j01tz: Agreed, it could make sense to using a more descriptive name about functionality rather than a prescriptive name about health.

j01tz:
Made a few cleanups based on feedback and removed the [WIP] tag. I think if we want to make more substantial changes beyond here we may want to revert back to the listener check and buffer only for this PR and move the **incoming** and **outgoing** connection changes to a separate PR.

antiochp:
@j01tz This is good to go right? I'll merge if so.

j01tz:
Yes I think so, although haven't heard from @hashmap since he started testing it

hashmap:
Works on my machine!

antiochp:
Just merged this.
Checked out latest `master` and running a node locally on my laptop with default config.
I'm seeing it hit the DNS seeds but I'm not seeing my node actually connect to any peers.

Can someone else see if it works for them (from a fresh node, no local peers db)?

@hashmap @j01tz 

I'm wondering if this is an issue with current seeds not accepting connections currently (and not directly related to this PR).

antiochp:
Current `master` appears to work fine if the node already has peers.
Previous release appears to work fine from empty peer db.

So it _looks_ like there is an issue here with this PR merged to master where we are not successfully getting peer connections after hitting the initial DNS seed list.

antiochp:
Going to hold off opening an issue for this until we can verify this is actually a problem.
And we should probably revert these changes on master and investigate on a branch if we do verify this as an issue.

j01tz:
@antiochp has anything changed with the seeds? I've been testing this for a few days and only started seeing this issue today.

lehnberg:
can confirm that building with this PR included had me connecting to 0 peers, 
and reverting and building without the PR had me connected to 14 peers within a few seconds,
so it feels highly likely that it's something in the PR that is causing this.

hashmap:
Hm, this PR still works for me, got peers (I reverted the revert), but much slower than usual, with 10-20 secs delay. It feels like we are  slower in opening outbound connections, perhaps waiting for some delay.

antiochp:
See #2993 for proposed fix and explanation.

PR #2986, opened 2019-08-01
cleanup and simplify how we generate the genesis block in user_testing

antiochp:
I suspect some recent changes inadvertently broke "user testing" due to how we set the timestamp on the genesis block.
This PR simplifies genesis block creation in user testing. We just hard-code the timestamp now (we did before but at some point we started overwriting it with `now()`.

We now also just hard-code the nonce to `0` on both testing modes - it is fast to mine the dev genesis block in "user_testing" so just do that each time. 
To make this consistent we simply call `pow::mine_genesis_block()` when setting up the genesis block during user testing.

### DIFF
--- a/core/src/genesis.rs
+++ b/core/src/genesis.rs
@@ -22,7 +22,6 @@
 use chrono::prelude::{TimeZone, Utc};
 
 use crate::core;
-use crate::global;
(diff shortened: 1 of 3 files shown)


antiochp:
This affects _only_ automated tests and user testing and these all still _appear_ to work.
Merging.

PR #2989, opened 2019-08-08
Protect peer store with RwLock

bmhjyu:
Peer store is updated when a peer is connected. This actually can happen in different threads and may causes race conditions. Protecting it with RwLock will prevent the issue.


hashmap:
`PeerStore` is  a wrapper of `Store` which is a wrapper of `lmdb::Database`, protected by `RwLock`. Could you elaborate on how race condition could happen?

bmhjyu:
Apparently I misunderstood how it works. Thanks and I am closing this.

PR #2991, opened 2019-08-14
Fix doc grin <subcommand> help syntax.

Bedotech:
Fix wrong documentation, the subcommand implements the flag `--help` not `help`


antiochp:
Thanks for this!

PR #2993, opened 2019-08-21
improve checking for p2p connection limits (revised)

antiochp:
Replacement PR for #2985 which we reverted in https://github.com/mimblewimble/grin/commit/17dddeeb0d047e761996b42cd3ef544dd2725702

Tweaked the limit in terms of how many peer connections we _attempt_ regardless of our desired outbound connections.

The key point here is the vast majority of our peer addresses in the db are not publicly accessible.
So if we only try a random 8 then we stand a good chance of failing to connect to any new ones.

This is particularly important during the initial seed step on node startup - we want to try _all_ the seeds retrieved via dns, not just 8 of them. If these 8 fail for any reason then we end up stuck with our node not making any further attempts.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -274,12 +274,18 @@ fn comments() -> HashMap<String, String> {
 #how long a banned peer should stay banned
 #ban_window = 10800
 
-#maximum number of peers
-#peer_max_count = 125
+#maximum number of inbound peer connections
+#peer_max_inbound_count = 128
 
-#preferred minimum number of peers (we'll actively keep trying to add peers
-#until we get to at least this number
-#peer_min_preferred_count = 8
+#maximum number of outbound peer connections
+#peer_max_outbound_count = 8
+
+#preferred minimum number of outbound peers (we'll actively keep trying to add peers
+#until we get to at least this number)
+#peer_min_preferred_outbound_count = 8
+
+#amount of incoming connections temporarily allowed to exceed peer_max_inbound_count
+#peer_listener_buffer_count = 8
 
 # 15 = Bit flags for FULL_NODE
 #This structure needs to be changed internally, to make it more configurable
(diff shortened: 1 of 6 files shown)


j01tz:
My issue cleared up locally when I reverted https://github.com/mimblewimble/grin/pull/2985/commits/fb9b06ecf6f86a8424aa92f7a991b8754a8a0302

antiochp:
We were originally taking max peers (125) each time off the address queue and attempting to connect to them.

```
for addr in addrs.into_iter().take(p2p.config.peer_max_count() as usize) {
```

#2985 modified this to only take max outbound peers (8) each time.

```
for addr in addrs
	.into_iter()
	.take(p2p.config.peer_max_outbound_count() as usize)
```

So every 30s we were only attempting to connect to 8 new peers, with a high possibility that all or most of those were not publicly accessible.

This is why initial seeding sometimes failed completely and why it seems to be slower to get new peer connections.

This revised PR relaxes this rule and simply takes a max of 128 new peer addresses each loop, regardless of the inbound/outbound limits.

```
let max_outbound_attempts = 128;
for addr in addrs.into_iter().take(max_outbound_attempts) {
```

Testing seems to indicate this is significantly faster at filling up the default 8 outbound slots as we have a reasonable chance of finding publicly accessible nodes every loop now.

We also have a far more reliable initial seeding as we allow our node to attempt to connect to _all_ the seeds up front.

antiochp:
This config seems to be pretty stable behind a firewall, with my node looking (and finding a couple of new ones) for publicly accessible peers every 30s -

```
peer_max_inbound_count = 0
peer_max_outbound_count = 64
peer_min_preferred_outbound_count = 32
```

Takes a few minutes but its churning through the full set of peers in the db, marking large numbers as `defunct` - 

```
20190821 19:50:46.094 DEBUG grin_servers::grin::seed - monitor_peers: on 0.0.0.0:3414, 
  29 connected (26 most_work). all 4389 = 1584 healthy + 0 banned + 2805 defunct
```

Issue #2460, opened 2019-01-23
Look up kernel status


jaspervdm:
In order to prove a particular transaction has actually taken place, there are 2 things you could do:
- Look up status of transaction inputs/outputs
- Look up status of kernel

The former method is flawed, since outputs can be spent, re-used, cut-through etc. So really the only valid way would be the latter, to check if the kernel is included in a block.

Currently there is no API method to look up if a kernel is in a block or not. I am thinking of adding such a method. I think there are 2 possible approaches:
- Add an index for kernels
- Suggestion by @DavidBurkett on Gitter:
  > If you’re just talking a rust api, you probably would have to get the minimum height of the kernel, find the kernel mmr index of the block at that height, and I believe there’s a txhashset api to return all kernels after a specific mmr index

Which method do you think would be most desirable? Would it be a big issue to add an index, does it increase the storage requirements a lot?

Would really appreciate your input, before I start writing code that is not optimal.. Thanks!

antiochp:
We're going to need something like this for "relative kernel lock heights".

We've discussed different approaches to this a few times.
I proposed an index like you are suggesting and I think @ignopeverell shot this down pretty quickly.
Kernels are ~100bytes. The index would add an additional ~40 bytes per kernel (commitment + height).

So one approach may be (as @DavidBurkett suggests) - 
* find the pos of the kernel in the kernel MMR (use earliest one in case of a duplicate?)
* then search over the headers in the header MMR to find first header where kernel_mmr_size > pos

We do something similar in the existing `chain.get_header_for_output()` for that second step.

antiochp:
Thinking a bit more abut this - not sure full blocks actually gain us much for kernels.
Full blocks are useful for outputs as we have them in the blocks regardless of if they are subsequently spent. But for kernels we know we have them all in the kernel MMR anyway.

If I recall, the main argument against a kernel index is "most of the time we don't need it". Most kernels will simply be added to the kernel MMR and we have no need to go look them up afterwards.
For the few kernels where we _do_ care about their pos, we can go search for them as necessary.

This is in contrast to outputs, where we care about their pos when spent in a subsequent tx, i.e. we _always_ need to go look them up after adding to the MMR, and this makes the index a) more useful and b) worth the storage and maintenance costs.

DavidBurkett:
[...]

PR #3000, opened 2019-08-25
Kernel lookup api method

jaspervdm:
This PR adds an API call to look up an on-chain kernel and the height of the block it is included in.
It has an optional minimum search height. If provided (and close to the actual height), the request will be really fast (<10ms). Without this parameter the request can currently take up to 2.5s. Seeing as in the worst case scenario this time will increase linearly with the chain head height (which keeps increasing forever), it is recommended to always set the minimum height parameter.

Requesting review from @antiochp since I am not very experienced with MMRs and as a result some implementation details might be sub-optimal.

Closes https://github.com/mimblewimble/grin/issues/2460

__Use cases:__
- Wallets can check the status of a transaction without having to rely on the change output, which may or may not exist or might already be spent (https://github.com/mimblewimble/grin-wallet/issues/169)
- The on-chain kernel signature is required in some scriptless scripts applications

__Usage:__
`v1/chain/kernels/XXX?min_height=YYY&max_height=ZZZ` where `XXX` is the excess commitment. The `min_height` and `max_height` parameters are both optional. If not supplied, `min_height` will be set to 0 and `max_height` will be set to the head of the chain. The method will start at the block height `max_height` and traverse the kernel MMR backwards, until either the kernel is found or `min_height` is reached.

__Example:__
Request:
```
/v1/chain/kernels/0939fe3dc6a35350da91c6288138b7a257e0c0322eae30bda3938229d649e2e642?max_height=324300
```

Response:
```json
{
  "tx_kernel": {
    "features": {
      "Plain": {
        "fee": 12000000
      }
    },
    "excess": "0939fe3dc6a35350da91c6288138b7a257e0c0322eae30bda3938229d649e2e642",
    "excess_sig": "c6f269b98f611c03cde361467daa5c27109dea0e8b28a3189646a4e1e9ad96fe92ea86b65dc76151686c5a4f3922202bd298b665ec80ba2b1e43fb46df34f6a4"
  },
  "height": 324016,
  "mmr_index": 1946151
}
```

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -260,19 +260,25 @@ impl TxHashSet {
 			.elements_from_insertion_index(start_index, max_count)
 	}
 
-	/// Find a kernel with a given excess
-	pub fn find_kernel(&self, excess: &Commitment, start_index: u64) -> Option<(u64, TxKernel)> {
-		let last_pos = self.kernel_pmmr_h.last_pos;
-		let pmmr = ReadonlyPMMR::at(&self.kernel_pmmr_h.backend, last_pos);
+	/// Find a kernel with a given excess. Work backwards from `max_index` to `min_index`
+	pub fn find_kernel(
+		&self,
+		excess: &Commitment,
+		min_index: Option<u64>,
+		max_index: Option<u64>,
+	) -> Option<(TxKernel, u64)> {
+		let min_index = min_index.unwrap_or(1);
+		let max_index = max_index.unwrap_or(self.kernel_pmmr_h.last_pos);
 
-		let mut index = start_index;
-		while index <= last_pos {
+		let pmmr = ReadonlyPMMR::at(&self.kernel_pmmr_h.backend, self.kernel_pmmr_h.last_pos);
+		let mut index = max_index + 1;
+		while index > min_index {
+			index -= 1;
 			if let Some(t) = pmmr.get_data(index) {
 				if &t.kernel.excess == excess {
-					return Some((index, t.kernel));
+					return Some((t.kernel, index));
 				}
 			}
-			index += 1;
 		}
 		None
 	}
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -260,6 +260,23 @@ impl TxHashSet {
 			.elements_from_insertion_index(start_index, max_count)
 	}
 
+	/// Find a kernel with a given excess
+	pub fn find_kernel(&self, excess: &Commitment, start_index: u64) -> Option<(u64, TxKernel)> {
+		let last_pos = self.kernel_pmmr_h.last_pos;
+		let pmmr = ReadonlyPMMR::at(&self.kernel_pmmr_h.backend, last_pos);
+
+		let mut index = start_index;
+		while index <= last_pos {
+			if let Some(t) = pmmr.get_data(index) {
+				if &t.kernel.excess == excess {
+					return Some((index, t.kernel));
+				}
+			}
+			index += 1;
+		}
+		None
+	}
+
 	/// Get MMR roots.
 	pub fn roots(&self) -> TxHashSetRoots {
 		let header_pmmr =
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -26,6 +26,7 @@ use self::blocks_api::HeaderHandler;
 use self::chain_api::ChainCompactHandler;
 use self::chain_api::ChainHandler;
 use self::chain_api::ChainValidationHandler;
+use self::chain_api::KernelHandler;
 use self::chain_api::OutputHandler;
 use self::peers_api::PeerHandler;
 use self::peers_api::PeersAllHandler;
@@ -119,7 +120,9 @@ pub fn build_router(
 	let output_handler = OutputHandler {
 		chain: Arc::downgrade(&chain),
 	};
-
+	let kernel_handler = KernelHandler {
+		chain: Arc::downgrade(&chain),
+	};
 	let block_handler = BlockHandler {
 		chain: Arc::downgrade(&chain),
 	};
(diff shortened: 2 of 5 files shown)


DavidBurkett:
Why not index these in lmdb?

antiochp:
> Why not index these in lmdb?

We do this for outputs. There is also a related PR https://github.com/mimblewimble/grin/pull/2903 that proposes storing the height per output in lmdb.

We came to the conclusion in previous conversations that - 
* we need fast random access for _all_ outputs (node needs to verify spent outputs for every tx/block)
* we do not necessarily need fast random access for  _all_ kernels
    * a node only cares about a limited subset of kernels (based on connected wallets)
    * relative kernels and scriptless script use cases are assumed to be relatively rare

So given the tradeoff between storage costs for lots of kernels and slower lookups - we erred on the side of slower lookups.

We are comfortable maintaining an index of all _unspent_ outputs in lmdb as the UTXO set will not grow indefinitely.
We are less comfortable indexing all kernels as this set will grow indefinitely (and vast majority of this index will never be accessed or used).

That said - a given node probably has a reasonable idea of which kernels it is interested in locally. so it may make sense to store/cache these in lmdb based on prior access.
Or maybe it makes more sense to cache these at the wallet level (in the wallet lmdb instance).

antiochp:
Thinking a bit more about this. I'm torn about the "index or not to index" question.

yeastplume:
Couple of things (that ties in with a few thoughts I've had on wallet improvements)

1) Caching a min height in a wallet along with each transaction should be trivial, but this will obviously only be useful where the tx log and data is intact. (Think something will have to be done here anyhow as I don't think it's currently storing kernel IDs in a handy place). When restoring, checking the chain, etc, the wallet should still be retrieving by output anyhow so shouldn't make much of a difference.
2) Another approach to the original issue is to simply have the wallet scan outputs as part of normal operation (i.e. perform the `wallet_check` command, like most other wallets do for their chains). Wallets could store a birthdate block, keep track of the last block scanned between operations and start checks from last_block - sync horizon). I think this will be needed at some point anyhow for 100% wallet data robustness, so we will have to figure out how to keep this operation efficient.

jaspervdm:
@antiochp Hmm I hadn't considered the relative kernel locks usecase. But even then increasing the storage requirement of kernels by ~40% might be a lot, especially since we can never prune them.

Do you want me to already accomodate for this usecase? That means moving some of the logic into the `Chain`, and probably providing a flag that allows to scan backwards in the MMR instead of forwards.

@yeastplume Right, the kernel lookup would only be used for unconfirmed transactions, not for restores as they happen on the output level rather than the tx level.
> Another approach to the original issue is to simply have the wallet scan outputs as part of normal operation

Yes, I think incremental scanning of the utxo set is a good idea anyway. But I'm not sure if checking the tx confirmation status by looking at the presence of the inputs in the utxo set is ideal. If the user is using the seed on multiple different wallets at the same time, it could happen that the wrong tx gets marked as confirmed.

yeastplume:
Yep, agreed that kernel lookup is the best way to confirm transactions, we can get this implemented then look at incremental scanning as a near-term improvement.

antiochp:
> Do you want me to already accomodate for this usecase? That means moving some of the logic into the Chain, and probably providing a flag that allows to scan backwards in the MMR instead of forwards.

Do we want to always scan backwards through the MMR? Presumably if we are confirming txs they will be younger txs, present in more recent blocks?

antiochp:
> But even then increasing the storage requirement of kernels by ~40% might be a lot, especially since we can never prune them.

I'm not sure I follow what you are saying here - you mean in terms of the lmdb index/storage (if we stored the index in the db)?

jaspervdm:
> Do we want to always scan backwards through the MMR? Presumably if we are confirming txs they will be younger txs, present in more recent blocks?

Makes sense, will change it.

jaspervdm:
> I'm not sure I follow what you are saying here - you mean in terms of the lmdb index/storage (if we stored the index in the db)?

Yes if we store an index we'd need 40% more storage per kernel.

Thinking about it some more, a malicious user could significantly slow down initial sync by creating lots of relative lock height kernels that rely on very old kernels. What if we only create the index during initial sync and delete it afterwards?

antiochp:
Yeah I think there may be reasons why we would want to temporarily create an index during initial sync. But I don't think we want to assume this index is available for general usage after sync.

I would not be overly concerned about supporting relative kernels today - we can optimize initial fast sync later as you suggest.

We don't want to add 40% to the permanent storage costs today in anticipation of something that may be very rare.

garyyu:
> Thinking a bit more about this. I'm torn about the "index or not to index" question.

Same here :-) My left hand vote Yes for index but right hand vote for No index 😸 

There could be a middle way on this:
- Use index for recent kernels (pruning those indexes when blocks pruned, and no pruning if it's an archived mode node).
- Use slow searching for far old kernels (if the estimated height is beyond the `STATE_SYNC_THRESHOLD` and not found in indexed kernels).

The fast indexed kernel querying will cover 99% transaction status querying requirement, since in most cases people will query status immediately (or at least in two days) after a transaction posted.

(As reference, here is my trying on a pure indexing solution: https://github.com/gottstech/grin/pull/3 and https://github.com/gottstech/grin/pull/13. if working as a wallet service backend, the config item `pruning_kernel_index` can be changed as  `false`.)

antiochp:
@jaspervdm Is this ready for final review?

jaspervdm:
@antiochp I think it is! I just updated the description to reflect the latest changes.

chain/src/txhashset/txhashset.rs
```diff
@@ -260,6 +260,29 @@ impl TxHashSet {
 			.elements_from_insertion_index(start_index, max_count)
 	}
 
+	/// Find a kernel with a given excess. Work backwards from `max_index` to `min_index`
```
antiochp: This might make more sense in the `rewindable_kernel_view` (rather than the `txhashset`).
Maybe a separate PR to refactor this.

antiochp:
Looks pretty fast to me. This is the coinbase kernel for the genesis block - 

```
time curl "http://127.0.0.1:3413/v1/chain/kernels/096385d86c5cfda718aa0b7295be0adf7e5ac051edfe130593a2a257f09f78a3b1"
{
  "tx_kernel": {
    "features": "Coinbase",
    "excess": "096385d86c5cfda718aa0b7295be0adf7e5ac051edfe130593a2a257f09f78a3b1",
    "excess_sig": "142a6a482a0c64ba71ba216ba5361612696fc76fe8d5c03c79fae01cab29d050ed7567e9a6d4fbbc1b6c0bf4434b8450dafa3f29b31612fda815f6e4b2bcfd43"
  },
  "height": 0,
  "mmr_index": 1
}
curl   0.01s user 0.01s system 1% cpu 1.049 total
```

PR #3001, opened 2019-08-26
cleanup rebuild sync MMR logic

antiochp:
This PR cleans up (and optimizes) how we rebuild the sync MMR during node startup.

We do not need to truncate the sync MMR and rebuild it from the genesis header, we can simply "rewind and apply header fork" based on current sync MMR state and the current head of the header chain. In the common case we will not need to rewind any headers and we can simply "fast forward" the recent headers to allow the sync MMR to catch up to the header MMR.

This saves a significant amount of time when starting a node up if shutdown overnight for example.
Rebuilding a sync MMR from scratch was taking 5s or so on a reasonably fast laptop.

As part of this we have now made the "reset head" logic more explicit, passing in the new head updating as necessary.

This PR also cleans up and simplifies how we apply headers to the MMR when processing, either as part of sync or when processing "header first". This makes processing new headers and processing headers from the local db as part of a fork more consistent.

Previous behavior -
* rewind MMR to fork point if necessary
* apply headers to MMR up to and including previous header
* apply new header to MMR
* validate and store new header in the db

Simplified behavior -
* validate and store new header in the db
* rewind MMR to fork point if necessary
* apply headers to MMR up to and including new header

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1423,9 +1423,8 @@ fn setup_head(
 			head.last_block_h,
 			head.height,
 		);
-		batch.reset_header_head()?;
-		let header_head = batch.header_head()?;
-		batch.save_sync_head(&header_head)?;
+		batch.save_header_head(&head)?;
+		batch.save_sync_head(&head)?;
 	}
 
 	batch.commit()?;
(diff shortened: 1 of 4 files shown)


antiochp:
Closing as this is rolled up in #3004

PR #3007, opened 2019-08-29
verify the ignored fee is still 0

antiochp:
Resolves #2859.

While we continue to support kernel features using protocol version 1 - 
* read but ignore fee on coinbase kernels
* but verify these ignored bytes are indeed 0


antiochp:
@garyyu @tromp This is all that should be required here right?
When we read the bytes we still discard the fee and lock_height if not applicable, but we need to verify the bytes are all 0 before we discard them.

PR #3014, opened 2019-09-01
rewording

jackzhp:
I think it is better


quentinlesceller:
Hi @jackzhp thanks for the PR. However, I don't think this needs a PR by itself (considering it's just for a word and doesn't really change the meaning of the sentence).

PR #3019, opened 2019-09-03
[WIP] Chain recovery mechanism

### LINKED ISSUE #3018: Allow nodes to (manually) recover from dreaded "Already Spent" error
We still have scenarios and edge cases where we see the dreaded "Already Spent" error in the logs and a node fails to accept new blocks.

Assuming this is not a double spend attempt this indicates the local UTXO set was corrupted somehow. The UTXO set consists of - 

* the output MMR (the TXO set)
* the output leafset (which leaf pos are unspent in the output MMR)

Related #3016.

It would be _really_ useful if a node could recover from this somehow.
Particularly a node being used to test dev branches etc (where we are maybe more susceptible to borking the data).

Potential steps to recover - 

* select a new "horizon" to rewind to (say 1440 blocks earlier)
* request a fresh leafset from a peer for this horizon
* rewind the txhashset back to this horizon and drop the leafset in place
* sync blocks from this horizon

aka Minimal fast-sync given preexisting txhashet MMRs and header MMRs

----

This would not need to be a realtime or automatic resolution. Node operators could explicitly start a node in recovery mode or something similar - this should primarily be used by devs/qa messing with code on branches etc.

This should _not_ be something that triggers automatically on regular nodes (risking bigger underlying issues getting hidden and not surfacing).

antiochp:
Resolves #3018.

Leverages our "snapshot" functionality used during fast sync to allow a local node to revert to a prior chain state and resync full blocks from that point.

Our UTXO set is relatively fragile. It is derived from a pruned output and rangeproof MMR and a corresponding "leaf set" defining which MMR positions are non-removed leaves.

If a single MMR pos is incorrectly marked as removed or non-removed the UTXO set is invalid (for obvious reasons). __Anything__ that results in the leafset getting out of sync with the MMR results in an invalid UTXO set. 

A leafset file is not "rewindable" as it does not track history, it is simply a snapshot in time.

But - we can have multiple snapshots (representing previous UTXO state) present on disk at any given point in time.

This allows us to select a previous "snapshot", revert the UTXO set to be consistent with that snapshot and resolve any "Already Spent" errors resulting from a set of corrupted txhashset backend files.

This PR is a "proof of concept" that we can robustly revert to any previous chain state as long as -
* we have the necessary snapshot files for this header
* the snapshot is not too old (cannot be beyond the 7 day prune/compaction horizon)

TODO - 
- [x] Take a snapshot of the leafset files (UTXO set) on -
  - [x] successful startup 
  - [x] daily compaction
- [ ] Remove hardcoded block hash...
- [ ] Add support for `--recover <hash>` flag on `grin server run`
- [ ] More testing in various scenarios

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -165,16 +165,21 @@ impl Chain {
 		db_root: String,
 		adapter: Arc<dyn ChainAdapter + Send + Sync>,
 		genesis: Block,
+		header: Option<&BlockHeader>,
 		pow_verifier: fn(&BlockHeader) -> Result<(), pow::Error>,
 		verifier_cache: Arc<RwLock<dyn VerifierCache>>,
 		archive_mode: bool,
 	) -> Result<Chain, Error> {
 		let store = Arc::new(store::ChainStore::new(&db_root)?);
 
-		// open the txhashset, creating a new one if necessary
-		let mut txhashset = txhashset::TxHashSet::open(db_root.clone(), store.clone(), None)?;
+		let hash = Hash::from_hex("0000018c118c48d32fecb3e586d5a053f18d82963def6a0cc7f4d1d569fa4186").unwrap();
+		let foo_header = store.get_block_header(&hash)?;
+		let header = Some(&foo_header);
 
-		setup_head(&genesis, &store, &mut txhashset)?;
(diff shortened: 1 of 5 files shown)


antiochp:
Closing for now. Needs some more thinking through.

PR #3023, opened 2019-09-06
Verify headers and blocks only when needed

### LINKED ISSUE #1642: Do not parse block body before header has been validated
The current code and data structures read a full block all at once. This could make a short-lived (because banning would still be relatively quick) DDoS attack.

hashmap:
Currently we have some lightweight validation implemented as part of
entity deserialization, which is safer and allows us to not parse the
entire object if some part is invalid. At the same time this logic
always applies when we read an entity, eg when reading from DB.

This PR introduces UntrustedHeader/Block which is used when we read from
the network. It does partial validation during read, then it is supposed
to be converted into regular header/block which doesn't validate itself.

Also this PR adds "lightweight" validation to block header read like we have
for block body, so we don't parse block body if the header is invalid.

Fixes #1642

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -25,8 +25,6 @@ use crate::store;
 use crate::txhashset;
 use crate::types::{Options, Tip};
 use crate::util::RwLock;
-use chrono::prelude::Utc;
-use chrono::Duration;
 use grin_store;
 use std::sync::Arc;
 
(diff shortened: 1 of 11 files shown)


core/src/core/block.rs
```diff
@@ -289,48 +296,52 @@ impl Writeable for BlockHeader {
...
+		|| timestamp < MIN_DATE.and_hms(0, 0, 0).timestamp()
+	{
+		return Err(ser::Error::CorruptedData);
+	}
+
+	// Check the block version before proceeding any further.
+	// We want to do this here because blocks can be pretty large
+	// and we want to halt processing as early as possible.
+	// If we receive an invalid block version then the peer is not on our hard-fork.
+	if !consensus::valid_header_version(height, version) {
```
antiochp: Is the `valid_header_version` check worth moving into `UntrustedBlockHeader`?
It takes a read lock on the global CHAIN_TYPE internally and we presumably don't need to do this check every time we read a header from the db.

hashmap: Sounds reasonable, will move it there

hashmap: done

hashmap:
> Actually the _only_ time we do anything with compact blocks is when we receive them via p2p so maybe these are always untrusted?

I ended up having an additional Untrusted type, for symmetry, also some code relies on the existing structure of CompactBlock, so it's cleaner to read UntrustedCompactBlock and then convert it

core/fuzz/Cargo.lock
```diff
@@ -365,7 +365,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "grin_core"
-version = "2.1.0-beta.1"
+version = "3.0.0-alpha.1"
```
antiochp: These changes should not be necessary on `master`.

core/src/core/block.rs
```diff
@@ -397,6 +400,59 @@ impl BlockHeader {
...
+	}
+}
+
+/// Block header which does lightweight validation as part of deserialization,
+/// it supposed to be used when we can't trust the channel (eg network)
+pub struct UntrustedBlockHeader(BlockHeader);
+
+/// Deserialization of an untrusted block header
+impl Readable for UntrustedBlockHeader {
+	fn read(reader: &mut dyn Reader) -> Result<UntrustedBlockHeader, ser::Error> {
```
antiochp: Wondering if we never actually pass an `UntrustedBlockHeader` around what this might look like if we modified this signature to actually return the trusted version once we had successfully read it - 

```
impl Readable for UntrustedBlockHeader {
	fn read(reader: &mut dyn Reader) -> Result<BlockHeader, ser::Error> {
        ...
        }
    ...
}
```

I assume we cannot do this easily because of how `Readable` is defined but it would potentially allow us to not need the additonal explicit `.into()` to do conversions later.

Maybe being explicit like this is better though.

chain/src/pipe.rs
```diff
@@ -314,35 +312,6 @@ fn prev_header_store(
...
-		// TODO add warning in p2p code if local time is too different from peers
-		return Err(ErrorKind::InvalidBlockTime.into());
-	}
-
-	if !ctx.opts.contains(Options::SKIP_POW) {
-		if !header.pow.is_primary() && !header.pow.is_secondary() {
-			return Err(ErrorKind::LowEdgebits.into());
-		}
-		let edge_bits = header.pow.edge_bits();
-		if !(ctx.pow_verifier)(header).is_ok() {
```
antiochp: @hashmap Where does this check happen now?

antiochp: The only place I see similar code is in `validate_pow_only()` but this is only used for orphan blocks.

PR #3031, opened 2019-09-08
fix: body sync seamless infinite loop on 'no pending block request, asking more'

garyyu:
In some situations, the `block_sync` will request 1 block only; and in exception case, it will request none block.  These will trigger the seamless infinite loop of `body_sync`.

The typical log in this case will look like:
```sh
 20190907 20:26:38.032 DEBUG grin_servers::grin::sync::body_sync - body_sync: no pending block request, asking more
 20190907 20:26:38.032 DEBUG grin_chain::chain - body_sync: body_head - 000002121a63, 336847, header_head - 000000606f52, 336853, sync_head - 000000606f52, 336853
 20190907 20:26:38.043 DEBUG grin_servers::grin::sync::body_sync - body_sync: no pending block request, asking more
 20190907 20:26:38.043 DEBUG grin_chain::chain - body_sync: body_head - 000002121a63, 336847, header_head - 000000606f52, 336853, sync_head - 000000606f52, 336853
 20190907 20:26:38.056 DEBUG grin_servers::grin::sync::body_sync - body_sync: no pending block request, asking more
      ... ...
      # infinite logs like that at about every 10 ms, and never stop
```

The root cause is that we have a condition `<2` here: 
```Rust
		// off by one to account for broadcast adding a couple orphans
		if self.blocks_requested < 2 {
			debug!("body_sync: no pending block request, asking more");
			return Ok(true);
		}
```

I can't remember why @ignopeverell wrote this `off by one to account for broadcast adding a couple orphans`, but looks like it's better to give a simple logic here to judge 0, and leaving the orphan cases to be handled by the 1 second timeout:
```Rust
	// some blocks have been requested
	if self.blocks_requested > 0 {
		// but none received since timeout, ask again
		let timeout = Utc::now() > self.receive_timeout;    # << here is the 1 second timeout I said above
		if timeout && blocks_received <= self.prev_blocks_received {
			debug!(
				"body_sync: expecting {} more blocks and none received for a while",
				self.blocks_requested,
			);
			return Ok(true);
		}
	}
```

### DIFF
--- a/servers/src/grin/sync/body_sync.rs
+++ b/servers/src/grin/sync/body_sync.rs
@@ -176,9 +176,8 @@ impl BodySync {
 			self.prev_blocks_received = blocks_received;
 		}
 
-		// off by one to account for broadcast adding a couple orphans
-		if self.blocks_requested < 2 {
-			// no pending block requests, ask more
+		// no pending block requests, ask more
+		if self.blocks_requested == 0 {
 			debug!("body_sync: no pending block request, asking more");
 			return Ok(true);
 		}


garyyu:
Oops, this can only solve the issue when `blocks_requested` is 1, and can't help if nothing is requested in exception case.  Need more thinking about how to fix that, or it could not deserve a "fix" at all.

Issue #3030, opened 2019-09-08
Node out of sync and can't recover


garyyu:
A long standing bug come up to the surface.

If the Body Head happen to be a fork, there's no chance to recover by itself except removing the whole `chain_data` and restarting the node.

```Rust
DEBUG grin_servers::common::hooks - Received block header 000002121a63 at 336847 from 139.162.168.18:3414, going to process.
DEBUG grin_chain::pipe - header_head updated to 000002121a63 at 336847

DEBUG grin_chain::pipe - pipe: process_block 000002121a63 at 336847 [in/out/kern: 3/5/3]
DEBUG grin_chain::pipe - head updated to 000002121a63 at 336847

    # now the Body Head is on the fork block 336847@000002121a63
    # then it will reject all the new blocks because it didn't see the main chain block 336847@0000018e109d, before it start a new Syncing.

DEBUG grin_servers::common::hooks - Received block header 000000393005 at 336848 from 139.162.168.18:3414, going to process.
DEBUG grin_servers::common::adapters - Block header 000000393005 refused by chain: StoreErr(NotFoundErr("BLOCK HEADER: 0000018e109d"), "NotFoundErr(\"BLOCK HEADER: 0000018e109d\")")

DEBUG grin_servers::common::hooks - Received block header 000002a11b76 at 336849 from 139.162.168.18:3414, going to process.
DEBUG grin_servers::common::adapters - Block header 000002a11b76 refused by chain: StoreErr(NotFoundErr("BLOCK HEADER: 000000393005"), "NotFoundErr(\"BLOCK HEADER: 000000393005\")")

DEBUG grin_servers::common::hooks - Received block header 000020b1eb94 at 336850 from 139.162.168.18:3414, going to process.
DEBUG grin_servers::common::adapters - Block header 000020b1eb94 refused by chain: StoreErr(NotFoundErr("BLOCK HEADER: 000002a11b76"), "NotFoundErr(\"BLOCK HEADER: 000002a11b76\")")

        ... ...

DEBUG grin_servers::common::hooks - Received block header 000000606f52 at 336853 from 217.8.61.200:3414, going to process.
DEBUG grin_servers::common::adapters - Block header 000000606f52 refused by chain: StoreErr(NotFoundErr("BLOCK HEADER: 000000d2f919"), "NotFoundErr(\"BLOCK HEADER: 000000d2f919\")")

    # now, it enable the Syncing as designed.

[...]

PR #3033, opened 2019-09-09
fix: find common header between the header chain and the current body chain

garyyu:
Fix of #3030 

The current `is_on_current_chain()` function has an implicit meaning of **current `header` chain**, don't remember when we did this change, anyway it impacts the logic here to find the header common between the header chain and the current body chain.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -791,6 +791,7 @@ impl Chain {
 		let mut oldest_height = 0;
 		let mut oldest_hash = ZERO_HASH;
 
+		// Start with body_head (head of the full block chain)
 		let mut current = self.get_block_header(&body_head.last_block_h);
 		if current.is_err() {
 			error!(
@@ -804,6 +805,9 @@ impl Chain {
 		// TODO - Investigate finding the "common header" by comparing header_mmr and
 		// sync_mmr (bytes will be identical up to the common header).
 		//
+		// Traverse back through the full block chain from body head until we find a header
+		// that "is on current chain", which is the "fork point" between existing header chain
+		// and full block chain.
 		while let Ok(header) = current {
 			// break out of the while loop when we find a header common
 			// between the header chain and the current body chain
@@ -816,15 +820,16 @@ impl Chain {
 			current = self.get_previous_header(&header);
 		}
 
-		// Go through the header chain reversely to get all those headers after oldest_height.
+		// Traverse back through the header chain from header_head back to this fork point.
+		// These are the blocks that we need to request in body sync (we have the header but not the full block).
 		if let Some(hs) = hashes {
-			let mut current = self.get_block_header(&header_head.last_block_h);
-			while let Ok(header) = current {
+			let mut h = self.get_block_header(&header_head.last_block_h);
+			while let Ok(header) = h {
 				if header.height <= oldest_height {
 					break;
 				}
 				hs.push(header.hash());
-				current = self.get_previous_header(&header);
+				h = self.get_previous_header(&header);
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


antiochp:
> The current is_on_current_chain() function has an implicit meaning of current header chain, don't remember when we did this change, anyway it impacts the logic here to find the header common between the header chain and the current body chain.

I don't fully understand what you mean here by "implicit meaning of current header chain".
This is the _only_ way of knowing if we are on the current chain or not based on our header MMR.

We check for `is_on_current_chain` with the following logic - 

* put header MMR in correct state via `rewind_and_reapply_header_fork`
  * header MMR data file now contains sequential headers up to and including current height
  * this may have diverged from the files on disk if this is a fork (we only persist the most work chain)
* for the header in question
  * take its height and lookup the header in our MMR at this height
  * if its the same header its on the "current" chain
  * otherwise it is on a different fork

i.e. If block `abc` has height 1,000 and on the current chain the header at height 1,000 is `abc` then we know `abc` is on the current chain.
Which feels a little circular but if we find block `def` at height 1,000 then we know `abc` cannot also be at height 1,000 and _must_ be on a different fork.

chain/src/chain.rs
```diff
@@ -791,11 +791,11 @@ impl Chain {
 		let mut oldest_height = 0;
 		let mut oldest_hash = ZERO_HASH;
 
-		let mut current = self.get_block_header(&header_head.last_block_h);
```
antiochp: I don't think this change makes sense.
We are checking to see if a `header` exists in the db and we should do this based on the current `header_head` which is the tip of the current header chain.

We are in `check_txhashset_needed` which (confusingly) returns a vec of block hashes to go request (if we do not need to request a txhashset archive).
So we need to determine which blocks we do not have, based on the headers we do have. So we should be looking at the headers in both our db and our header MMR. Headers up to and including the current tip of the header chain (`header_head`).

chain/src/chain.rs
```diff
@@ -807,34 +807,41 @@ impl Chain {
...
-			oldest_hash = header.hash();
-			if let Some(hs) = hashes {
-				hs.push(oldest_hash);
-			}
 			current = self.get_previous_header(&header);
 		}
 
+		// Go through the header chain reversely to get all those headers after oldest_height.
+		if let Some(hs) = hashes {
+			let mut current = self.get_block_header(&header_head.last_block_h);
```
antiochp: I see we are redefining current in this PR here, but I don't think we need to do this. Earlier this is `body_head` and now this is redefined based on `header_head`. 

I think it should consistently be based on a header from the db starting at `header_head` and working back until we find one that also refers to a full block in our db.

chain/src/chain.rs
```diff
@@ -791,11 +791,11 @@ impl Chain {
 		let mut oldest_height = 0;
 		let mut oldest_hash = ZERO_HASH;
 
-		let mut current = self.get_block_header(&header_head.last_block_h);
+		let mut current = self.get_block_header(&body_head.last_block_h);
```
antiochp: Lets rename _this_ usage of current to make it clearer that we want to check for headers in our db based on `body_head`.

This is a 2 step process.

* start with `head` (head of the full block chain)
  * if we do not have a _full block_ in the db for this `head` then its an error
  * if we do not have the header in the db for this `head` then its an error
* traverse back through the full block chain from `head` until we find a header that "is on current chain"
  * this is our "fork point" between existing header chain and full block chain
* now we want to traverse back through our header chain from `header_head` back to this fork point
  * these are the blocks that we need to request (we have the header but not the full block)

garyyu: I put some of yours comments to the source code comments, and renamed the 2nd `current`. 
pls let me whether it's better now.

PR #3035, opened 2019-09-10
Optimize POW read 

hashmap:
This functionality is used to deserialize header (from network or from
DB), it was taking up to 40% cpu time during initial header sync. This
PR brings it down to 3-4%. Function read_number would look better as
closure, unfortunately the compliler doesn't inline it in this case, so
it would be 2x slower.

Before:
Screenshot 2019-09-07 at 00 15 06

after:
Screenshot 2019-09-07 at 00 07 15

### DIFF
--- a/core/tests/block.rs
+++ b/core/tests/block.rs
@@ -269,7 +269,7 @@ fn empty_block_serialized_size() {
 	let b = new_block(vec![], &keychain, &builder, &prev, &key_id);
 	let mut vec = Vec::new();
 	ser::serialize_default(&mut vec, &b).expect("serialization failed");
-	let target_len = 1_111;
+	let target_len = 1_112;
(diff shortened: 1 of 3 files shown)


core/tests/block.rs
```diff
@@ -269,7 +269,7 @@ fn empty_block_serialized_size() {
 	let b = new_block(vec![], &keychain, &builder, &prev, &key_id);
 	let mut vec = Vec::new();
 	ser::serialize_default(&mut vec, &b).expect("serialization failed");
-	let target_len = 1_107;
```
antiochp: Are these different because of the updated `AUTOMATED_TESTING` constants above?

hashmap: Yes, this method has a limitation - a compressed POW must be at least 64 bits long because we read into u64, our test parameters were too low.

PR #3036, opened 2019-09-10
Optimize Option to Error conversion

hashmap:
To convert option to error we generate an error message. In some places
it contains header or block hash code or other data which is costly to
produce. So during the initial header sync we spend 12% of all time on
generating those messages (in 99% cases we don't use it). This PR
introduces a lazy generation of error messages which completely
eliminates CPU load during the header sync.

Before:
Screenshot 2019-09-10 at 12 33 59

after:
Screenshot 2019-09-10 at 11 54 24

### DIFF
--- a/store/src/lmdb.rs
+++ b/store/src/lmdb.rs
@@ -54,9 +54,12 @@ impl From<lmdb::error::Error> for Error {
 }
 
 /// unwraps the inner option by converting the none case to a not found error
-pub fn option_to_not_found<T>(res: Result<Option<T>, Error>, field_name: &str) -> Result<T, Error> {
+pub fn option_to_not_found<T, F>(res: Result<Option<T>, Error>, field_name: F) -> Result<T, Error>
+where
+	F: Fn() -> String,
+{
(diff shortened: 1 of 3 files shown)


store/src/lmdb.rs
```diff
@@ -54,9 +54,12 @@ impl From<lmdb::error::Error> for Error {
...
 
 /// unwraps the inner option by converting the none case to a not found error
-pub fn option_to_not_found<T>(res: Result<Option<T>, Error>, field_name: &str) -> Result<T, Error> {
+pub fn option_to_not_found<T, F>(res: Result<Option<T>, Error>, field_name: F) -> Result<T, Error>
+where
+	F: Fn() -> String,
+{
 	match res {
-		Ok(None) => Err(Error::NotFoundErr(field_name.to_owned())),
+		Ok(None) => Err(Error::NotFoundErr(field_name())),
```
garyyu: 👍

quentinlesceller:
Nice find!

PR #3045, opened 2019-09-17
Read header_head and sync_head from header MMRs directly

antiochp:
~__Will hold of merging this until 2.1.0 is officially released__~

We cannot revert easily to a prior version of the code and reuse an existing db with these changes.

----

We separated the header MMR from the txhashset output|rangeproof|kernel trifecta in #3004.

We now reliably keep the header MMR aligned with the most work header.
Similarly we keep the sync MMR aligned with the most work header during header sync.

There is no longer any need to maintain `header_head` and `sync_head` entries in the local db.
The `header_head` should always be aligned with the rightmost entry in the header MMR.
Similarly for `sync_head` and the rightmost entry in the sync MMR.

If we simply take the header MMR and read the rightmost entry we can determine the current `header_head`.

This gives us one less thing to keep synchronized when flushing updates to disk (across both the MMR files and the db files).

This will be marginally more expensive to read `header_head` as this now involves a single db read and a single MMR read (was previously just a db read). But this is worth it as we no longer need to write an updated `header_head` every time we apply a new header to the MMR.
The MMR files are also very fast to read from as they are mmap'd.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -201,7 +201,6 @@ impl Chain {
 			&mut sync_pmmr,
 			&mut txhashset,
 		)?;
-		Chain::log_heads(&store)?;
 
 		let chain = Chain {
 			db_root,
@@ -226,6 +225,8 @@ impl Chain {
 			chain.rebuild_height_for_pos()?;
 		}
 
+		chain.log_heads()?;
+
 		Ok(chain)
 	}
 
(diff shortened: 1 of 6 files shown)


yeastplume:
Decent optimisation and looks straightfoward enough, LGTM

antiochp:
Going to merge this to master. 2.1.0 is out.

chain/src/chain.rs
```diff
@@ -1217,9 +1193,9 @@ impl Chain {
 
 	/// Tip (head) of the header chain.
 	pub fn header_head(&self) -> Result<Tip, Error> {
-		self.store
-			.header_head()
-			.map_err(|e| ErrorKind::StoreErr(e, "chain header head".to_owned()).into())
+		let hash = self.header_pmmr.read().head_hash()?;
```
JosephGoulden: I think this change breaks the TUI hashset validation percentage. Because that calls this function which now requires a read lock, which is not available as the validation itself has a write lock on header_pmmr.

PR #3046, opened 2019-09-17
fix: Add some more stats to basic status page of TUI (disk usage, chain timestamp, tx pool size)

JosephGoulden:
Hi. I added some more stats to the basic status page of the TUI to help node operators.
- Disk usage measured in GB
- Chain tip timestamp
- Tx pool size
- Stem pool size

I've run this against mainnet and everything appears to be updating in real time as expected.

image

### DIFF
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -25,7 +25,6 @@ use crate::core::ser::ProtocolVersion;
 
 use chrono::prelude::*;
 
-use crate::chain;
 use crate::chain::SyncStatus;
 use crate::p2p;
 use grin_core::pow::Difficulty;
@@ -54,7 +53,7 @@ pub struct ServerStats {
 	/// Chain head
 	pub chain_stats: ChainStats,
 	/// sync header head
-	pub header_head: chain::Tip,
+	pub header_stats: ChainStats,
 	/// Whether we're currently syncing
 	pub sync_status: SyncStatus,
 	/// Handle to current stratum server stats
@@ -78,8 +77,8 @@ pub struct ChainStats {
 	pub last_block_h: Hash,
 	/// Total difficulty accumulated on that fork
 	pub total_difficulty: Difficulty,
-	/// Timestamp of last block
-	pub last_block_time: DateTime<Utc>,
+	/// Timestamp of highest block or header
+	pub latest_timestamp: DateTime<Utc>,
 }
 /// Transaction Statistics
 #[derive(Clone, Serialize, Debug)]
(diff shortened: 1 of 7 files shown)


DavidBurkett:
This is awesome! One small comment in the code, but otherwise LGTM.

JosephGoulden:
Thanks @DavidBurkett but I can't see your comment for some reason.

servers/src/common/stats.rs
```diff
@@ -51,7 +52,7 @@ pub struct ServerStats {
 	/// Number of peers
 	pub peer_count: u32,
 	/// Chain head
-	pub head: chain::Tip,
+	pub chain_stats: ChainStats,
```
antiochp: Want to replace `header_head` with a `ChainStats` as well? 
It would be useful to see the timestamp for the header chain during the initial header sync.

servers/src/grin/server.rs
```diff
@@ -489,14 +491,45 @@ impl Server {
...
+
+		let head = self.chain.head_header().unwrap();
+		let head_stats = ChainStats {
+			last_block_time: head.timestamp,
+			height: head.height,
+			last_block_h: head.prev_hash,
+			total_difficulty: head.total_difficulty(),
+		};
+
+		let disk_usage_gb = {
```
antiochp: I see this is measuring disk usage for the main lmdb database.
We also use significant storage for the various MMR files as well (and the peers lmdb database, but less so).

```
chain_data/lmdb
chain_data/peer
chain_data/header/header_head
chain_data/header/sync_head
chain_data/txhashset/kernel
chain_data/txhashset/output
chain_data/txhashset/rangeproof
```

servers/src/common/stats.rs
JosephGoulden: yep will do

servers/src/grin/server.rs
JosephGoulden: These looked very small which is why I left them out. But I realise now I'm not running full archive and network activity is only low at the moment. I'll see about adding them to the usage statistic.

antiochp: The kernel and rangeproof MMRs should be approx 150MB each on mainnet, regardless of archive mode.

JosephGoulden: Ok I was not looking hard enough.. I see them now. I added the Walkdir crate to go through the whole db_root dir, it's small has a nice API, do you think it's ok?

antiochp: Dependency on walkdir is fine I think. We already use it in the util crate. :+1:

antiochp:
`Transaction Pool Size` and `Stem Pool Size` are awesome on the summary screen by the way. 🎉

PR #3047, opened 2019-09-18
Gracefully handle WouldBlock and TimedOut when downloading txhashset

antiochp:
Turns out `WouldBlock` can be raised in rust even when using blocking IO.

```
20190902 22:17:46.291 ERROR grin_p2p::protocol - handle_payload: 
txhashset archive save to file fail. err=Connection(Os 
{ code: 35, kind: WouldBlock, message: "Resource temporarily unavailable" })
```

We occasionally see this after getting most of a txhashset dowloaded and it is _very_ frustrating.

Intuitively this should never happen as blocking IO would simply _block_ with no need for a `WouldBlock` error...

Except the docs indicate the following - 

https://doc.rust-lang.org/std/net/struct.TcpStream.html#method.set_read_timeout

> Platforms may return a different error code whenever a read times out as a result of setting this option. For example Unix typically returns an error of the kind WouldBlock, but Windows may return TimedOut.

tl;dr We set a read timeout on our blocking tcp stream. Rust tries its hardest to confuse us by surfacing this as a `WouldBlock` error when we call `read_exact`.

We handle this in the `try_break!` macro for the majority of our p2p message handling.
But we do not do this in `copy_attachment` which is used when downloading the txhashset zip file.

This PR adds similar "retry" logic allowing the txhashset download to continue downloading if we experience either a `WouldBlock` or a `TimedOut` error.
Also included `Interrupted` for completeness as the docs indicate this can be safely retried.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -113,7 +113,17 @@ impl<'a> Message<'a> {
 		while written < len {
 			let read_len = cmp::min(8000, len - written);
 			let mut buf = vec![0u8; read_len];
-			self.stream.read_exact(&mut buf[..])?;
+			if let Err(e) = self.stream.read_exact(&mut buf[..]) {
+				match e.kind() {
+					io::ErrorKind::WouldBlock
+					| io::ErrorKind::TimedOut
+					| io::ErrorKind::Interrupted => {
+						thread::sleep(Duration::from_millis(10));
+						return Ok(0);
+					}
+					_ => return Err(e.into()),
+				}
+			}
 			writer.write_all(&mut buf)?;
 			written += read_len;
 		}
--- a/servers/src/grin/sync/state_sync.rs
+++ b/servers/src/grin/sync/state_sync.rs
@@ -117,7 +117,7 @@ impl StateSync {
 
 			if let SyncStatus::TxHashsetDownload { .. } = self.sync_state.status() {
 				if download_timeout {
-					error!("state_sync: TxHashsetDownload status timeout in 10 minutes!");
+					error!("state_sync: TxHashsetDownload download timed out, restarting.");
 					self.sync_state.set_sync_error(
 						chain::ErrorKind::SyncError(format!("{:?}", p2p::Error::Timeout)).into(),
 					);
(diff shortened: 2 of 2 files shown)


antiochp:
Actually we use `read_exact` in a handful of other places, particularly `read_header` and `read_body`.

I'm not actually convinced this works as intended.

A single `read_exact` internally loops repeatedly, calling `read` until all the bytes are read.
But on a tcp stream it can _still_ raise a `WouldBlock` or a `TimedOut` after a partial read. We may have read some of the bytes and then hit a `WouldBlock`.
Its not enough in this scenario to simply retry the entire `read_exact` call, we actually need to retry the individual `read` call internally.

The only error handling in `read_exact` is to squash `Interrupted` and continue looping until all the bytes are read. There is no such error handling for `WouldBlock` or `TimedOut`, both of which can also occur during an individual `read` call.

We know that `read_exact` does not work for non-blocking IO. 
I don't think it works robustly for blocking IO either...

hashmap:
Hm I thought @garyyu fixed it recently. I agree with the last comment, perhaps we need to bring back our impl of `read-exact`, let me think about it.

antiochp:
Closing this for now - I don't fell like I've got a good grasp of what our expectations around tcp read timeouts are here. And this sort of feels like a bandaid solution without resolving the underlying issue.

Things I think I understand - 
* We use blocking IO
* We set 1000ms read and write timeout on the tcp stream
* these timeouts allow the node to be responsive during shutdown (we don't block for too long waiting for peer threads to stop)
* internal calls to `read` can timeout (and surface as a `WouldBlock`)
* so `read_exact` calls can actually fail with partially filled buffer
* we use `try_break!` in various places to "retry" `read_exact` (but we do not account for partially filled buffers here)
* txhashset file attachment probably wants handling as a special case - unclear how to handle timeouts here

From a UI responsiveness perspective these timeouts make a lot of sense.
But from a networking perspective we basically ignore them. If we timeout reading from a peer we simply attempt to keep reading. But run the risk of seeing a partially filled buffer (and presumably close the connection downstream of this with corrupted or unexpected data).

Issue #3002, opened 2019-08-26
Config file - inconsistent behavior with missing values


josef-v:
I was trying to minimize my config file, as I usually only keep key-value pairs that are necessary (e. g. in config files of other coin daemons). I am used to believe that missing keys in config files fallback to default values. I would also expect that the default config file contains the default values.
I have following issues with grin behavior:
* Some options can't be removed from config file at all or the app crashes at start. I would be willing to accept this behavior with options that are security related, but `log_to_file`? C'mon!
* I can't be sure about the default values. I couldn't find any reference which states, that values in config file are the default ones. Well, obviously, because it is not true. You enable/disable API authentication by commenting `api_secret_path`! It would be nice to have documented default values in the config file (or at least somewhere).

Of course this is not a big issue, but I believe I am not the only one who likes to glance at the config file and see where I have chosen to use non-default values and also have a compact config. Also, the described behavior is unpredictable, which is probably never a good thing. (Do I use the defaults by commenting this out? What is the default, is it this value?)

For context, [here is link](https://www.grin-forum.org/t/config-fallback-to-defaults/5914) to my whine on grin forum.

reagentoo:
Config options api_secret_path and foreign_api_secret_path does not work anyway.

Anynomouss:
@reagentoo Will be fixed in upcoming pull request, I will tag this one. 
The default values being loaded if missing, not so sure if that has been fixed and whether that is actually desirable.

Anynomouss:
> Config options api_secret_path and foreign_api_secret_path don't work. This files are saving to ~/.grin/main anyway.

This issue has been fixed by this pull:
https://github.com/mimblewimble/grin-wallet/pull/732

Future work might include an option to load default options when parameters are not specified in _grin-wallet.toml._

I will leave this open for now since I did not test how grin-wallet deals with missing values. 
Also indeed loading default values might lead to unexpected behavior. If implemented, the user should at least get a warning for key value pairs that are not loaded because of typos in the parameter names.

PR #3053, opened 2019-09-22
Provide default values information in generated config file, better handle logging default parameters

glemercier:
Improvements linked to request #3002 

This pull request provides information on default values for each config parameter as comments of the auto-generated `grin-server.toml` file, as per requested in issue mentioned above.
It also better handles cases where some logging parameters are not provided, leading to the program exiting. Instead we use optional parameters set to their default value when absent from the config file.

This PR improves overall configuration experience based on @josef-v feedback but it could be amended to better improve simple configuration inconsistencies coming from anyone else's request.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -477,7 +477,7 @@ fn comments() -> HashMap<String, String> {
 }
 
 fn get_key(line: &str) -> String {
-	if line.contains("[") && line.contains("]") {
+	if line.contains("[") && line.contains("]") && !line.contains("=") {
 		return line.to_owned();
 	} else if line.contains("=") {
 		return line.split("=").collect::<Vec<&str>>()[0].trim().to_owned();
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -258,8 +258,58 @@ fn comments() -> HashMap<String, String> {
 	);
 
 	retval.insert(
-		"[server.p2p_config.capabilities]".to_string(),
-		"#If the seeding type is List, the list of peers to connect to can
+		"ban_window".to_string(),
+		"
+#how long a banned peer should stay banned
+"
+		.to_string(),
+	);
+
+	retval.insert(
+		"peer_max_inbound_count".to_string(),
+		"
+#maximum number of inbound peer connections
+"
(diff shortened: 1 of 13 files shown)


glemercier:
Ok I get your point, fair request indeed. I can confirm "serde defaults" are not used everywhere (mostly in one place if I remember well). I'll look into improving this first.

I don't think @josef-v was concerned with the size of the default configuration file as long as he can trim it himself without breaking things. If you think there is a better place to extract and expose the default config values, I'll be happy to look into it.

josef-v:
> #3002 is about having less config file and relying on defaults. Whereas these changes actually increases the size of the generated config file.

I agree with @glemercier that adding another comment is not an issue as you can simply delete it later without any consequences. So having defaults in a comment is ok, maybe unnecessary as we can have the config file generated with defaults and some statement "config file generated with default values" on top of it. However hardcoding these values into config is not a way to go.

> Ideally all this config, specifically the default config would be sourced from one single "source of truth".

That's exactly what I meant.

glemercier:
Thank you both for your feedback and implementation details. I'll work on this and propose an update to this PR.

glemercier:
Update to the PR:
- Use serde defaults on all the config structures. Removed `Option<>` where relevant for cleaner code
- Put all the default values as constants a the top of concerned files. They may be moved to a single dedicated file for a central default config place
- Explicitely implement Serialize/Deserialize for `Capabilities` bitflags
This has for effect that the parameter appears like this in the config file:

```
[server.p2p_config]
capabilities = 15
```

instead of previously

```
bits = 15
```

however it is possible to comment the line in the new case and it will take the default value, which was not possible the old way (using derived Serialize/Deserialize)
 - When generating default config, commenting out all values, since default will be used
 - An (almost) empty config file will work, only the root titles are required to avoid crashing at startup:

```
[server]
[logging]
```

Feel free to provide feedback on this PR and suggest any change.

antiochp:
Planning to take a closer look over this over the next couple of days. But :+1: on the approach.

phyro:
@antiochp this one has accumulated quite a bit of conflicts. Is the proposed change still relevant?

PR #3054, opened 2019-09-23
Timeout Client requests after 20 seconds

quentinlesceller:
All clients requests now timeout after 20 seconds: read or write.
Might be a good idea to revisit the implementation with hyper 0.13 (when released).


antiochp:
Are we comfortable merging this for 2.1.0?

bladedoyle:
Thank you @quentinlesceller - very much appreciated.

PR #3057, opened 2019-09-25
Refactor transaction building combinators 

antiochp:
This allows for more robust error handling as we can now propagate errors up to the top level.

Specifically the `Append` type which takes a Result and returns a Result.

```
pub type Append<K, B> = dyn for<'a> Fn(
	&'a mut Context<'_, K, B>,
	Result<(Transaction, TxKernel, BlindSum), Error>,
) -> Result<(Transaction, TxKernel, BlindSum), Error>;
```

Also replaced `with_fee()` and `with_lock_height()` combinators with simply passing in kernel features to `build::transaction()` and `build::partial_transaction()`. It was confusing (and fragile) setting `fee` and `lock_height` individually via combinators. We now explicitly specify the features for the kernel of the tx being built.

```
let tx = transaction(
    KernelFeatures::Plain { fee: 4 },
    vec![
        input(6, key_id1),
        output(2, key_id2),
    ],
    &keychain,
    &builder,
)
```

----

__NOTE:__ Wallet uses `build::transaction()` and `build::partial_transaction()` so we need to be careful merging this to make sure the wallet project is updated to reflect this.

### DIFF
--- a/chain/tests/mine_simple_chain.rs
+++ b/chain/tests/mine_simple_chain.rs
@@ -16,7 +16,7 @@ use self::chain::types::{NoopAdapter, Tip};
 use self::chain::Chain;
 use self::core::core::hash::Hashed;
 use self::core::core::verifier_cache::LruVerifierCache;
-use self::core::core::{Block, BlockHeader, OutputIdentifier, Transaction};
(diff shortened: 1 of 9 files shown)


antiochp:
Merging this. grin-wallet will no longer build until the related PR is also merged.

PR #3060, opened 2019-09-26
add block_hash to head_info so we can easily use them later

antiochp:
I suspect the `header_mmr.read()` lock in get_server_stats is causing issues as _every_ peer thread needs write access to the header_mmr for processing blocks and block headers.

I have seen anecdotal evidence that something is causing a deadlock and gdb indicates the stuck tui is blocked waiting on a read() lock. Just not 100% sure _which_ read lock, but this is a likely candidate).

There is no need to lock the header_mmr here just to read several headers by height to find their block hashes. It makes more sense to simply track these block hashes when we build the difficulty iterator initially.
We also update the tui stats every 1s or so, which introduces the potential for lock contention on this header_mmr via the tui thread.

Currently running a node on mainnet with these changes applied and TUI enabled. I'd like to let this run for a bit to see if we can still reproduce the deadlock scenario or if these changes do appear to alleviate the issue.

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -562,6 +562,7 @@ impl<'a> Iterator for DifficultyIter<'a> {
 			let scaling = header.pow.secondary_scaling;
 
 			Some(HeaderInfo::new(
+				header.hash(),
 				header.timestamp.timestamp() as u64,
 				difficulty,
 				scaling,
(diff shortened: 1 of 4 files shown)


antiochp:
This has not made anything _worse_ and it should definitely have some benefits with tui updates being less resource intensive. So I'm going to merge this.

PR #3062, opened 2019-09-26
Update zeroize to 0.9.3

quentinlesceller:


antiochp:
Looks like we need this as well - 

```
cargo update -p zeroize_derive
```

PR #3063, opened 2019-09-28
2.1.0-beta.1 deadlock fix

yeastplume:
Can't entirely explain why this would have caused a threading deadlock, but debugging a stuck node showed 2 threads stuck trying to get a lock on the transaction pool at:

server.rs:480: `stem_pool_size: self.tx_pool.read().stempool.entries.len()`
dandelion_monitor.rs:159 - `let mut tx_pool = tx_pool.write();`

I'd conjecture it occurred for some Rusty reason when the write lock happened to be taken just after the first read lock.

Current fix:

* Changes the tui `read' to `try_read`
* Only attempt to read once (was no need to lock twice in the existing code)

I'd guess that only taking the lock once would have been enough to resolve, but no harm just using a `try_read` for tui purposes. Ran with these modifications overnight seems to resolve the freezing issues.

### DIFF
--- a/servers/src/grin/server.rs
+++ b/servers/src/grin/server.rs
@@ -475,9 +475,17 @@ impl Server {
 			.map(|p| PeerStats::from_peer(&p))
 			.collect();
 
+		let (tx_pool_size, stem_pool_size) = {
+			let tx_pool_lock = self.tx_pool.try_read();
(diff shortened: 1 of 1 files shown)


antiochp:
Nice catch with this.

:+1:

Issue #821, opened 2018-03-20
TUI Log/Error handling


yeastplume:
Just putting there here to keep me reminded. Currently, if there's a panic on any thread the panic is written to the log, but not into the TUI, therefore a user looking at it has no idea there was a problem (or has to guess that there's been a problem from the way the stats aren't updating properly. To mitigate this and make it more user friendly, we need a way of communicating from the logger (which also handles panics) to the TUI.

Current thinking is:

* Create an slog drain that writes its output to an mpsc channel
* TUI can instantiate and provide a listener to the logger on initiation
* logger can then send messages back to the TUI that contain log information as well as panic status. Warnings for panics can be displayed on screen (popped-up nicely, even), and as an added feature we can add a log viewer screen right in the TUI.

JosephGoulden:
I'm gonna have a go at this.

quentinlesceller:
Can we consider this fixed by #3064?

PR #3064, opened 2019-09-29
feat: TUI logs view

JosephGoulden:
Adding a logs tab to TUI as described in issue #821 

I used different colours for Info/Warn/Error log levels. I didn't make the panics pop up in any way, as described in the issue, but they are logged with full stack trace in red.

I've manually tested stopping/starting grin with and without the TUI, various log levels and logging to a file on and off. All is working well. There are some small code changes around the logging when stop starting the server that came from this testing and I believe all the relevant information is logged in the right places.

image


src/bin/grin.rs
```diff
@@ -136,14 +138,11 @@ fn real_main() -> i32 {
 		}
 	}
 
+	let (logs_tx, logs_rx) = mpsc::channel::<LogEntry>();
```
antiochp: I have not finished reading this PR fully but posting this here (mainly to remind me to look at this again) - this is an async channel. If the receiver is too slow for any reason then the queue will grow indefinitely.

util/src/types.rs
```diff
@@ -1,104 +1 @@
...
-			tui_running: None,
-		}
-	}
-}
-
-use std::ops::Deref;
-use zeroize::Zeroize;
-/// Zeroing string, mainly useful for password
-#[derive(Clone, PartialEq, PartialOrd)]
-pub struct ZeroingString(String);
```
antiochp: This seems unrelated? And its password related (masking out passwords in the logs?) So probably best to have this in a separate PR if this is indeed unused and to be cleaned up.

src/bin/tui/ui.rs
```diff
@@ -124,6 +133,10 @@ impl UI {
 			return false;
 		}
 
+		while let Some(message) = self.logs_rx.try_iter().next() {
```
antiochp: We may want to rethink some of the tui refresh logic. I think we update the _entire_ tui on every step, regardless of what view or screen in the tui we are actually in.

Particularly for something as heavy as the full screen of log messages, we may want to look into limiting updates here only if we are actually viewing this screen. (Maybe this is a bad idea, I have not thought this through fully).

But it feels like updating many lines of streaming log msg output might be overly expensive if the tui is not actually viewing these at the time. Particularly if the node is logging debug (or trace...) level logs.

JosephGoulden: We do up date the whole thing every second yes. But we could update the logs view only when its the active tab as you suggest. When moving to a separate tab we would clear the logs view and consume but ignore the log messages. So when you went back to the logs view it would start scrolling up from the bottom with the latest messages.

util/src/types.rs
JosephGoulden: Yes it's unrelated and unused. Will put it back for now.

src/bin/grin.rs
JosephGoulden: Ok, could try the sync_channel instead with a reasonable buffer? I'm not sure how this would affect the sender (logging) threads if the buffer filled and they were blocked.

JosephGoulden:
Thanks @antiochp and thanks for the review. I'll have a look at those issues you raised and wait for any further comments.

src/bin/grin.rs
antiochp: There's a `try_send` on a sync sender that we can take advantage of here.
Basically just drop log messages on the floor if we cannot safely send it via the buffer.

https://doc.rust-lang.org/std/sync/mpsc/struct.SyncSender.html#method.try_send

src/bin/tui/ui.rs
antiochp: Might be worth getting some performance numbers and a rough idea of how costly this is to update before making these changes. Its possible this is not actually an issue and fine for now.

JosephGoulden: So I did some testing and observational measurement between this branch and master and I cannot discern any noticeable impact on resources or responsiveness of the TUI.

Typically grin uses between 10-20% CPU and ~23MB RAM on my machine and this didn't change with TUI trace logging enabled. Also tested with log to file and TUI logging enabled. And tested with chain sync from scratch, again there was no noticeable difference.

src/bin/grin.rs
JosephGoulden: done this now

src/bin/tui/ui.rs
```diff
@@ -173,8 +187,8 @@ impl Controller {
 			while let Some(message) = self.rx.try_iter().next() {
 				match message {
 					ControllerMessage::Shutdown => {
+						warn!("Shutdown in progress, please wait");
 						self.ui.stop();
-						println!("Shutdown in progress, please wait");
```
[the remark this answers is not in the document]
JosephGoulden: yeah I don't like the text appearing randomly in the middle of the TUI, let me see if I can find a better way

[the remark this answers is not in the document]
hashmap: I hacked it, unfortunately could not find a way to show a dialog window there

servers/src/grin/server.rs
```diff
@@ -80,38 +80,31 @@ pub struct Server {
...
-	pub fn start<F>(config: ServerConfig, mut info_callback: F) -> Result<(), Error>
-	where
-		F: FnMut(Server),
-	{
-		let mining_config = config.stratum_mining_config.clone();
-		let enable_test_miner = config.run_test_miner;
-		let test_miner_wallet_url = config.test_miner_wallet_url.clone();
-		let serv = Server::new(config)?;
+	/// Start ancillary services
+	pub fn start_ancillary(&self) -> Result<(), Error> {
```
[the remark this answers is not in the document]
JosephGoulden: I changed the name because I moved the Server::new(config) out to the calling function, this is where the server is actually started.
And the reason I did that is because the Receiver is not clonable so has to be passed as another parameter in the callback which did not look very pretty.
I can do it this way if you want though, I don't really mind. Will change it so you can see.

src/bin/grin.rs
```diff
@@ -136,14 +138,11 @@ fn real_main() -> i32 {
 		}
 	}
 
+	let (logs_tx, logs_rx) = mpsc::sync_channel::<LogEntry>(200);
```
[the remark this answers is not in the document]
JosephGoulden: yeah I think I can do that, and wrap Sender and Receiver with Option.

src/bin/tui/ui.rs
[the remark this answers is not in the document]
JosephGoulden: ok.. actually I'll revert this for now and see about doing it in a separate MR.

util/src/lib.rs
```diff
@@ -42,8 +42,9 @@ pub use crate::logger::{init_logger, init_test_logger};
 pub mod secp_static;
 pub use crate::secp_static::static_secp_instance;
 
+/// Zeroing string for passwords
```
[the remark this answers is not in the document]
JosephGoulden: Fixed now. Accidentally removed the comments in util/src/types.rs

JosephGoulden:
Thanks for the review @quentinlesceller . I've fixed all your suggestions, let me know what you think.

quentinlesceller:
Currently looking at the changes now. Is there a way to wrap the text when reducing the windows size ?

JosephGoulden:
> Currently looking at the changes now. Is there a way to wrap the text when reducing the windows size ?

Yes there is.. and I've now done this.

quentinlesceller:
@JosephGoulden  Just tried and somehow the colors are not working anymore with the logs.
See attached screenshot (might be something with my terminal):
Capture d’écran 2019-10-30 à 14 07 38

JosephGoulden:
@quentinlesceller  fixed that now.

util/src/types.rs
```diff
@@ -12,64 +12,7 @@
...
-			file_log_level: LogLevel::Info,
-			log_file_path: String::from("grin.log"),
-			log_file_append: true,
-			log_max_size: Some(1024 * 1024 * 16), // 16 megabytes default
-			log_max_files: Some(DEFAULT_ROTATE_LOG_FILES),
-			tui_running: None,
-		}
-	}
-}
+//! Zeroing String
```
quentinlesceller: Weird comment here?

JosephGoulden: It's module documentation and it seems to be required for all public modules. Without it we get a warning due to the #![warn(missing_docs)] in lib.rs.

After I removed the logging code from this file I just updated the comment to say ZeroingString as that is all that is left in there. Maybe the file should be renamed to zeroing_string.rs. Or a better comment?

quentinlesceller: Okay let's leave it like that for now.

antiochp:
I think something may have slipped through here.

When we parse the config file we accept either `Info` or `INFO` as the log level.
Both appear to be parsed correctly.

But - when we _generate_ the config file we default to `INFO` which is not supported in previous code.

PR #3069, opened 2019-10-01
Fix reading POW for edge bits 59 and higher

hashmap:
It requires reading more than 8 bytes

### DIFF
--- a/core/src/pow/types.rs
+++ b/core/src/pow/types.rs
@@ -391,21 +391,54 @@ impl Proof {
 	}
 }
 
-#[inline(always)]
+fn extract_bits(bits: &Vec<u8>, bit_start: usize, bit_count: usize, read_from: usize) -> u64 {
+	let mut buf: [u8; 8] = [0; 8];
+	// read 8 bytes
+	buf.copy_from_slice(&bits[read_from..read_from + 8]);
+	// convert litl endian bits to u64
+	let mut nonce = u64::from_le_bytes(buf);
+	// shake off high bits (they are part of prev number)
+	nonce <<= 64 - (bit_start - read_from * 8) - bit_count;
+	// shake of low bits (next number if any)
+	nonce >> 64 - bit_count
+}
+
 fn read_number(bits: &Vec<u8>, bit_start: usize, bit_count: usize) -> u64 {
 	if bit_count == 0 {
 		return 0;
 	}
-	let mut buf: [u8; 8] = [0; 8];
-	let mut byte_start = bit_start / 8;
-	if byte_start + 8 > bits.len() {
-		byte_start = bits.len() - 8;
+	// find where the first byte to read starts
+	let mut read_from = bit_start / 8;
+	// move back if we are too close to the end of bits
+	if read_from + 8 > bits.len() {
+		read_from = bits.len() - 8;
 	}
-	buf.copy_from_slice(&bits[byte_start..byte_start + 8]);
-	buf.reverse();
-	let mut nonce = u64::from_be_bytes(buf);
-	nonce = nonce << 64 - (bit_start - byte_start * 8) - bit_count;
-	nonce >> 64 - bit_count
+	// calculate max bit we can read up to (+64 bits from the start)
+	let max_read_pos = (read_from + 8) * 8;
+	// calculate max bit we want to read
+	let max_pos = bit_start + bit_count;
+	// check if we can read it all at once
+	if max_read_pos >= max_pos {
+		extract_bits(bits, bit_start, bit_count, read_from)
+	} else {
+		// we need to read more than 8 bytes, we need to read twice
+		// how many valuable bits we can read now
+		let mut read_size = max_read_pos - bit_start;
+		// read it, it's little endian, so we get the low part of the number
+		let low = extract_bits(bits, bit_start, read_size, read_from);
+
+		// how many bits is left to read
+		read_size = max_pos - max_read_pos;
+		// find the suitable byte to read from
+		read_from = max_read_pos / 8;
+		if read_from + 8 > bits.len() {
+			read_from = bits.len() - 8;
+		}
+		// read the second part
+		let high = extract_bits(bits, max_read_pos, read_size, read_from);
+		// put high bits on the right spot in the number
+		(high << bit_count - read_size) + low
+	}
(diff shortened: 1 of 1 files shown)


core/src/pow/types.rs
```diff
@@ -391,21 +391,54 @@ impl Proof {
...
+fn extract_bits(bits: &Vec<u8>, bit_start: usize, bit_count: usize, read_from: usize) -> u64 {
+	let mut buf: [u8; 8] = [0; 8];
+	// read 8 bytes
+	buf.copy_from_slice(&bits[read_from..read_from + 8]);
+	// convert litl endian bits to u64
+	let mut nonce = u64::from_le_bytes(buf);
+	// shake off high bits (they are part of prev number)
+	nonce <<= 64 - (bit_start - read_from * 8) - bit_count;
+	// shake of low bits (next number if any)
+	nonce >> 64 - bit_count
```
tromp: Unwanted bits should be masked out rather than shifted out on the left:
let skip_bits = bit_start - read_from * 8;
let bit_mask = (1 << bit_count) - 1;
u64::from_le_bytes(buf) >> skip_bits & bit_mask;

core/src/pow/types.rs
```diff
@@ -391,21 +391,54 @@ impl Proof {
...
-	nonce = nonce << 64 - (bit_start - byte_start * 8) - bit_count;
-	nonce >> 64 - bit_count
+	// find where the first byte to read starts
+	let mut read_from = bit_start / 8;
+	// move back if we are too close to the end of bits
+	if read_from + 8 > bits.len() {
+		read_from = bits.len() - 8;
+	}
+	// calculate max bit we can read up to (+64 bits from the start)
+	let max_read_pos = (read_from + 8) * 8;
```
tromp: if prefer "end" instead of "max" since it's an exclusive position.

core/src/pow/types.rs
```diff
@@ -391,21 +391,54 @@ impl Proof {
...
+	// calculate max bit we want to read
+	let max_pos = bit_start + bit_count;
+	// check if we can read it all at once
+	if max_read_pos >= max_pos {
```
tromp: i prefer max_pos <= max_read_pos  :-)

core/src/pow/types.rs
```diff
@@ -391,21 +391,54 @@ impl Proof {
...
+		extract_bits(bits, bit_start, bit_count, read_from)
+	} else {
+		// we need to read more than 8 bytes, we need to read twice
```
tromp: how about simply reading one byte separately with

let low = bits[read_from];
let high =  extract_bits(bits, bit_start+8, bit_count, read_from+1);
(high << 8) + low;

core/src/pow/types.rs
```diff
@@ -478,3 +514,40 @@ impl BitVec {
...
+				Ok(p) => assert_eq!(p, proof),
+			}
+		}
+	}
+
+	fn gen_proof(bits: u32) -> Vec<u64> {
+		let mut rng = rand::thread_rng();
+		let mut v = Vec::with_capacity(42);
+		for _ in 0..42 {
+			v.push(rng.gen_range(u64::pow(2, bits - 1), u64::pow(2, bits)));
```
tromp: u64::pow(2, 64) should be 0.
does rng.gen_range(1<<63, 0) really work to generate random numbers between 2^63 and 2^64 ?

core/src/pow/types.rs
tromp: That avoid the need to compare read_from+8 against bits.len()

core/src/pow/types.rs
hashmap: applied the change, but we need an additional check -  if `bit_count == 64` the second line will panic

core/src/pow/types.rs
hashmap: done

core/src/pow/types.rs
hashmap: done

core/src/pow/types.rs
hashmap: very elegant, thanks

core/src/pow/types.rs
hashmap: I believe it would cause a panic in Rust, but somehow works here, perhaps because he second point is exclusive. Anyway i added a special handling for 64 bits

core/src/pow/types.rs
```diff
@@ -391,21 +391,39 @@ impl Proof {
 	}
 }
 
-#[inline(always)]
+fn extract_bits(bits: &Vec<u8>, bit_start: usize, bit_count: usize, read_from: usize) -> u64 {
+	let mut buf: [u8; 8] = [0; 8];
```
tromp: Is it possible in rust to directly assign

let bits64 = u64::from_le_bytes(&bits[read_from..read_from + 8]) ?

if so, that would be much preferable over the extra buffer and copying....

tromp: Other than that, this PR looks fine to me.

core/src/pow/types.rs
```diff
@@ -478,3 +499,47 @@ impl BitVec {
...
+			}
+		}
+	}
+
+	fn gen_proof(bits: u32) -> Vec<u64> {
+		let mut rng = rand::thread_rng();
+		let mut v = Vec::with_capacity(42);
+		for _ in 0..42 {
+			v.push(rng.gen_range(
+				u64::pow(2, bits - 1),
```
tromp: btw, isn't it easier to write 1 << (bits - 1) than u64::pow(2, bits - 1) ?

DavidBurkett: I think it's easier to read u64::pow(2, bits - 1), so personally prefer it iff there's no performance hit.

core/src/pow/types.rs
hashmap: Array and slice are different types in Rust, so we can't directly use `bits` in `from_le_bytes`

tromp: and there's no function or method we can call on the slice to convert to array?

tromp: btw, remaining comments are just about aesthetics (can always be reconsidered later) and should not hold up this PR...

PR #3072, opened 2019-10-01
Change 2018 to 2019 in copyright files

quentinlesceller:
Not sure if that's really needed. But if yes should be done one day or another.


yeastplume:
Sure, why not?

PR #3078, opened 2019-10-03
chore: pre-commit hook - format all staged files, re-add then commit

JosephGoulden:
The pre-commit hook isn't working so well at the moment. Any files that require it are formatted but they are then left un-staged after the modification and the commit proceeds regardless, leaving out those formatted files.

I suggest all staged files that need it are formatted, re-staged and committed, which is what I have implemented in this PR.

I've tested commits with and without multiple files that require formatting.

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -43,7 +43,7 @@ else
   for file in ${problem_files[@]}; do
     rustfmt $file
     git add $file
-		printf "\033[0;31m    $file\033[0m \n"
+		printf "\033[0;32m    $file\033[0m \n"
   done
 fi
 
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -22,38 +22,29 @@ if [ $? != 0 ]; then
 	exit 1
 fi
 
-result=0
 problem_files=()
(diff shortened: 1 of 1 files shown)


antiochp:
:+1: Let me play around with this a bit. 

We were intentionally _not_ staging and reformatting as part of the commit to allow devs to see what the hook was doing. The rough idea was you could always `git add . && git commit --amend` if you were happy with what it was doing.
The downside (as you're experiencing) is it needs some manual steps to actually do its thing.

JosephGoulden:
Sure. Assuming the format isn't going to break the code in some way I prefer to just let it do its thing. Even if I don't personally like the choices it makes sometimes it just removes that subjectivity all together. If it did do something weird hopefully that would be picked up in the review or tests.

antiochp:
Pinged people in the grincoin.team.node_dev channel on keybase to see if anyone has any thoughts for/against this. If not we can get this merged.
I'm :+1: on this but I'd like to solicit more thoughts/feedback as it does affect workflow.

antiochp:
Merged - lets give this a try over the next few days.

PR #3085, opened 2019-10-04
Retrieves OutputPrintable instead of Output

quentinlesceller:
Simplify the API v1 output endpoint. We now always return a single object called OutputPrintable instead of different object when it is by id or by height.
Add method `get_unspent_output_at_pos`.

Before:

```
[
  {
    "commit": "08392e391abc679f6a66478fab4febb783d11904fc9b32cdeabf442ff6026f4baa",
    "height": 375611,
    "mmr_index": 4102967
  }
]
```

After:
```
[
  {
    "output_type": "Transaction",
    "commit": "08392e391abc679f6a66478fab4febb783d11904fc9b32cdeabf442ff6026f4baa",
    "spent": false,
    "proof": "1e93a916bccf08746147c9310954c720140faab878c73558f9a515ef1db314b57affa22d080fc8c0bbb8efa578089bd38025ac347f423648b71adf19bf93416c0cb98805a9eb9e27403621201fa47c89b4b44c864edc8c9123d8890d6194aeb0e13ad0a99d87b265ef79c7940b6d237c6da0b7cdc87badfca921d0598f33c09a3c9f72f3e909afb75d1e06334ab3f0c8a23374f7101d7bd03906f393cc9b1a4de04b561ca6750bffa4c2fbc94f8c8874e289e908665f3d25fdc8fd1707d90e135c640c590b20f00d57f96251dbe84a1cadad8fd06fc4f1556b0da934c5a096eed9788b96464399ce49aeabedf66f8685e6dacef0df781512313fe67e49c27ab47b9c7ee8c2f8d7810d9ec354113da563999e357d754d25750a22dfc10fe9f5482689aee0ac24e64ca48c9639564e6324160591a9ebb8320e13842856f6022ee8361d15f8f1d618a654f9c7fa012e3ba51e188418fdd89b16ae1fd6390051090497810131de7b15898d811bd35efb3ed9f6629ed56f7c25d4aeb669c7dc76077926ac70a1805d0390c0cd13de626034517772d87c43289d9c8ee665fa62837f9b9fc47a9a9e5aff421c4c7c5c5150645e74855cc437bbd9381fbf504dbf22c8c1296c34df202d921b1ba539c113d49247c698d53f7115ee8a202cfb4d15017b9a027a8f51d57b2582eb976bb60593a17e3534b97711c038256e83d801b0b4100c5312ae559e87782687c53861718dbf6367962fe5fbef77ceed9c799df25e2a4f9a795963d2e86f4c09c3c45da87e98c2b4aaabb47a4457d639371ea10ce7649c6434a825115e6452e7f241ddeb15b76145d8c682c6efe185b1cb0467e2b3964fa149a4c4865de58ad5b18edaa93670040e0e9ab997f56d585b07cd58379edb9dc5773e37ed809441ee28b48635b9aaf97d1456e753be6d822d5ee724059afa67ab1e97",
    "proof_hash": "a51aa12fba6316518368f045de5b4658cea859d6da386900e646d0445c5254bb",
    "block_height": 375611,
    "merkle_proof": null,
    "mmr_index": 4102967
  }
]
```

### DIFF
--- a/api/src/handlers/utils.rs
+++ b/api/src/handlers/utils.rs
@@ -33,7 +33,7 @@ pub fn w<T>(weak: &Weak<T>) -> Result<Arc<T>, Error> {
 pub fn get_output(
 	chain: &Weak<chain::Chain>,
 	id: &str,
-) -> Result<(Output, OutputIdentifier), Error> {
+) -> Result<(OutputPrintable, OutputIdentifier), Error> {
 	let c = util::from_hex(String::from(id)).context(ErrorKind::Argument(format!(
 		"Not a valid commitment: {}",
 		id
@@ -54,12 +54,22 @@ pub fn get_output(
 	for x in outputs.iter() {
 		let res = chain.is_unspent(x);
 		match res {
-			Ok(output_pos) => {
-				return Ok((
-					Output::new(&commit, output_pos.height, output_pos.position),
-					x.clone(),
-				));
-			}
+			Ok(output_pos) => match chain.get_unspent_output_at(output_pos.position) {
+				Ok(output) => {
+					match OutputPrintable::from_output(&output, chain.clone(), None, true, true) {
+						Ok(output_printable) => return Ok((output_printable, x.clone())),
+						Err(e) => {
+							trace!(
+								"get_output: err: {} for commit: {:?} with feature: {:?}",
+								e.to_string(),
+								x.commit,
+								x.features
+							);
+						}
+					}
+				}
+				Err(_) => return Err(ErrorKind::NotFound)?,
+			},
 			Err(e) => {
 				trace!(
 					"get_output: err: {} for commit: {:?} with feature: {:?}",
--- a/api/src/handlers/utils.rs
+++ b/api/src/handlers/utils.rs
@@ -26,13 +26,14 @@ use std::sync::{Arc, Weak};
 // boilerplate of dealing with `Weak`.
 pub fn w<T>(weak: &Weak<T>) -> Result<Arc<T>, Error> {
 	weak.upgrade()
-		.ok_or_else(|| ErrorKind::Internal("failed to upgrade weak refernce".to_owned()).into())
+		.ok_or_else(|| ErrorKind::Internal("failed to upgrade weak reference".to_owned()).into())
 }
 
 /// Retrieves an output from the chain given a commit id (a tiny bit iteratively)
 pub fn get_output(
 	chain: &Weak<chain::Chain>,
 	id: &str,
+	include_proof: bool,
 ) -> Result<(OutputPrintable, OutputIdentifier), Error> {
 	let c = util::from_hex(String::from(id)).context(ErrorKind::Argument(format!(
 		"Not a valid commitment: {}",
@@ -56,7 +57,13 @@ pub fn get_output(
 		match res {
 			Ok(output_pos) => match chain.get_unspent_output_at(output_pos.position) {
 				Ok(output) => {
-					match OutputPrintable::from_output(&output, chain.clone(), None, true, true) {
+					match OutputPrintable::from_output(
+						&output,
+						chain.clone(),
+						None,
+						include_proof,
+						false,
+					) {
 						Ok(output_printable) => return Ok((output_printable, x.clone())),
 						Err(e) => {
 							trace!(
--- a/chain/src/txhashset/utxo_view.rs
+++ b/chain/src/txhashset/utxo_view.rs
@@ -21,12 +21,14 @@ use crate::core::global;
 use crate::core::ser::PMMRIndexHashable;
 use crate::error::{Error, ErrorKind};
 use crate::store::Batch;
+use crate::util::secp::pedersen::RangeProof;
 use grin_store::pmmr::PMMRBackend;
 
 /// Readonly view of the UTXO set (based on output MMR).
 pub struct UTXOView<'a> {
 	output_pmmr: ReadonlyPMMR<'a, Output, PMMRBackend<Output>>,
 	header_pmmr: ReadonlyPMMR<'a, BlockHeader, PMMRBackend<BlockHeader>>,
+	rproof_pmmr: ReadonlyPMMR<'a, RangeProof, PMMRBackend<RangeProof>>,
 	batch: &'a Batch<'a>,
 }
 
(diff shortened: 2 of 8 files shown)


DavidBurkett:
This is used under the hood during wallet updates here: https://github.com/mimblewimble/grin-wallet/blob/11481669ef14f2d36ba86631ca6464b79a03098a/libwallet/src/internal/updater.rs#L357

So we may need some performance numbers before making this change, since it will inevitably slow that down.

chain/src/txhashset/utxo_view.rs
```diff
@@ -21,12 +21,14 @@ use crate::core::global;
...
 use crate::error::{Error, ErrorKind};
 use crate::store::Batch;
+use crate::util::secp::pedersen::RangeProof;
 use grin_store::pmmr::PMMRBackend;
 
 /// Readonly view of the UTXO set (based on output MMR).
 pub struct UTXOView<'a> {
 	output_pmmr: ReadonlyPMMR<'a, Output, PMMRBackend<Output>>,
 	header_pmmr: ReadonlyPMMR<'a, BlockHeader, PMMRBackend<BlockHeader>>,
+	rproof_pmmr: ReadonlyPMMR<'a, RangeProof, PMMRBackend<RangeProof>>,
```
antiochp: Hmm. This makes the `utxo_view` even more similar to the `txhashset` itself. I _suspect_ we're going to want to get rid of the `utxo_view` at some point and just rely on the `txhashset` itself for more use cases like this.

The limiting factor is the writeable nature of the `txhashset` and the fact we need to take a write lock on it even if just reading from it.

I have a branch locally messing around with an attempt to `try_clone` the MMR backend itself. The idea being we could clone the txhashset to get a readonly view of the files, relying on the internal mmap implementation to give a consistent view of the data.

api/src/handlers/utils.rs
```diff
@@ -54,12 +54,22 @@ pub fn get_output(
...
 		match res {
-			Ok(output_pos) => {
-				return Ok((
-					Output::new(&commit, output_pos.height, output_pos.position),
-					x.clone(),
-				));
-			}
+			Ok(output_pos) => match chain.get_unspent_output_at(output_pos.position) {
+				Ok(output) => {
+					match OutputPrintable::from_output(&output, chain.clone(), None, true, true) {
```
antiochp: We may want to parameterize the `include_proof` and `include_merkle_proof` here.
In many cases we don't really care about these and this is where the additional cost would be incurred or avoided.

quentinlesceller: Yeah I'll do. Was thinking that the user probably don't really care about whether having the chance or not. But since it is used by the wallet we will certainly need the performance boost herE.

antiochp: Yeah I think it would be fine to default to both `true`. And any client that cares about performance can opt out of these.

DavidBurkett: > just written that out I'm actually inclined to suggest we default to `false` for both of these and force clients to opt-in to this optional data when they know they really want it...
> 
> We should default to the api being fast and lightweight. Particularly with the size of the rangeproof and the cost of generating the (currently not used by _anybody_) Merkle proof.

The problem is we used to return the rangeproof and merkle proof, depending on how this was called. So defaulting to anything other than `true` would be a non-passive API change, and would require a "major" release. So for v2 APIs, we can default to `false`, but for v1, we probably should default to `true`

quentinlesceller: Okay I have added the `include_rp` and new `include_mp` parameters for outputs by id and by height.

chain/src/txhashset/utxo_view.rs
quentinlesceller: That'd make sense and could significantly reduce the code complexity.

api/src/handlers/utils.rs
quentinlesceller: I'm hesitant to default `include_merkle_proof` to `true`. I think adding it something about in the release note would be fine.

quentinlesceller:
For the records with 3800 outputs.
With this PR:
`grin-wallet --floonet -p "" outputs` => 1.1s
`grin-wallet --floonet -p "" info` => 0.4 ms.
Without:
`grin-wallet --floonet -p "" outputs` => 0.5 ms
`grin-wallet --floonet -p "" info` => 0.2 ms.

So there is a strong performance hit here....

api/src/handlers/utils.rs
antiochp: > The problem is we used to return the rangeproof and merkle proof

In which api endpoint?
My understanding is `byids` only ever returned minimal `outputs` as described in this PR and `byheight` defaulted to `false` for `include_rp`.

quentinlesceller: By height was always including the `merkle_proof` https://github.com/mimblewimble/grin/pull/3085/files#diff-6f4987d3855712fd885db9713f06652cR149. Rangeproof was always off by default.

quentinlesceller: I mean there was no field `proof` at that point.

DavidBurkett: > > The problem is we used to return the rangeproof and merkle proof
> 
> In which api endpoint?
> My understanding is `byids` only ever returned minimal `outputs` as described in this PR and `byheight` defaulted to `false` for `include_rp`.

You're right about rangeproofs, but it looks like we included merkle_proofs by default for byheight. I'm ok with changing that during a non-major version, since the odds are low that anybody is actually using merkle proofs. But it's not following the standard we voted on, so we should at least make it clear in the release docs, as mentioned.

https://github.com/mimblewimble/grin/blob/master/api/src/handlers/chain_api.rs#L180
https://github.com/mimblewimble/grin/blob/master/api/src/handlers/chain_api.rs#L146

quentinlesceller: However I agree that it must be for 3.0 since we are moving to a completely different structure.

antiochp:
> grin-wallet --floonet -p "" outputs => 500 s

That's "500 ms" not "500 s"?

quentinlesceller:
Indeed edited.

antiochp:
Out of the loop for a bit here - this is backward compatible right? We're simply adding fields to the existing json?
Or is there a difference here with `height` vs `block_height`? Is this something we need to support for compatibility?

quentinlesceller:
Because of the `height`/`block_height` thing this is not backward compatible. That's why there is https://github.com/mimblewimble/grin-wallet/pull/235. But if you think this is not worth it (to introduce such breaking change), we might as well simply leave the v1 as is and keep the improvements for the v2.

quentinlesceller:
As much as I think this is a good improvement, I don't see any urgent to introduce backward incompatibility in the v1 API right now. I'll keep the improvements for v2. Closing.

PR #3086, opened 2019-10-04
More robust peer banning

quentinlesceller:
Handle the errors and propagate them instead of just logging.

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -253,50 +253,43 @@ impl Peers {
 		}
 		false
 	}
-
 	/// Ban a peer, disconnecting it if we're currently connected
-	pub fn ban_peer(&self, peer_addr: PeerAddr, ban_reason: ReasonForBan) {
-		if let Err(e) = self.update_state(peer_addr, State::Banned) {
-			error!("Couldn't ban {}: {:?}", peer_addr, e);
-			return;
-		}
+	pub fn ban_peer(&self, peer_addr: PeerAddr, ban_reason: ReasonForBan) -> Result<(), Error> {
+		self.update_state(peer_addr, State::Banned)?;
 
-		if let Some(peer) = self.get_connected_peer(peer_addr) {
-			debug!("Banning peer {}", peer_addr);
-			// setting peer status will get it removed at the next clean_peer
-			match peer.send_ban_reason(ban_reason) {
-				Err(e) => error!("failed to send a ban reason to{}: {:?}", peer_addr, e),
-				Ok(_) => debug!("ban reason {:?} was sent to {}", ban_reason, peer_addr),
-			};
-			peer.set_banned();
-			peer.stop();
-
-			let mut peers = match self.peers.try_write_for(LOCK_TIMEOUT) {
-				Some(peers) => peers,
-				None => {
-					error!("ban_peer: failed to get peers lock");
-					return;
-				}
-			};
-			peers.remove(&peer.info.addr);
+		match self.get_connected_peer(peer_addr) {
+			Some(peer) => {
+				debug!("Banning peer {}", peer_addr);
+				// setting peer status will get it removed at the next clean_peer
+				peer.send_ban_reason(ban_reason)?;
+				peer.set_banned();
+				peer.stop();
+				let mut peers = match self.peers.try_write_for(LOCK_TIMEOUT) {
+					Some(peers) => peers,
+					None => {
+						error!("ban_peer: failed to get peers lock");
+						return Err(Error::PeerException);
+					}
+				};
+				peers.remove(&peer.info.addr);
+				Ok(())
+			}
+			None => return Err(Error::PeerNotFound),
 		}
 	}
 
 	/// Unban a peer, checks if it exists and banned then unban
-	pub fn unban_peer(&self, peer_addr: PeerAddr) {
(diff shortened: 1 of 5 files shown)


servers/src/grin/seed.rs
```diff
@@ -157,7 +157,9 @@ fn monitor_peers(
 				let interval = Utc::now().timestamp() - x.last_banned;
 				// Unban peer
 				if interval >= config.ban_window() {
-					peers.unban_peer(x.addr);
+					if let Err(e) = peers.unban_peer(x.addr) {
+						error!("failed to  unban peer {}: {:?}", x.addr, e);
```
antiochp: Excessive whitespace.

quentinlesceller: Good catch !

p2p/src/peers.rs
```diff
@@ -605,7 +598,12 @@ impl ChainAdapter for Peers {
 				"Received a bad block {} from  {}, the peer will be banned",
 				hash, peer_info.addr,
 			);
-			self.ban_peer(peer_info.addr, ReasonForBan::BadBlock);
+			self.ban_peer(peer_info.addr, ReasonForBan::BadBlock)
+				.map_err(|e| {
+					let err: chain::Error =
```
hashmap: Here and below - perhaps just return error `chain::ErrorKind::Other(format!("ban peer error :{:?}", e)).into()` instaead of defining a temp var and returnning it?

p2p/src/peers.rs
```diff
@@ -253,50 +253,43 @@ impl Peers {
...
-			}
-			Err(e) => error!("Couldn't unban {}: {:?}", peer_addr, e),
-		};
+		// check if peer exist
+		self.get_peer(peer_addr)?;
+		if self.is_banned(peer_addr) {
+			self.update_state(peer_addr, State::Healthy)?;
+			Ok(())
+		} else {
+			error!("Couldn't unban {}: peer is not banned", peer_addr);
```
hashmap: I'm not sure we should return error and log it in the same place, assuming we log the error higher on the call stack

p2p/src/peers.rs
```diff
@@ -253,50 +253,43 @@ impl Peers {
...
-					error!("Couldn't unban {}: peer is not banned", peer_addr);
-				}
-			}
-			Err(e) => error!("Couldn't unban {}: {:?}", peer_addr, e),
-		};
+		// check if peer exist
+		self.get_peer(peer_addr)?;
+		if self.is_banned(peer_addr) {
+			self.update_state(peer_addr, State::Healthy)?;
+			Ok(())
```
hashmap: I'd prefer to return on the previous line, I remember @antiochp like to be more explicit though

p2p/src/peers.rs
```diff
@@ -253,50 +253,43 @@ impl Peers {
...
+		self.update_state(peer_addr, State::Banned)?;
+
+		match self.get_connected_peer(peer_addr) {
+			Some(peer) => {
+				debug!("Banning peer {}", peer_addr);
+				// setting peer status will get it removed at the next clean_peer
+				peer.send_ban_reason(ban_reason)?;
+				peer.set_banned();
+				peer.stop();
+				let mut peers = match self.peers.try_write_for(LOCK_TIMEOUT) {
```
hashmap: Wdyt about writing it like:
```
let mut peers = self.peers.try_write_for(LOCK_TIMEOUT).ok_or_else(|| {
  error!("ban_peer: failed to get peers lock");
  Error::PeerException
})?;
```

p2p/src/peers.rs
quentinlesceller: I have no chance otherwise Rust complains (cannot infer type thing).

p2p/src/peers.rs
quentinlesceller: :+1:

p2p/src/peers.rs
quentinlesceller: indeed

p2p/src/peers.rs
quentinlesceller: Unfortunately cannot be done. Rust need type annotations for that one.

p2p/src/peers.rs
antiochp: [totally off-topic]

My concern is mainly around multiple calls in a block of code - 

```
// inconsistent across multiple lines (easy to miss that `three()` returns a Result)
{
    one()?;
    two()?;
    three()
}

// more consistency, less line noise if we introduce additional lines or reorder lines etc.
{
    one()?;
    two()?;
    three()?;
    Ok(())
}
```

Also we definitely have cases where you need the `?` to translate errors appropriately.
Not sure this is an issue here though.

I'm definitely coming around to single line blocks of code returning on that single line as you suggest @hashmap. As long as the errors convert correctly.

```
// Also an option if necessary
{
    Ok(one()?)
}
```

p2p/src/peers.rs
quentinlesceller: 👍

quentinlesceller:
Will merge once tests pass.

p2p/src/peers.rs
hashmap: Yeah, but requires more lines, may require to introduce a temp var and last but not least - we unwrap Result on line 3 and then wrap it on line 4, instead directly returning it as is. Same with the last approach. I hope the compiler produces the same code though.

PR #3094, opened 2019-10-11
Node API v2

quentinlesceller:
*Replacement of https://github.com/mimblewimble/grin/pull/3077 does not introduce any backward incompatibility.*

RFC here with more details https://github.com/mimblewimble/grin-rfcs/pull/28 
TLDR: Introduce a Node JSON-RPC API v2.

TODO:
- [x] `get_status`
- [x] `get_version`
- [x] `get_block`
- [x] `get_header`
- [x] `chain` methods
- [x] `pool` methods
- [x] `peers` methods
- [x] `txhashset` methods
- [x] Working documentation on docs.rs
- [x] ~~Doctest~~ In a different PR
- [x] ~~Internal use of API v2 for Grin~~In a different PR

BONUS:
`ban_peer`/`unban_peer` can be port specific.
`peers.ban_peer`/`peers.unban_peer` now return `Result<(),Error>`

## how to test this PR

```bash
git remote add quentinlesceller git://github.com/mimblewimble/grin.git
git fetch quentinlesceller
git checkout quentinlesceller/nodeapi
cargo build --release
```

Run grin and do POST queries with JSON-RPC body to http://127.0.0.1:3413/v2/owner and  http://127.0.0.1:3413/v2/foreign.


quentinlesceller:
Latest commit introduces a lot of things:
- Owner API: 
  - `ban_peer`
  - `unban_peer`
  - `get_connected_peers`
  - `get_peers`
  - `compact_chain`
  - `validate_chain`
  - `get_status`
- Foreign API:
  - `push_transaction`
  - `get_unconfirmed_transactions `
  - `get_stempool_size `
  - `get_pool_size `
  - `get_pmmr_indices `
  - `get_unspent_outputs `
  - `get_outputs `
  - `get_kernel `
  - `get_tip `
  - `get_version `
  - `get_block `
  - `get_header `

- Basic auth for Owner API is the same as the V1 basic auth (specified in `.api_secret`)
- Basic auth for Foreign API is different and set in `.foreign_api_secret`.

api/src/types.rs
```diff
@@ -266,7 +266,7 @@ pub struct OutputPrintable {
 	/// Rangeproof hash (as hex string)
 	pub proof_hash: String,
 	/// Block height at which the output is found
-	pub block_height: Option<u64>,
```
yeastplume: Why the change here? Just wondering if anything in the wallet will be expecting a ‘null’ here

yeastplume:
First, all looks very good and well done. I’m glad to see the approach done in the wallet duplicated here cause I think it works quite well.

Am okay with this in general, particularly as it runs alongside the v1 api which should give us time to adapt and asjust. There’s the obvious caveat that there may be some further changes required as I haven’t had any chance to test the behaviour of needed wallet functions against the V2 API. 

Also, definitely missing a trick here by not having doctest functions with sample JSON for each function. This should be easy to do, however it would likely require test framework code that currently only exists in the wallet. We really need to figure our our overall integration strategy, but that’s another discussion.

Good job here, and look forward to porting the wallet functions at some point between v3.0.0 and v4.0.0

api/src/types.rs
quentinlesceller: I don't remember why I changed the structure from `Option<u64>`to `u64`. Reverting.

PR #3095, opened 2019-10-14
mmr simplified Chinese translation

AurevoirXavier:
- add simplified Chinese translation for [mmr.md](https://github.com/mimblewimble/grin/blob/master/doc/mmr.md)
- add some missing star

Is the [line 106](https://github.com/mimblewimble/grin/blob/master/doc/mmr.md#L106) should be `Taking its left child is
subtracting 2^h.` instead of the `Taking its left sibling is subtracting 2^h.`?


quentinlesceller:
Hi @AurevoirXavier. Thank you for the PR!

> Is the line 106 should be Taking its left child is subtracting 2^h. instead of the Taking its left sibling is subtracting 2^h.?

Indeed it is! If you can fix in both docs (English and Chinese), that'd be great!

AurevoirXavier:
@quentinlesceller Done, :)

PR #3096, opened 2019-10-20
fix: split state validation status into kernel and rproof updates.

### LINKED ISSUE #3084: [TUI] Initial sync status/progress could be improved
The TUI is not doing a great job of displaying status and progress once we get into "validating chain state" after downloading the txhashset zip file.

It kind of just sits here, slowly creeping up in single digit % for what feels like a _long time_ -  

```
Validating chain state: 8%, step 3/4
```

Then it jumps from around 12% suddenly to step 4/4.

```
Downloading blocks: 99%, step 4/4
```

And here it just stays at 99% the _entire_ time until we are fully synchronized...

I'm' guessing we're downloading less than 1% of the overall blocks, which is great. But not great in terms of sync progress... We should probably use a % of blocks from the sync horizon, not since genesis.

tl;dr

It would be nice to have - 
* Improved % progress during state validation
* Improved % progress during final block download

JosephGoulden:
Fixes #3084 

I split out the sync validation status into kernel and range proof validation and added these as extra sync steps. Maybe this is too much information but I feel its nice to see progress being made when syncing and to help understand what's going on.
Also the percentage completion for range proofs and kernels is accurate from 0-100%.

There are a couple of things I'm not sure about though and could use some help.

1. The way I fixed the status was to add a function to the backend PMMR to get the leaf_set length, instead of using pmmr::n_leaves. Is this okay? I don't understand why the two functions give different results. Can't we just use the length of the leaf_set in all places instead of calculating leafs based on size in pmmr::n_leaves.

2. Breaking change in the server API - TxHashsetValidation replaced by TxHashsetRangeProofsValidation and TxHashsetKernelsValidation. Is this okay for a major version increment like 3.0.0 or do we need to depreciate first?


antiochp:
> The way I fixed the status was to add a function to the backend PMMR to get the leaf_set length, instead of using pmmr::n_leaves. Is this okay? I don't understand why the two functions give different results. 

Notice that `pmmr::n_leaves()` does not operate on any specific pmmr instance. It is simply a way of determining how many leaves are in an MMR (any MMR) of a particular size.
i.e. 
* An MMR with 3 elements in it will _always_ have 2 leaves (peak at height 1 with 2 leaves under it). 
* An MMR with 4 elements in it will _always_ have 3 leaves (peak at height 1, peak at height 0).

The `leaf_set` is not generic in this way - it represents a particular MMR instance, specifically one that can be pruned, with leaves being removed. 
So the `leaf_set` represents a sub set of _all_ leaves for that specific MMR that have not been removed. For the output MMR this represents the "unspent" outputs. All other outputs have been spent (removed) and potentially pruned away.

In the simple examples above we may have an MMR of size 4 (3 leaves) where we _only_ have a single leaf position in the `leaf_set`. Say at the last pos 4 in the MMR, with outputs at pos 1 and pos 2 having been removed.

```
   3                      3
  /\            ->       /\
 1  2  4                .  .  4
```

Note: outputs can be removed but not yet pruned because we need to support "rewind" in a fork situation where we undo a block and reapply a potentially different set of transactions to the chain state.
We may have pruned the outputs and pos 1 and pos 2 in the example above, leaving only the peak hash at pos 3. But we may have not yet pruned them, they may simply be removed (removed from the `leaf_set` but the data is still present in the underlying MMR files).

> Can't we just use the length of the leaf_set in all places instead of calculating leafs based on size in pmmr::n_leaves.

Not necessarily because we do not always want to know the number of leaves in the entire MMR. We have many instances where we need to know the number of leaves in a particular subtree or beneath a particular peak within an MMR.
And for the entire MMR we often need to know what the number of leaves would be if the MMR was not pruned (so we can calculate offsets etc into the underlying files) and we cannot use the `leaf_set` for this.

Hope this makes sense. The internals of the MMR data structure are pretty complex. But this complexity is worth it. The immutable, append-only semantics that it gives us though are _very_ well aligned with the data we need to store on disk.

core/src/core/pmmr/backend.rs
```diff
@@ -55,6 +55,9 @@ pub trait Backend<T: PMMRable> {
 	/// Iterator over current (unpruned, unremoved) leaf positions.
 	fn leaf_pos_iter(&self) -> Box<dyn Iterator<Item = u64> + '_>;
 
+	/// Number of leaves
+	fn n_leafs(&self) -> u64;
```
antiochp: We should think about a better name for this. Its not the number of leaves. Its the number of "not removed" leaf positions but that's a pretty awkward name.

servers/src/grin/sync/syncer.rs
```diff
@@ -192,7 +192,8 @@ impl SyncRunner {
 			match self.sync_state.status() {
 				SyncStatus::TxHashsetDownload { .. }
 				| SyncStatus::TxHashsetSetup
-				| SyncStatus::TxHashsetValidation { .. }
```
antiochp: @quentinlesceller What implications are there for changing the set of sync states here? Is this something we can safely do or should we maintain the existing set of states?

store/src/pmmr.rs
```diff
@@ -140,6 +140,10 @@ impl<T: PMMRable> Backend<T> for PMMRBackend<T> {
 		}
 	}
 
+	fn n_leafs(&self) -> u64 {
+		self.leaf_set.len() as u64
```
antiochp: This works for a "prunable" MMR like the output and rangeproof MMR where we remove (and then eventually prune) leaf positions. But I don't think this will work correctly for other MMRs like the kernel MMR and the header MMR where we only ever append and never remove. I don't think we use the `leaf_set` in these MMRs as we do not need to write this additional file to disk. _All_ leaves are "not removed" in the case of kernels, for example.

servers/src/grin/sync/syncer.rs
JosephGoulden: could add the TxHashsetValidation status back to maintain compatibility with any consumers?

antiochp:
Just testing this locally and the tui is just showing the following during rangeproof and signature validation - 
```
Sync step 3/7 - Preparing chain state for validation
```

I don't see it progress to `4/7` or `5/7`.

servers/src/grin/sync/syncer.rs
quentinlesceller: If think it's pretty safe to move forward and add this two new state. That could potentially break wallet using the status if they are doing strong type check but I presume it's okay for 3.0.0.

JosephGoulden:
> Just testing this locally and the tui is just showing the following during rangeproof and signature validation -
> 
> ```
> Sync step 3/7 - Preparing chain state for validation
> ```
> 
> I don't see it progress to `4/7` or `5/7`.

I think this might be broken on master. Seems to be around getting header stats whilst txhashset validating/rebuilding is going on. Looking into a fix.

JosephGoulden:
I think it was broke by this https://github.com/mimblewimble/grin/pull/3045

Not sure how to get the status now there is a quite long lived write lock on pmmr_header.

JosephGoulden:
@antiochp this is working now.

Another potential future improvement would be to have a separate state for catching up on the latest blocks, rather than flicking between sync state 1 and 7 (or 1 and 4 as it is now).

api/src/handlers/server_api.rs
```diff
@@ -113,16 +113,19 @@ fn sync_status_to_api(sync_status: SyncStatus) -> (String, Option<serde_json::Va
...
 		),
-		SyncStatus::TxHashsetValidation {
-			kernels,
-			kernel_total,
+		SyncStatus::TxHashsetRangeProofsValidation {
 			rproofs,
-			rproof_total,
+			rproofs_total,
+		} => (
+			"txhashset_rangeproof_validation".to_string(),
```
quentinlesceller: txhashset_rangeproof_validation -> txhashset_rangeproofs_validation

chain/src/types.rs
```diff
@@ -65,12 +65,15 @@ pub enum SyncStatus {
...
 	TxHashsetSetup,
-	/// Validating the full state
-	TxHashsetValidation {
+	/// Validating the kernels
+	TxHashsetKernelsValidation {
 		kernels: u64,
-		kernel_total: u64,
+		kernels_total: u64,
+	},
+	/// Validating the proofs
```
quentinlesceller: /// Validating the proofs -> /// Validating the range proofs ?

JosephGoulden:
Fixed those, thanks @quentinlesceller .

PR #3101, opened 2019-10-27
fix: add shutting down dialog to TUI

JosephGoulden:
Add dialog to let user know that server is shutting down. It appears as soon as 'q' is pressed and remains for a second or two until the process quits. This is instead of the current behaviour where the "Shutdown in progress, please wait" text appears awkwardly.


hashmap:
Looks great, thanks!

antiochp:
Yay! Just saw this and it looks really good. 🎉

PR #3102, opened 2019-10-29
optimize: avoid `insert` into `Vec`; use `into_iter` for less derefer…

AurevoirXavier:
- avoid `insert` into `Vec`
- use `into_iter` for less dereference


core/src/core/pmmr/pmmr.rs
```diff
@@ -99,10 +99,11 @@ where
 			.filter(|x| *x < peak_pos)
 			.filter_map(|x| self.backend.get_from_file(x))
 			.collect::<Vec<_>>();
-		res.reverse();
 		if let Some(rhs) = rhs {
-			res.insert(0, rhs);
+			res.push(rhs);
 		}
+		res.reverse();
```
antiochp: Nice. Good catch. :+1:

PR #3103, opened 2019-10-29
Retrieve outputs within a block height range

yeastplume:
Related to https://github.com/mimblewimble/grin-wallet/pull/246

The idea is to ensure there's an API to scan outputs committed within a specified range of blocks. This will be used during the wallet update process to scan the output set (similarly to how the current `check_repair` function works,) and and re-create any outputs that are missing due to forks or multiple wallets using the same seed.

This PR simply adds the function to the API, WIP until I've verified and tweaked against the current wallet PR.

### DIFF
--- a/api/src/handlers/transactions_api.rs
+++ b/api/src/handlers/transactions_api.rs
@@ -35,7 +35,7 @@ use std::sync::Weak;
 
 // UTXO traversal::
 // GET /v1/txhashset/outputs?start_index=1&max=100
-// GET /v1/txhashset/outputsbyheight?start_height=1&end_height=1000&max=100
+// GET /v1/txhashset/heightstopmmr?start_height=1&end_height=1000
 //
 // Build a merkle proof for a given pos
 // GET /v1/txhashset/merkleproof?n=1
@@ -81,7 +81,7 @@ impl TxHashSetHandler {
 		}
 		let chain = w(&self.chain)?;
 		let outputs = chain
-			.unspent_outputs_by_insertion_index(start_index, max, end_index)
+			.unspent_outputs_by_pmmr_index(start_index, max, end_index)
 			.context(ErrorKind::NotFound)?;
 		let out = OutputListing {
 			last_retrieved_index: outputs.0,
--- a/api/src/handlers/transactions_api.rs
+++ b/api/src/handlers/transactions_api.rs
@@ -69,14 +69,19 @@ impl TxHashSetHandler {
 	}
 
 	// allows traversal of utxo set
-	fn outputs(&self, start_index: u64, mut max: u64) -> Result<OutputListing, Error> {
+	fn outputs(
+		&self,
+		start_index: u64,
+		end_index: Option<u64>,
+		mut max: u64,
+	) -> Result<OutputListing, Error> {
 		//set a limit here
 		if max > 10_000 {
 			max = 10_000;
 		}
 		let chain = w(&self.chain)?;
 		let outputs = chain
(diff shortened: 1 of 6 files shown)


api/src/handlers/transactions_api.rs
```diff
@@ -121,15 +145,27 @@ impl Handler for TxHashSetHandler {
...
+		};
 
 		match right_path_element!(req) {
 			"roots" => result_to_response(self.get_roots()),
 			"lastoutputs" => result_to_response(self.get_last_n_output(last_n)),
 			"lastrangeproofs" => result_to_response(self.get_last_n_rangeproof(last_n)),
 			"lastkernels" => result_to_response(self.get_last_n_kernel(last_n)),
-			"outputs" => result_to_response(self.outputs(start_index, max)),
+			"outputs" => result_to_response(self.outputs(start_index, end_index, max)),
+			"heightstopmmr" => result_to_response(
```
antiochp: Subjective comment but "heightstopmmr" is a confusing name... "height stop mmr"?

Also not convinced we want to leak implementation details like this?
Seems like (at least intuitively) that you'd want to get outputs based on height directly, rather than retrieving the intermediate mmr positions.

Once we start doing things based on positions we inevitably go down the path of "is this 1-indexed or 0-index" and "is this leaf insertion indices or actual positions in the MMR"? And it is potientially confusing here because we are looking for leaves, the outputs themselves but via mmr positions, and presumably 1-indexed.

I'd much rather expose this via the api based on block height and keep the MMR implementation hidden behind the api.

yeastplume:
Going to merge based on discussions below. Note this change actually makes the PMMR scan a bit more consistent as it uses actual position exclusively as opposed to index. Can come back to this later on during a general cleanup on what the API exposes:

```
"heights to pmmr", it's a crap name but I didn't have anything better
So it sounds like you prefer the earlier approach, where the current 'outputs' function is changed to accept block heights instead?
antiochp
2:48 PM
I think so yeah (personal preference) - just hesitant to expose more of the MMR internals via the api, specifically as we are inconsistent in various usage in terms of 0-indexed and 1-indexed and leaf index vs position.
EDITED
exposing this via the api will potentially complicate reworking this to be more consistent (and will be a source of confusion for others)
yeastplume
2:48 PM
this was basically done this way to be as compatible as possible with the existing pmmr scan, which confusingly accepted a leaf index position but returned an actual position to continue the scan from on the next chunk of scan
antiochp
2:48 PM
that said - willing to be ignored convinced otherwise
EDITED
which confusingly accepted a leaf index position but returned an actual position to continue the scan from on the next chunk of scan
that's even worse than my examples...
EDITED
yeastplume
2:50 PM
heh.. right.. well that's because it was returning, say chunks of 1000 at a time, and it was more efficient to just return the last index and continue from there
antiochp
2:51 PM
maybe we don't overthink it right now and revisit it once we start getting more serious about consistent usage
yeastplume
2:51 PM
but I can get the point about exposing pmmr index details being confusing. I think just changing the output function to accept height instead should work just as well, it may be just slightly less efficient
antiochp
2:52 PM
I'm good with leaving it as you propose right now - and we can tackle the bigger picture later
(I suspect the ideal approach is to always use leaf indices for data itself, and only ever use positions for hashes)
EDITED
leaves by position is just an implementation detail
yeastplume
2:53 PM
yeah.. hmm would probably appreciate that as what I have is working as intended and there's a ton of other stuff to attend to. That function call is relatively well encapsulated anyhow so we can revisit at a later date on a pass that cleans up the API somewhat
👍

```

PR #3108, opened 2019-11-04
Enable faster sync

antiochp:
Non-consensus related implementation for RFC https://github.com/mimblewimble/grin-rfcs/pull/29

* refactor `vec_backend` so we can use it outside of tests
* introduce a "hash only" `vec_backend` for the accumulator
* introduce `bitmap_accumulator` -
  * MMR backed by a hash only vec backend
  * leaves are 1024 bit chunks of the leafset bitmap
  * it is relatively small, keep it in memory only
  * no need to store data in leaves (hash only) as data is maintained in the existing leafset
* implement "rewind" logic so we handle bitmap_accumulator correctly during a fork/reorg
  * it is not append-only as the bitmap updates over time
* build new bitmap accumulator root for each new block
  * log only (no consensus changes)

We will eventually be able to use this bitmap accumulator root to verify fragments of the leafset against subtrees of the existing output PMMR. This will allow us to verify "fast sync" data in chunks, without needing to download the full `txhashset.zip` which is increasingly becoming a bottleneck and single point of failure during initial fast sync.

----

TODO - 

- [x] test coverage for various bitmap accumulator scenarios (particularly rewind)
- [x] test coverage for bitmap fragment serialization (endianness etc)

### DIFF
--- a/chain/src/txhashset/bitmap_accumulator.rs
+++ b/chain/src/txhashset/bitmap_accumulator.rs
@@ -20,9 +20,7 @@ use croaring::Bitmap;
 
 use crate::core::core::hash::{DefaultHashable, Hash};
 use crate::core::core::pmmr::{self, ReadonlyPMMR, VecBackend, PMMR};
-use crate::core::ser::{
-	self, FixedLength, PMMRIndexHashable, PMMRable, Readable, Reader, Writeable, Writer,
-};
+use crate::core::ser::{self, FixedLength, PMMRable, Readable, Reader, Writeable, Writer};
 use crate::error::{Error, ErrorKind};
 
 /// The "bitmap accumulator" allows us to commit to a specific bitmap by splitting it into
@@ -57,27 +55,44 @@ impl BitmapAccumulator {
 	}
 
 	/// Initialize a bitmap accumulator given the provided idx iterator.
-	pub fn init<T: IntoIterator<Item = u64>>(&mut self, idx: T) -> Result<(), Error> {
-		self.apply_from(idx, 0)
+	pub fn init<T: IntoIterator<Item = u64>>(&mut self, idx: T, size: u64) -> Result<(), Error> {
+		self.apply_from(idx, 0, size)
 	}
 
-	/// Apply the provided idx iterator starting at the provided chunk idx.
-	fn apply_from<T>(&mut self, idx: T, from_chunk_idx: u64) -> Result<(), Error>
+	/// Find the start of the first "chunk" of 1024 bits from the provided idx.
+	/// Zero the last 10 bits to round down to multiple of 1024.
+	pub fn chunk_start_idx(idx: u64) -> u64 {
+		idx & !0x3ff
+	}
+
+	/// The first 1024 belong to chunk 0, the next 1024 to chunk 1 etc.
+	fn chunk_idx(idx: u64) -> u64 {
+		idx / 1024
+	}
+
+	/// Apply the provided idx iterator to our bitmap accumulator.
+	/// We start at the chunk containing from_idx and rebuild chunks as necessary
+	/// for the bitmap, limiting it to size (in bits).
+	/// If from_idx is 1023 and size is 1024 then we rebuild a single chunk.
+	fn apply_from<T>(&mut self, idx: T, from_idx: u64, size: u64) -> Result<(), Error>
 	where
 		T: IntoIterator<Item = u64>,
 	{
 		let now = Instant::now();
 
+		// Find the (1024 bit chunk) chunk_idx for the (individual bit) from_idx.
+		let from_chunk_idx = BitmapAccumulator::chunk_idx(from_idx);
 		let mut chunk_idx = from_chunk_idx;
+
 		let mut chunk = BitmapChunk::new();
 
-		let mut leaves = idx.into_iter().peekable();
-		while let Some(x) = leaves.peek() {
+		let mut idx_iter = idx.into_iter().filter(|&x| x < size).peekable();
+		while let Some(x) = idx_iter.peek() {
 			if *x < chunk_idx * 1024 {
 				// skip until we reach our first chunk
-				leaves.next();
+				idx_iter.next();
 			} else if *x < (chunk_idx + 1) * 1024 {
-				let idx = leaves.next().expect("next after peek");
+				let idx = idx_iter.next().expect("next after peek");
 				chunk.set(idx % 1024, true);
 			} else {
 				self.append_chunk(chunk)?;
@@ -99,10 +114,14 @@ impl BitmapAccumulator {
 	}
 
 	/// Apply updates to the bitmap accumulator given an iterator of invalidated idx and
-	/// an additional iterator of new idx to be added.
+	/// an iterator of idx to be set to true.
 	/// We determine the existing chunks to be rebuilt given the invalidated idx.
-	/// We then add new idx, extending the accumulator with new chunk(s) as necessary.
-	pub fn apply<T, U>(&mut self, invalidated_idx: T, new_idx: U) -> Result<(), Error>
+	/// We then rebuild given idx, extending the accumulator with new chunk(s) as necessary.
+	/// Resulting bitmap accumulator will contain sufficient bitmap chunks to cover size.
+	/// If size is 1 then we will have a single chunk.
+	/// If size is 1023 then we will have a single chunk (bits 0 to 1023 inclusive).
+	/// If the size is 1024 then we will have two chunks.
+	pub fn apply<T, U>(&mut self, invalidated_idx: T, idx: U, size: u64) -> Result<(), Error>
 	where
 		T: IntoIterator<Item = u64>,
 		U: IntoIterator<Item = u64>,
@@ -112,21 +131,19 @@ impl BitmapAccumulator {
 		// Note: We rebuild everything after rewind point but much of the bitmap may be
 		// unchanged. This can be further optimized by only rebuilding necessary chunks and
 		// rehashing.
-		if let Some(first_chunk_idx) = invalidated_idx.into_iter().next().map(|x| x / 1024) {
-			self.rewind_prior(first_chunk_idx)?;
-			self.pad_left(first_chunk_idx)?;
-			self.apply_from(new_idx, first_chunk_idx)?;
+		if let Some(from_idx) = invalidated_idx.into_iter().next() {
+			self.rewind_prior(from_idx)?;
+			self.pad_left(from_idx)?;
+			self.apply_from(idx, from_idx, size)?;
 		}
-
-		// Trim any "empty" chunks from the right to keep everything consistent.
-		self.trim_right()?;
 
 		Ok(())
 	}
 
-	/// Given the provided chunk idx rewind the bitmap accumulator to the end of the
+	/// Given the provided (bit) idx rewind the bitmap accumulator to the end of the
 	/// previous chunk ready for the updated chunk to be appended.
-	fn rewind_prior(&mut self, chunk_idx: u64) -> Result<(), Error> {
+	fn rewind_prior(&mut self, from_idx: u64) -> Result<(), Error> {
+		let chunk_idx = BitmapAccumulator::chunk_idx(from_idx);
 		let last_pos = self.backend.size();
 		let mut pmmr = PMMR::at(&mut self.backend, last_pos);
 		let chunk_pos = pmmr::insertion_to_pmmr_index(chunk_idx + 1);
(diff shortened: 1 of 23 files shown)


antiochp:
We now log this for every new block - 

```
20191115 11:48:41.272 DEBUG grin_chain::types - validate roots: 00000c3194d8 at 435707, 
output_root: 4856008b7c31, output pmmr: 4856008b7c31 (bitmap: 7a488eb915b8, merged: 454f61a2a6df)
```

It would be good to get this merged and have multiple nodes deployed with this code so we can verify the `merged` root looks acceptable across a wider sample of nodes on the network.

Note: We are not currently using the `merged` root for validation, simply logging it for visibility.
The proposal is to begin validating this root at the HF.

quentinlesceller:
@antiochp taking a look today.

tromp:
will look over this weekend

chain/src/txhashset/bitmap_accumulator.rs
```diff
@@ -0,0 +1,241 @@
...
+
+		Ok(())
+	}
+
+	/// Given the provided chunk idx rewind the bitmap accumulator to the end of the
+	/// previous chunk ready for the updated chunk to be appended.
+	fn rewind_prior(&mut self, chunk_idx: u64) -> Result<(), Error> {
+		let last_pos = self.backend.size();
+		let mut pmmr = PMMR::at(&mut self.backend, last_pos);
+		let chunk_pos = pmmr::insertion_to_pmmr_index(chunk_idx + 1);
```
tromp: doesn't a chunk index need to be multiplies by 1024 to get an MMR insertion index?

chain/src/txhashset/bitmap_accumulator.rs
```diff
@@ -0,0 +1,241 @@
...
+			.map_err(|e| ErrorKind::Other(e))?;
+		Ok(())
+	}
+
+	/// Make sure we append empty chunks to fill in any gap before we append the chunk
+	/// we actually care about. This effectively pads the bitmap with 1024 chunks of 0s
+	/// as necessary to put the new chunk at the correct place.
+	fn pad_left(&mut self, chunk_idx: u64) -> Result<(), Error> {
+		let last_pos = self.backend.size();
+		let chunk_pos = pmmr::insertion_to_pmmr_index(chunk_idx);
```
tromp: ditto

chain/src/txhashset/bitmap_accumulator.rs
```diff
@@ -0,0 +1,241 @@
...
+	/// Create a new bitmap chunk, defaulting all bits in the chunk to false.
+	pub fn new() -> BitmapChunk {
+		BitmapChunk(BitVec::from_elem(Self::LEN_BITS, false))
+	}
+
+	/// Set a single bit in this chunk.
+	/// 0-indexed from start of chunk.
+	pub fn set(&mut self, idx: u64, value: bool) {
+		let idx = usize::try_from(idx).expect("usize from u64");
+		if idx < Self::LEN_BITS {
```
tromp: what if idx >= LEN_BITS? if that should never happen, then we should assert that instead.
silently ignoring it looks bad.

chain/src/txhashset/txhashset.rs
```diff
@@ -148,6 +151,13 @@ impl TxHashSet {
 			header,
 		)?;
 
+		let bitmap_accumulator = {
+			let output_pmmr = ReadonlyPMMR::at(&output_pmmr_h.backend, output_pmmr_h.last_pos);
+			let mut bitmap_accumulator = BitmapAccumulator::new();
+			bitmap_accumulator.init(&mut output_pmmr.leaf_idx_iter(0))?;
+			bitmap_accumulator
+		};
+
```
tromp: Is this code block equivalent to just lines 155-157 ?

chain/src/txhashset/bitmap_accumulator.rs
```diff
@@ -0,0 +1,241 @@
...
+		let last_pos = self.backend.size();
+		let chunk_pos = pmmr::insertion_to_pmmr_index(chunk_idx);
+		for _ in last_pos..chunk_pos {
+			self.append_chunk(BitmapChunk::new())?;
+		}
+		Ok(())
+	}
+
+	/// Recursively trim the last chunk if empty. Stop when rightmost chunk is non-empty
+	/// or the accumulator itself is empty.
```
tromp: What does "empty" mean? Is that all 0s? I.e. all spent outputs? But for every spent output there's a later unspent output, so how does this trimming work?

chain/src/txhashset/txhashset.rs
```diff
@@ -901,28 +923,46 @@ impl<'a> Extension<'a> {
...
 
-	fn apply_input(&mut self, input: &Input) -> Result<(), Error> {
+	fn apply_to_bitmap_accumulator(&mut self, output_pos: &[u64]) -> Result<(), Error> {
+		let mut output_idx: Vec<_> = output_pos
+			.iter()
+			.map(|x| pmmr::n_leaves(*x).saturating_sub(1))
+			.collect();
+		output_idx.sort_unstable();
+		let min_idx = output_idx.first().unwrap_or(&0);
+		let chunk_start_idx = (min_idx / 1024) * 1024;
```
tromp: equivalent to min_idx & ~0x3ff; // zero last 10 bits

chain/tests/bitmap_accumulator.rs
```diff
@@ -0,0 +1,142 @@
...
+
+	let mut bit_vec = BitVec::from_elem(1024, false);
+	let mut accumulator = BitmapAccumulator::new();
+	assert_eq!(accumulator.root(), Hash::default());
+
+	// 1000... (rebuild from 0, setting [0] true)
+	accumulator.apply(vec![0], vec![0]).unwrap();
+	bit_vec.set(0, true);
+	let expected_bytes = bit_vec.to_bytes();
+	let expected_hash = expected_bytes.hash_with_index(0);
```
tromp: would't expected_bytes have to be LEN_BYTES long?

core/src/ser.rs
```diff
@@ -860,8 +860,14 @@ pub trait AsFixedBytes: Sized + AsRef<[u8]> {
 	/// The length in bytes
 	fn len(&self) -> usize;
 }
-
-impl<'a> AsFixedBytes for &'a [u8] {
+// impl<'a> AsFixedBytes for &'a [u8] {
+// 	// TODO - This does not feel right?
```
tromp: shouldn't this be the array length??

core/src/ser.rs
```diff
@@ -860,8 +860,14 @@ pub trait AsFixedBytes: Sized + AsRef<[u8]> {
...
+// 	fn len(&self) -> usize {
+// 		1
+// 	}
+// }
+impl AsFixedBytes for &[u8] {
+	// TODO - This does not feel right?
```
tromp: Then what would be right?
Why do you need to implement a length function for byte arrays?

chain/src/txhashset/bitmap_accumulator.rs
antiochp: Not here no. We are rewinding the MMR that backs the "accumulator" so each pos here represents a 1024 bit chunk.

chain/src/txhashset/bitmap_accumulator.rs
antiochp: See above.

chain/src/txhashset/bitmap_accumulator.rs
antiochp: Yes agreed. The silent ignore is bad practice.

chain/src/txhashset/bitmap_accumulator.rs
tromp: ah, yes, you're right. this MMR is the UBMT...

chain/src/txhashset/txhashset.rs
antiochp: Yes but with some caveats. The reason it was wrapped in a block was we need `bitmap_accumulator` to be `mut` so we can call `init()` on it. but once it is initialized we don't need it to be `mut`.
So the block wraps this `mut` requirement up, hides it and returns an immutable instance.
Makes the intent of the code clearer.

We also only need the `output_pmmr` for building the `bitmap_accumulator` so no need to expose it outside the block of code. 

Its a common rust pattern to limit scope as much as possible (both for variables and `mut`).

chain/src/txhashset/bitmap_accumulator.rs
antiochp: Yes exactly - empty here means all zero values.

I actually went back and forth between these two competing (and somewhat ambiguous) interpretations -
a) bitmap is the txo size rounded up to multiple of 1024
b) bitmap is minimum required to cover the final 1 value, rounded up to multiple of 1024

I ended up going with (b) for a couple of reasons - 
1) roaring bitmap actually only represents 1 values, so without explicitly providing the txo size its undefined exactly how big the bitmap needs to be
2) in practice this question is moot - the txo set always includes a final 1 value because the situation where we create an output and then spend it in the _same_ block would not result in a 0 value due to block level cut-through. So the last pos in the bitmap is always a 1.

So keeping the bitmap minimum size is easier to implement and does not require the txo size to be explicitly specified when building bitmaps.

chain/src/txhashset/txhashset.rs
tromp: isn't it possible to make the constructor accept the argument of init BitmapAccumulator::new(output_pmmr.leaf_idx_iter(0)) so we don't need to play these mutable games?

chain/src/txhashset/bitmap_accumulator.rs
tromp: but does 0 mean spent or unspent? if 0 is spent then a) and b) come out the same, since the very last txo is always unspent.

antiochp: 1 is unspent, 0 spent

chain/tests/bitmap_accumulator.rs
antiochp: Are you saying we should just assert this in the test? Or something else?

chain/src/txhashset/bitmap_accumulator.rs
tromp: If the last pos is always a 1 (unspent), then why is there a need to trim at all?

chain/tests/bitmap_accumulator.rs
tromp: No; I'm puzzled how this test can work, since I expect the UBMT to compute hashes of input length LEN_BYTES at the leaves, while this test hashes expected_bytes whose length I'm not sure how to determine but it appears to be smaller than that, with only bit 0 set.

antiochp: `expected_bytes` comes from `bit_vec` which itself is initialized as 1024 false values - 

https://github.com/mimblewimble/grin/blob/a3bae98b3f306c9d65caa3b2616a0beb2046c5c9/chain/tests/bitmap_accumulator.rs#L27

So `expected_bytes` is 1024 bits in length which corresponds to `LEN_BYTES`.

chain/src/txhashset/bitmap_accumulator.rs
antiochp: I think I did this so we did not have undefined behavior around the bitmap itself. We only have a "last pos is always 1" due to our block specific behavior. There is nothing intrinsic to the bitmap accumulator that would ensure the last pos will always be a 1.

Its possible to write test cases where this "last pos is always 1" does not hold and its nice to have defined unambiguous behavior here.

core/src/ser.rs
antiochp: Yeah this was just something I noticed when looking at this. Its not related to this PR so I'm going to revert these changes and discuss elsewhere - but something feels wrong with this code to me.

antiochp: Moved this to a separate PR. See #3131 for wider cleanup of `AsFixedBytes`.

chain/src/txhashset/bitmap_accumulator.rs
antiochp: Rethinking this. Lets make the `size` simply the index of the rightmost 1 value in the bitmap. 

So we rely on the assumption that our usage involves an unspent output at the rightmost position of the output MMR. But this should simplify how we go about making sure the bitmap is in a canonical state with no "empty" chunks incorrectly appended at the end.

Our usage guarantees this - a new block always contains at least one unspent output (the coinbase output). Any other outputs that appear after this in canonical output ordering must always be unspent also, otherwise they would have been cut-through.

And we can reflect this in the various tests by simply making sure the rightmost bit is always set to true.

tromp: > the index of the rightmost 1 value in the bitmap.

Isn't that 1 less than the number of txos? If so, I would just use that instead, as it's an actual size.
No need for special casing pre-genesis either, when there is no rightmost 1 value.

tromp: When increasing this value, we can simply do an assert than the prior bit is 1...

antiochp:
I'm going to merge this now. It adds the accumulator without relying on it yet. And adds logging around the accumulator which I'd like to start seeing on multiple nodes on `mainnet` and `floonet` sooner rather than later.
There are no consensus changes in this PR - these are coming soon as part of the scheduled HF2.

PR #3109, opened 2019-11-11
Increase IO_TIMEOUT to allow nodes on high-latency connections to sync

mmgen:
Commit d3dbafa80b "Use blocking IO in P2P to reduce CPU load" (merged into v2.1.0) introduced the constant IO_TIMEOUT, setting it to 1 second.

On nodes with high-latency connections, this short timeout causes the txhashset archive download during step 2 of the IBD process to invariably fail before it completes.  Since there's no mechanism for resuming a failed download, this means the node gets stuck at this stage and never syncs.

Increasing IO_TIMEOUT to 10 seconds solves the issue on my node; others might suggest a more optimal value for the constant.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -38,7 +38,7 @@ use std::{
 	thread::{self, JoinHandle},
 };
 
-const IO_TIMEOUT: Duration = Duration::from_millis(1000);
+const IO_TIMEOUT: Duration = Duration::from_millis(10000);
 
 /// A trait to be implemented in order to receive messages from the
 /// connection. Allows providing an optional response.


DavidBurkett:
Does this change affect shutdown times?

mmgen:
Doesn't seem to.

hashmap:
Yes, it would increase shutdown time. Also we need to figure out why it fails the download, timeout is not an exceptional situation, it doesn't stop reading from a peer. Perhaps we partially read, get timeout and drop the buffer, I'll check it. This would be a short-term fix.
In long term we need to think about using async io on top of tokio or async-std. We used tokio for p2p and got rid of it in early 2018 because it was hard to maintain and reason about (we still have tokio in api layer and it's a pain), but with the latest stable rust brings us `async/.await` which makes async io is much more pleasant to work with. However this change would require some redesign in message processing area, we need to separate network io from header/block processing.

mmgen:
Agree this is a short-term fix.  But it does solve the problem and doesn't seem to cause any serious issues of its own.

hashmap:
Interesting. And I was wrong about longer shutdown, it would be the case with the original PR, but then we changed the behavior, now we just close remaining threads after short period of time

mmgen:
I've been unable to observe any difference in the patched node's behavior. Aside from the fact that it now syncs :-)

PR #3111, opened 2019-11-13
add tx pool kernel counts to tui status page

antiochp:
We added tx pool size to the tui status page previously.
This PR adds the kernel count in addition to this.
This provides a rough approximation of how much tx aggregation is occurring by comparing pool size (in transactions) to the size in kernels.


antiochp:
Good catch. Will fix.

PR #3119, opened 2019-11-14
fix: for TUI status try to acquire read lock on pmmr_header but if no…

JosephGoulden:
TUI status is currently broken when syncing the chain. It stops responding and does not show the percentage validation status of the txhashset. This is because during validation there is long lived write lock on pmmr_header. So when we try to get a read lock to update the status it blocks.

This fix adds a try_header_head() function which might return a Tip or not if the lock is not available right away. For TUI status this is okay as we can just update based on the available information and do not need to block.

### DIFF
--- a/servers/src/grin/server.rs
+++ b/servers/src/grin/server.rs
@@ -506,17 +506,17 @@ impl Server {
 			total_difficulty: head.total_difficulty(),
 		};
 
-		let header_tip = self.chain.try_header_head()?;
-		let header_stats = if let Some(tip) = header_tip {
-			let header = self.chain.get_block_header(&tip.hash())?;
-			Some(ChainStats {
-				latest_timestamp: header.timestamp,
-				height: header.height,
-				last_block_h: header.prev_hash,
-				total_difficulty: header.total_difficulty(),
-			})
-		} else {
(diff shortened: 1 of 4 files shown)


servers/src/grin/server.rs
```diff
@@ -506,13 +506,17 @@ impl Server {
...
-		let header_tip = self.chain.header_head()?;
-		let header = self.chain.get_block_header(&header_tip.hash())?;
-		let header_stats = ChainStats {
-			latest_timestamp: header.timestamp,
-			height: header.height,
-			last_block_h: header.prev_hash,
-			total_difficulty: header.total_difficulty(),
+		let header_tip = self.chain.try_header_head()?;
+		let header_stats = if let Some(tip) = header_tip {
+			let header = self.chain.get_block_header(&tip.hash())?;
```
hashmap: Nitpick - perhaps `match` would be more readable here, `if let` shines when we care about one variant

JosephGoulden: You're right that is better :)

antiochp:
Some day we'll actually have useful read locks around some of this that actually allow validation to happen without blocking.

PR #3125, opened 2019-11-18
feat: allow DNS names in peers/seeds list and resolve them

### LINKED ISSUE #3120: allow FQDN in peers section
**Is your feature request related to a problem? Please describe.**
adding domain names FQDN instead of IP addresses when connection to preferred or allowed peers

**Describe the solution you'd like**
example:
peers_preferred = ["DOMAIN.tld"]

JosephGoulden:
**This change is for after 3.0 release/branch**

Enhancement to allow DNS names in the config for seeds, allow/deny/preferred peers and dandelion peer. Fixes #3120 

Port is always required otherwise a panic will occur when attempting to parse the domain name.
There is a slight confusion caused by the seed type of DNSSeed, as the List type can now contain DNS seeds as well. Maybe DNSSeed could be deprecated in favour of DefaultSeed.

I've manually tested adding DNS names peers list and works as expected.


p2p/src/types.rs
```diff
@@ -223,18 +223,18 @@ pub struct P2PConfig {
...
 	/// The list of seed nodes, if using Seeding as a seed type
-	pub seeds: Option<Vec<PeerAddr>>,
+	pub seeds: Option<Vec<String>>,
 
 	/// Capabilities expose by this node, also conditions which other peers this
 	/// node will have an affinity toward when connection.
 	pub capabilities: Capabilities,
 
-	pub peers_allow: Option<Vec<PeerAddr>>,
+	pub peers_allow: Option<Vec<String>>,
```
hashmap: It reduces type safety, can we somehow keep it while supporting the new functionality? I haven't checked the code carefully, but perhaps smth like implementing TryFrom/TryInto would help to hide the logic of conversion.

JosephGoulden: Could potentially implement a Serde deserializer for Vec<PeerAddr>. But it wouldn't be just deserializing it would be resolving the DNS names as well. And there are potentially multiple IP addresses for each DNS names which complicates things. I can give it a go.

servers/src/grin/seed.rs
```diff
@@ -361,35 +361,42 @@ fn listen_for_addrs(
...
+		)
 	})
 }
 
+pub fn resolve_dns_to_addrs(dns_records: &Vec<String>) -> Vec<PeerAddr> {
+	let mut addresses: Vec<PeerAddr> = vec![];
+	for dns in dns_records {
+		debug!("Retrieving addresses from dns {}", dns);
+		match dns.to_socket_addrs() {
+			Ok(addrs) => addresses.append(&mut addrs.map(|addr| PeerAddr(addr)).collect()),
```
quentinlesceller: Seems like you are not removing the duplicates addresses anymore?

p2p/src/types.rs
```diff
@@ -161,6 +166,45 @@ impl Readable for PeerAddr {
 	}
 }
 
+impl<'de> Visitor<'de> for PeerAddrs {
```
quentinlesceller: I'm *not sure* you need to implement the deserializer manually. Why not simply add the serde `Deserialize` trait to `PeerAddrs`?

JosephGoulden: Because it's not just deserializing, its resolving the DNS name into one or many IP addresses by calling to_socket_addrs(). Serde wouldn't know how to do that as far as I'm aware.

servers/src/grin/seed.rs
JosephGoulden: Not sure why I removed that, have added it back.

antiochp:
We should hold off merging this until after 3.0.0 is released. This is potentially wide-reaching in terms of its effect on the network as a whole.

@JosephGoulden Can you add something to the top of the PR description to remind us?

p2p/src/types.rs
quentinlesceller: 👍

p2p/src/types.rs
quentinlesceller: Outdated, resolved.

DavidBurkett:
3.0.0 has long been released. Should we get this merged now?

PR #3127, opened 2019-11-20
Fixing misleading privacy claim

DavidBurkett:
It was recently pointed out that, although our privacy model was correct, our readme was misleading. This replaces the misleading claim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 Grin is an in-progress implementation of the Mimblewimble protocol. Many characteristics are still undefined but the following constitutes a first set of choices:
 
   * Clean and minimal implementation, and aiming to stay as such.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,7 +11,7 @@
 Grin is an in-progress implementation of the Mimblewimble protocol. Many characteristics are still undefined but the following constitutes a first set of choices:
 
   * Clean and minimal implementation, and aiming to stay as such.
-  * Follows the Mimblewimble protocol, which provides great anonymity and scaling characteristics.
+  * Follows the Mimblewimble protocol, which provides confidential transactions and scaling advantages.
```
mably: CT won't probably say much to most people, I would add something like `(hidden amounts)` behind.

DavidBurkett: Fixed.

PR #3129, opened 2019-11-20
use try_read_for() with timeout for tui status updates (header and pool stats)

### LINKED ISSUE #3128: header sync stats no longer reflected in tui accurately
Looks like the recent changes to how we read the `header_head` to update the tui during header sync has broken the actual header stats update. On the tui these no longer reflect the actual state of the sync and get stale, showing old values.

I'm 99% sure its related to these changes where we no longer block on reading the header - 

```
let header_stats = match self.chain.try_header_head()?
```

The issue is if we don't block then we don't actually reliably read the data...

So we're kind of stuck between having the TUI block if unable to read the header and not blocking and not having the TUI update at all during something intensive like the header sync where its hard to obtain the lock to read the header here.

antiochp:
Resolves #3128.

Use `try_read_for()` so we block when obtaining the read lock _but_ we only block for a short period of time.
If we do not block then we risk having the TUI fail to update when updating stats (and displaying stale stats).

https://docs.rs/lock_api/0.3.1/lock_api/struct.RwLock.html#method.try_read_for

We already use `try_read_for()` internally when reading our list of connected peers in p2p layer.

Refactored `tx_stats` to be `Option<>` so we do not revert to 0 values in the case where we are unable to obtain the read lock to update the stats. Only update stats if we have `Some` and just leave stats unchanged on `None`.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1216,15 +1216,15 @@ impl Chain {
 			.map_err(|e| ErrorKind::StoreErr(e, "chain tail".to_owned()).into())
 	}
 
-	/// Tip (head) of the header chain if read lock can be acquired right now.
+	/// Tip (head) of the header chain if read lock can be acquired reasonably quickly.
+	/// Used by the TUI when updating stats to avoid locking the TUI up.
 	pub fn try_header_head(&self) -> Result<Option<Tip>, Error> {
-		match self.header_pmmr.try_read() {
-			Some(lock) => {
-				let hash = lock.head_hash()?;
-				let header = self.store.get_block_header(&hash)?;
-				Ok(Some(Tip::from_header(&header)))
-			}
-			None => Ok(None),
+		if let Some(pmmr) = self.header_pmmr.try_read_for(Duration::from_millis(500)) {
+			let hash = pmmr.head_hash()?;
+			let header = self.store.get_block_header(&hash)?;
+			Ok(Some(Tip::from_header(&header)))
+		} else {
+			Ok(None)
 		}
 	}
 
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1218,19 +1218,21 @@ impl Chain {
 
 	/// Tip (head) of the header chain if read lock can be acquired reasonably quickly.
 	/// Used by the TUI when updating stats to avoid locking the TUI up.
-	pub fn try_header_head(&self) -> Result<Option<Tip>, Error> {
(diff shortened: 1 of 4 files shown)


chain/src/chain.rs
```diff
@@ -1216,15 +1216,15 @@ impl Chain {
...
+	/// Used by the TUI when updating stats to avoid locking the TUI up.
 	pub fn try_header_head(&self) -> Result<Option<Tip>, Error> {
-		match self.header_pmmr.try_read() {
-			Some(lock) => {
-				let hash = lock.head_hash()?;
-				let header = self.store.get_block_header(&hash)?;
-				Ok(Some(Tip::from_header(&header)))
-			}
-			None => Ok(None),
+		if let Some(pmmr) = self.header_pmmr.try_read_for(Duration::from_millis(500)) {
```
hashmap: Perhaps externalize 500 as a constant?

servers/src/grin/server.rs
```diff
@@ -506,17 +498,16 @@ impl Server {
...
-				let header = self.chain.get_block_header(&tip.hash())?;
-				Some(ChainStats {
-					latest_timestamp: header.timestamp,
-					height: header.height,
-					last_block_h: header.prev_hash,
-					total_difficulty: header.total_difficulty(),
-				})
-			}
-			_ => None,
+		let header_stats = if let Some(head) = self.chain.try_header_head()? {
```
hashmap: I personally find `match` more readable than exhaustive `if let /else`, but it's a matter of taste

servers/src/grin/server.rs
```diff
@@ -480,23 +480,15 @@ impl Server {
...
-					tx_pool_kernels: 0,
-					stem_pool_size: 0,
-					stem_pool_kernels: 0,
-				},
-			}
-		};
+		let tx_stats = self
+			.tx_pool
+			.try_read_for(Duration::from_millis(500))
+			.map(|pool| TxStats {
```
hashmap: Really nice

servers/src/grin/server.rs
antiochp: Yeah maybe I'll revert this back to `match`. I'm not really sure which I prefer in cases like this.

chain/src/chain.rs
antiochp: Actually moved it out and we now pass it in as a parameter.

servers/src/grin/server.rs
antiochp: Now using `match` and a `map` like we do for the tx_stats above. I think this reads better now.

PR #3131, opened 2019-11-20
Cleanup redundant AsFixedBytes and FixedLength traits

antiochp:
We had two relatively awkward traits `AsFixedBytes` and `FixedLength` traits that basically encapsulated a way of determining how many bytes a "fixed size" struct would serialize to.

`AsFixedBytes` basically wrapped `AsRef<u8>` and acted as a marker trait. The `len()` functon was never actually used anywhere as we always converted to a byte slice via `to_ref()` and then called `len()` on the byte slice directly. I suspect this has simply been refactored over time and ended up being unused without us realizing it.

`FixedLength` was used for the PMMR backend impl to determine how many bytes to use for each leaf entry in the PMMR. It ends up being far simpler to replace this with an `elmt_size()` fn on `PMMRable` for the few structs we actually care about maintaining in PMMRs. And call this directly when instantiating the PMMR backend.

This all came about because we spotted the following code and something felt wrong here - 

```
impl<'a> AsFixedBytes for &'a [u8] {
	fn len(&self) -> usize {
		1
	}
}
```

The length in bytes of a slice of u8 values should not be 1.
So this specific instance of `AsFixedBytes` appeared to be unused (or we had bigger problems during serialization/deserialization).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ test_output
 wallet_data
 wallet/db
 .idea/
+/node*
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -26,14 +26,14 @@ use crate::core::{
 use crate::global;
 use crate::pow::{verify_size, Difficulty, Proof, ProofOfWork};
 use crate::ser::{
-	self, deserialize_default, serialize_default, PMMRable, Readable, Reader,
-	Writeable, Writer,
+	self, deserialize_default, serialize_default, PMMRable, Readable, Reader, Writeable, Writer,
 };
 use chrono::naive::{MAX_DATE, MIN_DATE};
 use chrono::prelude::{DateTime, NaiveDateTime, Utc};
 use chrono::Duration;
 use keychain::{self, BlindingFactor};
 use std::collections::HashSet;
+use std::convert::TryInto;
 use std::fmt;
 use std::iter::FromIterator;
 use std::sync::Arc;
@@ -168,11 +168,6 @@ impl Writeable for HeaderEntry {
 	}
 }
 
-impl HeaderEntry {
-	/// Length of a HeaderEntry in bytes.
-	pub const LEN: u16 = 32 + 8 + Difficulty::LEN + 4 + 1;
-}
-
 impl Hashed for HeaderEntry {
 	/// The hash of the underlying block.
 	fn hash(&self) -> Hash {
@@ -267,8 +262,10 @@ impl PMMRable for BlockHeader {
 		}
 	}
 
+	// Size is hash + u64 + difficulty + u32 + u8.
 	fn elmt_size() -> Option<u16> {
-		Some(HeaderEntry::LEN)
+		const LEN: usize = Hash::LEN + 8 + Difficulty::LEN + 4 + 1;
+		Some(LEN.try_into().unwrap())
 	}
 }
 
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -26,7 +26,7 @@ use crate::core::{
 use crate::global;
 use crate::pow::{verify_size, Difficulty, Proof, ProofOfWork};
 use crate::ser::{
-	self, deserialize_default, serialize_default, FixedLength, PMMRable, Readable, Reader,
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -12,3 +12,4 @@ test_output
 wallet_data
 wallet/db
 .idea/
+/node*
```
quentinlesceller: Out of curiosity why?

core/src/core/block.rs
```diff
@@ -261,6 +262,10 @@ impl PMMRable for BlockHeader {
 			is_secondary: self.pow.is_secondary(),
 		}
 	}
+
+	fn elmt_size() -> Option<u16> {
+		Some(HeaderEntry::LEN)
```
quentinlesceller: Is the const HeaderEntry ::LEN used anywhere else? If not doesn't moving it inside the `elmt_size`makes sense?

core/src/ser.rs
```diff
@@ -609,24 +604,23 @@ impl Readable for RangeProof {
...
-
 impl PMMRable for RangeProof {
 	type E = Self;
 
 	fn as_elmt(&self) -> Self::E {
 		self.clone()
 	}
+
+	fn elmt_size() -> Option<u16> {
+		Some((8 + MAX_PROOF_SIZE).try_into().unwrap())
```
quentinlesceller: Might be useful to keep the length prefix comment here.

store/tests/pmmr.rs
```diff
@@ -35,7 +34,7 @@ fn pmmr_leaf_idx_iter() {
 		let mut backend = store::pmmr::PMMRBackend::new(
 			data_dir.to_string(),
 			true,
-			false,
+			// false,
```
quentinlesceller: Might as well remove that line?

.gitignore
antiochp: Oh I have a bunch of dirs locally for `node_mainnet`, `node_floonet` etc. And they are all under the root dir of grin...

antiochp:
> Also this seems to introduce more "hardcoded" value. Wouldn't it be safer to use [size_of](https://doc.rust-lang.org/core/mem/fn.size_of.html) function?

For primitives, rather than hard-coding 8 for a u64 for example? 
Yes it probably would, but then I'm not sure how hard-coded this actually is, a u64 is _always_ 8 bytes, so do we really need to call a fn to do this, given that the fn is going to return a hard-coded value itself.
Let me take another pass at this and see which approach reads better.

quentinlesceller:
> Yes it probably would, but then I'm not sure how hard-coded this actually is, a u64 is always 8 bytes, so do we really need to call a fn to do this, given that the fn is going to return a hard-coded value itself.

Yep fair point too 👍 .

Also just wanted to point that you can do stuff like `size_of::<(u64, u64, u64)>()` which *might* be easier to read .

antiochp:
> Also just wanted to point that you can do stuff like `size_of::<(u64, u64, u64)>()` which _might_ be easier to read .

Oh that's interesting. I wonder if we can make `size_of` do the right thing for our structs (like `Hash` etc.)

That way we could do things like `size_of::<(Hash, u64, Difficulty, u32, u8)>()` which would be nice.

quentinlesceller:
Yeah good point. Probably simpler to go with what you did then 👍

quentinlesceller:
>If we start to use size_of for primitives like u64 its going to be easy for someone to accidentally use it for a more complex struct where the size is not stable.
>`size_of::<(u16, u64)>()`
>Is different to (due to padding consideration) -
>`size_of::<(u64, u16)>()`

Not sure I am following that part.
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5de22fe1bd249e43d4199528e51e1274

antiochp:
> > If we start to use size_of for primitives like u64 its going to be easy for someone to accidentally use it for a more complex struct where the size is not stable.
> > `size_of::<(u16, u64)>()`
> > Is different to (due to padding consideration) -
> > `size_of::<(u64, u16)>()`

Oh I misread the docs. i think this only applies for structs when using `#[repr(C)]`

https://doc.rust-lang.org/std/mem/fn.size_of.html#size-of-reprc-items

PR #3135, opened 2019-11-25
[DNM] Readonly PMMR backend via try_clone()

antiochp:
Still WIP as we need to think through some edge cases.

One of the bottlenecks in the code today is the need to obtain a _write_ lock on the txhashset even if we only want to read from a "readonly" view of the txhashset. This is because we need a `mutable` instance of the various PMMR backends due to the way we handle "rewind" via mutability of `last_pos`.

This PR makes the PMMR backend cloneable by introducing `try_clone()` on a couple of structs, specifically the `PMMRHandle` and the backend hash file and data file. 

[tbd - describe what can be cloned and why]
* files are mmap'd and "immutable"
* leafset (what do we do with this and why?)
* prunelist is readonly during normal operation

Q) What happens here during the occasional prune operation? What happens if we are reading a cloned instance of a PMMR backend and it is pruned and compacted from beneath us?

Todo - Explain the issue of our `batch` implementation. All lmdb interaction is done via a batch, reads _and_ writes. Explain why we still have an implicit global lock around the txhashset.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -230,14 +230,12 @@ impl Chain {
 		Ok(chain)
 	}
 
-	/// Return our shared header MMR handle.
-	pub fn header_pmmr(&self) -> Arc<RwLock<PMMRHandle<BlockHeader>>> {
-		self.header_pmmr.clone()
+	pub fn header_pmmr_cloned(&self) -> Result<PMMRHandle<BlockHeader>, Error> {
+		self.header_pmmr.read().try_clone()
 	}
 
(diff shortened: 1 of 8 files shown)


antiochp:
Closing - pretty sure we can do this with `clone()` rather than `try_clone()` as the append only file does not actually need to maintain a reference to the underlying file on disk.

Also want to explore flipping this around and using clone-on-write rather than cloning on every read.
https://doc.rust-lang.org/std/borrow/enum.Cow.html

PR #3136, opened 2019-11-25
Hardfork2

tromp:
---
name: HardFork2
about: This hard-fork allows blocks with YEAR_HEIGHT/FLOONET_SECOND_HARD_FORK <= height < YEAR_HEIGHT\*3/2 with header version 3 and Cuckaroom29 as secondary PoW. 
title: HardFork2
labels: consensus breaking

---
Currently, consensus::valid_header_version(height, version) returns false for all height >= YEAR_HEIGHT.
This PR changes that to return true for version 3 at either FLOONET_SECOND_HARD_FORK <= height < YEAR_HEIGHT\*3/2 on floonet or YEAR_HEIGHT <= height < YEAR_HEIGHT\*3/2 on mainnet.

This PR also changes global::create_pow_context to return new_cuckaroom_ctx(edge_bits, proof_size) when consensus::valid_header_version(height, version) holds for version 3, which means that Cuckaroom29 [1] is the only valid secondary PoW after the hard fork.

Successfully tested on a mainnet fast sync.

[1] https://www.grin-forum.org/t/next-pow-cuckaroom-unveiled-at-grincon1

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -130,6 +130,9 @@ pub const HARD_FORK_INTERVAL: u64 = YEAR_HEIGHT / 2;
 /// Floonet first hard fork height, set to happen around 2019-06-20
 pub const FLOONET_FIRST_HARD_FORK: u64 = 185_040;
 
+/// Floonet second hard fork height, set to happen around 2019-12-19
+pub const FLOONET_SECOND_HARD_FORK: u64 = 303_696;
+
 /// Check whether the block version is valid at a given height, implements
 /// 6 months interval scheduled hard forks for the first 2 years.
 pub fn valid_header_version(height: u64, version: HeaderVersion) -> bool {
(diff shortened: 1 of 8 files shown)


core/src/consensus.rs
```diff
@@ -130,6 +130,9 @@ pub const HARD_FORK_INTERVAL: u64 = YEAR_HEIGHT / 2;
 /// Floonet first hard fork height, set to happen around 2019-06-20
 pub const FLOONET_FIRST_HARD_FORK: u64 = 185_040;
 
+/// Floonet second hard fork height, set to happen around 2019-12-19
+pub const FLOONET_SECOND_HARD_FORK: u64 = 298_080;
+
 /// Check whether the block version is valid at a given height, implements
 /// 6 months interval scheduled hard forks for the first 2 years.
 pub fn valid_header_version(height: u64, version: HeaderVersion) -> bool {
```
antiochp: Not directly related to this PR but it would be _awesome_ if we had a function that took a block height and returned the valid header version. We could then refactor `valid_header_version()` to use this internally. We would _also_ be able to leverage it for block height specific logic elsewhere.

tromp: Yes; I've also been thinking about adding that, just to simplify the logic in valid_header_version.
The latter does have the advantage in making certain heights invalid regardless of version, which is a nice way to stop obsolete clients in their track.

tromp: see latest commit...

antiochp:
@tromp Is this good to go now?

@yeastplume Want to take a look before we merge this?

tromp:
should be good to go...

PR #3144, opened 2019-11-27
Increase sync speed by favouring faster peers

JosephGoulden:
Two modifications here in an attempt to speed up sync speed. 

First is to reduce header sync timeout from 120 seconds to 20. This means if a peer is stalling or can't send the headers for some reason we don't wait very long (looks like Grin++ only waits ~12 seconds).

Second is to order most_work_peers by the duration of the last ping/pong exchange with the connected peers. Also added the ping duration to the TUI peers screen. You may notice when running that during header sync the peer that the headers are being synced from might not be the lowest ping, I think this is just because it is now under load from sending headers. 

There are potentially wider implications to this change, interested to hear other opinions.


servers/src/grin/sync/header_sync.rs
```diff
@@ -142,7 +142,7 @@ impl HeaderSync {
 						match self.sync_state.status() {
 							SyncStatus::HeaderSync { .. } | SyncStatus::BodySync { .. } => {
 								// Ban this fraud peer which claims a higher work but can't send us the real headers
-								if now > *stalling_ts + Duration::seconds(120)
+								if now > *stalling_ts + Duration::seconds(20)
```
hashmap: I need to check the code in details, but I saw today that:
* we re-request headers after 10 seconds
* we don't ban a peers which doesn't reply to get headers request
so probably this code doesn't work as expected

DavidBurkett:
Haven't reviewed the code yet, but I like the idea of finding the fastest peer for header sync, except you'll probably end up overburdening the few seed peers with your approach, since they start by sending a bunch of data others don't (list of peers).

servers/src/grin/sync/header_sync.rs
JosephGoulden: My understanding of it is that if the peer is not sending the requested headers we will keep asking every 10 seconds up till 120 seconds at which point they are banned.

antiochp:
Would it be beneficial to prioritize `outbound` peers when selecting a peer to sync from?
These are public nodes (as they accept inbound connections) and are presumably more likely to have faster connections and presumably be more stable and long lived.

This would only be an issue if your node itself accepted inbound connection - but might be a nice heuristic to apply if you are attempting to bring up a new public node.

antiochp:
Unrelated to this PR but its got me thinking. If we are going to choose a single peer and stick to it (until we decide otherwise) why create all these roundtrip requests?

-> I have headers up to 1, go ask for headers with locator [1].
<- receive 512 headers from 1 onwards
-> I have headers 1 thru 512, go ask for headers with locator [x, y, z, 512]
<- receive 512 headers from 513 onwards
...

Why not ask for headers _once_ given a particular locator, then the peer simply starts streaming _all_ headers back from the locator position.

The peer is free to limit how many headers to return.
The receiving end is free to go ask another peer for headers at any time (but no need if still receiving headers from the original peer).
But by default asking for headers will result in the peer sending _all_ headers back, but "streaming" them chunk by chunk.

The receiving peer will need to be aware of the possibility the sending peer is on a fork, but this is the case currently. The receiving peer needs to know to switch to a different peer advertising "most work".

JosephGoulden:
Thanks for the comments, very useful. I'm currently experimenting with using ping duration to select a peer instead of the bytes sent, will update this MR when I have something.

JosephGoulden:
Think this is ready for another review. Now using ping duration to order the most_work_peers. Seems to work well in testing. As long as there is a syncing peer already being used to request headers it will be reused unless it stops responding.

I didn't prioritise outbound peers, perhaps that could be turned on when you are setting a public node, I didn't try that.

DavidBurkett:
Anytime we intentionally favor 1 peer over another, we increase the chances a sophisticated attacker can use that to game the system in some way. But since this is only during initial sync, and since we're a long way from dealing with the sophisticated attacks bitcoin has to worry about, the pros far outweigh the cons. Concept ACK.

JosephGoulden:
I did look at Bitcoin for inspiration, as far as I can tell it uses a single peer for initial header sync as well. I'm sure there are lots of other checks it does as well that we can implement to mitigate attacks.

https://github.com/bitcoin/bitcoin/blob/d9a45500018fa4fd52c9c9326f79521d93d99abb/src/net_processing.cpp#L3620

DavidBurkett:
Right, Grin++ does as well. My only concern is with how we choose that peer, but like I said, during initial sync, it's not a big deal. The worst thing a malicious peer could do is intentionally slow down syncing.

It was more just a general caution about using absolute criteria to choose a peer.

servers/src/grin/sync/header_sync.rs
```diff
@@ -157,14 +158,14 @@ impl HeaderSync {
 										peer.info.height(),
 										peer.info.total_difficulty(),
 									);
+									self.syncing_peer = None
```
quentinlesceller: Could you clarify why you moved this line?

phyro:
@quentinlesceller @antiochp @DavidBurkett if any of you finds the time, please make a quick pass on this one to re-evaluate the changes and whether they should be considered

DavidBurkett:
I'm still in favor of the change, but considering how long it's been, and how many conflicts there now are, we should probably just close this until someone comes along and decides to tackle syncing performance again.

yeastplume:
Taking an executive decision to close, not because the concept is bad but because of how long it's been and how out-of date the changes are. Initial sync times will be revisited with PIBD integration, and alternate ideas for speeding up the header portion of the sync might stem from it

PR #3145, opened 2019-11-29
Set longer timeout for msg body read/write operations

hashmap:


p2p/src/conn.rs
```diff
@@ -322,8 +330,11 @@ where
 		.name("peer_write".to_string())
 		.spawn(move || {
 			let mut retry_send = Err(());
+			writer
+				.set_write_timeout(Some(BODY_IO_TIMEOUT))
+				.expect("set timeout");
 			loop {
-				let maybe_data = retry_send.or_else(|_| send_rx.recv_timeout(IO_TIMEOUT));
+				let maybe_data = retry_send.or_else(|_| send_rx.recv_timeout(HEADER_IO_TIMEOUT));
```
antiochp: This is actually reading a msg from our internal msg queue, not the tcp stream.
So I think we should keep a specific timeout for this.

hashmap: done

hashmap:
@antiochp I don't see it in `perf` output, so must be fast enough. In p2p-server we spend most of the time calling non-blocking accept and then sleep (to allow responsive shutdown), but a synced grin node consumes 0.3-1% cpu on my server, so no real issue.

PR #3147, opened 2019-11-29
height (header version) specific output PMMR root rules

antiochp:
This PR introduces the necessary consensus rule change to support "enable faster sync" https://github.com/mimblewimble/grin-rfcs/pull/29.

HF2 will change the meaning of the `output_root` field on block headers.
* Prior to HF2 the `output_root` is simply to root of the output PMMR.
* Post HF2 the `output_root` is a combined root of both the output PMMR and the associated "bitmap accumulator" representation of the PMMR `leafset`.

This PR introduces the following changes - 
* add block height aware `output_root` rules (base logic on validated header version)
* `pipe::validate_header()` now explicitly validates the header version for the given block height
  * this was implicit before (and not exercised at all in the tests or locally in `user_testing`)
* add testing specific fork heights for use in `user_testing` and `automated_testing` to exercise HF rules in tests
  * we now exercise HF1 at height 3 and HF2 at height 6 in tests and locally in `user_testing`

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -30,8 +30,7 @@ use crate::store;
 use crate::txhashset;
 use crate::txhashset::{PMMRHandle, TxHashSet};
 use crate::types::{
-	BlockStatus, ChainAdapter, NoStatus, Options, OutputMMRPosition, Tip, TxHashSetRoots,
-	TxHashsetWriteStatus,
+	BlockStatus, ChainAdapter, NoStatus, Options, OutputMMRPosition, Tip, TxHashsetWriteStatus,
 };
 use crate::util::secp::pedersen::{Commitment, RangeProof};
 use crate::util::RwLock;
@@ -604,7 +603,7 @@ impl Chain {
 		b.header.prev_root = prev_root;
 
 		// Set the output, rangeproof and kernel MMR roots.
-		b.header.output_root = roots.output_root(&b.header)?;
+		b.header.output_root = roots.output_root(&b.header);
 		b.header.range_proof_root = roots.rproof_root;
 		b.header.kernel_root = roots.kernel_root;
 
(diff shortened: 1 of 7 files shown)


chain/src/chain.rs
```diff
@@ -634,9 +636,10 @@ impl Chain {
 		txhashset.merkle_proof(commit)
 	}
 
-	/// Returns current txhashset roots.
-	pub fn get_txhashset_roots(&self) -> TxHashSetRoots {
-		self.txhashset.read().roots()
+	/// Returns current txhashset roots along with current head header.
+	pub fn get_txhashset_roots(&self) -> Result<(TxHashSetRoots, BlockHeader), Error> {
```
tromp: shouldn't this change name to get_txhashset_roots_and_header ?

chain/src/types.rs
```diff
@@ -193,24 +194,24 @@ pub struct TxHashSetRoots {
 }
 
 impl TxHashSetRoots {
-	/// Accessor for the underlying output PMMR root
-	pub fn output_root(&self) -> Hash {
-		self.output_roots.output_root()
+	/// Accessor for the output PMMR root (rules here are block height dependent).
+	pub fn output_root(&self, header: &BlockHeader) -> Result<Hash, Error> {
```
tromp: shouldn't it depend only on header version, rather than whole header?

chain/src/types.rs
```diff
@@ -232,15 +233,27 @@ pub struct OutputRoots {
 }
 
 impl OutputRoots {
+	/// The root of our output PMMR. The rules here are block height specific.
+	/// We use the merged root here for header version 3 and later.
+	pub fn root(&self, header: &BlockHeader) -> Result<Hash, Error> {
+		if !consensus::valid_header_version(header.height, header.version) {
```
tromp: i don't see why all these utility functions need to error on incorrect header version. ideally there is only one place where correct header version is checked, and other functions just assume it correct.

but in this case, I don't even see why the root function depends on the header. It should only depend on header version, in which case it can not error either.

antiochp: Fair point - I can clean this up.

chain/src/chain.rs
antiochp: Actually realised this can be cleaned up - its only used by the api in one place and there is an easier way of getting the "txhashset" roots. We just need the header itself, no need to recalculate the roots off the actual PMMRs.

chain/src/types.rs
```diff
@@ -193,24 +193,26 @@ pub struct TxHashSetRoots {
 }
 
 impl TxHashSetRoots {
-	/// Accessor for the underlying output PMMR root
-	pub fn output_root(&self) -> Hash {
-		self.output_roots.output_root()
+	/// Accessor for the output PMMR root (rules here are block height dependent).
+	/// We assume the header version is consistent with the block height and that this
+	/// has been validated elsewhere.
```
tromp: Better to specifically state that it's validate_header()'s responsibility to check for correct header version.

antiochp:
Will merge once CI passes.

PR #3155, opened 2019-12-03
Rename API V1 push_tx -> push

yeastplume:
Basically, the reverse of this:

https://github.com/mimblewimble/grin/pull/2897

For a quick reminder, this will ensure that 2.x.x wallets won't be able to operate against a 3.0.0 node, forcing everyone to upgrade the wallet as well as the node.

Hopefully this should be taken care of already in HF3 if the v1 api is removed and the 4.0 code uses the V2 node API exclusively.


yeastplume:
Closed after gitter discussion

PR #3157, opened 2019-12-05
pass stopped into consume so we can halt txhashset mid download

antiochp:
We check `stopped` before every msg consumed to see if the the node is in the process of shutting down, so we can stop the peer read threads robustly.

This PR passes `stopped` _into_ `consume()` so we can check it during the txhashset download in the read loop. This allows us to halt the txhashset mid download, making it possible to stop and restart the node easily if the download is from a slow peer.


antiochp:
I'm pretty sure we used to do something similar to this but it was removed during one of the many rewrites of the p2p layer.

PR #3160, opened 2019-12-06
fix(TUI): Fixed panic when changing order of items in TableView

JosephGoulden:
This is awkward to reproduce but can be done reliably if you get the timing right.
- Remove peers database (rm -r .grin/main/chain_data/peer/), just to force us to go looking for more.
- Select Peers tab in TUI and order peers by Address.
- Whenever the list of peers updates reorder again straight away, so the row with focus should always be at the bottom of the list.
- At some point the number of peers will go to a large number then back to 8 (if you have the default desired peers). It's at this point the panic occurs, because the focus might be on peer 15 but the new list only has 8 peers.

thread 'main' panicked at 'index out of bounds: the len is 8 but the index is 14': 
...
   6: grin::tui::table::TableView<T,H>::sort_by::haee3b0fe6388d824 (0x55986b9dd817)
   7: grin::tui::table::TableView<T,H>::set_items::h6712930b43f95168 (0x55986b9de079)

Have changed it so that sort function does not try to set_selected_item if it is outside the range of the new peers list and the panic no longer occurs.

### DIFF
--- a/src/bin/tui/table.rs
+++ b/src/bin/tui/table.rs
@@ -492,7 +492,7 @@ impl<T: TableViewItem<H>, H: Eq + Hash + Copy + Clone + 'static> TableView<T, H>
 	/// Returns the index of the currently selected item within the underlying
 	/// storage vector.
 	pub fn item(&self) -> Option<usize> {
-		if self.items.is_empty() {
+		if self.items.is_empty() || self.focus > self.rows_to_items.len() {
 			None
 		} else {
 			Some(self.rows_to_items[self.focus])
@@ -604,7 +604,7 @@ impl<T: TableViewItem<H>, H: Eq + Hash + Copy + Clone + 'static> TableView<T, H>
 
 	fn sort_items(&mut self, column: H, order: Ordering) {
 		if !self.is_empty() {
-			let old_item = self.item().unwrap();
+			let old_item = self.item();
 
 			let mut rows_to_items = self.rows_to_items.clone();
 			rows_to_items.sort_by(|a, b| {
@@ -616,7 +616,7 @@ impl<T: TableViewItem<H>, H: Eq + Hash + Copy + Clone + 'static> TableView<T, H>
 			});
 			self.rows_to_items = rows_to_items;
 
-			self.set_selected_item(old_item);
+			old_item.map(|o| self.set_selected_item(o));
 		}
 	}
 


antiochp:
Nice.

magp3wd912:
@@ -1,6 +1,6 @@ 
   [package] 
   name = "grin" 
   -version = "3.0.0-alpha.1" 
   +version = "3.0.0-beta.1" 
    authors = ["Grin Developers <mimblewimble@lists.launchpad.net>"] 
    description = "Simple, private and scalable cryptocurrency implementation based on the Mimblewimble chain format." 
   license = "Apache-2.0" 

 

 
 @@ -32,14 +32,14 @@ term = "0.5" 
    failure = "0.1" 
    failure_derive = "0.1" 
    
 
    -grin_api = { path = "./api", version = "3.0.0-alpha.1" } 
    -grin_config = { path = "./config", version = "3.0.0-alpha.1" } 
    -grin_chain = { path = "./chain", version = "3.0.0-alpha.1" } 
    -grin_core = { path = "./core", version = "3.0.0-alpha.1" } 
   -grin_keychain = { path = "./keychain", version = "3.0.0-alpha.1" } 
-grin_p2p = { path = "./p2p", version = "3.0.0-alpha.1" } 
 -grin_servers = { path = "./servers", version = "3.0.0-alpha.1" } 
 -grin_util = { path = "./util", version = "3.0.0-alpha.1" } 
   +grin_api = { path = "./api", version = "3.0.0-beta.1" } 
  +grin_config = { path = "./config", version = "3.0.0-beta.1" } 
   +grin_chain = { path = "./chain", version = "3.0.0-beta.1" } 
+grin_core = { path = "./core", version = "3.0.0-beta.1" }
   +grin_keychain = { path = "./keychain", version = "3.0.0-beta.1" } 
  +grin_p2p = { path = "./p2p", version = "3.0.0-beta.1" } 
  +grin_servers = { path = "./servers", version = "3.0.0-beta.1" } 
  +grin_util = { path = "./util", version = "3.0.0-beta.1" } 
    
 
   [target.'cfg(windows)'.dependencies] 
   cursive = { version = "0.12", default-features = false, features = ["pancurses-backend"] } 

 
 @@ -53,5 +53,5 @@ cursive = "0.12" 
    built = "0.3" 
    
 
    [dev-dependencies] 
  -grin_chain = { path = "./chain", version = "3.0.0-alpha.1" } 
 -grin_store = { path = "./store", version = "3.0.0-alpha.1" }  +grin_chain = { path = "./chain", version = "3.0.0-beta.1" } 
 +grin_store = { path = "./store", version = "3.0.0-beta.1" }

antiochp:
I've definitely encountered this before - and always failed to investigate it...

src/bin/tui/table.rs
```diff
@@ -616,7 +616,7 @@ impl<T: TableViewItem<H>, H: Eq + Hash + Copy + Clone + 'static> TableView<T, H>
 			});
 			self.rows_to_items = rows_to_items;
 
-			self.set_selected_item(old_item);
+			old_item.map(|o| self.set_selected_item(o));
```
antiochp: Oh that's really nice - just using `map()` rather than an `if let Some(x) = old_item { ... }`.

src/bin/tui/table.rs
```diff
@@ -492,7 +492,7 @@ impl<T: TableViewItem<H>, H: Eq + Hash + Copy + Clone + 'static> TableView<T, H>
 	/// Returns the index of the currently selected item within the underlying
 	/// storage vector.
 	pub fn item(&self) -> Option<usize> {
-		if self.items.is_empty() {
+		if self.items.is_empty() || self.focus > self.rows_to_items.len() {
```
antiochp: Should this actually be `>=`?

I just happened to see the following panic locally - 

```
thread 'main' panicked at 'index out of bounds: the len is 15 but the index is 15': /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/slice/mod.rs:2717stack backtrace:
```

JosephGoulden: Yes you're right I think it should. I can do a fix for that now.

antiochp: If its just this one line change I think we can get it in for `beta.2`.
:+1:

PR #3161, opened 2019-12-07
fix(TUI): selected column and row of table are now preserved during update

JosephGoulden:
Currently if you move around a TUI table (e.g. peers) with the arrow keys your selection is reset every time the table is updated with a call to set_items(). This change is to preserve the selected row and column which should make is easier to monitor a selected peer for example.

### DIFF
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -193,6 +193,24 @@ pub struct PeerStats {
 	pub received_bytes_per_sec: u64,
 }
 
+impl PartialEq for PeerStats {
+	fn eq(&self, other: &PeerStats) -> bool {
+		*self.addr == other.addr
+	}
+}
+
+impl PartialEq for WorkerStats {
+	fn eq(&self, other: &WorkerStats) -> bool {
+		*self.id == other.id
+	}
+}
+
+impl PartialEq for DiffBlock {
+	fn eq(&self, other: &DiffBlock) -> bool {
+		self.block_height == other.block_height
+	}
+}
+
 impl StratumStats {
 	/// Calculate network hashrate
 	pub fn network_hashrate(&self, height: u64) -> f64 {
--- a/src/bin/tui/peers.rs
+++ b/src/bin/tui/peers.rs
@@ -33,7 +33,7 @@ use crate::tui::table::{TableView, TableViewItem};
 use crate::tui::types::TUIStatusListener;
 
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
-enum PeerColumn {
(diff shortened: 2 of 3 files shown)


antiochp:
:fire: Let me test this out locally. :+1: on what we are fixing here.

antiochp:
We do need to make a decision around the trade-offs involved in merging this (and any other fixes) between now (beta.1) and 3.0.0 release.
I would lean toward merging this but we should have some consensus here around the decision making process itself.

The trade-off being fixing issues and improvements vs. the risk of introducing other (potentially subtle) issues inadvertently.

lehnberg:
As per the timeline in our [process doc](https://www.grin-forum.org/t/grin-v3-0-0-hard-fork-upgrade-jan-2020/6036), `code freeze` is scheduled for Dec 12, defined as:

> No changes are permitted on the branch to be released except critical bugs.

Today is the 9th. So technically, this could get merged in before Thursday and it would be fine. After Thursday it would not be fine.

Personally however, I lean towards not merging anything that's not in scope and not critical at this point, because I prefer a more solid quality, less capable build, over a more capable but less performant release. And I worry about our testing capabilities. I suspect it's quite easy for problematic bugs to sneak in without us noticing it, but I have no basis for this claim. We'd add  more risk, at a point where I feel we have enough risk as is already.

That's just my 2 grins worth. But yes, might be something to talk about in tmrw's dev meeting, if you'd like to add a point to the agenda?

hashmap:
what about aiming for 3.1 in a month after 3.0?

yeastplume:
There's one or two things I wouldn't mind squeezing in this week, (mostly very superficial), let's discuss at dev meeting tomorrow.

JosephGoulden:
Sorry I might not be able to available to discuss tomorrow. This is relatively low risk in my opinion, over a month of testing seems like plenty of time. But I've not been around for previous releases so if consensus is it's too risky then that's fine. I'll have some time the next few weeks to help fix any issues (particularly with things I worked on).

antiochp:
> So technically, this could get merged in before Thursday and it would be fine.

> Personally however, I lean towards not merging anything that's not in scope and not critical

Yes. That's kind of what I mean - "code freeze" does not necessarily mean anything goes prior to that.
That said I agree with @JosephGoulden here, this is low risk and I think likely fine (pending some extra eyes on it and some local testing).

I just wanted to bring this up _somewhere_ and this was as good a place as any.

antiochp:
Tested this locally with #3160 and it all looks good. I think its worth the minimal risk involved in merging this close to `beta.2`

PR #3164, opened 2019-12-10
drop msg if sync_sender buffer is full (do not close peer connection)

### LINKED ISSUE #3162: "txhashset save to file fail" with corresponding "sending on a full channel"
I was trying to track down the deadlock issue we are still seeing with "step 6/7" getting locked up and came across this with a node configured to sync against a single peer.

This was on the end _sending_ the txhashet file - 

```
20191209 17:14:46.012 DEBUG grin_p2p::peers - Error pinging peer PeerAddr(V4(<redacted>:3414)): Send("sending on a full channel")
20191209 17:14:46.012 DEBUG grin_p2p::peer - Stopping peer PeerAddr(V4(<redacted>:3414))
20191209 17:14:46.864 DEBUG grin_p2p::conn - Shutting down reader connection with <redacted>:59172
```

Which caused the following on the _receiving_ end - 

```
20191209 17:14:31.605 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 330624000/506602362
20191209 17:14:42.793 DEBUG grin_p2p::protocol - handle_payload: txhashset archive: 337152000/506602362
20191209 17:14:47.812 ERROR grin_p2p::protocol - handle_payload: txhashset archive save to file fail. err=Connection(Custom { kind: UnexpectedEof, error: "failed to fill whole buffer" })
20191209 17:14:47.812 DEBUG grin_p2p::conn - try_break: exit the loop: Connection(Custom { kind: UnexpectedEof, error: "failed to fill whole buffer" })
20191209 17:14:47.812 DEBUG grin_p2p::conn - Shutting down reader connection with <redacted>:3414
```

So I think this explains at least one scenario where the receiving side sees a "txhashset save to file fail" and it is because the mpsc sync_sender (our internal per-peer outbound msg queue) is full and we aggressively shut the connection down even if we are in the middle of sending a txhashset.

If we are on a relatively slow connection and we are sending a txhashet to a peer then all subsequent msgs (to that peer) will be queued up behind it. 

We put a hard limit of 100 msgs on the per-peer msg queue - 
```
pub const SEND_CHANNEL_CAP: usize = 100;
```

The limit is probably fine under normal broadcast situations, but during a txhashet download this may not be correct behavior. We send regular pings and potentially other msgs if we are low on peers and looking for more.

Intuitively we may just want to simply drop msgs on the floor if we see a "sending on a full channel", but leave the connection intact. And only close the connection in the case where we actually encounter an error/exception around the connection itself (and not the msg queue).

antiochp:
Resolves #3162.

We do not necessarily want to treat a "sending on a full channel" as an error serious enough to warrant closing the peer connection.
There are various scenarios where the `sync_sender` buffer can fill up with "yet to be sent" msgs. One example is during a long running txhashset download.

This PR splits out the error handling during `try_send`. A `Disconnected` will propagate up and cause the caller to close the peer connection. A `Full` will simply cause us to drop the msg and continue.

The assumption here is that if the buffer filling up is a symptom of a more serious issue with the peer connection this will be handled at the peer level, closing the connection as necessary. It is not the responsibility of the msg queue to identify and handle this.

### DIFF
--- a/p2p/src/conn.rs
+++ b/p2p/src/conn.rs
@@ -38,6 +38,8 @@ use std::{
 	thread::{self, JoinHandle},
 };
 
+pub const SEND_CHANNEL_CAP: usize = 100;
+
 const HEADER_IO_TIMEOUT: Duration = Duration::from_millis(2000);
 const CHANNEL_TIMEOUT: Duration = Duration::from_millis(1000);
 const BODY_IO_TIMEOUT: Duration = Duration::from_millis(60000);
@@ -130,8 +132,6 @@ impl<'a> Message<'a> {
 	}
 }
 
-pub const SEND_CHANNEL_CAP: usize = 100;
-
 pub struct StopHandle {
 	/// Channel to close the connection
 	stopped: Arc<AtomicBool>,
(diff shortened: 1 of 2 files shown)


p2p/src/conn.rs
```diff
@@ -32,12 +32,15 @@ use std::io::{self, Read, Write};
 use std::net::{Shutdown, TcpStream};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{mpsc, Arc};
+
```
hashmap: care to remove it?

antiochp: Gone. :+1:

PR #3165, opened 2019-12-10
try_header_head in the syncer loop with a short timeout

antiochp:
We see "fast sync" sometimes locking up after "Step 6/7" which is the transition from txhashset download and validation to the final full block sync.

This does not appear to be a deadlock as it does (at least occasionally) resolve itself.
From thread backtraces via lldb it also does not appear to involve any deadlock between threads as the _only_ thread that appears to be involved is the `sync` thread. No peer threads appear to be blocked during this time.

This PR is a workaround and not a permanent fix - we use `try_header_head` with a 1 sec timeout and let the syncer loop retry via the existing `unwrap_or_restart_loop!` mechanism.

This has been tested multiple times, each time with a successful "fast sync" ending in a "running" node synchronized to the latest block height.

From the RwLock docs - 

> This lock uses a task-fair locking policy which avoids both reader and writer starvation. This means that readers trying to acquire the lock will block even if the lock is unlocked when there are writers waiting to acquire the lock. Because of this, attempts to recursively acquire a read lock within a single thread may result in a deadlock.

I suspect this bit is related somehow (but have no evidence for this) - 

> This means that readers trying to acquire the lock will block even if the lock is unlocked when there are writers waiting to acquire the lock.

Alternatively this is related but its not obvious from our code where this (recursive locking) is occurring - 

> Because of this, attempts to recursively acquire a read lock within a single thread may result in a deadlock.

https://docs.rs/parking_lot/0.10.0/parking_lot/type.RwLock.html

### DIFF
--- a/servers/src/grin/sync/syncer.rs
+++ b/servers/src/grin/sync/syncer.rs
@@ -182,7 +182,18 @@ impl SyncRunner {
 			// if syncing is needed
 			let head = unwrap_or_restart_loop!(self.chain.head());
 			let tail = self.chain.tail().unwrap_or_else(|_| head.clone());
-			let header_head = unwrap_or_restart_loop!(self.chain.header_head());
+
+			// We still do not fully understand what is blocking this but if this blocks here after
+			// we download and validate the txhashet we do not reliably proceed to block_sync,
+			// potentially blocking for an extended period of time (> 10 mins).
+			// Does not appear to be deadlock as it does resolve itself eventually.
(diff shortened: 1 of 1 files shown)


antiochp:
Just doing one final mainnet test for fast sync - if it syncs successfully will merge.

antiochp:
```
20191210 16:15:11.357 INFO grin_servers::grin::sync::syncer - synchronized at 1335770779969550 @ 471826 [00001902457e]
```

Merging.

PR #3168, opened 2019-12-12
tests: split consensus tests into separate files for each chain type …

JosephGoulden:
CHAIN_TYPE in global.rs is shared global state and we have multiple tests that modify this state through global::set_mining_mode(). Each Rust test target (file) is run with multiple threads, so the tests with shared state cannot be sure that the state is correct (chain_type could be modified by another test thread) causing intermittent failures.

I suggest we separate tests into different files depending on the chain type they need as I have done in this change. That way we can keep the global state, which is convenient and still run tests for the same chain type in parallel.


JosephGoulden:
I fixed the conflicts.

What exactly would you suggest to put in consensus.rs? As far as I can see it would be empty (there are no consensus integration tests that don't call global::set_mining_mode().

PR #3169, opened 2019-12-12
fix: panic in TUI table selected item. focus is zero indexed

JosephGoulden:
Additional fix for #3160 
Focus is zero indexed so should not be equal to the length of rows.


antiochp:
This is actually important to fix now that we preserve the selected row (https://github.com/mimblewimble/grin/pull/3161).

```
thread 'main' panicked at 'index out of bounds: the len is 8 but the index is 8': /rustc/4560ea788cb760f0a34127156c78e2552949f734/src/libcore/slice/mod.rs:2717stack backtrace:
```

Now that the selection is preserved we can actually end up with the now-non-existent row selected without any user interaction.

PR #3175, opened 2019-12-20
Fix compiler warning 'unnecessary parentheses around type'

quentinlesceller:
Probably related to Rust 1.40.0.
Remove unnecessary parentheses such as:
```
warning: unnecessary parentheses around type
   --> store/src/pmmr.rs:113:41
    |
113 |     fn get_hash(&self, pos: u64) -> Option<(Hash)> {
    |                                            ^^^^^^ help: remove these parentheses
    |
    = note: `#[warn(unused_parens)]` on by default
```

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -1008,7 +1008,7 @@ impl<'a> Extension<'a> {
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


JosephGoulden:
I think it's this change https://github.com/rust-lang/rust/issues/64169

PR #3178, opened 2019-12-28
Build block header from pre pow data and proof

### LINKED ISSUE #2595: Please add function for building block header to grin-core
**Is your feature request related to a problem? Please describe.**
I have pre-pow, nonce, edge_bits, and proof and want to verify that its valid.  Currently there is no way to do this without writing (and maintaining) custom code.   Pools (and probably other use cases) need this ability.

**Describe the solution you'd like**
Provide a function for building a block header.

**Describe alternatives you've considered**
Write and maintain it myself.

austinabell:
- Resolves #2595 (just picked up issue to learn a bit about grin)

Couple notes about the changes:
- I could add another function which takes a vector of pre pow bytes including the serialized nonce and the proof (or replace if you want a specific function signature)
- I made some changes to the signature mentioned in the issue since it just seemed cleaner and more usable
- I kept the pre_pow parameter a `String` although there is no need for it to be mutable in this function.
  - I am happy to change the from_hex function you have to instead take a `&str` since that does not need to be mutable either as it is just reading (that function is the main reason I kept as String, because it would just be converted back to a String anyway)


core/src/core/block.rs
```diff
@@ -109,6 +111,12 @@ impl From<keychain::Error> for Error {
 	}
 }
 
+impl From<std::num::ParseIntError> for Error {
```
antiochp: I suspect this ties a pretty general `ParseIntError` to a very specific `HexError` impl.
We could potentially fail to "parse an int" in a multitude of different scenarios (now or in the future) and not all of them will be caused by hex serialization related problems.

It may make more sense to just explicitly convert one error to another closer to where the error actually occurs.

austinabell: oops lol yeah that's probably better for future proofing

core/src/core/block.rs
```diff
@@ -350,6 +354,23 @@ impl BlockHeader {
...
 
+	/// Constructs a header given pre_pow string, nonce, and proof
+	pub fn from_pre_pow_and_proof(
+		pre_pow: String,
+		nonce: u64,
+		proof: Proof,
+	) -> Result<Self, Error> {
+		// Convert hex pre pow string
+		let mut header_bytes = from_hex(pre_pow)
+			.map_err(|e| Error::Serialization(ser::Error::HexError(e.to_string())))?;
```
antiochp: :+1:

Issue #3182, opened 2020-01-06
TUI peer sort order consistency


antiochp:
We can currently sort the list of peers by selecting a column header, for example by `Direction` or `User Agent`.

The issue is that within this selected sort field there is no consistent sub-sorting resulting in many peers with say the same `User Agent` being effectively unsorted and arbitrarily re-ordered on each refresh.

We should sort by selected field _then_ by `Address` each time to keep the ordering consistent.

antiochp:
Resolved in #3184.

PR #3184, opened 2020-01-07
TUI peer sort order

cadwgan0:
Potential fix for: https://github.com/mimblewimble/grin/issues/3182

When TUI peers are sorted by a specific column the results are not consistent and change on each refresh. This PR introduces a fallback using the address column whenever specific cmp values are equal.  

I tested this on floonet terminal and the peers display consistently.  I am new to rust so any feedback is welcome.


cadwgan0:
I'll try this out after work and resubmit, thanks

cadwgan0:
Removed the function I was using prior, replaced with closure and then utilized that with the .then chain to clean the code up.

PR #3185, opened 2020-01-07
Remove unused crates

hashmap:
also bump minor versions in Cargo.lock


quentinlesceller:
What's the command to know which crates are unused? 
Also:
```
15 | use env_logger;
   |     ^^^^^^^^^^ no `env_logger` external crate
```

hashmap:
@quentinlesceller I used https://crates.io/crates/cargo-udeps
Thanks, missed this test dependency, fixed

Issue #3188, opened 2020-01-14
cleanup accidental file checked in


antiochp:
Looks like a garbage file was checked in accidentally as part of https://github.com/mimblewimble/grin/pull/1565

```
api/tests/localhost+1.p12
```

See https://www.grin-forum.org/t/tls-support-localhost-1-p12/6907.

hashmap:
It wasn't garbage at that time, it was required to compile tests, but since that tls code was changed a bit so now it's not needed

PR #3189, opened 2020-01-15
Removed binary blob: localhost+1.p12.

mplsgrant:
---
name: Removed binary blob: localhost+1.p12.
about: See issue: https://github.com/mimblewimble/grin/issues/3188
---

**A binary blob exists in Grin**
The current Grim implementation includes a binary blob: https://github.com/mimblewimble/grin/blob/master/api/tests/localhost%2B1.p12

**Binary blobs harm Grin**
Binary blobs harm open source projects because they take agency away from users and developers by obscuring the operations and goings-on of the software project.

**Effects of removal**
Removing the binary blob does not appear to break building the Grin implementation. Cargo build, test, and run all come back clean.

> running 4 tests
> test tui::table::test::test_set_items_preserves_order ... ok
> test tui::status::test_status_txhashset_rproofs ... ok
> test tui::status::test_status_txhashset_kernels ... ok
> test tui::table::test::test_set_items_preserves_selected_item ... ok
> 
> test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

**Floonet**
I have not tested this PR against floonet.


cadwgan0:
I think the reason no tests fail is because the test case is ignored: https://github.com/mimblewimble/grin/blob/master/api/tests/rest.rs#L87 either way that test case advises to retrieve your own certificates and the test case fails if you remove the ignore whether or not the file is present.  

```
"tests/fullchain.pem".to_string(),
"tests/privkey.pem".to_string(),
```

hashmap:
Thanks for your contibution!
@cadwgan0  is right,  the test is disabled because it requires an extra setup, this file was needed to get tests at least compile but it's not the case anymore.

@mplsgrant unrelated hint -  you need to run `cargo test --all` or go to each crate and run tests there, in your output you run tests only for `grin`

Issue #3199, opened 2020-01-19
running grin-wallet check against grin >= 3.0.0 using older versions of grin-wallet results in infinite loops


cgilliard:
**Describe the bug**
When using older versions of the grin-wallet or wallet713 against grin 3.0.0, 'check' command fails. The API is supposed to return highest_index and last_retrieved_index, but appears to return a different value. This is related to a change in txhashset.rs. Making this change fixes the issue:

        /// highest output insertion index available
        pub fn highest_output_insertion_index(&self) -> u64 {
-                pmmr::n_leaves(self.output_pmmr_h.last_pos)
+                //pmmr::n_leaves(self.output_pmmr_h.last_pos)
+                 // API expected the max vali in mmr_index units that are sparce
+                self.output_pmmr_h.last_pos
        }

Not only does this error produce issues with running older wallets that rely on 'highest_index', but also the API is inaccurate.

**To Reproduce**
1. Run '...' using grin-wallet < 3.0
grin-wallet check
The result is an infinite loop.

quentinlesceller:
Fixed by https://github.com/mimblewimble/grin/pull/3200 closing.

PR #3200, opened 2020-01-19
Bug fix for highest_index in output API.

cgilliard:
Bug fix for the following issue: https://github.com/mimblewimble/grin/issues/3199


yeastplume:
Verified the issue and the proposed fix, I think what happened here is in the pre-3.0.0 version of the node API code there were some confusing mismatches between pmmr index and leaf index, and most of the code was changed to use pmmr index only. However, the newer version of the wallet doesn't call `highest_output_insertion_index` so this function was missed.

Thanks for catching and fixing this!

Note this should also be cherry-picked into maintenance branches.

PR #3204, opened 2020-01-21
3.0.x: Bug fix for highest_index in output API #3200

quentinlesceller:
Equivalent of https://github.com/mimblewimble/grin/pull/3200 for `current/3.0.x`.


quentinlesceller:
Closing this PR since we agreed that if a 3.0.1 is needed we would cut it off from master.

PR #3209, opened 2020-01-28
#2149: Resolve JSON RPC 2.0 non-compliance

### LINKED ISSUE #2149: Stratum server accepts only string ids

WhoNeedszZz:
Currently, the JSON RPC requests and responses are not compliant with the spec as the server only accepts strings.
This change is *non-breaking* as it simply adds the ability to understand integers as expected.
Any old code continuing to only send string requests will continue to work as before.

Resolves #2149

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -52,9 +52,17 @@ type Tx = mpsc::UnboundedSender<String>;
 // http://www.jsonrpc.org/specification
 // RPC Methods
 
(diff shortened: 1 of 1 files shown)


hashmap:
Test would be beneficial for this PR not just to test the current impl but also to make sure that this case is not broken in the future

WhoNeedszZz:
Ordinarily, yes, but this was discussed in Keybase. If that doesn't satisfy, then feel free to add to the discussion.

PR #3213, opened 2020-01-29
Resolve JSON RPC 2.0 non-compliance (redo)

### LINKED ISSUE #2149: Stratum server accepts only string ids

WhoNeedszZz:
Currently, the JSON RPC requests and responses are not compliant with the spec as the server only accepts strings.
This change is non-breaking as it simply adds the ability to understand integers as expected.
Any old code continuing to only send string requests will continue to work as before.

Unit tests are included to ensure (de)serialization works on both strings and integers.

Resolves #2149

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -52,17 +52,26 @@ type Tx = mpsc::UnboundedSender<String>;
 // http://www.jsonrpc.org/specification
 // RPC Methods
 
-#[derive(Serialize, Deserialize, Debug)]
+/// Represents a compliant JSON RPC 2.0 id.
+/// Valid id: Integer, String.
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
+#[serde(untagged)]
+enum JsonId {
+	IntId(u32),
+	StrId(String),
+}
+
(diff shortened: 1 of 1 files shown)


antiochp:
Not to open a big can of worms but wondering if there is a more `serde` like way of handling this?

https://serde.rs/impl-deserialize.html#the-visitor-trait

> Here is a Visitor that is able to deserialize a primitive i32 from a variety of types.

servers/src/mining/stratumserver.rs
```diff
@@ -501,7 +510,7 @@ impl Handler {
 		// Issue #1159 - use a serde_json Value type to avoid extra quoting
 		let job_template_value: Value = serde_json::from_str(&job_template_json).unwrap();
 		let job_request = RpcRequest {
-			id: String::from("Stratum"),
+			id: JsonId::StrId(String::from("Stratum")),
```
antiochp: To my point about an additional explicit layer - we need to remember to wrap values like this.

quentinlesceller: One way would be to use the `#[serde(with = "secp_ser::string_or_u64")]` like grin-wallet does.
https://github.com/mimblewimble/grin-wallet/blob/master/libwallet/src/slate_versions/v3.rs#L60

PR #3216, opened 2020-01-30
Less cloning and pattern simplifications

quentinlesceller:
Relatively large PR. 
- Reduce the number of cloning that we do.
- Replace `!format` by `to_string()` when possible.
- Remove needless return and variable allocation.
- Other simplifications

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -63,7 +63,7 @@ fn validate_pow_only(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result
 	if !header.pow.is_primary() && !header.pow.is_secondary() {
 		return Err(ErrorKind::LowEdgebits.into());
 	}
-	if !(ctx.pow_verifier)(header).is_ok() {
+	if (ctx.pow_verifier)(header).is_err() {
 		error!(
 			"pipe: error validating header with cuckoo edge_bits {}",
 			header.pow.edge_bits(),
@@ -385,7 +385,7 @@ fn validate_block(block: &Block, ctx: &mut BlockContext<'_>) -> Result<(), Error
 	let prev = ctx.batch.get_previous_header(&block.header)?;
 	block
 		.validate(&prev.total_kernel_offset, ctx.verifier_cache.clone())
-		.map_err(|e| ErrorKind::InvalidBlockProof(e))?;
+		.map_err(ErrorKind::InvalidBlockProof)?;
 	Ok(())
 }
 
@@ -489,7 +489,7 @@ pub fn rewind_and_apply_header_fork(
 ) -> Result<(), Error> {
 	let mut fork_hashes = vec![];
 	let mut current = header.clone();
-	while current.height > 0 && !ext.is_on_current_chain(&current, batch).is_ok() {
+	while current.height > 0 && ext.is_on_current_chain(&current, batch).is_err() {
 		fork_hashes.push(current.hash());
 		current = batch.get_previous_header(&current)?;
 	}
@@ -530,9 +530,9 @@ pub fn rewind_and_apply_fork(
 	// Rewind the txhashset extension back to common ancestor based on header MMR.
 	let mut current = batch.head_header()?;
 	while current.height > 0
-		&& !header_extension
+		&& header_extension
 			.is_on_current_chain(&current, batch)
-			.is_ok()
+			.is_err()
 	{
 		current = batch.get_previous_header(&current)?;
 	}
@@ -552,7 +552,7 @@ pub fn rewind_and_apply_fork(
 	for h in fork_hashes {
 		let fb = batch
 			.get_block(&h)
-			.map_err(|e| ErrorKind::StoreErr(e, format!("getting forked blocks")))?;
+			.map_err(|e| ErrorKind::StoreErr(e, "getting forked blocks".to_string()))?;
 
 		// Re-verify coinbase maturity along this fork.
 		verify_coinbase_maturity(&fb, ext, batch)?;
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -26,7 +26,7 @@ use grin_store as store;
 use grin_store::{option_to_not_found, to_key, Error, SerIterator};
 use std::sync::Arc;
 
-const STORE_SUBPATH: &'static str = "chain";
+const STORE_SUBPATH: &str = "chain";
 
 const BLOCK_HEADER_PREFIX: u8 = 'h' as u8;
 const BLOCK_PREFIX: u8 = 'b' as u8;
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -67,7 +67,7 @@ impl<T: PMMRable> PMMRHandle<T> {
 		fs::create_dir_all(path.clone())?;
 		let path_str = path
 			.to_str()
-			.ok_or_else(|| Error::from(ErrorKind::Other("invalid file path".to_owned())))?;
+			.ok_or_else(|| ErrorKind::Other("invalid file path".to_owned()))?;
 		let backend = PMMRBackend::new(path_str.to_string(), prunable, version, header)?;
 		let last_pos = backend.unpruned_size();
 		Ok(PMMRHandle { backend, last_pos })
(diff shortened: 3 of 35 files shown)


quentinlesceller:
Test comment for Keybase bot.

chain/src/txhashset/txhashset.rs
```diff
@@ -864,7 +864,7 @@ impl<'a> Committed for Extension<'a> {
 
 	fn kernels_committed(&self) -> Vec<Commitment> {
 		let mut commitments = vec![];
-		for n in 1..self.kernel_pmmr.unpruned_size() + 1 {
+		for n in 1..=self.kernel_pmmr.unpruned_size() {
```
jaspervdm: According to a discussion on Reddit [here](https://www.reddit.com/r/rust/comments/eiwhkn/why_is_iterator_so_much_faster/fctvmt4/), for loops with inclusive ranges might actually be significantly slower. So maybe we want to keep it like this?

(Same comment for the other spots this appears)

chain/src/txhashset/txhashset.rs
```diff
@@ -1307,12 +1307,12 @@ impl<'a> Extension<'a> {
...
 		let total_kernels = pmmr::n_leaves(self.kernel_pmmr.unpruned_size());
 		let mut tx_kernels: Vec<TxKernel> = Vec::with_capacity(KERNEL_BATCH_SIZE);
-		for n in 1..self.kernel_pmmr.unpruned_size() + 1 {
+		for n in 1..=self.kernel_pmmr.unpruned_size() {
 			if pmmr::is_leaf(n) {
 				let kernel = self
 					.kernel_pmmr
 					.get_data(n)
-					.ok_or::<Error>(ErrorKind::TxKernelNotFound.into())?;
+					.ok_or_else(|| Error::from(ErrorKind::TxKernelNotFound))?;
```
jaspervdm: Do we actually need `Error::from` here? We have a `From<ErrorKind>` impl on `Error`, so
```
.ok_or(ErrorKind::TxKernelNotFound)?;
```
might suffice

chain/src/txhashset/txhashset.rs
quentinlesceller: Wow nice find!

chain/src/pipe.rs
```diff
@@ -63,7 +63,7 @@ fn validate_pow_only(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result
 	if !header.pow.is_primary() && !header.pow.is_secondary() {
 		return Err(ErrorKind::LowEdgebits.into());
 	}
-	if !(ctx.pow_verifier)(header).is_ok() {
+	if (ctx.pow_verifier)(header).is_err() {
```
antiochp: :+1:

chain/src/store.rs
```diff
@@ -26,7 +26,7 @@ use grin_store as store;
 use grin_store::{option_to_not_found, to_key, Error, SerIterator};
 use std::sync::Arc;
 
-const STORE_SUBPATH: &'static str = "chain";
+const STORE_SUBPATH: &str = "chain";
```
antiochp: :rocket:

core/src/core/pmmr/pmmr.rs
```diff
@@ -302,7 +302,7 @@ where
 	/// Walks all unpruned nodes in the MMR and revalidate all parent hashes
 	pub fn validate(&self) -> Result<(), String> {
 		// iterate on all parent nodes
-		for n in 1..(self.last_pos + 1) {
+		for n in 1..self.last_pos + 1 {
```
antiochp: Personal preference here is the original version reads better, even if the parens are not required.

core/src/libtx/proof.rs
```diff
@@ -680,24 +632,9 @@ mod tests {
...
-				&keychain,
-				&builder,
-				amount,
-				&id,
-				&switch,
-				commit.clone(),
-				None,
-			)
-			.unwrap();
+			let proof = create(&keychain, &builder, amount, &id, &switch, commit, None).unwrap();
```
antiochp: Is rustfmt going to reformat these back to multiline fn calls (not sure when it decides to do this)?
Or was this actually the output of rustfmt?

core/src/pow/types.rs
```diff
@@ -389,7 +389,7 @@ impl Proof {
 	}
 }
 
-fn extract_bits(bits: &Vec<u8>, bit_start: usize, bit_count: usize, read_from: usize) -> u64 {
+fn extract_bits(bits: &[u8], bit_start: usize, bit_count: usize, read_from: usize) -> u64 {
```
antiochp: :+1:

keychain/src/mnemonic.rs
```diff
@@ -124,7 +124,7 @@ pub fn from_entropy(entropy: &Vec<u8>) -> Result<String, Error> {
 
 	let mut hash = [0; 32];
 	let mut sha2sum = Sha256::default();
-	sha2sum.input(&entropy.clone());
+	sha2sum.input(&(*entropy));
```
antiochp: This `&(*foo)` looks weird - what's this actually doing?
I'm guessing `entropy` cannot just be used directly, but not immediately clear to me why not.

util/src/rate_counter.rs
```diff
@@ -105,6 +105,6 @@ impl RateCounter {
 fn millis_since_epoch() -> u64 {
 	let since_epoch = SystemTime::now()
 		.duration_since(SystemTime::UNIX_EPOCH)
-		.unwrap_or(Duration::new(0, 0));
+		.unwrap_or_else(|_| Duration::new(0, 0));
```
antiochp: Trivial comment but why do this? Is it to avoid calling Duration::new(0,0) each time regardless?
If so maybe better to turn this into a constant and reuse it in a cleaner `unwrap_or()`?

core/src/pow/types.rs
quentinlesceller: Also there is several of place we could do something similar that I did not include in this PR.

core/src/core/pmmr/pmmr.rs
quentinlesceller: 👍

core/src/libtx/proof.rs
quentinlesceller: This is the output of rustfmt. I'm guessing that removing the `.clone()` gave us a line < MAX_CHAR_LINE or something like that.

keychain/src/mnemonic.rs
quentinlesceller: Indeed, fixed.

util/src/rate_counter.rs
quentinlesceller: Good point! 
From rust doc:

>Arguments passed to `unwrap_or` are eagerly evaluated; if you are passing the result of a function call, it is recommended to use `unwrap_or_else`, which is lazily evaluated

Actually we can simplify this since Rust 1.33. Will update. (https://stackoverflow.com/questions/26593387/how-can-i-get-the-current-time-in-milliseconds)

quentinlesceller:
@antiochp updated. See https://github.com/mimblewimble/grin/pull/3216/files#diff-cba101f43397c905b40c554acc3f2c05R107 for comment about `unwrap_or_else` and this particular function.

util/src/rate_counter.rs
```diff
@@ -103,8 +104,8 @@ impl RateCounter {
 // turns out getting the millisecs since epoch in Rust isn't as easy as it
 // could be
 fn millis_since_epoch() -> u64 {
-	let since_epoch = SystemTime::now()
-		.duration_since(SystemTime::UNIX_EPOCH)
-		.unwrap_or(Duration::new(0, 0));
-	since_epoch.as_secs() * 1000 + u64::from(since_epoch.subsec_millis())
+	match SystemTime::now().duration_since(SystemTime::UNIX_EPOCH) {
+		Ok(since_epoch) => since_epoch.as_millis().try_into().unwrap_or(0),
+		Err(_) => 0,
```
antiochp: `.ok_or(0)` ?

quentinlesceller: Good point. Addressed it and potentially created a monster 👹.

quentinlesceller: Looks like this is a bit to cutting edge for the CI. `map_or` was introduced in Rust 1.41.

quentinlesceller: Can be replaced by:
```
// turns out getting the millisecs since epoch in Rust isn't as easy as it
// could be
fn millis_since_epoch() -> u64 {
	SystemTime::now()
		.duration_since(SystemTime::UNIX_EPOCH)
		.map(|since_epoch| since_epoch.as_millis().try_into().unwrap_or(0))
		.unwrap_or(0)
}
```

PR #3217, opened 2020-01-31
Optimization: Pass SwitchCommitment by value instead of reference

quentinlesceller:
As the title says, this PR changes the way we pass the `SwitchCommitment` enum: by value instead of reference. 
Why? An enum is 1 byte in memory but a reference is 8 bytes.

### DIFF
--- a/keychain/src/keychain.rs
+++ b/keychain/src/keychain.rs
@@ -100,7 +100,7 @@ impl Keychain for ExtKeychain {
 		&self,
 		amount: u64,
 		id: &Identifier,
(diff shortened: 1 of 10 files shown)


keychain/src/keychain.rs
```diff
@@ -109,7 +109,7 @@ impl Keychain for ExtKeychain {
 			ext_key = ext_key.ckd_priv(&self.secp, &mut h, p.path[i as usize])?;
 		}
 
-		match *switch {
```
antiochp: :+1:

PR #3221, opened 2020-02-05
Attempt to read ipv4-mapped ipv6 to ipv4 if possible

quentinlesceller:
A potential additional operation during read some maybe not the best idea.

### DIFF
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -143,13 +143,15 @@ impl Readable for PeerAddr {
 			))))
 		} else {
 			let ip = try_iter_map_vec!(0..8, |_| reader.read_u16());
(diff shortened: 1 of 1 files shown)


quentinlesceller:
Only case that I have seen is 3. This PR do the job in that case.

antiochp:
So I think we also need similar logic when we compare ip addrs when looking to see if we are _already_ connected to a peer. 
We do this when accepting a connection (and maybe when trying to connect). If we are already connected then we want to identify the duplicate attempt and close the new connection. And I believe we do this based on a comparison of the ip address.

antiochp:
Related - #3208.

antiochp:
If we can get to point where we are confident this improves the overall situation regarding ipv4-mapped addresses on the network then we should aim to get this merged in for a potential `3.1.0` release.

PR #3224, opened 2020-02-06
keep output_pos index consistent with the output MMR file

### LINKED ISSUE #3222: COMMIT_POS_HGT index issue - potentially causing AlreadySpent errors
I think we have an issue with the `COMMIT_POS_HGT` index in lmbd.
This index is used to lookup (unspent) output MMR pos and height based on commitment.
We use this to look the output up when "spending" an output via `apply_input()`.

__Note:__ If we do not find the pos via the index then we assume the output is spent and raise an `AlreadySpent` error. So it is _critical_ this index is accurate.

We only need to maintain this index for the UTXO as there is no need to maintain index entries of historical spent outputs. We clean this index up during our regular chain compaction.

The issue arises because we are not considering the possibility of "rewind" for fork/reorg handling when we clean this index up. We clear the index based on the _current_ state of the UTXO. So _all_ spent outputs will be removed from the index, even those that were spent only 1 or 2 blocks ago.

If we compact the chain, then immediately rewind even a single block, we are going to find the index out of sync with the UTXO set in this rewound state. We can potentially "unspend" an output during rewind but this will not be guaranteed to be reflected in the index if it was recently cleared out.

Robust index cleaning - 
1. first rewind the UTXO set back to the horizon
1. note the size of the output MMR at the horizon
1. clear "old" spent entries in the index - 
    * remove outputs based on _rewound_ UTXO
    * but do not remove any _future_ UTXO entries based on size of rewound output MMR

It is hard to tell how often this might happen in practice but it may explain some of the `AlreadySpent` errors we see. These may not be "corruption" of the chain state from unclean shutdown, they _may_ be "corruption" from bad timing with chain compaction and fork processing.

antiochp:
Resolves #3222 

* we need to support rewind so index must maintain recently removed output pos
* cleanup the output_pos index migration code (no longer needed)
* cleanup the code for the old index and consolidate
* rework duplicate code in `get_output_pos()` and `get_output_pos_height()`

TODO - 
- [ ] Still need to test this under various scenarios locally and on mainnet.
- [ ] Probably good idea to split the fix out from the cleanup?

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -273,11 +273,6 @@ impl TxHashSet {
 		Ok(self.commit_index.get_block_header(&hash)?)
 	}
 
-	/// Get all outputs MMR pos
-	pub fn get_all_output_pos(&self) -> Result<Vec<(Commitment, u64)>, Error> {
-		Ok(self.commit_index.get_all_output_pos()?)
-	}
-
 	/// returns outputs from the given pmmr index up to the
 	/// specified limit. Also returns the last index actually populated
 	/// max index is the last PMMR index to consider, not leaf index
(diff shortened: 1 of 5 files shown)


chain/src/txhashset/txhashset.rs
```diff
@@ -368,61 +363,72 @@ impl TxHashSet {
...
+			.output_pos_iter()?
+			.filter(|(_, (pos, _))| {
+				// Note we use get_from_file() here as we want to ensure we have an entry
+				// in the index for *every* output still in the file, not just the "unspent"
+				// outputs. This is because we need to support rewind to handle fork/reorg.
+				// Rewind may "unspend" recently spent, but not yet pruned, outputs and the
+				// index must be consistent in this situation.
+				*pos <= last_pos && output_pmmr.get_from_file(*pos).is_none()
+			})
+			.map(|(key, _)| batch.delete(&key))
```
antiochp: This is the actual fix for #3222. 
The rest of the PR is cleanup.

antiochp:
Closing. Will reopen two separate PRs to separate the fix and the cleanup.

PR #3225, opened 2020-02-06
Convert ipv4-mapped ipv6 when accepting peer

quentinlesceller:
Linked to #3221.
Maybe a bit naive. This will attempt to convert ipv4-mapped ipv6 when accepting peer, so we don't accept twice the same IP (ipv4 and ipv4-mapped ipv6).

### DIFF
--- a/p2p/src/serv.rs
+++ b/p2p/src/serv.rs
@@ -14,7 +14,7 @@
 
 use std::fs::File;
 use std::io::{self, Read};
-use std::net::{Shutdown, SocketAddr, TcpListener, TcpStream};
+use std::net::{IpAddr, Shutdown, SocketAddr, SocketAddrV4, TcpListener, TcpStream};
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


quentinlesceller:
I'm not even sure `TcpListener` accept will give us a some ipv4 mapped ipv6.

antiochp:
> I'm not even sure `TcpListener` `accept()` will give us an ipv4 mapped ipv6.

My understanding is it will _if_ the local node is listening on `::` (rather than `0.0.0.0`).
i.e. If we are listening for both ipv4 and ipv6 connections then any ipv4 connections will themselves be mapped as we need a common way of representing them internally.

Which is making me think about why we are suddenly seeing a large number of these mapped addresses on the network.

* _single_ node on the network listening on `::`
* _every_ ipv4 address connecting to it will be seen as a mapped address by this node
* it will then send out a peer addrs msg containing many mapped addresses to its peers
* these then propagate through the network

PR #3227, opened 2020-02-10
Cleanup output pos idx

antiochp:
Related #3226. 

- [x] Merge #3226 first then rebase to clean the diff up in this PR.

We have two indices in the db -
 * COMMIT_POS_PREFIX
 * COMMIT_POS_HGT_PREFIX

The original `COMMIT_POS_PREFIX` (pos only) was replaced with `COMMIT_POS_HGT_PREFIX` (pos and height) and a migration process was used to convert from one to the other.
Now all nodes are on `3.0.0` this migration is no longer required and the old legacy index can be safely removed.

Also took the opportunity to rename it to `OUTPUT_POS_PREFIX` to make it clearer this index is specific to the output MMR.

Refactored `get_output_pos()` for less code duplication with `get_output_pos_height()`.

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -113,17 +113,7 @@ impl ChainStore {
 
 	/// Get PMMR pos for the given output commitment.
 	pub fn get_output_pos(&self, commit: &Commitment) -> Result<u64, Error> {
-		let res: Result<Option<(u64, u64)>, Error> = self
-			.db
-			.get_ser(&to_key(OUTPUT_POS_PREFIX, &mut commit.as_ref().to_vec()));
-		match res {
-			Ok(None) => Err(Error::NotFoundErr(format!(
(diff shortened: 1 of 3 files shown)


chain/src/store.rs
```diff
@@ -32,8 +32,7 @@ const BLOCK_HEADER_PREFIX: u8 = b'h';
 const BLOCK_PREFIX: u8 = b'b';
 const HEAD_PREFIX: u8 = b'H';
 const TAIL_PREFIX: u8 = b'T';
-const COMMIT_POS_PREFIX: u8 = b'c';
-const COMMIT_POS_HGT_PREFIX: u8 = b'p';
+const OUTPUT_POS_PREFIX: u8 = b'p';
```
quentinlesceller: 👍 for changing the name

PR #3228, opened 2020-02-10
kernel pos index

antiochp:
Maintain index of "recent" kernels so we can quickly lookup a kernel MMR pos (and block height) by kernel excess commitment.

This is similar to the existing `output_pos` index.
Important difference is outputs can be removed and pruned (after being spent), kernels are never removed. But we only want to maintain an index of "recent" kernels sufficient for the proposed NSKR relative kernel locks.

Compact the kernel pos index when we compact the txhashset. We maintain 2 weeks of kernel history in the index. This covers 7 days of full block history back to the horizon (for rewind purposes) and a further 7 days of kernel history to cover the max NSKR relative kernel lock period.

Populate the index as we go when calling `apply_kernel()` during processing of new full blocks.

Add test coverage around chain compaction and interaction with the kernel pos index.

Track an "undo list" for each full block so we can rewind the kernel_pos index as required when rewinding a block.

----

We do not need to track the kernel_pos index currently as it is unused.
But we need this implemented for efficient handling of NRD kernel lookups.
Once we have the NRD kernel feature variant we can modify the kernel_pos index to only index recent NRD kernels (currently we index all recent kernels).

----

Related #3273.

If we track `kernel_pos` via the "linked list" impl then we no longer need to maintain the "undo list" per block.
We can simply pop entries off the head of the list during rewind, rewinding each kernel in the block, leaving the index in the previous state in the case of duplicates.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1107,7 +1107,7 @@ impl Chain {
 
 		// Remove historical blocks from the db unless we are running in archive mode.
 		if !self.archive_mode {
-			self.remove_historical_blocks(&header_pmmr, &mut batch)?;
+			self.remove_historical_blocks(&header_pmmr, &batch)?;
 		}
 
 		// Compact the txhashset itself (rewriting the pruned backend files).
@@ -1154,7 +1154,7 @@ impl Chain {
 	/// Get the position of the kernel if it exists in the kernel_pos index.
 	/// The index is limited to 14 days of recent kernels.
 	pub fn get_kernel_pos(&self, excess: Commitment) -> Result<u64, Error> {
-		Ok(self.txhashset.read().get_kernel_pos(excess)?)
+		self.txhashset.read().get_kernel_pos(excess)
 	}
 
 	/// outputs by insertion index
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1090,7 +1090,7 @@ impl Chain {
 			} else {
 				global::cut_through_horizon()
 			};
-			let next_compact = tail.height.saturating_add(threshold);
+			let next_compact = tail.height.saturating_add(threshold.into());
 			if next_compact > head.height {
 				debug!(
 					"compact: skipping startup compaction (next at {})",
@@ -1152,7 +1152,7 @@ impl Chain {
 	}
 
 	/// Get the position of the kernel if it exists in the kernel_pos index.
-	/// The index is limited to 7 days of recent kernels.
+	/// The index is limited to 14 days of recent kernels.
 	pub fn get_kernel_pos(&self, excess: Commitment) -> Result<u64, Error> {
 		Ok(self.txhashset.read().get_kernel_pos(excess)?)
 	}
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -980,8 +980,11 @@ impl Chain {
 			batch.save_body_tail(&tip)?;
 		}
 
-		// Rebuild our output_pos index in the db based on fresh UTXO set.
+		// Initialize output_pos index in the db based on current UTXO set.
 		txhashset.init_output_pos_index(&header_pmmr, &batch)?;
+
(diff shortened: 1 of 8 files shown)


chain/src/txhashset/txhashset.rs
```diff
@@ -474,6 +488,69 @@ impl TxHashSet {
...
+				if let Some(kernel) = self.kernel_pmmr_h.backend.get_data(pos) {
+					batch.save_kernel_pos_height(&kernel.excess, pos, current_header.height)?;
+					kernel_count += 1;
+				}
+			}
+		}
+		debug!(
+			"init_kernel_pos_index: {} kernels, took {}s",
+			kernel_count,
+			now.elapsed().as_secs()
```
quentinlesceller: Would be interesting to see if there is any difference with the comment above.

chain/src/txhashset/txhashset.rs
```diff
@@ -474,6 +488,69 @@ impl TxHashSet {
...
+		let now = Instant::now();
+		let head_header = batch.head_header()?;
+		let cutoff_height = head_header
+			.height
+			.saturating_sub(global::kernel_index_horizon().into());
+		let deleted = batch
+			.kernel_pos_iter()?
+			.filter(|(_, (_, height))| *height < cutoff_height)
+			.map(|(key, _)| batch.delete(&key))
+			.count();
```
quentinlesceller: Out of curiosity I'm wondering if there is a more efficient way to do that. Like some kind of `filter_map`.
 Also we are loosing the error of the batch delete without any logging?

chain/src/chain.rs
```diff
@@ -1140,6 +1147,12 @@ impl Chain {
 		Ok(self.txhashset.read().get_output_pos(commit)?)
 	}
 
+	/// Get the position of the kernel if it exists in the kernel_pos index.
+	/// The index is limited to 7 days of recent kernels.
```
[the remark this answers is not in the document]
quentinlesceller: This is outdated.

chain/src/txhashset/txhashset.rs
```diff
@@ -474,6 +488,69 @@ impl TxHashSet {
...
+		let cutoff_hash = header_pmmr.get_header_hash_by_height(cutoff_height)?;
+		let cutoff_header = batch.get_block_header(&cutoff_hash)?;
+		let cutoff_pos = if cutoff_header.height == 0 {
+			1
+		} else {
+			batch.get_previous_header(&cutoff_header)?.kernel_mmr_size + 1
+		};
+		let mut kernel_count = 0;
+		let mut current_header = cutoff_header;
+		for pos in cutoff_pos..=self.kernel_pmmr_h.last_pos {
```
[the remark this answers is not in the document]
antiochp: Ah good catch. :+1:

chain/src/txhashset/txhashset.rs
```diff
@@ -354,6 +355,16 @@ impl TxHashSet {
 		Ok(self.commit_index.get_output_pos(&commit)?)
 	}
 
+	/// Get the position of the kernel if it exists in the kernel_pos index.
+	/// The index is limited to 7 days of recent kernels.
```
quentinlesceller: Same comment.

antiochp:
Flagging as [DNM] for now - discussions under way around potentially maintaining the `kernel_pos` index purely in memory.

chain/src/chain.rs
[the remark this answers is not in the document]
antiochp: :+1: updated.

chain/src/txhashset/txhashset.rs
antiochp: What do you mean in terms of "efficient"? 
We need to iterate over the full set of entries in the db.

chain/src/txhashset/txhashset.rs
antiochp: :+1:

antiochp:
Closing, replaced with #3273 (which is a more general solution).

PR #3233, opened 2020-02-18
Update croaring

quentinlesceller:
Let's see if it fix the CI issues.

PR #3236, opened 2020-02-21
Block input bitmap rework

### LINKED ISSUE #3235: Rework output_pos index (better txhashset transactional support)
Recent exploration into the implementation of an index of "recently seen" kernels led us down the path of maintaining an "undo" list per block. This way we can robustly handle duplicate kernels, maintaining the pos of the most recent instance of the kernel in the index, while being able to handle rewind scenarios involving previous instances of the kernel.

Related - https://github.com/mimblewimble/grin/pull/3228

We do something similar to this for the `output_pos` index currently where we track a bitmap of spent output pos per block. And during rewind we use this per-block bitmap to ensure the UTXO set is consistent with rewound chain state. Note that this allows the utxo set to be rewound reliably but leaves "false-positives" in the `output_pos` index as we do not rewind the index itself.

So we have always used the `output_pos` index as non-authoritative. During rewind we can (and will expect to) encounter "false-positive" results from the index. We may find an entry in the index for a given output, but we do not yet know for sure if this index entry is accurate. So we go to the PMMR itself for authoritative view of chain state.

The process is roughly - 
* look in output_pos for possible result
  * if nothing in index then output is spent
  * if we find a result in the index, look in the output PMMR
* compare output commitment between index and PMMR
  * if match then output is unspent
  * otherwise output is spent

It became apparent during the kernel index exploration that there is a better way to do this that can also be applied to the `output_pos` index.
Rather than storing a bitmap of spent output positions which is by definition unsorted, we can store a vec of output positions where the sort order is consistent with the inputs to the block.
This provides the ability to map positions to input commitments themselves. Rewind can now take advantage of this to ensure the output_pos index remains consistent with the rewound chain state. The `output_pos` index can be maintained transactionally alongside block processing, in both regular (apply new block) and rewound directions.
This eliminates the possibility for "false positives" in the index lookup. If we find an entry in the index then we know the output is unspent. If we do not find an entry, the output is spent, or never existed.

This involves a change to what we store in the db. We need to store a vec of output_pos (and associated block heights) rather than the existing serialized bitmap per block. 
This also involves changes to both `apply_block()` and the `rewind_block` impls to update the `output_pos` as necessary, within the txhashset extension (and therefore transactionally).

Reworking the `output_pos` index to handle transactional rewind in this way gives us a clear path forward to use the same approach for the `kernel_pos` index. Rather than "spending" outputs we replace kernels with more recent instances (in the "recently seen" index).

We have this working on a branch. PR coming shortly for these changes. Ideally we can get this into `3.1.0-beta` and allow it to be sufficiently tested and released as part of `3.1.0`.
Tracking this here so we can tag it as scheduled for inclusion in `3.1.0`.

antiochp:
Resolves #3235.
Extracted the "undo list" concept from the kernel index exploration (#3228) and took advantage of it for outputs (where undo == unspend during rewind).

This allows the `output_pos` index (quick lookup of output pos by output commitment) to more closely track the current utxo set in a transactional manner.
We now keep the `output_pos` index updated when applying new blocks _and_ when rewinding existing blocks under a fork scenario.

This eliminates the edge-case of needing to account for "false positive" results in the index.
We still do not treat the `output_pos` as authoritative but there are currently believed to be no cases where it will temporarily diverge from the set of unspent output pos.

We rebuild (via optimized `init`) the output_pos_index on - 
  1. startup, to ensure the node starts up in a consistent state
  1. compact (once outputs are removed and pruned/compacted we no longer need to keep them in the index)
  1. during fast sync, after receiving txhashset from a peer

We log discrepencies between the index and set of outputs during index rebuild.
This should give us some insight into whether our assumptions are correct here.
If we are comfortable that we can begin treating this index as authoritative then we will only need to rebuild the index on sync (3) above.

This PR is a relatively large change in terms of code changes. But conceptually it is pretty straightforward.

* Update `output_pos` index when we call `apply_block()`
  * add new index entries for new outputs
  * remove index entries for spent outputs
* Update `output_pos` on rewind (per block)
  * remove "future" outputs that no longer exist
  * re-add "unspent" entries to the index based on per-block "spent_index" (aka undo list)
* Store a spent_index per block when we store the associated block in the db 

We used to do something similar but not exactly the same - 
* we stored a bitmap of spent pos per block
  * this allowed us to rewind the utxo set itself, but not the associated index
* we did not undo these in the index itself when rewinding
  * this led to false positives when reading the index, so we always went to the PMMR to corroborate what the index said (if index was inconsistent with PMMR then treat as spent)

The majority of the code change involves breaking the `rewind` functionality out to do it per iteratively block. Before we could simply take the union of all the per-block bitmaps. This has been reworked to rewind block by block to take advantage of our ability to rewind the index itself.

----

We were originally hoping to get the `kernel_index` impl into `3.1.0` but this was not possible given the timing of the `3.1.0` release. This PR gives us a way of confirming the approach works on the existing `output_pos` index.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -23,7 +23,7 @@ use crate::core::pow;
 use crate::error::{Error, ErrorKind};
 use crate::store;
 use crate::txhashset;
-use crate::types::{Options, OutputPos, Tip};
+use crate::types::{CommitPos, Options, Tip};
 use crate::util::RwLock;
 use grin_store;
 use std::sync::Arc;
@@ -431,7 +431,7 @@ fn apply_block_to_txhashset(
 	block: &Block,
 	ext: &mut txhashset::ExtensionPair<'_>,
 	batch: &store::Batch<'_>,
-) -> Result<Vec<OutputPos>, Error> {
+) -> Result<Vec<CommitPos>, Error> {
 	let spent = ext.extension.apply_block(block, batch)?;
 	ext.extension.validate_roots(&block.header)?;
 	ext.extension.validate_sizes(&block.header)?;
@@ -444,7 +444,7 @@ fn apply_block_to_txhashset(
 fn add_block(
 	b: &Block,
 	block_sums: &BlockSums,
-	spent: &Vec<OutputPos>,
+	spent: &Vec<CommitPos>,
 	batch: &store::Batch<'_>,
 ) -> Result<(), Error> {
 	batch.save_block(b)?;
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -23,7 +23,7 @@ use crate::core::pow;
 use crate::error::{Error, ErrorKind};
 use crate::store;
 use crate::txhashset;
-use crate::types::{Options, Tip};
+use crate::types::{Options, OutputPos, Tip};
 use crate::util::RwLock;
 use grin_store;
 use std::sync::Arc;
@@ -121,7 +121,7 @@ pub fn process_block(b: &Block, ctx: &mut BlockContext<'_>) -> Result<Option<Tip
 	let ref mut header_pmmr = &mut ctx.header_pmmr;
 	let ref mut txhashset = &mut ctx.txhashset;
 	let ref mut batch = &mut ctx.batch;
-	let block_sums = txhashset::extending(header_pmmr, txhashset, batch, |ext, batch| {
+	let (block_sums, spent) = txhashset::extending(header_pmmr, txhashset, batch, |ext, batch| {
 		rewind_and_apply_fork(&prev, ext, batch)?;
 
 		// Check any coinbase being spent have matured sufficiently.
@@ -143,22 +143,24 @@ pub fn process_block(b: &Block, ctx: &mut BlockContext<'_>) -> Result<Option<Tip
 		// Apply the block to the txhashset state.
 		// Validate the txhashset roots and sizes against the block header.
 		// Block is invalid if there are any discrepencies.
-		apply_block_to_txhashset(b, ext, batch)?;
+		let spent = apply_block_to_txhashset(b, ext, batch)?;
 
 		// If applying this block does not increase the work on the chain then
 		// we know we have not yet updated the chain to produce a new chain head.
+		// We discard the "child" batch used in this extension (original ctx batch still active).
+		// We discard any MMR modifications applied in this extension.
 		let head = batch.head()?;
 		if !has_more_work(&b.header, &head) {
 			ext.extension.force_rollback();
 		}
 
-		Ok(block_sums)
+		Ok((block_sums, spent))
 	})?;
 
 	// Add the validated block to the db along with the corresponding block_sums.
 	// We do this even if we have not increased the total cumulative work
 	// so we can maintain multiple (in progress) forks.
-	add_block(b, &block_sums, &ctx.batch)?;
+	add_block(b, &block_sums, &spent, &ctx.batch)?;
 
 	// If we have no "tail" then set it now.
 	if ctx.batch.tail().is_err() {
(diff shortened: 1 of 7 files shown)


antiochp:
Will rebase once #3233 has been merged to fix the croaring issue.

chain/src/txhashset/txhashset.rs
```diff
@@ -378,35 +376,63 @@ impl TxHashSet {
...
 
+		// Iterate over the current output_pos index, removing any entries that
+		// do not point to to the correct output.
+		let mut removed_count = 0;
+		for (key, (pos, _)) in batch.output_pos_iter()? {
+			if let Some(out) = output_pmmr.get_data(pos) {
+				if let Ok(pos_via_mmr) = batch.get_output_pos(&out.commitment()) {
+					if pos == pos_via_mmr {
+						continue;
+					}
```
jaspervdm: Is this cleanup foolproof? Lets imagine the following situation
- `output_pos` entry for commit `A` points to pos `x`
- `output_pos` entry for commit `B` points to pos `x`
- output MMR at pos `x` contains commit `B`

I don't think this loop will delete the incorrect entry for commit `A` from the `output_pos` db, because `pos == pos_via_mmr`.

jaspervdm:
I think these changes behave as expected, however I am not an expert in this area of the code. Would appreciate it if we have one additional review

chain/src/txhashset/txhashset.rs
antiochp: :+1: Yes I think you are right. We need to check both the pos matches _and_ the commitment matches here to `continue`.

chain/src/types.rs
```diff
@@ -258,16 +259,41 @@ impl OutputRoots {
...
 
-/// A helper to hold the output pmmr position of the txhashset in order to keep them
-/// readable.
+/// Minimal struct representing an output commitment at a known position (and block height)
+/// in an MMR.
 #[derive(Debug)]
-pub struct OutputMMRPosition {
-	/// The hash at the output position in the MMR.
-	pub output_mmr_hash: Hash,
+pub struct OutputPos {
```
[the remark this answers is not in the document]
antiochp: :+1: I thought we were actually using it in `api` but its implicit and uses `CommitPos` now.

chain/src/txhashset/txhashset.rs
antiochp: I had managed to convince myself that this was not necessary but I was not considering your scenario there.

chain/src/pipe.rs
```diff
@@ -429,20 +431,25 @@ fn apply_block_to_txhashset(
...
 	ext: &mut txhashset::ExtensionPair<'_>,
 	batch: &store::Batch<'_>,
-) -> Result<(), Error> {
-	ext.extension.apply_block(block, batch)?;
+) -> Result<Vec<CommitPos>, Error> {
+	let spent = ext.extension.apply_block(block, batch)?;
 	ext.extension.validate_roots(&block.header)?;
 	ext.extension.validate_sizes(&block.header)?;
-	Ok(())
+	Ok(spent)
```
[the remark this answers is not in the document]
antiochp: Yes intended behavior, albeit a little convoluted and confusing.

Every _new_ block processed goes through `process_block()` and if successful we store the spent_list and the block_sums along with the full block.
A call to `rewind_and_apply_fork()` reprocesses existing blocks, reading them from the db. And for all these existing blocks we will have already stored the spent_list in the db.

This whole "what gets committed to the db and when" area of block processing needs a cleanup to simplify some of this logic. We handle both the MMR backed files and the lmdb transaction/batch and there is some complexity around transactional semantics in there.
This is for a separate PR though.

antiochp:
Making an executive decision to merge this now. Feel free to continue review post merge.
I'd like to get the initial `3.1.0-beta` tagged today.

PR #3249, opened 2020-02-26
explicit read_timeout and write_timeout during hand/shake

### LINKED ISSUE #3248: no read/write timeout on initial hand/shake messages
Once a peer is up and running in connected state - 
 * We explicitly set a read_timeout for reading msg headers and msg bodies.
 * We also explicitly set a write_tiemout for writing msgs.

But during the initial handshake when accepting a new peer or initiating a new peer connection we do not set any read or write timeout.
This will cause the connection to block indefinitely.

Each new peer connection spawns a new thread `peer_connect` and we have seen situations where we have a large number of these threads, blocked on `read_exact`.

Screen Shot 2020-02-26 at 11 12 30 AM

We need to consider adding a read_timeout and a write_timeout when doing this initial handshake. It may be reasonable to simply cancel this handshake if we do cannot write a msg to the peer or if we cannot read a msg within the timeout period.

This is probably another situation where now that we reliably use blocking IO on these connections the error handling is different and we now block in places where we did not do before (due to accidental usage of non-blocking IO).

antiochp:
Resolves #3248.

We were blocking indefinitely on the initial hand/shake.
Once the peer is up and running for broadcast we will replace these timeout values with the existing settings for msg header and body reads and msg writes.

without these timeouts we see `peer_connect` threads stuck, blocked indefinitely waiting for `read_exact` on connections that may have closed suddenly.

### DIFF
--- a/p2p/src/handshake.rs
+++ b/p2p/src/handshake.rs
@@ -35,8 +35,20 @@ const NONCES_CAP: usize = 100;
 /// 10 should be enough since most of servers don't have more than 10 IP addresses.
 const ADDRS_CAP: usize = 10;
 
-/// The read_timeout and write_timeout for initial hand/shake messages.
-const HANDSHAKE_TIMEOUT: Duration = Duration::from_millis(2_000);
+/// The initial Hand message should come in immediately after the connection is initiated.
+/// But for consistency use the same timeout for reading both Hand and Shake messages.
+const HAND_READ_TIMEOUT: Duration = Duration::from_millis(10_000);
+
+/// We need to allow time for the peer to receive our Hand message and send back a Shake reply.
+const SHAKE_READ_TIMEOUT: Duration = Duration::from_millis(10_000);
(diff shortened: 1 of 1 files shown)


jaspervdm:
Given that the handshake reading is not in a loop like in the reader thread of `poll()`, it would mean that the connection is dropped if no message is received within `HANDSHAKE_TIMEOUT`. So in that case 2 seconds is maybe a bit too short.

antiochp:
> Given that the handshake reading is not in a loop like in the reader thread of `poll()`, it would mean that the connection is dropped if no message is received within `HANDSHAKE_TIMEOUT`. So in that case 2 seconds is maybe a bit too short.

That's a good point - this would need to cover the period waiting for a `shake` after first sending a `hand` msg to a peer as we immediately try to read from the stream after sending.
I'll set these both to 10 seconds and see how it behaves.

PR #3253, opened 2020-02-28
simplify when block_sums and spent_index are added to the db

antiochp:
Store `spent_index` when calling `apply_block()` and `block_sums` when calling `verify_block_sums()`. 

This lets us simplify `process_block` and `add_block` and makes it a little clearer when/where these are added to the db (via the current batch). We do not need to pass the data around and return it from various fns.

Conceptually we have 2 (nested) places where we commit data to the db - 
1. We commit the block itself to the db if we process the block successfully (fork or otherwise)
1. We _only_ commit the following data iff this block also increases total work - 
    * chain head
    * various index entries

We do this via a child batch - this child batch includes the update to chain head, indices etc. and is rolled back along with the MMR changes if the block does not increase total work.
We apply everything transactionally (db + MMRs via txhashset extension) then discard this transaction.
The outer db transaction is then committed to commit the block itself to the db (unless validation failed or other error).

This PR moves the block_sums and spent_index into the child batch. This is safe to do we rebuild both for all blocks along a fork being processed. And we only make assumptions about them being present in the db for the main chain itself.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -156,9 +156,10 @@ pub fn process_block(b: &Block, ctx: &mut BlockContext<'_>) -> Result<Option<Tip
 		Ok(())
 	})?;
 
-	// Add the validated block to the db along with the corresponding block_sums.
-	// We do this even if we have not increased the total cumulative work
-	// so we can maintain multiple (in progress) forks.
+	// Add the validated block to the db.
+	// Note we do this in the outer batch, not the child batch from the extension
+	// as we only commit the child batch if the extension increases total work.
+	// We want to save the block to the db regardless.
 	add_block(b, &ctx.batch)?;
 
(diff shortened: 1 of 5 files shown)


chain/src/pipe.rs
```diff
@@ -154,13 +153,13 @@ pub fn process_block(b: &Block, ctx: &mut BlockContext<'_>) -> Result<Option<Tip
 			ext.extension.force_rollback();
 		}
 
-		Ok((block_sums, spent))
+		Ok(())
 	})?;
 
 	// Add the validated block to the db along with the corresponding block_sums.
```
jaspervdm: Could you update this comment, since it no longer stores the block_sums?

antiochp:
Related - #3267 (in terms of the outer and child lmdb transactions).

PR #3256, opened 2020-02-29
API: don't error on missing output

jaspervdm:
When the wallet is querying the api for a list of commitment containing unconfirmed outputs, they show up as errors in the log like this:
```
20190614 21:37:47.937 ERROR grin_api::handlers::chain_api - Failure to get output for commitment 0951860ff6a74dc41902f08cbe7614b440af0fff2ad629d2847cd0a8c4a9211fd4 with error Not found.
20190614 21:37:47.937 ERROR grin_api::handlers::chain_api - Failure to get output for commitment 08707b918dacbc8e5e4990dc1e47ce0c28b879297f3440959953cce7954683d21f with error Not found.
20190614 21:37:47.938 ERROR grin_api::handlers::chain_api - Failure to get output for commitment 087dd75eaa1f0e43bf54f0773ccde17005683c4fd02aa9b0e19fac51727b10c74a with error Not found.
```
This is not really necessary, since it is expected behaviour that some of the commitments are not in the unspent set. Furthermore, those errors were discarded higher up in the call chain, meaning real errors were not propagated to the wallet response properly.

This PR fixes both these behaviours by changing some internal functions from `Result<.., Error>` to `Result<Option<..>, Error>`.

### DIFF
--- a/api/src/handlers/utils.rs
+++ b/api/src/handlers/utils.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::chain;
+use crate::chain::types::CommitPos;
 use crate::core::core::{OutputFeatures, OutputIdentifier};
 use crate::rest::*;
 use crate::types::*;
@@ -29,11 +30,11 @@ pub fn w<T>(weak: &Weak<T>) -> Result<Arc<T>, Error> {
 		.ok_or_else(|| ErrorKind::Internal("failed to upgrade weak refernce".to_owned()).into())
 }
 
-/// Retrieves an output from the chain given a commit id (a tiny bit iteratively)
-pub fn get_output(
-	chain: &Weak<chain::Chain>,
+/// Internal function to retrieves an output by a given commitment
+fn retrieve_output(
+	chain: &Arc<chain::Chain>,
 	id: &str,
-) -> Result<(Output, OutputIdentifier), Error> {
+) -> Result<Option<(Commitment, CommitPos, OutputIdentifier)>, Error> {
 	let c = util::from_hex(String::from(id)).context(ErrorKind::Argument(format!(
 		"Not a valid commitment: {}",
 		id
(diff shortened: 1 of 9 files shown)


api/src/handlers/utils.rs
```diff
@@ -79,63 +82,27 @@ pub fn get_output_v2(
...
+	};
 
-	// We need the features here to be able to generate the necessary hash
-	// to compare against the hash in the output MMR.
-	// For now we can just try both (but this probably needs to be part of the api
-	// params)
-	let outputs = [
-		OutputIdentifier::new(OutputFeatures::Plain, &commit),
-		OutputIdentifier::new(OutputFeatures::Coinbase, &commit),
-	];
```
antiochp: This always felt kind of awkward. Are we still doing two lookups to see if either match? Or is this not necessary?

api/src/handlers/utils.rs
```diff
@@ -49,28 +50,30 @@ pub fn get_output(
...
+		None => return Ok(None),
+	};
+
+	Ok(Some((
+		Output::new(&commit, output_pos.height, output_pos.pos),
+		identifier,
+	)))
 }
 
 /// Retrieves an output from the chain given a commit id (a tiny bit iteratively)
```
antiochp: See comment below - is this still "iteratively"?

chain/src/chain.rs
```diff
@@ -504,7 +504,7 @@ impl Chain {
 	/// spent. This querying is done in a way that is consistent with the
 	/// current chain state, specifically the current winning (valid, most
 	/// work) fork.
-	pub fn is_unspent(&self, output_ref: &OutputIdentifier) -> Result<CommitPos, Error> {
+	pub fn is_unspent(&self, output_ref: &OutputIdentifier) -> Result<Option<CommitPos>, Error> {
```
antiochp: Does this now return `None` if an output is spent? 
Might be worth refactoring this and naming it `get_unspent()` rather than `is_unspent()`?

chain/src/txhashset/txhashset.rs
```diff
@@ -223,23 +223,23 @@ impl TxHashSet {
 	/// Check if an output is unspent.
 	/// We look in the index to find the output MMR pos.
 	/// Then we check the entry in the output MMR and confirm the hash matches.
-	pub fn is_unspent(&self, output_id: &OutputIdentifier) -> Result<CommitPos, Error> {
+	pub fn is_unspent(&self, output_id: &OutputIdentifier) -> Result<Option<CommitPos>, Error> {
```
antiochp: Consider renaming this `get_unspent()`. It was never returning true/false anyway...

api/src/handlers/utils.rs
jaspervdm: It's still there, just moved to a different function (`retrieve_output()`). I agree its awkward, we can get rid of it by changing the `is_unspent()` on txhashset to only take a commitment and ignore the output features from the output mmr

chain/src/chain.rs
jaspervdm: Yes, agreed. Will update

api/src/handlers/utils.rs
antiochp: I think that might actually make sense (to ignore the features).
Let me think about this a bit more. There is a weird edge case with this where we do not allow duplicate outputs, but we can spend one and replace it with another, with the new one having different features. At that point is the "output" spent or not?
From the wallet perspective they are different, but from the node impl we spend based purely on commitment.

antiochp: We should leave this as-is for now - its a material change to ignore the features here (and may have other implications). Either way it should be a separate change to this PR.

chain/src/store.rs
```diff
@@ -115,16 +115,20 @@ impl ChainStore {
...
 	}
 
 	/// Get PMMR pos and block height for the given output commitment.
-	pub fn get_output_pos_height(&self, commit: &Commitment) -> Result<(u64, u64), Error> {
-		option_to_not_found(
-			self.db
-				.get_ser(&to_key(OUTPUT_POS_PREFIX, &mut commit.as_ref().to_vec())),
-			|| format!("Output position for: {:?}", commit),
-		)
+	pub fn get_output_pos_height(&self, commit: &Commitment) -> Result<Option<(u64, u64)>, Error> {
```
antiochp: We have `get_output_pos_height()` on the batch and on the store itself. We should keep this fns consistent with each other if possible.

And we call this in a few places in txhashset.rs and I think we then need to account for `None` when we call it here -

https://github.com/mimblewimble/grin/blob/db6fb3a9922b970d7cadc2b87ec1783e6c825b4f/chain/src/txhashset/txhashset.rs#L428

jaspervdm: Good call, forgot about the batch. Updating it now

jaspervdm: Having to keep them consistent makes me think we should put those functions in a trait and have both the store and batch implement them. But we can do that later in a new PR :)

PR #3260, opened 2020-03-02
no need to rehash with index to compare output with input spending it

antiochp:
When validating an input (that it spends an unspent output) we do not need to re-hash the features+commit and compare against the hash in the output MMR. 

We can simply compare the OutputIdentifier (features+commit) directly.

Comparing 32 bytes of a commitment will be cheaper than hashing the commitment and then comparing 32 bytes of the resulting hash.
Reading the commitment from the data file will be no more expensive than reading the hash from the hash file.

Implement `From<Input> for OutputIdentifier` to keep the code clean here.

TODO -
- [x] implement the same equality check in `is_unspent()`

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -288,7 +288,7 @@ impl OutputPrintable {
 			OutputType::Transaction
 		};
 
-		let out_id = core::OutputIdentifier::from_output(&output);
+		let out_id = core::OutputIdentifier::from(*output);
 		let res = chain.is_unspent(&out_id);
 		let (spent, block_height) = if let Ok(output_pos) = res {
 			(false, Some(output_pos.height))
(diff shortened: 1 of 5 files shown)


antiochp:
Related - https://github.com/mimblewimble/grin/pull/3256

We also do a `hash_with_index()` in `txhashet.is_unspent()`. We should make the same change there.

api/src/types.rs
```diff
@@ -288,7 +288,7 @@ impl OutputPrintable {
 			OutputType::Transaction
 		};
 
-		let out_id = core::OutputIdentifier::from_output(&output);
+		let out_id = core::OutputIdentifier::from(*output);
```
jaspervdm: Maybe its better to change the `From<Output>` impl on `OutputIdentifier` to `From<&Output>`. This would avoid having to copy over the proof

antiochp: :+1:
Yes. I wanted to do that but could not remember what needed to be done to achieve this.
I was forgetting you could do `From<&Output>`.

PR #3262, opened 2020-03-07
[WIP] reopen header_pmmr every time, no lock required

antiochp:
Turns out it is really fast to open the header pmmr as it is non-prunable and fixed size.
It is just backed by two mmap'd files (data and hash files).
These are both strictly append-only. There is a possible edge case where we may truncate and rewrite data during a reorg but the size of the file will never decrease.

Every time we accept a new block header we add the block header to the data file and the hash to the hash file along with any necessary parent hashes.

So rather than jumping through a bunch of hoops trying to manage the `RwLock` around it across multiple peer threads and the syncer thread we can just reopen it based on the data on disk each time.

Getting rid of the `RwLock` intuitively feels wrong. But if we only ever append a new header to the data file and one or more hashes to the hash file then we can safely read the files off disk (via mmap) at _any time_ regardless of any other "readers" or "writers".

And the OS lets us do this really efficiently via mmap as the pages remain cached in memory across multiple reads of the file, without needing to maintain a reference to the file. We can simply mmap the file repeatedly and we do not start with an empty cache each time, the data is _already_ memory mapped if the file has not changed on disk.

So we can open the header PMMR and read the `header_head` lock free.
Then we can open the header PMMR and append a new header to it, lock free.
Then if we open it again we will read the new `header_head`, again lock free.

We are basically being optimistic and assume the data will _always_ be consistent and in a readable state. And the append-only semantics of the MMR backend files allow this.
This would be expensive if we were reopening and rereading the backend files each time but `mmap` makes this efficient to do repeatedly.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -148,8 +148,8 @@ pub struct Chain {
 	adapter: Arc<dyn ChainAdapter + Send + Sync>,
 	orphans: Arc<OrphanBlockPool>,
 	txhashset: Arc<RwLock<txhashset::TxHashSet>>,
-	header_pmmr: Arc<RwLock<txhashset::PMMRHandle<BlockHeader>>>,
-	sync_pmmr: Arc<RwLock<txhashset::PMMRHandle<BlockHeader>>>,
+	// header_pmmr: Arc<RwLock<txhashset::PMMRHandle<BlockHeader>>>,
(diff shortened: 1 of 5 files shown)


antiochp:
Closing. I'd like to continue experimenting with this but closing the draft PR to clean things up a bit.

PR #3264, opened 2020-03-09
Use smaller LMDB Allocation Size in non-production modes

yeastplume:
Adds a new parameter `alloc_chunk_size` to the creation of a new LMDB handle. The intention is to allow calling tests to provide a custom allocation size to reduce the use of disk space (particularly on windows).

Should not have any effect on grin itself at the moment, this is intended for use by the wallet's integration tests.

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -46,7 +46,7 @@ pub struct ChainStore {
 impl ChainStore {
 	/// Create new chain store
 	pub fn new(db_root: &str) -> Result<ChainStore, Error> {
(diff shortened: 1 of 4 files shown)


yeastplume:
Edit, tried to be too clever. Just simplified so that non-production modes use the smaller allocation size in all cases.

jaspervdm:
I'm wondering if there is some negative effect these changes will have on the node tests, I guess in the worst case it will just result in a bunch of db resizes right?

yeastplume:
I thought the same and came to the same conclusion, we can tweak/modify if we run across any testing issues (seeing as how we need to rebuild a lot of integration tests anyhow)

PR #3266, opened 2020-03-10
Write header hash (aka checkpoint) to disk on successful txhashset extension (based on rewind point)

antiochp:
We perform a "rewind" before processing each block (and block header).
The "happy path" where we process the _next_ block based on current chain head still "rewinds" but is optimized to be very cheap and is effectively a no-op.

We take advantage of the "rewind" to identify the "fork point" in the chain.
This fork point represents the earliest block that is subject to being rewritten in the context of processing the new block. Everything earlier than (and including this block) is immutable and the data is guaranteed not to change. Everything later than this may be rewritten to files on disk and the local db.

The immutable data _cannot_ be corrupted regardless of how abruptly the nodes shuts down. The backend files are truncated (at the fork point) then new bytes are written to them. 

This PR notes the "fork point" and stores the block header hash to disk in a file `chain_data/grin.checkpoint` in human readable hex format.

This is done every time the node processes a new block and happens prior to writing anything to disk or the local db. Anything written to disk subsequently will be ignored on startup and reprocessed, until the _next_ block is then processed.

On node startup we read the hash from the checkpoint file and rewind to this point before continuing.
In the majority of cases this will simply represent the header at `h-1` relative to the current chain head. So we end up reprocessing the latest full block. Conceptually this simply forces the node to reprocess the latest block, ensuring that any data partially written before shutdown reprocessed and fixed if corrupted.

Related - #3265 (more robust parsing of a header hash in hex form in a file on disk).

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1406,6 +1406,8 @@ fn setup_head(
 	sync_pmmr: &mut txhashset::PMMRHandle<BlockHeader>,
 	txhashset: &mut txhashset::TxHashSet,
 ) -> Result<(), Error> {
+	let checkpoint = txhashset.last_checkpoint();
+
 	let mut batch = store.batch()?;
 
 	// Apply the genesis header to header and sync MMRs.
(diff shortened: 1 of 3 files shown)


antiochp:
For reference, we can test "corruption" via something along there lines (remove the final hash from the output MMR hash file) - 

```
truncate -s -32 chain_data/txhashset/output/pmmr_hash.bin
```

antiochp:
Proposal here is a simple grin.checkpoint file that contains a block header hash in hex format - 

```
cat chain_data/grin.checkpoint 
000012bacde26540693e9cb85867cdc671e02ac1f6981fdd7a221c5b3f26ceda
```

Bonus feature is you can edit this file manually and force the node to rewind and replay blocks from a user specified checkpoint. This _must_ be a block on the current chain, it will not work on a fork for example.

antiochp:
This is not a perfect solution as we do not handle the `leafset` transactionally and there is still room for data corruption if that file is out of sync with the other backend files during a hard shutdown (power failure etc.)

Tracking a checkpoint on disk allows for several scenarios to be robustly recovered from and simplifies the recovery process during init.

antiochp:
> Looks good! Could we add a test for the checkpoint? Create a checkpoint file, add some blocks, rewind to checkpoint and reapply blocks.

Yeah good point - let me take a look and see what's involved there.

antiochp:
Tests failing on windows as we are trying to reopen the chain without releasing the reference to the old chain instance.
I need to rework that tests so windows is happy with it.

antiochp:
The more I think about it the more I think this is the wrong approach.
We have some inconsistencies in our code around `fsync` and the assumptions made around writing files to disk. Going to write some of this up in a separate issue for reference later.

PR #3268, opened 2020-03-11
go back to storing header_head in the db

antiochp:
This PR reverts the "read head from the header MMR directly" changes, simplifies `header_head` handling and maintains the `header_head` in the db.

We changed the way we handle the `header_head` in #3045. We used to maintain the header_head in the db and that PR changed it to simply read the head of the header MMR from the header MMR files directly.

The rationale for the change was to reduce the number of things we need to keep consistent (MMR file + db index).

I'm now convinced this was a bad choice and the wrong direction to go in.

The RwLock around the header MMR means we cannot read the current head without taking a read lock and this blocks if anything (block processing, header processing, sync process) currently has a write lock on the MMR.

This necessitated the whole `try_header_head()` timeout based workaround.

The fact that writing the `header_head` to the db is transactional is actually beneficial here - 
* we can read from the db regardless of what other threads are doing
* once the header_head is committed to the db we know this is good (and we are not at risk of reading inconsistent state from the non-transactional MMR files)

So being forced to keep the MMR and the db "in sync" and consistent is a _good thing_. The transactional semantics of the db means we only update the `header_head` once we commit the db transaction, vs the MMR backend files that can (and will) be inconsistent at times.

In concrete terms this will get rid of the large volume of "failed to obtain lock for try_header_head" errors during fast sync.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1212,25 +1212,11 @@ impl Chain {
 			.map_err(|e| ErrorKind::StoreErr(e, "chain tail".to_owned()).into())
 	}
 
-	/// Tip (head) of the header chain if read lock can be acquired reasonably quickly.
-	/// Used by the TUI when updating stats to avoid locking the TUI up.
-	pub fn try_header_head(&self, timeout: Duration) -> Result<Option<Tip>, Error> {
-		self.header_pmmr
-			.try_read_for(timeout)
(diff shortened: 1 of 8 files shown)


antiochp:
This has been pretty heavily tested now in various situations.
Migrating from currently released version to this code correctly adds the missing db entry on startup etc.

Merging.

PR #3273, opened 2020-03-18
[WIP] Output pos linked list

antiochp:
Initial experiment with some primitives in `store.rs` toward a "linked list" maintained in lmdb for the `output_pos` index.

This is likely throwaway code. We still need to get the RFC together to decide on the final approach.

But I needed to sit down and write some code just to focus on something today for a couple of hours.


antiochp:
Thinking about how we make this more generic, to support both `output_pos` and `kernel_pos` indexes. Presumably via a trait and associated type. Outputs will use `OutputPos` (with output features) and kernels the simpler `CommitPos`.

antiochp:
The general approach here (once it works) feels like a good candidate for fully replacing #3228.
Rather than tracking "undo list" per block and rebuilding the index based on "undo" we track per kernel via the "linked list" pushing and popping off the stack of positions.

Proposed approach:
* Get this working here for kernel pos and cleanup
* Think through the prune/compact logic - 
  * during compaction we need to prune the kernel_pos index (keep 2 weeks of data)
* Implement for coinbase kernels initially (as proof of concept)?
  * We want to do this for NoRecentDuplicate kernels eventually but they do not yet exist

antiochp:
Closing. Replaced with #3302.

Issue #3261, opened 2020-03-07
Grin 3.1.0 crash on start


shrikus:
Just updated grin node from 3.1.0-beta.2 to 3.1.0 (amd64 release binaries), and it's crashing on start:

```
# gdb --args grin "server" "run"
GNU gdb (Ubuntu 8.1-0ubuntu3.2) 8.1.0.20180409-git
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from grin...done.
(gdb) r
Starting program: /usr/local/bin/grin server run
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGILL, Illegal instruction.
0x0000555555d8a3cb in ra_portable_deserialize ()
(gdb) bt
#0  0x0000555555d8a3cb in ra_portable_deserialize ()
#1  0x0000555555d8a811 in roaring_bitmap_portable_deserialize_safe ()
#2  0x0000555555cc7f28 in grin_store::read_bitmap ()
#3  0x0000555555cc830d in grin_store::leaf_set::LeafSet::open ()
#4  0x0000555555d0b443 in grin_store::pmmr::PMMRBackend<T>::new ()
#5  0x0000555555c9437b in grin_chain::txhashset::txhashset::TxHashSet::open ()
#6  0x0000555555ce4bad in grin_chain::chain::Chain::init ()
#7  0x00005555558b9db1 in grin_servers::grin::server::Server::new ()
#8  0x0000555555686e0d in grin_servers::grin::server::Server::start ()
#9  0x00005555556b0e3d in grin::cmd::server::start_server_tui ()
#10 0x00005555556b1917 in grin::cmd::server::server_command ()
#11 0x00005555556be8b1 in grin::real_main ()
[...]

PR #3274, opened 2020-03-23
Build croaring without optimizations

jaspervdm:
I have modified croaring-rs to support a flag that turns off certain optimizations. These optimizations can result in incompatibilities of our binaries with older cpus.

Because we are not sure if/when the changes will be merged upstream (see [this PR](https://github.com/saulius/croaring-rs/pull/60)), I have published a fork of the repository on crates.io under the name [croaring-mw](https://crates.io/crates/croaring-mw). We can in the future always switch back if needed.

Once this is merged I suggest we release a v3.1.1 so people have a working binary to use if they so choose, instead of being forced to build from source.

Related: https://github.com/mimblewimble/grin/issues/3261

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -333,18 +333,18 @@ dependencies = [
 ]
 
 [[package]]
-name = "croaring"
-version = "0.4.4"
+name = "croaring-mw"
+version = "0.4.5"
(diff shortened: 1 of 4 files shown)


antiochp:
Thanks for taking this on. 🙏

PR #3278, opened 2020-03-25
Use latest hosted agent for CI

quentinlesceller:
Which are:
macOS 10.15 Catalina
Ubuntu 18.04 LTS
Windows server 2019

- Use `ubuntu-latest`, `macos-latest` and `windows-latest`.
- Remove the shared install steps.

Build should be a bit faster.

more info here:
https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#software

### DIFF
--- a/.ci/install.yml
+++ b/.ci/install.yml
@@ -1,17 +1,5 @@
 steps:
   - script: |
       git apply .ci/win.patch
-      choco install -y llvm
-    displayName: Windows Install LLVM
+    displayName: Windows Apply Patch
     condition: eq( variables['Agent.OS'], 'Windows_NT' )
-  - script: |
-      clang --version
-      brew uninstall llvm
-      clang --version
-    displayName: macOS Uninstall LLVM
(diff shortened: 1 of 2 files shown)


jaspervdm:
Looks good, although I'm a bit careful about changing stuff like this since croaring is pretty picky about versions etc. Have you tried creating a binary on the new OS, does it run fine?

Actually on second thought: are we sure we want to use `-latest` for everything? I'm expecting it might break our CI randomly in the future when a new version is out

quentinlesceller:
I haven't tried to create a binary.

Regarding the `latest`, we can expect at most once a year some issue with the CI. This is using only stable release.

quentinlesceller:
I'll try a release on my repo later today to see how it goes.

quentinlesceller:
Release working well @jaspervdm https://github.com/quentinlesceller/grin/releases/tag/v3.1.2-test.1

PR #3282, opened 2020-03-29
Update seed.rs add seed-node

MCM-Mike:
added mainnet-seed.grinnode.live

By using [DNS Round-Robin](https://en.wikipedia.org/wiki/Round-robin_DNS) (3 nodes) on mainnet-seed.grinnode.live we can provide a high-available seed-node.

---
name: Pull Request
about: Pull Request checklist
title: ''
labels: ''
assignees: ''

---
If your PR is a work in progress, please feel free to create it and include a [WIP] tag in the PR name. We encourage everyone to PR early and often so that other developers know what you're working on.

Before submitting your PR for final review, please ensure that it:

* Updates any documentation that's affected by the PR

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -39,6 +39,7 @@ const MAINNET_DNS_SEEDS: &'static [&'static str] = &[
 	"mainnet.seed.grin.lesceller.com", // q.lesceller@gmail.com
 	"mainnet.seed.grin.prokapi.com",   // hendi@prokapi.com
 	"grinseed.yeastplume.org",         // yeastplume@protonmail.com
+	"mainnet-seed.grinnode.live",      // info@grinnode.live
 ];
 const FLOONET_DNS_SEEDS: &'static [&'static str] = &[
 	"floonet.seed.grin.icu",           // gary.peverell@protonmail.com


antiochp:
Follow up question - if we use round-robin here with a single DNS name resolving to multiple ip addresses. Is this actually giving us any benefits in terms of high-availability? We're going to just resolve it once, get one of many ip addresses and try that one (regardless of it being available or not).

Does the DNS here handle liveness of the underlying nodes behind the DNS?
And do we have any kind of "retry" functionality on our end when starting up and trying the seeds?

MCM-Mike:
> Follow up question - if we use round-robin here with a single DNS name resolving to multiple ip addresses. Is this actually giving us any benefits in terms of high-availability? We're going to just resolve it once, get one of many ip addresses and try that one (regardless of it being available or not).

**In parts it was my problem, when arguing DNS Round-Robin is a High-Available solution. Which technically it is not.** 
If one of the 3/3 IP addresses is down, 1/3 of the requests are not going to find an Grin-Node connection. It's just an easy way for distributing loads , but no real high-availability. 

> Does the DNS here handle liveness of the underlying nodes behind the DNS?
> And do we have any kind of "retry" functionality on our end (in the node) when starting up and trying the seeds?

**No, as we do not correlate the "liveness" of an IP address and add or delete it to the TLD A-Records.** 

Based on what I did read on the documentation, all Seed-Nodes are only normal Grin-Nodes. 
If you want we could change the IP address behind the DNS A-Record into our real High-Available but I thought by using a Round-Robin DNS setting this would be enough. 

We can always roll back to only one IP address or substitute the IP address with the HA-IP address. 

**Lastly as I was checking all Seed-Nodes , a couple of them where down, this was my motivation to open this pull request.**

antiochp:
@MCM-Mike thanks for the answers. Just to be clear, this was not in any way a criticism of what you are doing here. I just wanted to make sure I understood how the grin node was going to handle and respond to a single "seed" that resolved to multiple underlying grin nodes.

I'm :+1: on adding an additional seed here. Particularly one that is likely going to be at least as reliable as the existing seeds, if not more so.

PR #3291, opened 2020-04-09
Update Cursive to 0.14 and update backend.

quentinlesceller:
Update Cursive to 0.14.
Use pancurses-backend instead of ncurses-backend: 
>uses the pancurses library, which forwards calls to ncurses-rs on Linux/macOS or pdcurses-sys on Windows. Requires ncurses to be installed on the system.

I don't think we need the complexity for Windows in the Cargo.toml.


quentinlesceller:
Doing some additional tests on Windows

Issue #3138, opened 2019-11-26
tui logs tab causing panic


antiochp:
Not entirely sure what causes this to occur but I'm able to reproduce this locally reliably in debug mode.

```
thread 'main' panicked at 'attempt to subtract with overflow': src/bin/tui/logs.rs:97
```

Its happening here - 

```
p.print((0, p.size.y - 1 - i), span.content);
```

So it appears `p.size.y` is `0` and we try and subtract from it, going negative.

Fixable with a `saturating_sub()` but not sure what the actual reason for this happening is.
I'll put a PR up to fix this. But might be worth investigating what caused this to happen.

antiochp:
There must be a race condition or some kind of timing issue during startup - I can't reproduce it now using either a `debug` or `release` build.

It's really weird. 

Are we assuming something is initialized during tui startup (say the logging infrastructure) but it may not be guaranteed to be?

either way I think the `saturating_sub()` is worth adding regardless and is presumably safe either way.

da-kami:
I ran into a similar issue when syncing. I stumbled over this ticket when searching for a solution. Since my problem seems to be related I post here, please ping me if you want this in a new ticket.

I got this error repeatedly in the initial fast-sync, in step. Here are the logs of one run:

```
20200413 22:05:44.145 INFO grin_util::logger - log4rs is initialized, file level: Info, stdout level: Warn, min. level: Info
20200413 22:05:44.146 INFO grin - Using configuration file at [removed]
20200413 22:05:44.146 INFO grin - This is Grin version 3.1.1 (git v3.1.1), built for x86_64-apple-darwin by rustc 1.42.0 (b8cedc004 2020-03-09).
20200413 22:05:44.146 WARN grin::cmd::server - Starting GRIN in UI mode...
20200413 22:05:44.146 INFO grin_servers::grin::server - Starting server, genesis block: 40adad0aec27
20200413 22:05:44.227 INFO grin_servers::grin::server - Starting rest apis at: 127.0.0.1:3413
20200413 22:05:44.228 WARN grin_api::handlers - Starting HTTP Node APIs server at 127.0.0.1:3413.
20200413 22:05:44.228 WARN grin_api::handlers - HTTP Node listener started.
20200413 22:05:44.228 INFO grin_servers::grin::server - Starting dandelion monitor: 127.0.0.1:3413
20200413 22:05:44.228 WARN grin_servers::grin::server - Grin server started.
[...]

PR #3294, opened 2020-04-14
Fix rare panic in TUI

jaspervdm:
According to https://github.com/mimblewimble/grin/issues/3138#issuecomment-613357541, sometimes during IBD it could happen that the previous downloaded size is bigger than the downloaded size. This PR prevents the panic by not underflowing. It doesn't fix the underlying issue (which is that the downloaded size doesn't always increase monotonically) but since we're going to revamp the IBD anyway this fix should suffice for now.


antiochp:
Is this that downloaded size == previous downloaded size? Or that it actually _decreases_?

jaspervdm:
If they are equal subtracting them won't panic. So yes somehow the number actually decreases.

antiochp:
Maybe this involves a slow download timing out and replaced by a new download - that's the only way I can see the bytes decreasing.

PR #3297, opened 2020-04-16
Refactor SyncState

hashmap:
Method `sync_error()` return type was simplified.
`update_txhashset_download()` was  made type safe, which eliminates a runtime enum variant's  check. As a bonus matching code is shorter and simpler when we do a partial destructuing, which happens quite often.

### DIFF
--- a/chain/src/types.rs
+++ b/chain/src/types.rs
@@ -15,14 +15,13 @@
 //! Base types that the block chain pipeline requires.
 
 use chrono::prelude::{DateTime, Utc};
-use std::sync::Arc;
 
 use crate::core::core::hash::{Hash, Hashed, ZERO_HASH};
 use crate::core::core::{Block, BlockHeader, HeaderVersion};
 use crate::core::pow::Difficulty;
 use crate::core::ser::{self, PMMRIndexHashable, Readable, Reader, Writeable, Writer};
 use crate::error::{Error, ErrorKind};
-use crate::util::RwLock;
+use crate::util::{RwLock, RwLockReadGuard};
 
 bitflags! {
 /// Options for block validation
@@ -40,7 +39,6 @@ bitflags! {
 
 /// Various status sync can be in, whether it's fast sync or archival.
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Deserialize, Serialize)]
-#[allow(missing_docs)]
 pub enum SyncStatus {
 	/// Initial State (we do not yet know if we are/should be syncing)
 	Initial,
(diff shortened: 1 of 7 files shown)


chain/src/types.rs
```diff
@@ -143,8 +165,8 @@ impl SyncState {
 	}
 
 	/// Get sync error
-	pub fn sync_error(&self) -> Arc<RwLock<Option<Error>>> {
-		Arc::clone(&self.sync_error)
+	pub fn sync_error(&self) -> RwLockReadGuard<Option<Error>> {
+		self.sync_error.read()
 	}
```
antiochp, line 170: It would be nice if we could avoid the `RwLockReadGuard` here with something along the lines of what we do with `peers` in p2p -

```
	pub fn sync_error(&self) -> Option<Error> {
		self.sync_error.read().as_ref().cloned()
	}
```

But it looks like the error is not actually cloneable...

hashmap: This is a tricky thing, initially I went for the clone approach, because I agree that returning Guard may look strange, but if we return a detached copy the rwlock guarantees are lost for us, a client gets a snapshot of maybe obsolete state. This is not a big deal in the current code, but in general it doesn't sound great. We have the same situation with state filed, we return a copy, I'm not sure if it's the right approach, was thinking should we remove `Clone` impl or not. We may end up dealing with a wrong state for some period of time.

antiochp: Yeah - we either retain the lock and prevent the case of seeing stale state (or error).
Or we take a "snapshot" of the current state (or error) and return that, and allow the lock to be releases.

I'm not sure we gain much by retaining the lock here.

hashmap: `sync_error` may be harmless, but it all depends on the client behavior, which may change in the future. `status` is more interesting, eg here we have a racy check https://github.com/mimblewimble/grin/blob/master/servers/src/grin/sync/state_sync.rs#L141 `update` may run against a different state. At the same time exposing Guard in `status()` would lead to a deadlock.

chain/src/types.rs
```diff
@@ -114,37 +144,54 @@ impl SyncState {
...
 
 	/// Update the syncing status
-	pub fn update(&self, new_status: SyncStatus) {
-		if self.status() == new_status {
```
hashmap, line 118: This check is not great, we take a read lock, compare and then release the lock, so when we obtain a write lock the state may be different already.

chain/src/types.rs
```diff
@@ -114,37 +144,54 @@ impl SyncState {
...
-	/// Update txhashset downloading progress
-	pub fn update_txhashset_download(&self, new_status: SyncStatus) -> bool {
-		if let SyncStatus::TxHashsetDownload { .. } = new_status {
-			let mut status = self.current.write();
-			*status = new_status;
-			true
+	/// Update the syncing status if predicate f is satisfied
+	pub fn update_if<F>(&self, new_status: SyncStatus, f: F) -> bool
+	where
+		F: Fn(SyncStatus) -> bool,
```
hashmap, line 169: it's a bit awkward to have predicate here, but some variants of `SyncStatus` have state attached which makes matching painful

chain/src/types.rs
```diff
@@ -114,37 +144,54 @@ impl SyncState {
...
+
 	/// Communicate sync error
 	pub fn set_sync_error(&self, error: Error) {
 		*self.sync_error.write() = Some(error);
 	}
 
 	/// Get sync error
-	pub fn sync_error(&self) -> Arc<RwLock<Option<Error>>> {
-		Arc::clone(&self.sync_error)
+	pub fn sync_error(&self) -> Option<String> {
```
hashmap, line 190: We don't need Error per se, just the fact of its existence and description, returning String makes life much easier

antiochp: :+1: yeah that makes sense.

PR #3298, opened 2020-04-17
check_known now takes total_difficulty into consideration

antiochp:
During block processing we `check_known` and halt processing if the block is "already known".
This avoids wasting resources reprocessing blocks from the network.

We treat a block as known if it matches the current chain head (reprocessing a recent block) or if the full block already exists in the local db (reprocessing a historical block).

With the work on "checkpoints" in #3266 there is another case to consider - one of "future blocks". If the chain head is reset or reverted to an earlier block (for _any_ reason) the current block processing refuses to (re)process blocks that are "already known". 

We can get into tricky situations where the chain head will refuse to progress because we incorrectly believe a block is "known" yet not actually on the current chain.

This PR makes `check_known` take `total_difficulty` into account when comparing blocks against current chain head. If the block would increase total difficulty then we want to treat it as "unknown" regardless of local db.

This makes is significantly easier to reset the chain head and still have block processing pipeline work correctly for reprocessing subsequent blocks.

Note: This _only_ applies in situations where the chain head is explicitly reset. Under normal block processing the chain head only ever progresses in one direction, with total difficulty strictly increasing.

Added test coverage to demonstrate "forgetting" about an "already known" block, based on total difficulty relative to chain head.

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -45,12 +45,17 @@ pub struct BlockContext<'a> {
 	pub verifier_cache: Arc<RwLock<dyn VerifierCache>>,
 }
 
-// Check if we already know about this block for various reasons
-// from cheapest to most expensive (delay hitting the db until last).
-fn check_known(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(), Error> {
-	check_known_head(header, ctx)?;
-	check_known_store(header, ctx)?;
-	Ok(())
+// If this block has greater total difficulty than treat as unknown in current context.
+// If it matches current chain head (latest or previous hash) then we know about it.
+// If it exists in the local db then we know about it.
+fn check_known(header: &BlockHeader, head: &Tip, ctx: &BlockContext<'_>) -> Result<(), Error> {
+	if header.total_difficulty() > head.total_difficulty {
+		Ok(())
+	} else {
+		check_known_head(header, head)?;
+		check_known_store(header, head, ctx)?;
+		Ok(())
+	}
 }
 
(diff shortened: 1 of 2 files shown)


chain/src/pipe.rs
```diff
@@ -231,9 +239,13 @@ pub fn process_block_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) ->
 	// Check this header is not an orphan, we must know about the previous header to continue.
 	let prev_header = ctx.batch.get_previous_header(&header)?;
 
-	// Check if we know about the full block for this header.
-	if check_known(header, ctx).is_err() {
-		return Ok(());
+	// If we have already processed the full block for this header then done.
+	// Note: "already known" in this context is success so subsequent processing can continue.
+	{
```
hashmap, line 244: for my education - why do we need a lexical scope here?

chain/tests/test_block_known.rs
```diff
@@ -0,0 +1,81 @@
...
+		let genesis = chain
+			.get_block(&chain.get_header_by_height(0).unwrap().hash())
+			.unwrap();
+		let head = chain.head().unwrap();
+		let latest = chain.get_block(&head.last_block_h).unwrap();
+		(latest, genesis)
+	};
+
+	// attempt to reprocess latest block
+	{
```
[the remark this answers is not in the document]
antiochp: We cannot call `init_chain()` repeatedly on windows from the same dir. 
I believe it does not release some file handles or something.

chain/src/pipe.rs
antiochp: See above.

PR #3299, opened 2020-04-17
Refactor needs_syncing

hashmap:
We return false in some "error" cases, eg when we don't have peers. It may trigger chain compaction for no reason, because we treat sync -> no sync state change as successful completion of syncing. So we should not return false in this case, we may just get a network error in the middle of the sync.
Also small code simplification

### DIFF
--- a/servers/src/grin/sync/syncer.rs
+++ b/servers/src/grin/sync/syncer.rs
@@ -17,7 +17,6 @@ use std::thread;
 use std::time;
 
 use crate::chain::{self, SyncState, SyncStatus};
-use crate::core::global;
 use crate::core::pow::Difficulty;
 use crate::grin::sync::body_sync::BodySync;
 use crate::grin::sync::header_sync::HeaderSync;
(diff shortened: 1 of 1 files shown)


servers/src/grin/sync/syncer.rs
```diff
@@ -71,31 +70,12 @@ impl SyncRunner {
...
-		loop {
-			if self.stop_state.is_stopped() {
+		for _i in 0..wait_secs {
+			if self.stop_state.is_stopped() || self.peers.more_or_same_work_peers()? > MIN_PEERS {
 				break;
 			}
-			let wp = self.peers.more_or_same_work_peers()?;
-			// exit loop when:
-			// * we have more than MIN_PEERS more_or_same_work peers
-			// * we are synced already, e.g. grin was quickly restarted
```
hashmap, line 85: this part is already covered by `peers.more_or_same_work_peers()`

servers/src/grin/sync/syncer.rs
```diff
@@ -71,31 +70,12 @@ impl SyncRunner {
...
-			// * timeout
-			if wp > MIN_PEERS
-				|| (wp == 0
-					&& self.peers.enough_outbound_peers()
-					&& head.total_difficulty > Difficulty::zero())
-				|| n > wait_secs
-			{
-				if wp > 0 || !global::is_production_mode() {
```
hashmap, line 93: not sure why we have it

servers/src/grin/sync/syncer.rs
```diff
@@ -149,13 +125,8 @@ impl SyncRunner {
 			thread::sleep(time::Duration::from_millis(10));
 
 			let currently_syncing = self.sync_state.is_syncing();
-
 			// check whether syncing is generally needed, when we compare our state with others
-			let (needs_syncing, most_work_height) = unwrap_or_restart_loop!(self.needs_syncing());
-			if most_work_height > 0 {
-				// we can occasionally get a most work height of 0 if read locks fail
```
hashmap, line 156: not anymore

yeastplume:

PR #3300, opened 2020-04-17
Don't swallow errors in body_sync

hashmap:
We handle them inside but it makes the code more cluttered


antiochp:
Testing this locally now.

PR #3302, opened 2020-04-18
[wip] kernel pos index (linked list)

antiochp:
[tbd]


chain/src/chain.rs
```diff
@@ -296,6 +297,11 @@ impl Chain {
 	/// Returns true if it has been added to the longest chain
 	/// or false if it has added to a fork (or orphan?).
 	fn process_block_single(&self, b: Block, opts: Options) -> Result<Option<Tip>, Error> {
+		// Process the header first.
+		// If invalid then fail early.
+		// If valid then continue with block processing with header_head committed to db etc.
+		self.process_block_header(&b.header, opts)?;
```
tromp, line 303: in what cases would we call process_block_single without having validated/processed the header already?
(I would naively expect never)
Does header validation not commit it to db?

chain/src/linked_list.rs
```diff
@@ -0,0 +1,582 @@
...
+	}
+}
+
+enum_from_primitive! {
+	#[derive(Copy, Clone, Debug, PartialEq)]
+	enum ListEntryVariant {
+		// Start at 2 here to differentiate from ListWrapperVariant above.
+		Head = 2,
+		Tail = 3,
+		Middle = 4,
```
tromp, line 52: Do we really need a linked list with 5 different variants?
Doesn't the database readily support maps from 64-bit ints to 64-bit ints?
We could map the MMR leaf position of an NRD-kernel to the next lower leaf position of an NRD-kernel with the same top k bits, if any. Besides that, we have a 2^k entry map giving the largest MMR leaf position of any NRD-kernel with those top k bits. I guess we could start with k=8 and it would take a long time before NRD kernel usage increases to the point where we'd want to migrate to k=16.

Issue #3288, opened 2020-04-02
Tracking issue for `relative-kernels`


lehnberg:
This is a tracking issue for RFC000X relative-kernels, https://github.com/mimblewimble/grin-rfcs/pull/19

```
As this RFC has not been merged yet, this is a placeholder
issue for the time being, and will be updated if and when
the RFC has been accepted.
```

Once all the above have been implemented and deployed we can close this tracking issue.

PR #3303, opened 2020-04-20
Enable NRD kernel support (noop) feature flagged

antiochp:
NRD kernel serialization/deserialization support
This PR effectively treats NRD kernels as NO-OP (but only valid after HF3).

RFC - https://github.com/mimblewimble/grin-rfcs/pull/47
Tracking issue - https://github.com/mimblewimble/grin/issues/3288

Global/local chain_type config PR - https://github.com/mimblewimble/grin/pull/3327/files (feature flag support based on this).

----

TODO - 

- [x] add new NRD kernel variant
- [x] add v1 and v2 ser/deser for new kernel variant
  - [x] test coverage for basic tx creation + ser/deser
  - [x] NRD kernel relative height within range 1 to WEEK_HEIGHT inclusive
- [x] add block validation rule for NRD kernels based on HF3 (header version 4)
  - [x] add core tests for block validation
- [x] add block processing pipeline test to cover valid block containing NRD kernel
- [x] add txpool validation rule (reject NRD txs if earlier than HF3 block height)
  - [x] add txpool test to cover accept/reject rule based on NRD kernels
- [x] fuzz tests (to be covered in separate issue/PR)
- [x] "feature flag" (global/local) for NRD kernel support
  - [x] during block validation
  - [x] during txpool/stempool tx validation

Note: these validation rules only cover serialization/deserialization of individual NRD kernels.
These rules do not cover "relative locks" between pairs of NRD kernels.

### DIFF
--- a/chain/tests/mine_nrd_kernel.rs
+++ b/chain/tests/mine_nrd_kernel.rs
@@ -21,7 +21,6 @@ use grin_util as util;
 
 use self::chain_test_helper::{clean_output_dir, genesis_block, init_chain};
 use crate::chain::{Chain, Options};
-use crate::core::core::hash::Hashed;
 use crate::core::core::{Block, KernelFeatures, NRDRelativeHeight, Transaction};
 use crate::core::libtx::{build, reward, ProofBuilder};
 use crate::core::{consensus, global, pow};
--- a/chain/tests/mine_nrd_kernel.rs
+++ b/chain/tests/mine_nrd_kernel.rs
@@ -14,17 +14,91 @@
 
 mod chain_test_helper;
 
-use self::chain_test_helper::{clean_output_dir, mine_chain};
+use grin_chain as chain;
+use grin_core as core;
+use grin_keychain as keychain;
+use grin_util as util;
+
+use self::chain_test_helper::{clean_output_dir, genesis_block, init_chain};
+use crate::chain::{Chain, Options};
+use crate::core::core::hash::Hashed;
+use crate::core::core::{Block, KernelFeatures, NRDRelativeHeight, Transaction};
+use crate::core::libtx::{build, reward, ProofBuilder};
+use crate::core::{consensus, global, pow};
+use crate::keychain::{ExtKeychain, ExtKeychainPath, Identifier, Keychain};
+use chrono::Duration;
+
+fn build_block<K>(chain: &Chain, keychain: &K, key_id: &Identifier, txs: Vec<Transaction>) -> Block
+where
+	K: Keychain,
+{
+	let prev = chain.head_header().unwrap();
+	let next_header_info = consensus::next_difficulty(1, chain.difficulty_iter().unwrap());
+	let fee = txs.iter().map(|x| x.fee()).sum();
+	let reward =
+		reward::output(keychain, &ProofBuilder::new(keychain), key_id, fee, false).unwrap();
+
+	let mut block = Block::new(&prev, txs, next_header_info.clone().difficulty, reward).unwrap();
+
+	block.header.timestamp = prev.timestamp + Duration::seconds(60);
+	block.header.pow.secondary_scaling = next_header_info.secondary_scaling;
+
+	chain.set_txhashset_roots(&mut block).unwrap();
+
+	let edge_bits = global::min_edge_bits();
+	block.header.pow.proof.edge_bits = edge_bits;
+	pow::pow_size(
+		&mut block.header,
+		next_header_info.difficulty,
+		global::proofsize(),
+		edge_bits,
+	)
+	.unwrap();
+
+	block
+}
 
 #[test]
 fn mine_block_with_nrd_kernel() {
+	global::set_mining_mode(global::ChainTypes::AutomatedTesting);
+	util::init_test_logger();
+
 	let chain_dir = ".grin.nrd_kernel";
 	clean_output_dir(chain_dir);
-	let chain = mine_chain(chain_dir, 9);
+
+	let keychain = ExtKeychain::from_random_seed(false).unwrap();
+	let pb = ProofBuilder::new(&keychain);
+	let genesis = genesis_block(&keychain);
+	let chain = init_chain(chain_dir, genesis.clone());
+
+	for n in 1..9 {
+		let key_id = ExtKeychainPath::new(1, n, 0, 0, 0).to_identifier();
+		let block = build_block(&chain, &keychain, &key_id, vec![]);
+		chain.process_block(block, Options::MINE).unwrap();
+	}
+
 	assert_eq!(chain.head().unwrap().height, 8);
 
-	// now mine a block containing a tx with an NRD kernel.
+	let key_id1 = ExtKeychainPath::new(1, 1, 0, 0, 0).to_identifier();
+	let key_id2 = ExtKeychainPath::new(1, 2, 0, 0, 0).to_identifier();
+	let tx = build::transaction(
+		KernelFeatures::NoRecentDuplicate {
+			fee: 20000,
+			relative_height: NRDRelativeHeight::new(1440).unwrap(),
+		},
+		vec![
+			build::coinbase_input(consensus::REWARD, key_id1.clone()),
+			build::output(consensus::REWARD - 20000, key_id2.clone()),
+		],
+		&keychain,
+		&pb,
+	)
+	.unwrap();
+
+	let key_id9 = ExtKeychainPath::new(1, 9, 0, 0, 0).to_identifier();
+	let block = build_block(&chain, &keychain, &key_id9, vec![tx]);
+	chain.process_block(block, Options::MINE).unwrap();
+	chain.validate(false).unwrap();
 
 	clean_output_dir(chain_dir);
-	panic!("wat");
 }
--- a/chain/tests/mine_nrd_kernel.rs
+++ b/chain/tests/mine_nrd_kernel.rs
@@ -1,0 +1,30 @@
+// Copyright 2020 The Grin Developers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod chain_test_helper;
+
+use self::chain_test_helper::{clean_output_dir, mine_chain};
+
+#[test]
+fn mine_block_with_nrd_kernel() {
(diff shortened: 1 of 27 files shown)


antiochp:
This got me thinking - presumably we make this new kernel variant valid starting at HF block height, i.e. they are valid a block with height >= h and invalid at height < h.

What do we do for tx validity? Do we base this on _current_ block height also? A tx containing one of these kernels is invalid if current latest block height is < h?
It feels like there is an edge case with a rewind/reorg/fork at HF block height.
Do we need to handle this edge case? The txpool/stempool could contain _potentially_ invalid txs if we rewind back earlier than the HF block height.

More of a HF question in general and not specific to this kernel variant - it would apply to any consensus breaking change involving transaction data.

antiochp:
Related https://github.com/mimblewimble/grin/pull/3327

chain/tests/mine_nrd_kernel.rs
```diff
@@ -0,0 +1,105 @@
...
+	let genesis = genesis_block(&keychain);
+	let chain = init_chain(chain_dir, genesis.clone());
+
+	for n in 1..9 {
+		let key_id = ExtKeychainPath::new(1, n, 0, 0, 0).to_identifier();
+		let block = build_block(&chain, &keychain, &key_id, vec![]);
+		chain.process_block(block, Options::MINE).unwrap();
+	}
+
+	assert_eq!(chain.head().unwrap().height, 8);
```
tromp, line 81: why is height not 9 ?

antiochp: Maybe I misunderstand your question?

We want to mine 8 blocks to set the test up. Which we do in the loop above -

```
for n in 1..9 {
```

Then we mine the block at height 9 and check it is valid based on presence of NRD kernel.

tromp: D'oh! I thought the loop went over 1..9 inclusive
In Haskell, the list [1..9] is [1,2,3,4,5,6,7,8,9]
Ok, So in Rust the upper bound is exclusive...

core/src/core/block.rs
```diff
@@ -63,6 +63,10 @@ pub enum Error {
 	InvalidPow,
 	/// Kernel not valid due to lock_height exceeding block header height
 	KernelLockHeight(u64),
+	/// NRD kernels are not valid prior to HF3.
+	NRDKernelPreHF3,
+	/// NRD kernels are not valid if disabled locally via "feature flag".
+	NRDKernelNotEnabled,
```
tromp, line 69: why not flip value and call it NRDKernelEnabled ?
or if you really want to keep same value, use NRDKernelDisabled (I prefer above though).

core/src/core/transaction.rs
```diff
@@ -97,14 +176,28 @@ impl KernelFeatures {
...
+			KernelFeatures::HeightLocked { fee, lock_height } => {
+				writer.write_u64(*fee)?;
+				writer.write_u64(*lock_height)?;
+			}
+			KernelFeatures::NoRecentDuplicate {
+				fee,
+				relative_height,
+			} => {
+				writer.write_u64(*fee)?;
+				writer.write_empty_bytes(6)?;
```
tromp, line 197: is the height written in big endian? if so, should add a comment to that effect.

core/src/core/transaction.rs
```diff
@@ -176,6 +285,19 @@ impl KernelFeatures {
 				let lock_height = reader.read_u64()?;
 				KernelFeatures::HeightLocked { fee, lock_height }
 			}
+			KernelFeatures::NO_RECENT_DUPLICATE_U8 => {
+				// NRD kernels are invalid if NRD feature flag is not enabled.
+				if !global::is_nrd_enabled() {
+					return Err(ser::Error::CorruptedData);
```
tromp, line 291: what happened with error NRDKernelNotEnabled ?

antiochp:
@quentinlesceller 
> 1 - I don't see any reference to the V1 Kernel serialisation in details in the RFC. There is indeed a `"V1 "fixed size kernels"` part but this does not details the serialisation of NRD kernel. I don't know if that's important or not.

Good point. I'll rework the "V1 fixed size kernels" to make NRD more explicit. The "extra data" is used for the relative height (and we only use some of the bytes, hence the empty bytes).

core/src/core/transaction.rs
antiochp: Yes. This is 6 bytes of 00, followed by 2 bytes representing the u16 relative height.
We were originally using a u64 to serialize the u16 relative height which felt too implicit.
Now that we are more explicit with the bytes we should explain why we are doing what we are doing. :+1:

core/src/core/transaction.rs
```diff
@@ -97,14 +176,28 @@ impl KernelFeatures {
...
+				writer.write_u64(*fee)?;
+				writer.write_u64(*lock_height)?;
+			}
+			KernelFeatures::NoRecentDuplicate {
+				fee,
+				relative_height,
+			} => {
+				writer.write_u64(*fee)?;
+				writer.write_empty_bytes(6)?;
+				relative_height.write(writer)?;
```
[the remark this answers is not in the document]
antiochp: Yes this definitely matters. See my response to @tromp comment above.
There are 8 bytes of "extra data" that we use for the full u64 absolute lock_height.
In NRD kernels we only need 2 bytes for the u16 relative_height and we want to explicitly use the last 2 bytes of these.
This is technically the same as a u64 as the left most bytes are 0x00 (bigendian).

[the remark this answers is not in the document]
antiochp: We should add a section in the "V1 fixed size kernels" section of the RFC to make this clear.
Its bigendian so 6 bytes will always be 0x00 followed by 2 bytes representing the relative_height.

core/src/core/transaction.rs
```diff
@@ -113,48 +206,64 @@ impl KernelFeatures {
...
+			KernelFeatures::NO_RECENT_DUPLICATE_U8 => {
+				// NRD kernels are invalid if NRD feature flag is not enabled.
+				if !global::is_nrd_enabled() {
 					return Err(ser::Error::CorruptedData);
 				}
-				if lock_height != 0 {
-					return Err(ser::Error::CorruptedData);
+
+				let fee = reader.read_u64()?;
+				reader.read_empty_bytes(6)?;
```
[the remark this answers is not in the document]
antiochp: This is identical to reading the full 8 bytes into a u64 and then casting it to a u16.
This is what i was originally doing but it feels too implicit like this.

core/src/core/transaction.rs
antiochp: We don't have a single global set of errors.

This code needs to return a serialization error, and we use the pattern of returning a relatively generic `CorruptedData` for serialization errors like this.
`NRDKernelNotEnabled` is a block pipeline error and cannot be used here.

core/src/core/transaction.rs
```diff
@@ -1613,6 +1771,23 @@ mod test {
...
 		};
 
+		// Test explicit protocol version.
+		for version in vec![ProtocolVersion(1), ProtocolVersion(2)] {
+			let mut vec = vec![];
+			ser::serialize(&mut vec, version, &kernel).expect("serialized failed");
+			let kernel2: TxKernel = ser::deserialize(&mut &vec[..], version).unwrap();
+			assert_eq!(
+				kernel2.features,
+				KernelFeatures::HeightLocked {
```
tromp, line 1781: can't we write code to allow direct kernel comparison as in assert_eq!(kernel, kernel2); ?
Or failing that, can't we write assert_eq!(kernel.features, kernel2.features); ?

antiochp: Yes we can do the following - 

    assert_eq!(kernel.features, kernel2.features);

core/src/core/transaction.rs
```diff
@@ -1700,8 +1985,36 @@ mod test {
...
+				fee: 10,
+				relative_height: NRDRelativeHeight(100)
+			}
+		);
+
 		let mut vec = vec![];
-		ser::serialize_default(&mut vec, &(3u8, 0u64, 0u64)).expect("serialized failed");
+		ser::serialize_default(&mut vec, &(4u8)).expect("serialized failed");
 		let res: Result<KernelFeatures, _> = ser::deserialize_default(&mut &vec[..]);
 		assert_eq!(res.err(), Some(ser::Error::CorruptedData));
```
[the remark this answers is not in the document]
antiochp: :+1: I think I got bored of writing tests by this point. 😔 I'll add these for completeness.

core/src/core/transaction.rs
```diff
@@ -76,18 +149,24 @@ impl KernelFeatures {
...
 		}
 	}
 
-	/// msg = hash(features)                       for coinbase kernels
-	///       hash(features || fee)                for plain kernels
-	///       hash(features || fee || lock_height) for height locked kernels
+	/// msg = hash(features)                           for coinbase kernels
+	///       hash(features || fee)                    for plain kernels
+	///       hash(features || fee || lock_height)     for height locked kernels
+	///       hash(features || fee || relative_height) for height locked kernels
```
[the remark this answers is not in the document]
antiochp: Fixed.

chain/tests/mine_nrd_kernel.rs
```diff
@@ -0,0 +1,105 @@
...
+	)
+	.unwrap();
+
+	let key_id9 = ExtKeychainPath::new(1, 9, 0, 0, 0).to_identifier();
+	let block = build_block(&chain, &keychain, &key_id9, vec![tx]);
+	chain.process_block(block, Options::MINE).unwrap();
+	chain.validate(false).unwrap();
+
+	clean_output_dir(chain_dir);
+}
```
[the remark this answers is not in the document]
antiochp: :+1: on adding the test.

`chain.process_block()` actually returns an `Error::Other("...")` with the `InvalidBlockProof` backtrace serialized as the string in the Other error...

This is something I think we should investigate and revisit. Not just for NRD but for `process_block()` in general. But this is separate to this PR.

core/src/core/transaction.rs
```diff
@@ -36,6 +36,70 @@ use util::static_secp_instance;
...
+			> NRDRelativeHeight::MAX
+				.try_into()
+				.expect("WEEK_HEIGHT const should fit in u16")
+		{
+			Err(Error::InvalidNRDRelativeHeight)
+		} else {
+			Ok(Self(height))
+		}
+	}
+}
```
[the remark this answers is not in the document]
antiochp: I actually prefer keeping the two height checks in separate branches of the if/else statement for clarity. I find it too easy to miss the `||` in there.

antiochp:
> 2 - So NRD kernels are disabled by default? And the goal is to enabled them in a subsequent release?

Yes. I want the code in there and enabled on floonet, but disabled for now on mainnet.
This will make it a lot easier to test out and build on in floonet post HF3.

And we are not ready for this to go live on mainnet yet.

core/src/core/transaction.rs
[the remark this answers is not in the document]
quentinlesceller: Thanks for the explanation that makes sense!

PR #3305, opened 2020-04-20
Less cloning and pattern simplifications

quentinlesceller:
Remove a lot of cloning and weird formatting stuff like `format!(x).to_string()`.

### DIFF
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -183,8 +183,8 @@ impl<'de> Visitor<'de> for PeerAddrs {
 				Err(_) => {
 					let socket_addrs = entry
 						.to_socket_addrs()
-						.expect(format!("Unable to resolve DNS: {}", entry).as_str());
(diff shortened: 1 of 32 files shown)


p2p/src/types.rs
```diff
@@ -183,8 +183,8 @@ impl<'de> Visitor<'de> for PeerAddrs {
 				Err(_) => {
 					let socket_addrs = entry
 						.to_socket_addrs()
-						.expect(format!("Unable to resolve DNS: {}", entry).as_str());
-					peers.append(&mut socket_addrs.map(|addr| PeerAddr(addr)).collect());
+						.unwrap_or_else(|_| panic!("Unable to resolve DNS: {}", entry));
```
hashmap, line 186: can we get rid of expect/panic somehow?

quentinlesceller: I agree that this error must be handled. Will do in a subsequent PR.

Issue #3292, opened 2020-04-10
Loosing peers and troubles syncing. 


johanneshahn:
We found a bug in code where peers are banned for no reason.
This happens if you start syncing from zero or had a longer pause. After a while peers get banned for no reason.

The error is in header_sync.rs -> *stalling_ts
Which does not recognize by peer and ban any peer for no reason.
You can track this error with a small debug info in code.

```
let test = stalling_ts + Duration::seconds(120);
if let Some(ref peer) = self.syncing_peer {
	debug!(
		"sync_state: in ban loop {}, now {}, stuck max {}",
		peer.info.addr,
		now,
		test
	);
}
```
... stalling_ts never updated correctly

We track this issue for a while on epic cash too.

Here is the bugfix ->
https://gitlab.com/epiccash/epic/-/merge_requests/78/diffs#diff-content-725373f81536aa7f86f634ae14cfcb9b7878e709

greetz from epic dev Team.

**Describe the bug**

**To Reproduce**
1. Run '...'
2. Expect '....'

**Relevant Information**
An excerpt from the logs or a detailed error message.

**Screenshots**

**Desktop (please complete the following information):**
 - Version [e.g. 1.0.1]

**Additional context**

antiochp:
Hi thanks for bringing this to our attention. We will investigate.

> stalling_ts never updated correctly

Is this actually true though?

We (at least in theory) reset `stalling_ts` after receiving all the expected headers (can't vouch for this actually working as intended though) - 

```
if all_headers_received {
	// reset the stalling start time if syncing goes well
	self.stalling_ts = None;
} else {
```

johanneshahn:
ok lets explain why code does not work.

to get into the ban loop the peer is 2 seconds behind height.
This happens very often. 2 Seconds are to low.

```
if header_head.height > latest_height {
  self.prev_header_sync =
  (now + Duration::seconds(**2**), header_head.height, prev_height);

}
```


... so after this happens peer go into if statement
and the stalling time is set once and never reset only if we receive all headers from this peer and stalling is false.

```
if force_sync || all_headers_received || stalling {

   if stalling {
	if self.stalling_ts.is_none() { <-- stalling could never be update by this line.
		self.stalling_ts = Some(now);
	}
  } else {
	self.stalling_ts = None;
  }


...
} 
```
and why we often go into the if statement because the stalling time of 2 seconds is very low.
[...]

PR #3306, opened 2020-04-21
Fix fraud peer detection, introduce more agressive ban

hashmap:
Currently we ban a current syncing peer if we didn't get headers in 120s, however we send a new headers request every 10s and select a new syncing peer. That last peer is no more guilty than previous ones. Moreover, this situation is very unlikely and happens when we have some network issue on our side.

Usually we have one unresponsive peer which we never ban, so we just spent more time on syncing. NAT'd peer has up to 8 peers, so keeping one unresponsive peer affects header sync's performance sufficiently.

This PR removes 120s grace period and bans unresponsive peers after 10s timeout. I made a few full syncs and found that my node bans the same 3 peers, so this approach seems to be working well.

### DIFF
--- a/servers/src/grin/sync/header_sync.rs
+++ b/servers/src/grin/sync/header_sync.rs
@@ -30,7 +30,6 @@ pub struct HeaderSync {
 	prev_header_sync: (DateTime<Utc>, u64, u64),
 
 	syncing_peer: Option<Arc<Peer>>,
-	stalling_ts: Option<DateTime<Utc>>,
 }
 
 impl HeaderSync {
@@ -46,7 +45,6 @@ impl HeaderSync {
 			history_locator: vec![],
 			prev_header_sync: (Utc::now(), 0, 0),
 			syncing_peer: None,
-			stalling_ts: None,
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


antiochp:
Related - https://github.com/mimblewimble/grin/issues/3292

antiochp:
> Currently we ban a current syncing peer if we didn't get headers in 120s, however we send a new headers request every 10s and select a new syncing peer. That last peer is no more guilty than previous ones.

Oh is that what is happening here? We randomly choose a new peer every 10s and then eventually ban an arbitrary peer based on it being unlucky at the 120s threshold?

I'm guessing the intent was to eventually the ban the one non-responsive peer but now I'm not even sure this was fully thought through...

I need to go over the code again but yeah I think this potentially makes sense to simplify like this.

DavidBurkett:
> I'm guessing the intent was to eventually the ban the one non-responsive peer but now I'm not even sure this was fully thought through...

@garyyu wrote this IIRC, so he could provide more insight if he has anything to add. The original intention though was to just prevent peers from advertising a height they didn't have a block for, and therefore forcing nodes into header sync mode from which they never recover.

Igno always strongly cautioned against banning slow peers, which is probably why the original code didn't just ban everytime the peer failed to provide a requested header right away. I believe his concerns have been shown to be unwarranted, since the network has done fine for slower peers despite grin++'s ruthless banning policies.

antiochp:
Dropped the ball on getting this in for 4.0.0.
Lets pick this up again once we get through HF3 and get this merged.

bladedoyle:
I tested this change with a "slow disk" node (see https://github.com/mimblewimble/grin/issues/3561#issuecomment-779413163)

The issue of banning (healthy) peers at random is not improved. 
It seems like incorrect logic to ban a specific peer (whichever peer happens to be currently selected as syncing_peer) based on the global sync state.
Also, hardcoded 10 seconds may not be good for all situations.....it seems very short for nodes running on slow systems.

antiochp:
@bladedoyle thanks for testing this out locally with a slow disk. :+1:

yeastplume:

PR #3309, opened 2020-04-27
Refactor Readable trait

hashmap:
Currently `Writable` accepts trait `Write` as a type parameter but `Readable`
takes `Read` as a trait object, which is not symmetrical and also less performant. This PR changes Readable trait and all places where it's used

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -146,21 +146,21 @@ impl Msg {
 ///
 /// Note: We return a MsgHeaderWrapper here as we may encounter an unknown msg type.
 ///
-pub fn read_header(
-	stream: &mut dyn Read,
+pub fn read_header<R: Read>(
+	stream: &mut R,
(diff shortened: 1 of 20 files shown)


p2p/src/msg.rs
```diff
@@ -146,21 +146,21 @@ impl Msg {
...
-pub fn read_header(
-	stream: &mut dyn Read,
+pub fn read_header<R: Read>(
+	stream: &mut R,
 	version: ProtocolVersion,
 ) -> Result<MsgHeaderWrapper, Error> {
 	let mut head = vec![0u8; MsgHeader::LEN];
 	stream.read_exact(&mut head)?;
-	let header = ser::deserialize::<MsgHeaderWrapper>(&mut &head[..], version)?;
+	let header: MsgHeaderWrapper = ser::deserialize(&mut &head[..], version)?;
```
antiochp, line 155: :+1:

PR #3311, opened 2020-04-27
Reduce number of allocations in to_key calls

hashmap:
We have to make an extra allocation per db get request because key generation function to_key takes `Vec`. Taking byte slice (`AsRef<[u8]>` to be precise) also simplifies the code a bit.

### DIFF
--- a/p2p/src/store.rs
+++ b/p2p/src/store.rs
@@ -154,7 +154,7 @@ impl PeerStore {
 	) -> Result<Vec<PeerData>, Error> {
 		let mut peers = self
 			.db
-			.iter::<PeerData>(&to_key(PEER_PREFIX, &mut "".to_string().into_bytes()))?
+			.iter::<PeerData>(&to_key(PEER_PREFIX, ""))?
 			.map(|(_, v)| v)
 			.filter(|p| p.flags == state && p.capabilities.contains(cap))
 			.collect::<Vec<_>>();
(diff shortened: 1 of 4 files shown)


store/src/lib.rs
```diff
@@ -44,20 +44,22 @@ use byteorder::{BigEndian, WriteBytesExt};
...
 
 /// Build a db key from a prefix and a byte vector identifier.
-pub fn to_key(prefix: u8, k: &mut Vec<u8>) -> Vec<u8> {
+pub fn to_key<K: AsRef<[u8]>>(prefix: u8, k: K) -> Vec<u8> {
+	let k = k.as_ref();
 	let mut res = Vec::with_capacity(k.len() + 2);
 	res.push(prefix);
 	res.push(SEP);
-	res.append(k);
+	res.extend_from_slice(k);
```
quentinlesceller, line 52: Okay, interesting. So instead of emptying `k` we clone and append the slice. So we don't need a mutable reference 👍

p2p/src/store.rs
```diff
@@ -154,7 +154,7 @@ impl PeerStore {
 	) -> Result<Vec<PeerData>, Error> {
 		let mut peers = self
 			.db
-			.iter::<PeerData>(&to_key(PEER_PREFIX, &mut "".to_string().into_bytes()))?
+			.iter::<PeerData>(&to_key(PEER_PREFIX, ""))?
```
quentinlesceller, line 157: 👍

store/src/lib.rs
hashmap: I think in general it's preferable to use `extend` over `append`, because the later consumes the param. `extend_from_slice` is just an optimized version which may be removed when Rust gets specialization.

PR #3313, opened 2020-04-29
Remove some unwraps in LMDB client

hashmap:


store/src/lmdb.rs
```diff
@@ -248,19 +247,22 @@ impl Store {
...
-		self.get_ser_access(key, &access, db)
+		self.get_ser_access(key, &access, db.clone())
 	}
 
 	fn get_ser_access<T: ser::Readable>(
 		&self,
 		key: &[u8],
 		access: &lmdb::ConstAccessor<'_>,
-		db: RwLockReadGuard<'_, Option<Arc<lmdb::Database<'static>>>>,
+		db: Arc<lmdb::Database<'static>>,
```
antiochp, line 263: I like this. :+1: Still reviewing though, need to understand what we ware changing here.

antiochp:
Testing locally and looks good so far.

PR #3314, opened 2020-04-30
add test to demonstrate pair of "half" kernels sharing same public excess

antiochp:
* Added new `build_two_half_kernels` test
* Cleanup test transaction building code (so we can pass explicit excess in)

The test code used to simply generate a random offset but we can now force it to use the appropriate offset to compensate for a specific excess value during tx building and kernel signing.

We intend to do something similar to this for NRD kernels once we have support for multi-party outputs. In real-world usage both participants would provide the necessary offset to adjust the final public excess back to a value shared across two transactions.

### DIFF
--- a/core/src/libtx/build.rs
+++ b/core/src/libtx/build.rs
@@ -199,6 +199,8 @@ where
 }
 
 /// Builds a complete transaction.
+/// NOTE: We only use this in tests (for convenience).
+/// In the real world we use signature aggregation across multiple participants.
 pub fn transaction<K, B>(
 	features: KernelFeatures,
 	elems: Vec<Box<Append<K, B>>>,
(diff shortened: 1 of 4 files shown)


keychain/src/types.rs
```diff
@@ -281,22 +254,45 @@ impl BlindingFactor {
...
+			Ok(ZERO_KEY)
 		} else {
-			secp::key::SecretKey::from_slice(secp, &self.0).map_err(Error::Secp)
+			SecretKey::from_slice(secp, &self.0).map_err(Error::Secp)
+		}
+	}
+
+	// Convenient (and robust) way to add two blinding_factors together.
+	// Handles "zero" blinding_factors correctly.
+	pub fn add(&self, other: &BlindingFactor, secp: &Secp256k1) -> Result<BlindingFactor, Error> {
```
quentinlesceller, line 285: Out of curiosity why not keep the `Add` impl ? Also curious if grin wallet uses this somewhere.

pool/src/pool.rs
```diff
@@ -268,7 +269,12 @@ impl Pool {
 		header: &BlockHeader,
 	) -> Result<BlockSums, PoolError> {
 		let overage = tx.overage();
-		let offset = (header.total_kernel_offset() + tx.offset.clone())?;
+
+		let offset = {
+			let secp = static_secp_instance();
```
quentinlesceller, line 274: So it is my understanding (also related to previous comment), that you want to be able to choose the secp instance ?

keychain/src/types.rs
```diff
@@ -306,26 +302,19 @@ impl BlindingFactor {
...
-		// use blind_sum to subtract skey_1 from our key (to give k = k1 + k2)
+	pub fn split(
+		&self,
+		blind_1: &BlindingFactor,
+		secp: &Secp256k1,
+	) -> Result<BlindingFactor, Error> {
+		// use blind_sum to subtract skey_1 from our key such that skey = skey_1 + skey_2
 		let skey = self.secret_key(secp)?;
+		let skey_1 = blind_1.secret_key(secp)?;
 		let skey_2 = secp.blind_sum(vec![skey], vec![skey_1.clone()])?;
```
quentinlesceller, line 313: Cloning can be removed here.

keychain/src/types.rs
```diff
@@ -281,22 +254,45 @@ impl BlindingFactor {
...
+		BlindingFactor::from_secret_key(ZERO_KEY)
+	}
+
+	pub fn is_zero(&self) -> bool {
+		self.0 == ZERO_KEY.as_ref()
+	}
+
+	pub fn rand(secp: &Secp256k1) -> BlindingFactor {
+		BlindingFactor::from_secret_key(SecretKey::new(secp, &mut thread_rng()))
 	}
```
quentinlesceller, line 266: 👍

core/src/libtx/build.rs
```diff
@@ -199,47 +199,58 @@ where
 }
 
 /// Builds a complete transaction.
+/// NOTE: We only use this in tests (for convenience).
+/// In the real world we use signature aggregation across multiple participants.
 pub fn transaction<K, B>(
```
quentinlesceller, line 204: This function isn't used anywhere yet, right?

antiochp: No this is used extensively in our tests. Search for `build::transaction`.
It is also used in some tests in grin-wallet, but I think just legacy from when the repos were originally split (those tests should probably live in grin itself).

antiochp: The new part of this is where we call `build::transaction_with_kernel()` internally from `build::transaction()` allowing for more flexibility.

keychain/src/types.rs
antiochp: `Add` was really clunky as it consumed the two things being added, but `BlindingFactor` is `Clone` but not `Copy` so we needed to explicitly `clone()` when doing `A + B` (which is what `Add` permits you to do, it overrides `+`).
And we only did this in a single place (in pool) so it seemed easier to just have our own `add()` fn and call this.

antiochp: The `Add` trait also meant we had to jump through some hoops with `static_secp_instance` as there was no way to pass an instance in (see below).

pool/src/pool.rs
antiochp: We want to pass an secp instance in, just to keep the api consistent with the rest of the `BlindingFactor` fns. We don't _need_ to do this but I think it makes it cleaner.

antiochp: This impl detail was hidden behind the `Add` trait via the `A + B` syntax and I think this hid too much (in terms of needing to obtain an secp instance).

quentinlesceller: I see thanks for the clarification. 👍

PR #3315, opened 2020-05-01
Remove unnecessary newline in Stratum message

quentinlesceller:
Found out here https://github.com/mimblewimble/grin/issues/1215.
This line of code send an empty message. We don't the need the newline here.
Tested and this fix the issue.


quentinlesceller:
Can confirm this is causing some error on NiceHash too.

PR #3316, opened 2020-05-03
cleanup weird edge_bits handling in tests

antiochp:
Looks like some legacy refactoring left some conditional stuff in place that no longer makes any sense.

This PR simply cleans this up. All chain tests still pass.

### DIFF
--- a/chain/tests/chain_test_helper.rs
+++ b/chain/tests/chain_test_helper.rs
@@ -97,11 +97,7 @@ where
 
 		chain.set_txhashset_roots(&mut b).unwrap();
 
-		let edge_bits = if n == 2 {
-			global::min_edge_bits() + 1
-		} else {
-			global::min_edge_bits()
-		};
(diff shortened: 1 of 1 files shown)


chain/tests/chain_test_helper.rs
```diff
@@ -110,7 +106,6 @@ where
 			edge_bits,
 		)
 		.unwrap();
-		b.header.pow.proof.edge_bits = edge_bits;
```
antiochp, line 113: This is redundant as we already set this earlier.

chain/tests/chain_test_helper.rs
```diff
@@ -97,11 +97,7 @@ where
 
 		chain.set_txhashset_roots(&mut b).unwrap();
 
-		let edge_bits = if n == 2 {
```
antiochp, line 100: `n` here is block height and not sure why we would need special handling for height 2.

PR #3318, opened 2020-05-04
support explicit filtering of txs when preparing to mine

antiochp:
This PR adds functionality to explicitly filter txs (by bucket) when preparing a set of candidate transaction to mine a block.

We now explicitly filter by kernel variant, allowing buckets containing transactions with the following feature variants (and only these feature variants) - 

* Plain
* HeightLocked

HF3 will introduce the following new kernel variant (unsupported prior to HF3) - 

* NoRecentDuplicate

When we introduce the proposed NoRecentDuplicate kernel variant in HF3 we will use this new filtering logic to explicitly start allowing txs containing NoRecentDuplicate kernels to be included in a candidate block to be mined.

TODO - 

Later buckets can be dependent on earlier buckets due to fee calcs.

- [x] So we need to hook into the "reject" logic earlier to filter out invalid txs and their dependent txs.

Actually `tx_filter` closure can be used on the way in to the txpool also. The filtering logic is the same.

- [ ] Use `tx_filter` on way in to txpool/stempool also.

### DIFF
--- a/pool/src/pool.rs
+++ b/pool/src/pool.rs
@@ -339,21 +339,26 @@ where
 			// if single parent then we are good, we can bucket it with its parent
 			// if multiple parents then we need to combine buckets, but for now simply reject it (rare case)
 			let mut insert_pos = None;
-			let mut is_rejected = false;
 
-			for input in entry.tx.inputs() {
-				if rejected.contains(&input.commitment()) {
-					// Depends on a rejected tx, so reject this one.
-					is_rejected = true;
-					continue;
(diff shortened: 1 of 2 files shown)


antiochp:
Jumping through a lot of hoops here to resolve a theoretical edge case where the txpool is in invalid state (contains NRD kernel prior to HF3 due to rewind).

When "preparing a block to mine" we verify the block will indeed be valid and in the unlikely case it is invalid we already fallback to mining an empty block.
So all this hoop jumping is not actually necessary.

Closing.

PR #3320, opened 2020-05-08
Use multiple LMDB dbs (tables)

hashmap:
LMDB is a key-value DB, so each operation uses a key. We currently always allocate a key in the heap, which leads to millions of allocations which could be avoided.

We use one table per DB (there are 2 dbs - `chain` and `peers`), which is quite unusual, so we have to use prefixes to emulate table functionality. For example we store a few different types of entities in `chain` db - blocks, headers etc. To store or retrieve a block we call a function `to_key` which generates a key for a particular block `b/block-hash`. LMDB expects keys as byte slices, so we could use a reference to block's hash to store or retrieve it, but we need to allocate a `Vec` to add the prefix.

In our defense LMDB terms are confusing, db is called `environment`, table is called `database`, so we have 1 env and 1 db per a domain (chain, p2p). What we really want is 1 env + many db (tables), so in this case we can use already allocated data (block/header hash, commit etc) directly as keys.

This PR introduces multi db (aka multi tables) support to sufficiently reduce number of allocations and simplify the code a bit. This is also saner design, each entity type should have its own table.

This is a WIP, I susppose a bit unstable version, supports only sync from the scratch. Each singleton (head, tail etc) is saved into its own table, which is overkill, it should use one table for all such values. We also need to implement db migration and solve an UX problem - what options a user should have after migration is done - eg remove the old db, keep it for a potential downgrade, move it to backup etc.

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -29,18 +29,34 @@ use std::sync::Arc;
 
 const STORE_SUBPATH: &str = "chain";
 
-pub const SINGLE_KEY: &[u8] = b"K";
+/// Tables (dbs) names
+pub mod table {
+	/// Headers db name
+	pub const BLOCK_HEADERS: &str = "headers";
+	/// Blocks db name
+	pub const BLOCKS: &str = "blocks";
+	/// Db name for singleton values (like head, tail etc)
+	pub const VALUES: &str = "values";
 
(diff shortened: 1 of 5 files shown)


quentinlesceller:
I'm going to close this because of inactivity. However this would be nice to have so I opened #3461 to track it.

PR #3325, opened 2020-05-14
Some optimizsations in TUI

hashmap:
* Do not allocate String for static status, use Cow instead, for example currently we allocate "Running" string at every refresh on a synced node
* Use static dispatch for Views instead of Box<dyn View>
* Update only the current view
* Simplify Listener trait

### DIFF
--- a/src/bin/tui/status.rs
+++ b/src/bin/tui/status.rs
@@ -20,6 +20,7 @@ use cursive::traits::Identifiable;
 use cursive::view::View;
 use cursive::views::{LinearLayout, ResizedView, TextView};
 use cursive::Cursive;
+use std::borrow::Cow;
 
 use crate::tui::constants::VIEW_BASIC_STATUS;
 use crate::tui::types::TUIStatusListener;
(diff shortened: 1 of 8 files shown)


src/bin/tui/status.rs
```diff
@@ -20,6 +20,7 @@ use cursive::traits::Identifiable;
 use cursive::view::View;
 use cursive::views::{LinearLayout, ResizedView, TextView};
 use cursive::Cursive;
+use std::borrow::Cow;
```
antiochp, line 23: Ooh! First use of `Cow` in grin.

antiochp: I do not really understand what a `Cow` is.

hashmap: Cursive accepts smth like AsRef<str> (don’t remember exactly), but we need to return some type, we went with String because in some cases we generate a status. But in some cases we return a static string, so what if we could avoid an extra allocation? Cow does exactly that - it keeps an owned T or borrowed T (until we mutate it, then it switches to the  owned variant, doing copy on write or cow). In our case we don’t mutate the content and lifetime for borrowed str is static.

PR #3327, opened 2020-05-15
thread local chain type vs global chain type

antiochp:
Our global `CHAIN_TYPE` config has been driving me crazy for a while now.
We have tests that do not reliably run in parallel in separate threads due to the global `CHAIN_TYPE` being overwritten by other tests. Some tests require `Mainnet` as we test specific PoW params etc.

This PR breaks out the following - 
* GLOBAL_CHAIN_TYPE 
  * set once and _only once_ at node startup based on config
  * __not used in tests__
* thread local CHAIN_TYPE 
  * "inherits" GLOBAL_CHAIN_TYPE in threads via thread local
  * must be explicitly set on a per-test basis

Reads via `get_chain_type()` always go via the thread local `CHAIN_TYPE`. 
This will be lazily initialized based on `GLOBAL_CHAIN_TYPE` is not explicitly set.

* introduced `init_global_chain_type()` to be called (once) during node startup
  * this uses our `OneTime` util internally
* `get_chain_type()` reads the thread local, lazily inits it from global chain type
* renamed `set_mining_mode()` to `set_local_chain_type()`
  * tests are free to call `set_local_chain_type()` within the context of a single test thread
  * tests must explicitly call this and we panic with a useful msg if not set
* cleaned up `POW_CONTEXT_TYPE` and `PowContextType` (seemingly unused)

----

TODO

- [x] corresponding changes in grin-wallet (`set_mining_mode()` called in various tests)

Related: https://github.com/mimblewimble/grin-wallet/pull/409

### DIFF
--- a/core/src/global.rs
+++ b/core/src/global.rs
@@ -27,7 +27,8 @@ use crate::pow::{
 	self, new_cuckaroo_ctx, new_cuckarood_ctx, new_cuckaroom_ctx, new_cuckatoo_ctx, EdgeType,
 	PoWContext,
 };
-use util::RwLock;
+use std::cell::Cell;
+use util::OneTime;
 
 /// An enum collecting sets of parameters used throughout the
 /// code wherever mining is needed. This should allow for
@@ -104,7 +105,7 @@ pub const TXHASHSET_ARCHIVE_INTERVAL: u64 = 12 * 60;
 
 /// Types of chain a server can run with, dictates the genesis block and
 /// and mining parameters used.
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, PartialEq)]
 pub enum ChainTypes {
 	/// For CI testing
 	AutomatedTesting,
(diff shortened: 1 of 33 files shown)


core/src/global.rs
```diff
@@ -134,31 +135,43 @@ impl Default for ChainTypes {
 	}
 }
 
-/// PoW test mining and verifier context
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
-pub enum PoWContextTypes {
```
quentinlesceller, line 139: Good catch 👍

core/src/global.rs
```diff
@@ -134,31 +135,43 @@ impl Default for ChainTypes {
...
+	/// This is accessed via get_chain_type() which allows the global value
+	/// to be overridden on a per-thread basis (for testing).
+	pub static ref GLOBAL_CHAIN_TYPE: OneTime<ChainTypes> = OneTime::new();
 }
 
-lazy_static! {
-	/// The mining parameter mode
-	pub static ref CHAIN_TYPE: RwLock<ChainTypes> =
-			RwLock::new(ChainTypes::Mainnet);
+thread_local! {
```
quentinlesceller, line 145: nice didn't know about that macro.

core/tests/block.rs
```diff
@@ -34,13 +34,18 @@ use keychain::{BlindingFactor, ExtKeychain, Keychain};
...
+}
+
 fn verifier_cache() -> Arc<RwLock<dyn VerifierCache>> {
 	Arc::new(RwLock::new(LruVerifierCache::new()))
 }
 
 #[test]
 fn too_large_block() {
-	global::set_mining_mode(ChainTypes::AutomatedTesting);
+	global::set_local_chain_type(ChainTypes::AutomatedTesting);
```
quentinlesceller, line 48: `test_setup()`or `global::set_local_chain_type(global::ChainTypes::AutomatedTesting);`?

Overall `test_setup()` don't seems necessary no?

core/tests/block.rs
```diff
@@ -301,7 +313,7 @@ fn empty_compact_block_serialized_size() {
 
 #[test]
 fn compact_block_single_tx_serialized_size() {
-	global::set_mining_mode(ChainTypes::AutomatedTesting);
+	global::set_local_chain_type(ChainTypes::AutomatedTesting);
```
quentinlesceller, line 316: Same comment as above.

core/tests/block.rs
```diff
@@ -287,7 +299,7 @@ fn block_single_tx_serialized_size() {
 
 #[test]
 fn empty_compact_block_serialized_size() {
-	global::set_mining_mode(ChainTypes::AutomatedTesting);
+	global::set_local_chain_type(ChainTypes::AutomatedTesting);
```
quentinlesceller, line 302: Same comment as above.

pool/src/types.rs
```diff
@@ -144,7 +144,7 @@ fn default_max_stempool_size() -> usize {
 	50_000
 }
 fn default_mineable_max_weight() -> usize {
-	global::max_block_weight()
+	consensus::MAX_BLOCK_WEIGHT
```
quentinlesceller, line 147: Why?

core/src/global.rs
```diff
@@ -134,31 +135,43 @@ impl Default for ChainTypes {
...
-	/// PoW context type to instantiate
-	pub static ref POW_CONTEXT_TYPE: RwLock<PoWContextTypes> =
-			RwLock::new(PoWContextTypes::Cuckoo);
+/// Get the chain type via thread_local, fallback to global chain_type.
+pub fn get_chain_type() -> ChainTypes {
+	CHAIN_TYPE.with(|chain_type| match chain_type.get() {
+		None => {
+			if GLOBAL_CHAIN_TYPE.is_init() {
+				let chain_type = GLOBAL_CHAIN_TYPE.borrow();
+				set_local_chain_type(chain_type);
```
quentinlesceller, line 161: Out of curiosity, why do we need to set a per thread value here? Wouldn't returning `GLOBAL_CHAIN_TYPE.borrow()` enough?

pool/src/types.rs
antiochp: Ah good catch. This was necessary because max_block_weight() relies on `chain_type` internally. So we have a default value that cannot be used until we know (and have configured) the chain_type itself.
The default config needs to be genuine defaults, without any dependencies on other, possibly external, configuration.

core/src/global.rs
antiochp: Yeah we technically don't need to use the thread local in the case where the global is set.
But I think its better for -
* consistency (tests share the same code path for more code)
* _possibly_ performance as once the value is "cached" in the thread we no longer need to acquire a read lock on the global `RwLock` when reading this.

core/tests/block.rs
antiochp: Good catch. I just want to use `test_setup()` to avoid all the repeated calls to `set_local_chain_type()` as this should be common across all the tests.

core/src/global.rs
quentinlesceller: Makes sense 👍

PR #3330, opened 2020-05-19
Fixing get_unspent_outputs documentation

DavidBurkett:
params[1] is the (optional) end_index. params[2] is the (required) max number of outputs to return. The documentation mixed up those 2.


DavidBurkett:
It was `[1, 2, null, true]` which corresponds to: `[start_index: 1, end_index: 2, max: null, include_proof: true]`

But I'm changing it to `[1, null, 2, true]` which is: `[start_index: 1, end_index: null, max: 2, include_proof: true]`

PR #3334, opened 2020-05-26
HardFork3

tromp:
name: HardFork3
about: This hard-fork allows blocks with 3\*HARD_FORK_INTERVAL/FLOONET_THIRD_HARD_FORK <= height < 4\*HARD_FORK_INTERVAL with header version 4 and Cuckarooz29 as secondary PoW.
title: HardFork3
labels: consensus breaking

Currently, consensus::valid_header_version(height, version) returns false for all height >= 3\*HARD_FORK_INTERVAL.
This PR changes that to return true for version 4 at either FLOONET_THIRD_HARD_FORK <= height < 4\*HARD_FORK_INTERVAL on floonet or 3\*HARD_FORK_INTERVAL <= height < 4\*HARD_FORK_INTERVAL on mainnet.

This PR also changes global::create_pow_context to return new_cuckarooz_ctx(edge_bits, proof_size) when consensus::valid_header_version(height, version) holds for version 4, which means that Cuckarooz29 [1] is the only valid secondary PoW after the hard fork.

Successfully tested on a mainnet fast sync.

[1] https://forum.grin.mw/t/introducing-the-final-tweak-cuckarooz

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -133,11 +133,17 @@ pub const FLOONET_FIRST_HARD_FORK: u64 = 185_040;
 /// Floonet second hard fork height, set to happen around 2019-12-19
 pub const FLOONET_SECOND_HARD_FORK: u64 = 298_080;
 
+/// Floonet second hard fork height, set to happen around 2020-06-20
+pub const FLOONET_THIRD_HARD_FORK: u64 = 552_960;
+
 /// AutomatedTesting and UserTesting first hard fork height.
 pub const TESTING_FIRST_HARD_FORK: u64 = 3;
 
 /// AutomatedTesting and UserTesting second hard fork height.
 pub const TESTING_SECOND_HARD_FORK: u64 = 6;
+
+/// AutomatedTesting and UserTesting third hard fork height.
+pub const TESTING_THIRD_HARD_FORK: u64 = 9;
 
 /// Compute possible block version at a given height, implements
 /// 6 months interval scheduled hard forks for the first 2 years.
@@ -151,8 +157,10 @@ pub fn header_version(height: u64) -> HeaderVersion {
 				HeaderVersion(1)
 			} else if height < FLOONET_SECOND_HARD_FORK {
 				HeaderVersion(2)
-			} else if height < 3 * HARD_FORK_INTERVAL {
+			} else if height < FLOONET_THIRD_HARD_FORK {
 				HeaderVersion(3)
+			} else if height < 4 * HARD_FORK_INTERVAL {
+				HeaderVersion(4)
 			} else {
 				HeaderVersion(hf_interval)
 			}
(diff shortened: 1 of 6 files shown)


tromp:
NOTE that the FLOONET_THIRD_HARD_FORK value may need revising when building a beta or RC according to whether we still want to HF on June 20 and how fast floonet progressed.

core/src/consensus.rs
```diff
@@ -133,12 +133,18 @@ pub const FLOONET_FIRST_HARD_FORK: u64 = 185_040;
...
+pub const FLOONET_THIRD_HARD_FORK: u64 = 552_960;
+
 /// AutomatedTesting and UserTesting first hard fork height.
 pub const TESTING_FIRST_HARD_FORK: u64 = 3;
 
 /// AutomatedTesting and UserTesting second hard fork height.
 pub const TESTING_SECOND_HARD_FORK: u64 = 6;
 
+/// AutomatedTesting and UserTesting third hard fork height.
+pub const TESTING_THIRD_HARD_FORK: u64 = 9;
```
antiochp, line 146: Thanks!

core/src/consensus.rs
```diff
@@ -174,7 +184,7 @@ pub fn header_version(height: u64) -> HeaderVersion {
 /// Check whether the block version is valid at a given height, implements
 /// 6 months interval scheduled hard forks for the first 2 years.
 pub fn valid_header_version(height: u64, version: HeaderVersion) -> bool {
-	height < 3 * HARD_FORK_INTERVAL && version == header_version(height)
+	height < 4 * HARD_FORK_INTERVAL && version == header_version(height)
 }
```
antiochp, line 188: Just realized this logic is only correct for `mainnet`. 
It is not valid for either floonet or `usertesting` which do not use `HARD_FORK_INTERVAL` directly.

This is not specific to this PR and I don't think it has a material impact here - but we should revisit this.

Tracking here - https://github.com/mimblewimble/grin/issues/3335

core/src/global.rs
```diff
@@ -187,8 +187,11 @@ where
 {
 	let chain_type = get_chain_type();
 	match chain_type {
-		// Mainnet has Cuckaroo(d)29 for AR and Cuckatoo31+ for AF
+		// Mainnet has Cuckaroo{,d,m,z}29 for AR and Cuckatoo31+ for AF
```
antiochp, line 190: `{,`

tromp: The original AR PoW was simply called Cuckaroo, so the first of the 4 options has an empty string suffix, hence the {,

core/src/consensus.rs
tromp: Why can't floonet and usertesting be limited to 2 years worth of block just like mainnet is for version 4? Of course that will be changed downward at the next HF, but for now it looks fine to me.

antiochp: Yeah maybe that is fine actually.
I was originally thinking automatedtesting wants to be able to test an invalid block due to excessive block height. But its going to be based on header version anyway.

:+1:

quentinlesceller:
@tromp can you fix the merge conflict?

core/src/consensus.rs
```diff
@@ -133,12 +133,18 @@ pub const FLOONET_FIRST_HARD_FORK: u64 = 185_040;
 /// Floonet second hard fork height, set to happen around 2019-12-19
 pub const FLOONET_SECOND_HARD_FORK: u64 = 298_080;
 
+/// Floonet second hard fork height, set to happen around 2020-06-20
+pub const FLOONET_THIRD_HARD_FORK: u64 = 552_960;
```
quentinlesceller, line 137: For reference the exact time is currently 2020-06-19 18:49:36 +0000 UTC

anton679:
Payout wallet

PR #3342, opened 2020-06-05
Rework pool tests to use real chain (was mock chain)

antiochp:
This PR reworks the existing pool tests to setup and use a real chain instance.
We were mocking the chain out previously. 
This was introducing some inconsistencies in the chain block and tx validation rules.

This required reworking the `block_weight` tests to use `AutomatedTesting` chain type rather than `Mainnet`.

Changes in this PR are limited in scope to test code only. 
The only exception to this is an increase to the `global::TESTING_MAX_BLOCK_WEIGHT` to allow the block_weight tests to be run against `AutomatedTesting`.

This is refactoring required to test the NRD lock height rules (separate PR) so we can effectively test the rules without needing to mock them up in chain.


antiochp:
Going to merge this now so it does not block #3302.
Treating this as relatively low risk as it is limited in scope to test code with no functional changes.

@quentinlesceller please continue to review and we can address any feedback separately.

PR #3347, opened 2020-06-10
Remove wallet from CLI

quentinlesceller:
Simply remove the wallet command and argument not used since grin v1.1.0


antiochp:
:+1:

Issue #3352, opened 2020-06-12
fsync rabbit hole


antiochp:
I have been doing some investigation into our MMR backend files, how/when we fsync them to disk and a variety of assumptions we make around this, both implicit and explicit.

### Questionable Assumption 1
I (and presumably others) have been working under the assumption we call fsync on our leaf_set file every time we process and accept a new block. We do _not_ actually ever fsync the `leaf_set` or the `prune_list`. Out of all the files that we persist to disk, the `leaf_set` is probably the one we care about the most and should be fsync'd...

The `hash` and `data` files are fsynbc'd here via `sync_all()` - 

https://github.com/mimblewimble/grin/blob/dd88d28a7a22b509f01e20685fdf9a485ca82160/store/src/types.rs#L380-L381

In contrast the `leaf_set` is handled like this - 

https://github.com/mimblewimble/grin/blob/dd88d28a7a22b509f01e20685fdf9a485ca82160/store/src/leaf_set.rs#L174-L178

`save_via_temp_file` does a dance with creating a temp file, deleting the original file and renaming the temp file. But again, no fsync and no guarantee the bytes ever made it to disk. And I suspect with a variety of additional points in time where data loss can occur. For example between deleting the original file and renaming the temp file. 

The `prune_list` is handled similarly during compaction.
The call to `flush()` ensures the buffer is flushed to the file, but not that the file is fsync'd to disk.

Intuitively the fix here is simple. We want to ensure the `leaf_set` is fsync'd to disk after we successfully process and accept a new block.

### Questionable Assumption 2
tl;dr Still digesting this but fsync does not always provide the guarantees as commonly understood. There are failure conditions where a successful fsync will _still_ result in data being partially written to disk.

https://lwn.net/Articles/752063/

> In short, PostgreSQL assumes that a successful call to fsync() indicates that all data written since the last successful call made it safely to persistent storage. But that is not what the kernel actually does.

It appears this is what happens almost all the time, but not necessarily _guaranteed_ to be true 100% of the time.
And if this is not something that can be guaranteed 100% of the time - what is the point of relying on fsync at the app level?

### Questionable Assumption 3
We fsync 9 files to disk across the header, output, rangeproof and kernel MMRs when we successfully process and accept a new block. We do this to ensure the data is on disk and we _must_ do this for durability.

[...]

PR #3354, opened 2020-06-12
sync_all() on leaf_set and prune_list

antiochp:
4.0.0 has a dedicated branch now. Opening this up against master.

----

We should call `sync_all()` (aka `fsync`) when writing the leaf_set (and prune_list) with `save_via_temp_file`.

Renaming a file without first calling `fsync` is not safe.
See https://stackoverflow.com/questions/7433057/is-rename-without-fsync-safe

Also avoid removing the original file before renaming the temp file as this kind of defeats the atomicity of the "temp file and rename" approach.

Related #3352 (fsync rabbit hole).

### DIFF
--- a/store/src/leaf_set.rs
+++ b/store/src/leaf_set.rs
@@ -171,10 +171,8 @@ impl LeafSet {
 		self.bitmap.run_optimize();
 
 		// Write the updated bitmap file to disk.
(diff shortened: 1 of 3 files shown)


antiochp:
@quentinlesceller thanks for taking a look. I have not touched this branch for a bit so I'm going to pick it up again and do some more testing before merging.

Issue #3360, opened 2020-06-17
Blocks more than 12 mins into the future appear to lead to peer banning


tromp:
Block header deserialization fails in

https://github.com/mimblewimble/grin/blob/450d235358bb14016c1c8e26d036526e6bf3af36/core/src/core/block.rs#L422-L432

if the timestamp is more than 12 mins into the future, returning a CorruptedData error.
Presumably this leads to the banning of the sending peer.

But this happens almost randomly if a miner sets a timestamp nearly 12 mins into the future due to natural variation in local clock times, which would cause random peer banning left and right.

It would seem preferable to just ignore the block, rather than declare it corrupted?!

antiochp:
> Presumably this leads to the banning of the sending peer.

I may be wrong but I don't believe a `CorruptedData` results in peer banning.
We definitely ignore the block.
We _may_ drop the peer connection. But I'm relatively certain we do not ban the peer.

tromp:
I notice we return CorruptedData on future timestamp, which is not a bannable offense, and on invalid PoW, which does seem to be a bannable offense. Perhaps the header deserialize needs more informative error codes?

antiochp:
Also the distinct possibility that we do not ban peers due to the timestamp but then in the process of dropping the connection (and presumably reconnecting shortly afterwards) something happens that results in one end of that peer-peer connection getting banned.

antiochp:
From looking over this code again briefly - it does appear that bann-able offenses occur only once the block has been successfully deserialized and the block enters the processing pipeline.

Deserialization errors will only ever result in the msg and underlying peer connection getting dropped.

antiochp:
There are 5 places in peers.rs where we can explicitly ban a peer during processing - 

* block_received (processing a full block, after successfully deserializing it)
* compact_block_received (ditto for compact block)
* header_received (ditto for header)
* headers_received (ditto for sync headers)
* txhashset_write (when handling txhashset zip archive)

We can also explicitly ban a peer if we request headers during sync and the peer fails to provide them.

The only other place we ban peers is via the ban_peer api endpoint.

PR #3361, opened 2020-06-19
[WIP] Ban peer on specific deserialization errors

antiochp:
Currently a `CorruptedData` error during deserialization simply results in a peer connection getting dropped. 
Some errors during deserialization are more serious and should result in a peer being banned.

Specifically during header deserialization we should probably ban a peer if PoW related `verify_size()` fails.

Related #3360.

Several related cleanup/refactoring is also in this PR, revisiting the peer ban/unban impl.

* refactor ban/unban peer as errors here are not particularly useful
  * if we encounter an error when banning a peer there is not much we can do
  * it should not be an error if we attempt to unban a peer not currently banned
* move `ban_peer` to `NetAdapter` trait to make it available to `consume()` during msg parsing and deserialization

TODO - 
- [ ] expose a more serious version of `CorruptedData`
- [ ] handle this more serious version by banning the peer in `consume()`
- [ ] tests to ensure we are not deadlocking on `peers` or similar
- [ ] make `verify_size()` a bannable verification step
- [ ] investigate other "untrusted" bannable verification during deserialization

### DIFF
--- a/api/src/handlers/peers_api.rs
+++ b/api/src/handlers/peers_api.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use super::utils::w;
-use crate::p2p::types::{PeerAddr, PeerInfoDisplay, ReasonForBan};
+use crate::p2p::types::{NetAdapter, PeerAddr, PeerInfoDisplay, ReasonForBan};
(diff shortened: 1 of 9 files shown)


yeastplume:

PR #3365, opened 2020-06-24
add CuckooParams::node_mask, obsolete T:EdgeType and sipnode shift :bool arg

tromp:
This PR refactors the PoW code for greater simplicity.
Formerly, several data structures were templated by some T:EdgeType,
which was almost always instantiated as u64, except in the lean solver where it was u32.
Now it's simply hardwired to u64, many conversions are eliminated, and only a few added
in the lean solver.
The sipnode function had a third argument shift:bool which was always false and so could be deleted.
Finally, the node_mask which was set in the various cuckaroo* verify functions is now made a member
of CuckooParams and set through new constructor argument node_bits.

Code has been tested on a mainnet fast sync.

### DIFF
--- a/core/src/pow/common.rs
+++ b/core/src/pow/common.rs
@@ -14,7 +14,7 @@
 
 //! Common types and traits for cuckoo family of solvers
 
-use crate::pow::error::{Error, ErrorKind};
+use crate::pow::error::Error;
 use crate::pow::num::{PrimInt, ToPrimitive};
 use crate::pow::siphash::siphash24;
 use blake2::blake2b::blake2b;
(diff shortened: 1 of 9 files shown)


core/src/pow/common.rs
```diff
@@ -135,31 +113,29 @@ macro_rules! to_edge {
...
 
-impl<T> CuckooParams<T>
-where
-	T: EdgeType,
-{
+impl CuckooParams {
 	/// Instantiates new params and calculate edge mask, etc
-	pub fn new(edge_bits: u8, proof_size: usize) -> Result<CuckooParams<T>, Error> {
-		let num_edges = (1 as u64) << edge_bits;
-		let edge_mask = to_edge!(T, num_edges - 1);
```
antiochp, line 156: We can remove the `to_edge` macro up now I think as it is no longer used.
And are the other `to_u64`, `to_u32` and `to_usize` macros still in use? I suspect we no longer need these now that we are doing less conversions.

PR #3367, opened 2020-06-26
cleanup unused macros and error

antiochp:
These macros are no longer required as the conversion code was simplified and cleaned up in #3365.
Now the macros are gone we no longer need the `IntegerCast` error kind.

This PR cleans up the unused code.


antiochp:
Going to merge this. Just cleaning up unused code.

PR #3374, opened 2020-07-02
bump master to 4.0.1-alpha.1 to differentiate from 4.0.0 official tagged build

antiochp:
So we can identify `4.0.0` vs. `master` on mainnet.


antiochp:
Merging. This is limited to a version bump on `master`.

PR #3385, opened 2020-07-10
[WIP] v3 blocks and transactions (no redundant input features)

### LINKED ISSUE #3378: Transaction (and block) inputs need only include the commitment
We currently include the following in a transaction `input` - 
* output features
* commitment

So spending `(plain, C)` is different to spending `(coinbase, C)`.

We currently enforce uniqueness on output commitments in the utxo set. 
So this is not of material concern other than this being redundant data in a transaction.

If/when we support "duplicate outputs" then things get a little more complicated.
We want to build a transaction that spends the most recent commitment `C`, regardless of output features.
We _do not_ plan to support granular selection of the output being spent based on specific output features. The rule will simply be "last-in-first-out" and will be strictly enforced. The specific instance of multiple duplicate outputs (based on output commitment) will be the one with the highest output MMR position (right-most).

Related #3271.

We should consider simplifying transaction and block serialization sooner rather than later to simplify this and remove the output features on inputs.

Ideally we would remove the features bytes on _all_ inputs.
We may want to consider enforcing a zero features byte if necessary as an interim solution.
We do not want to simply ignore the feature byte as this would introduce potential malleability.

antiochp:
[WIP]

* v2 inputs are FeaturesAndCommit
* v3 inputs are CommitOnly

There is some complexity here as translating a v3 input to v2 requires "external" data (the output features). The output features must be read from the output MMR which must be in the correct state itself, based on current block being processed.

Additionally this messes with sort order as we sort based on hash and the data being hashed is different.
So converting v2->v3 and v3->v2 cannot be done purely on individual inputs. 
It must be handled at the collection level as the collection of inputs must be reordered to account for the new sort order.

Resolves #3378.

----

TODO - 

- [x] `Inputs` enum variants `CommitOnly, FeaturesAndCommit`
- [x] `CommitWrapper` for sorting commitments correctly by `hash()`
- [x] serialization support for Input variants (`v2` vs. `v3`)
- [ ] block support
  - [x] receive v2
  - [x] receive v3
  - [x] broadcast/send v2
  - [ ] broadcast/send v3
  - [x] convert v2 -> v3
  - [x] convert v3 -> v2
  - [x] store in db in v2 (for backward compatibility)
- [ ] transaction support
  - [x] convert v2 -> v3
  - [x] handle as v3 internally in txpool
  - [x] convert to v2 for backward compatibility for v2 peers
  - [x] relay/send v2
  - [ ] relay/send v3
- [ ] version negotiation (make sure this is working as expected)


antiochp:
Replaced with smaller more focused refactor PRs (various) and #3419.

PR #3395, opened 2020-07-16
Use mmr sizes in header to validate header weight

### LINKED ISSUE #3368: Consider "max weight" consensus rule for block headers (not just full blocks)
We have a "max block weight" consensus rule on full blocks. 
Weight is a function of the number of inputs, outputs and kernels in the block.

Proposal is to add a "max weight" consensus rule on _block headers_ also.
This would allow nodes to identify "too-heavy" blocks earlier and without necessarily having to receive full blocks.
This would prevent spam and other undesirable behavior during block propagation.

We commit to the `output_mmr_size` and `kernel_mmr_size` in block headers and from these we can determine the number of outputs and kernels. We unfortunately _cannot_ determine the number of inputs in a block given just the block headers.

One option would be to implement a similar but not identical "max weight" rule for headers, taking _only_ outputs and kernels into consideration. This would still allow headers for "too heavy" blocks to be accepted, but would restrict the composition of "too heavy" blocks. A block containing an excessive number of kernels or outputs would not have a valid header due to the "max weight" rule.
This can be implemented today without any changes to the header structure.

The other option would be to commit to _something_ in the block header that would allow the number of inputs to be determined. This would need to be a count or size related to either the stxo (spent outputs) or utxo (unspent outputs).
This would allow us to use the _same_ "max weight" consensus, in a way that could be applied to both headers and full blocks consistently.

The following discussion on keybase covers this - 

----

Keybase discussion involving @DavidBurkett, @tromp, @antiochp.
keybase://chat/grincoin.teams.node_dev#general/5220

> We should probably add consensus rules that limit the number of kernels and outputs per header. I know we have the size rule for blocks already, but technically a 1-week reorg could allow a miner to fill the chain with terabytes of data. Checking the output mmr size and kernel mmr size during header sync could prevent that.

> yeah that's actually an interesting idea - we don't need to add any new consensus rules, just apply them to headers in addition to full blocks (and use the mmr sizes as you suggest to determine counts for the block). So we can use the existing consensus size rules.

> Yep, we can probably just include that code whenever. It's technically a softfork, but it's so rare we wouldn't have to worry about activation or anything. Just let me know if/when you make the change and I'll do the same.

> ahh actually we can determine the number of outputs and kernels given just the sizes in the header, but not the number of inputs (which is annoying) still thinking about this
I guess one option would be to just consider outputs and kernels and check the block does not exceed max weight

> the other option is to add a number_of_spent_outputs header field

> I propose utxo_size as an alternative to number_of_spent_outputs in the header - its a "global" value, not per block and consistent with existing xxx_mmr_size and xxx_mmr_root etc. If we were to seriously consider adding a new header field.
and derive the number of spent inputs from this
utxo_size would potentially be useful to have - even if just for quick view into size of things over time

> number_of_spent_outputs is global too (cumulative)

> it's the size of the spent txo mmr if we had one

> do you remember argument against adding spent txo mmr to fix utxo commitment issue?

> it was significantly more complexity to implement vs the utxo bitmap approach we took - not sure if there were others
in terms of sync and p2p msgs and other related stuff
its subtly different as well right, thinking about it now - currently we care about utxo at a point in time (and do not care about the history in there) but the stxo would be strictly ordered

> i see. it defines a total order on (spent) outputs, which bitmap does not.

> to check for inputs being unspent, we need to check bitmap
and to update as well
which grows linearly with txo set size
but spent txo mmr is cheap to update, you only keep peaks
and you can choose any utxo set representation you want
which will be sublinear in txo set set

> depends on your utxo set impl. e.g. pick a radix tree or red black tree. it will all be O(log n) checks and updates
it will be decoupled from header commitment fields
which are only on mmrs

> yes, the txo mmr and stxo mmr define the commitments to utxo
and you're free to pick a utxo set representation to check membership
you just need to initialize it when starting up your node or syncing it

> i guess you could argue that the bitmap mmr need not really be linear in size

> you could still have your arbitrary utxo set representation that's linear in utxo set size, as long as you can efficiently compute bitmap mmr merkle root from it...
so i guess we're good

antiochp:
Resolves #3368.

We commit to output_mmr_size and kernel_mmr_size in the block header.
Given these we can approximate the full block weight (assuming 0 inputs).

This PR introduces a weight validation rule based on this when processing block headers.

My understanding is this is technically a softfork due to nodes fast-syncing headers (not full blocks) beyond the horizon? I believe @DavidBurkett stated this in keybase.

Otherwise we may accept a header as valid and then subsequently reject the full block as overweight (current behavior).

Took the opportunity to clean up `weight` to use `u64` consistently and not an awkward `usize`.

cc @tromp @DavidBurkett.

----

Suggestion from @tromp - we can go one better than this. 
This PR allows us to verify block weight per header. 
We can also verify the _global_ weight at any given header height based on the overall output MMR size.

----

DONE - 
- [x] validate "global" weight limits when deserializing "untrusted" block header
- [x] validate per-header weight limits based on mmr sizes
- [x] validate per-header "must have at least 1 output and 1 kernel for block reward" based on mmr sizes

### DIFF
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -18,7 +18,8 @@ use crate::core::consensus;
 use crate::core::core::hash::Hashed;
 use crate::core::core::verifier_cache::VerifierCache;
 use crate::core::core::Committed;
-use crate::core::core::{Block, BlockHeader, BlockSums, OutputIdentifier};
+use crate::core::core::{transaction, Block, BlockHeader, BlockSums, OutputIdentifier, TransactionBody};
+use crate::core::global;
 use crate::core::pow;
 use crate::error::{Error, ErrorKind};
 use crate::store;
@@ -334,6 +335,20 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(
 		// prevent time warp attacks and some timestamp manipulations by forcing strict
 		// time progression
 		return Err(ErrorKind::InvalidBlockTime.into());
+	}
+
+	// We can determine output and kernel counts for this block based on mmr sizes from previous header.
+	// Assume 0 inputs and estimate a lower bound on the full block weight.
+	// Block header is invalid (and block is invalid) if this lower bound is too heavy for a full block.
+	let num_outputs = header
+		.output_mmr_count()
+		.saturating_sub(prev.output_mmr_count());
+	let num_kernels = header
+		.kernel_mmr_count()
+		.saturating_sub(prev.kernel_mmr_count());
+	let weight = TransactionBody::weight_as_block(0, num_outputs, num_kernels);
+	if weight > global::max_block_weight() {
+		return Err(ErrorKind::Transaction(transaction::Error::TooHeavy).into());
 	}
 
 	// verify the proof of work and related parameters
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -18,9 +18,7 @@ use crate::core::consensus;
 use crate::core::core::hash::Hashed;
 use crate::core::core::verifier_cache::VerifierCache;
 use crate::core::core::Committed;
-use crate::core::core::{
-	transaction, Block, BlockHeader, BlockSums, OutputIdentifier, TransactionBody,
-};
+use crate::core::core::{block, Block, BlockHeader, BlockSums, OutputIdentifier, TransactionBody};
 use crate::core::global;
 use crate::core::pow;
 use crate::error::{Error, ErrorKind};
@@ -356,7 +354,7 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(
 	// Block header is invalid (and block is invalid) if this lower bound is too heavy for a full block.
 	let weight = TransactionBody::weight_as_block(0, num_outputs, num_kernels);
 	if weight > global::max_block_weight() {
-		return Err(ErrorKind::Transaction(transaction::Error::TooHeavy).into());
+		return Err(ErrorKind::Block(block::Error::TooHeavy).into());
 	}
 
 	// verify the proof of work and related parameters
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -349,7 +349,7 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(
 		.saturating_sub(prev.kernel_mmr_count());
 
 	// Each block must contain at least 1 kernel and 1 output for the block reward.
-	if header.height > 0 && (num_outputs == 0 || num_kernels == 0) {
+	if num_outputs == 0 || num_kernels == 0 {
 		return Err(ErrorKind::InvalidMMRSize.into());
 	}
 
--- a/chain/src/pipe.rs
+++ b/chain/src/pipe.rs
@@ -18,7 +18,9 @@ use crate::core::consensus;
 use crate::core::core::hash::Hashed;
 use crate::core::core::verifier_cache::VerifierCache;
 use crate::core::core::Committed;
-use crate::core::core::{transaction, Block, BlockHeader, BlockSums, OutputIdentifier, TransactionBody};
+use crate::core::core::{
+	transaction, Block, BlockHeader, BlockSums, OutputIdentifier, TransactionBody,
+};
 use crate::core::global;
 use crate::core::pow;
 use crate::error::{Error, ErrorKind};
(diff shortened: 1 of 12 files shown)


chain/src/pipe.rs
```diff
@@ -336,6 +339,26 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(
...
+	// Assume 0 inputs and estimate a lower bound on the full block weight.
+	let num_outputs = header
+		.output_mmr_count()
+		.saturating_sub(prev.output_mmr_count());
+	let num_kernels = header
+		.kernel_mmr_count()
+		.saturating_sub(prev.kernel_mmr_count());
+
+	// Each block must contain at least 1 kernel and 1 output for the block reward.
+	if header.height > 0 && (num_outputs == 0 || num_kernels == 0) {
```
tromp, line 352: validate_header is never called on genesis, so height test is redundant.
besides, even genesis has output and kernel:-)

antiochp: Good catch. I'll rework this and double check this assumption actually holds (even when running tests etc.)

antiochp: :+1:

chain/src/pipe.rs
```diff
@@ -336,6 +339,26 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(
...
+		return Err(ErrorKind::InvalidMMRSize.into());
+	}
+
+	// Block header is invalid (and block is invalid) if this lower bound is too heavy for a full block.
+	let weight = TransactionBody::weight_as_block(0, num_outputs, num_kernels);
+	if weight > global::max_block_weight() {
+		return Err(ErrorKind::Transaction(transaction::Error::TooHeavy).into());
```
tromp, line 359: Should we have a separate block::Error::TooHeavy for (inferring that) a single block exceeded the max weight?
We can reserve transaction::Error::TooHeavy for txs that are not block bodies...

antiochp: Yes we should, particularly as this is a result of us calling `weight_as_block` earlier.
:+1:

core/tests/block.rs
```diff
@@ -423,6 +424,59 @@ fn serialize_deserialize_block_header() {
...
+	let res: Result<UntrustedBlockHeader, _> = ser::deserialize_default(&mut &vec[..]);
+	assert_eq!(res.err(), Some(ser::Error::CorruptedData));
+
+	// Set reasonable mmr sizes on the header to confirm the header can now be read "untrusted".
+	b.header.output_mmr_size = 1;
+	b.header.kernel_mmr_size = 1;
+	set_pow(&mut b.header);
+
+	let mut vec = Vec::new();
+	ser::serialize_default(&mut vec, &b.header).expect("serialization failed");
```
tromp, line 475: why do you expect "serialization failed"  ?

core/tests/block.rs
```diff
@@ -423,6 +424,59 @@ fn serialize_deserialize_block_header() {
...
+
+	// Set excessively large output mmr size on the header.
+	b.header.output_mmr_size = 10_000;
+	b.header.kernel_mmr_size = 0;
+	set_pow(&mut b.header);
+
+	let mut vec = Vec::new();
+	ser::serialize_default(&mut vec, &b.header).expect("serialization failed");
+	let res: Result<UntrustedBlockHeader, _> = ser::deserialize_default(&mut &vec[..]);
+	assert_eq!(res.err(), Some(ser::Error::CorruptedData));
```
tromp, line 457: so the transaction::Error::TooHeavy gets converted into a ser::Error::CorruptedData ?

core/tests/block.rs
antiochp: Rust is confusing. This is saying we expect it _not_ to be an error here.
`expect("reason for error")` is an alternative to an opaque `unwrap()`

core/tests/block.rs
antiochp: Yes. We need to return a generic `ser::Error` error and in most cases we use `CorruptedData` for low level serialization issues.

tromp:
> We can also verify the global weight at any given header height based on the overall output MMR size.

That doesn't make sense in hindsight:-(

tromp (approved):

core/tests/block.rs
antiochp: Actually this is not happening because of `transaction::Error::TooHeavy`.
We have an additional validation check that happens during deserialization of a block header. This is where the "global weight" is used.

https://github.com/mimblewimble/grin/blob/96e8bfa34bc9718af4be99affbb7aa3df5414235/core/src/core/block.rs#L471-L479

This is a rough check that we can do very cheaply as part of deserialization.

> We can also verify the global weight at any given header height based on the overall output MMR size.
> That doesn't make sense in hindsight:-(

It's not perfect but it prevents somebody trying to broadcast a header with massively inflated MMR sizes.
And we can verify this without needing to read the previous header from the db.

We still do the per-header weight validation later to give us a more accurate validation.

antiochp:
> > We can also verify the global weight at any given header height based on the overall output MMR size.
> 
> That doesn't make sense in hindsight:-(

See comment above about `CorruptedData` error.
Still makes sense to verify this early (and cheaply).

PR #3396, opened 2020-07-17
Add TransactionPool fuzz driver

unseddd:
Add fuzz driver for fuzzing adding transactions to the main and stem `TransactionPool`

Optionally, generate a corpus of transactions:

```
$ cd grin/pool
$ GRIN_POOL_GEN_CORPUS=0 cargo fuzz run transaction_pool
```


unseddd:
Found some memory leaks (non-security bugs), and a potential "use of unitialized memory" that appears to actually be safe.

Can provide test inputs for reproducing leaks, if they are wanted.

antiochp:
Taking a look at this now. 
Does cargo-fuzz still require a nightly rust build?

unseddd:
> Does cargo-fuzz still require a nightly rust build?

It does. You can install on stable, but actually compiling/running fuzz tests requires nightly.

PR #3397, opened 2020-07-18
use sync pmmr for building locator

antiochp:
We don't need to repeatedly call `get_previous_header()` to build the locator.
We can take advantage of our sync pmmr and simply index directly into this to retrieve the necessary header hashes.
This is fast and we no longer need to jump through hoops maintaining a cached locator.


DavidBurkett:
image

My favorite kind of PR.

PR #3401, opened 2020-07-22
pass ref to chain around

antiochp:
Cleanup related to #3400. 
We were doing a lot of passing a cloned `Arc<chain::Chain>` around when we can simply be passing `&chain::Chain`.


antiochp:
Going to merge this. Just a refactor, no behavior changes.

PR #3404, opened 2020-07-26
pass slices around and not refs to vecs

antiochp:
It is preferable to pass `&[txs]` around rather than `&Vec<>`.

This refactoring is extracted from #3385 as that PR is getting a little out of hand.

No functional changes, purely refactoring.
Lots of test code was creating vecs of transactions and we now simply create slices directly.

Edit: The one large implementation change is in `cut_through()` with the "zero allocation" approach.
We cannot call `drain()` on the inputs and outputs to chop the cut-through non-copied middle section out. We now need to complete the "copy to start of slice" operation allowing us to then return the shortened slice.
We also return the shortened slices as we cannot simply update the slice refs. This requires some lifetimes.
Also use `swap()` here when moving non-cut-through elements to the start of the slice. This effectively resorts the elements of the full slice in a cleaner way, with non-cut-through elements at the start, ready to be sub-sliced.
This is an implementation detail but a larger change than majority of this PR.

Edit: I think there may be a way to actually update the refs to the passed in slices, but it does not appear to make the code any easier to read - https://stackoverflow.com/questions/34384089/how-can-i-modify-a-slice-that-is-a-function-parameter

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -212,8 +212,9 @@ impl<'a> Batch<'a> {
 
 	/// We maintain a "spent" index for each full block to allow the output_pos
 	/// to be easily reverted during rewind.
-	pub fn save_spent_index(&self, h: &Hash, spent: &Vec<CommitPos>) -> Result<(), Error> {
-		self.db.put_ser(&to_key(BLOCK_SPENT_PREFIX, h)[..], spent)?;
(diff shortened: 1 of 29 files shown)


antiochp:
Going to merge. This is just a refactor, no behavior changes.

Issue #495, opened 2017-12-14
make grin (the software) more accessible to new users


sesam:
Coordination here: https://etherpad.wikimedia.org/p/grin-welcome

sesam:
With the TUI implemented, I think the above is no longer needed :). Closing for now. Maybe there are more specific things that can still be improved, that can be handled as direct PRs to minimise bikeshedding.

PR #3407, opened 2020-07-29
Expose `max_tx_size` API

unseddd:
Related to grin-wallet [#495](https://github.com/mimblewimble/grin-wallet/pull/495)

### DIFF
--- a/core/src/global.rs
+++ b/core/src/global.rs
@@ -17,10 +17,10 @@
 //! should be used sparingly.
 
 use crate::consensus::{
-	graph_weight, valid_header_version, HeaderInfo, BASE_EDGE_BITS, BLOCK_TIME_SEC,
-	COINBASE_MATURITY, CUT_THROUGH_HORIZON, DAY_HEIGHT, DEFAULT_MIN_EDGE_BITS,
-	DIFFICULTY_ADJUST_WINDOW, INITIAL_DIFFICULTY, MAX_BLOCK_WEIGHT, PROOFSIZE,
-	SECOND_POW_EDGE_BITS, STATE_SYNC_THRESHOLD,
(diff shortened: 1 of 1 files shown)


antiochp:
Agree that we should have a limit on slate sizes. 
But I'm not convinced `max_block_size` is the correct thing to base this off.

unseddd:
> But I'm not convinced max_block_size is the correct thing to base this off.

Understand, response is in [#495](https://github.com/mimblewimble/grin-wallet/pull/495).

antiochp:
Maybe one option here would be to expose a new `max_tx_size()` (this can just be based off max_block_size internally).
This way the wallet would be using an api less likely to change unexpectedly.

May need to do additional calcs on the wallet side but this would be fine here.

unseddd:
> Even better if we could expose this as `global::max_tx_size()` and avoid needing to `require p2p` from wallet.

Exposing as `global::max_tx_size()` creates a cyclic dependency, so instead chose to expose as `api::max_tx_size()`.

Seems a logical place for the function, `api` already requires the needed sub-crates, and `api` is already required by `grin_wallet::util`. WDYT?

api/src/lib.rs
```diff
@@ -56,3 +56,9 @@ pub use crate::rest::*;
 pub use crate::router::*;
 pub use crate::types::*;
 pub use crate::web::*;
+
+/// Get max transaction size
+pub fn max_tx_size() -> u64 {
+	p2p::msg::max_block_size()
+		- std::mem::size_of_val(&core::core::block::BlockHeader::default()) as u64
```
antiochp, line 63: This does not necessarily correspond to a serialized size in bytes. Its simply the internal representation, which will likely be different.

PR #3409, opened 2020-07-29
refactor apply_inputs and support converting block for v2 compatibility

antiochp:
Refactor `apply_inputs()` and `validate_inputs()` to reduce duplicated code and simplify block conversion for future compatibility and v2 peer support.

Extracted some relatively isolated changes from #3385.

* `utxo_view` is now "source of truth" for identifying outputs being spent by inputs
* we now convert a block to v2 compatibility when processing new blocks
  * uses `utxo_view.validate_inputs()` to identify outputs being spent
* `apply_inputs()` now uses `utxo_view.validate_inputs()` internally
  * removed duplicate lookup code in `apply_inputs()`

We need to hook block conversion into the block processing pipeline early on, but _after_ orphan block identification. 
This required some refactoring of `chain.process_block_single()`.
Checking for orphan block is simplified with an explicit check for "previous block in db".

This is a relatively large PR but mainly refactoring with the addition of the "v2 block conversion" step (lookup features for each output being spent).

We have good test coverage of the block processing pipeline.
Testing this on mainnet also to ensure the node syncs and stays in sync.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -262,18 +262,9 @@ impl Chain {
 
 	/// Processes a single block, then checks for orphans, processing
 	/// those as well if they're found
-	pub fn process_block(&self, block: Block, opts: Options) -> Result<Option<Tip>, Error> {
-		let height = block.header.height;
-
-		// Process the header first.
-		// If invalid then fail early.
-		// If valid then continue with block processing with header_head committed to db etc.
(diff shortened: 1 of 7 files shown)


antiochp:
Merging this - tested thoroughly on floonet and mainnet (and mining on usertesting).

PR #3412, opened 2020-08-01
Tx conversion for v2 compatibility

antiochp:
Similar to converting blocks in #3409 this PR adds functionality to convert transactions.
We may receive (not yet implemented) transactons with "commit only" inputs, but we still need to support relaying this transactions to v2 peers that expect "features and commit" inputs.
This PR looks up outputs "to be spent" via both the stempool/txpool and current utxo to find the corresponding output features.
We leverage the improved cut_through functionality from #3410 in the lookup implementation.
We also take advantage of #3411 to allow cut_through on output identifiers rather than full outputs.

All transactions and blocks are now handled in v2 compatible mode internally.
This ensures we can subsequently relay or broadcast to v2 peers correctly without additional conversion or lookup.

Note: There is _no_ v3 transactions or blocks yet. So the conversion is simply v2 -> v2 in all cases.
But we are exercising the conversion code on all blocks and transactions.

The actual v3 support will come in a subsequent PR.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -607,12 +607,29 @@ impl Chain {
 		})
 	}
 
-	fn validate_tx_against_utxo(&self, tx: &Transaction) -> Result<(), Error> {
+	fn validate_tx_against_utxo(
+		&self,
+		tx: &Transaction,
+	) -> Result<Vec<(OutputIdentifier, CommitPos)>, Error> {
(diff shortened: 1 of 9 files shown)


antiochp:
Going to merge. Next step is a PR that will actually introduce v2/v3 inputs with alternative ser/deser impls.

PR #3416, opened 2020-08-07
"com" and "prf" in coinbase from wallet

antiochp:
Resolves: https://github.com/mimblewimble/grin-wallet/issues/499

We ask the wallet for a `CbData` (output and kernel for block reward).

The output looks like this - 

```
	"output": {
		"com": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
		"features": 1,
		"prf": "9d8488fc..."
	}
```

It used to look like this - 

```
	"output": {
		"commit": "08fe198e525a5937d0c5d01fa354394d2679be6df5d42064a0f7550c332fce3d9d",
		"features": "Coinbase",
		"proof": "9d8488fc..."
	}
```

This PR adds an `alias` to both `commit` and `proof` via a custom `CbOutput`.
For convenience we ignore the output features (and kernel features) as this wallet api will _only_ return coinbase specific output and kernel.
This code is only used for this single wallet api call.

Tested locally on `usertesting` and confirmed the node can talk to the wallet successfully (both on `master`).

### DIFF
--- a/servers/src/mining/mine_block.rs
+++ b/servers/src/mining/mine_block.rs
@@ -25,11 +25,13 @@ use std::time::Duration;
 use crate::api;
 use crate::chain;
 use crate::common::types::Error;
-use crate::core::core::{Output, TxKernel};
+use crate::core::core::{KernelFeatures, Output, OutputFeatures, TxKernel};
 use crate::core::libtx::secp_ser;
 use crate::core::libtx::ProofBuilder;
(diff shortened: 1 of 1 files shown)


antiochp:
Closing. We should fix this on wallet side to maintain compatibility between old nodes and new wallet.

PR #3418, opened 2020-08-08
orphan check needs to handle fast sync "edge case"

### LINKED ISSUE #3417: always get stuck at 99% of Downloading blocks
Log file at https://tromp.github.io/grin-server.log
TUI screen:

Screen Shot 2020-08-08 at 12 01 51 AM

antiochp:
Resolves #3417

We moved the "is_orphan" check earlier in the block processing pipeline.
As part of this refactoring we made an incorrect assumption about always having the previous block.
This check needs to handle the "edge case" during initial fast sync where we have no previous full block and only have the txhashset state at the horizon.

This is not the first time we have unintentionally broken block sync in this way and is unlikely to be the last.
We _always_ have the previous full block _except_ during fast sync > 1 week.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -342,8 +342,12 @@ impl Chain {
 
 	// Check if the provided block is an orphan.
 	// If block is an orphan add it to our orphan block pool for deferred processing.
+	// If this is the "next" block immediately following current head then not an orphan.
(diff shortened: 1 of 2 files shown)


antiochp:
Going to merge as `master` is currently broken for folks syncing from scratch.

PR #3424, opened 2020-08-16
verify_cut_through and test coverage

antiochp:
This PR makes the "cut-through" logic consistent across block validation "in isolation", aligning it with the _implicit_ "cut-through" rules applied later in the block processing pipeline.
This is related to a consensus rule but we have confirmed it does not impact existing consensus rules (see below).

This is a relatively large PR but vast majority of additional LOC is in new test coverage and greater flexibility in test setup.
The code changes themselves are minimal and limited to early block and transaction validation in `verify_cut_through()`.

----

In the block processing pipeline we `apply_block()` by iterating over the block outputs calling `apply_output()`

https://github.com/mimblewimble/grin/blob/01a300e68bd7efdccf513a813864e42bd4c9786f/chain/src/txhashset/txhashset.rs#L1125-L1127

Note that we compare the output _commitment_ and we do not allow duplicates by commitment.
We process block outputs _before_ block inputs - adding new outputs to the UTXO set before removing spent outputs.
So a block is not valid if it attempts to add an output to the UTXO set if an unspent output with that commitment already exists.

Rule 1: It is invalid to spend `C` to produce a new output `C'` if they share the same _commitment_.

When a node initially receives a block we validate it "in isolation". This happens outside of the context of the UTXO. The block is validated to ensure it is internally consistent. As part of this we verify the block is fully "cut-through", that it does not spend its own outputs.

https://github.com/mimblewimble/grin/blob/01a300e68bd7efdccf513a813864e42bd4c9786f/core/src/core/transaction.rs#L951-L970

Rule 2: It is invalid to spend `C` to produce a new output `C'` if they share the same _hash_.

Note that in both rule (1) and (2) we are verifying that the block is fully cut-through. But there is inconsistency in the way these rules are applied.

It is possible for two outputs to share the same commitment but differ in hash as we include output "features" in the data to be hashed.

https://github.com/mimblewimble/grin/blob/01a300e68bd7efdccf513a813864e42bd4c9786f/core/src/core/transaction.rs#L1472-L1478

https://github.com/mimblewimble/grin/blob/01a300e68bd7efdccf513a813864e42bd4c9786f/core/src/core/transaction.rs#L1725-L1731

We have two output features - 
* Plain
* Coinbase

So there is a form of transaction where an output of one feature variant is spent and could be replaced by a new output with the same commitment but different feature variant. This would involve spending a coinbase output and producing a plain output. In this way it would be possible to produce a block containing `C -> C'` where `C` is coinbase output and `C'` plain output, both sharing the same commitment.

The result of this inconsistency is that a block can be seen as valid (no cut-through) when validated internally, but then subsequently seen as invalid (due to cut-through) later in the block processing pipeline.

----

### Does this result in a consensus issue currently?

We have confirmed this inconsistency _does not and cannot_ lead to a consensus problem. The combination of the two cut-through checks (one explicit, one later implicit) ensures that a valid block will always be treated as valid and an invalid block always treated as invalid. We are letting invalid blocks through the pipeline further than they need to go, and catching them later in the process.

### Does fixing this risk introducing a consensus issue?

We want to make both cut-through checks consistent. A block should be caught as invalid early in the validation process. When we check for cut-through by comparing the block inputs and outputs internally we should apply the correct cut-through rules (comparing by commitment). The concern here is that we would be changing consensus sensitive code.
By making the earlier cut-through validation rule consistent with those applied later implicitly (when applying outputs to the UTXO) we are not changing the consensus rule in any way, we are simply catching an invalid block earlier.

----

The fix is relatively simple. When validating the block for cut-through we want to compare the inputs and outputs by commitment only. 

https://github.com/mimblewimble/grin/blob/6648904899bf88ca9dd07a30e17de91a8c53c442/core/src/core/transaction.rs#L990-L998

When comparing by hash we could take advantage of the enforced sort order of both inputs and outputs as they are both sorted by hash (verified during block validation). We took advantage of this to verify cut-through without needing to allocate. This is no longer possible as we need to resort the inputs and outputs by commitment only. The chosen implementation requires allocation due to this resorting but is significantly simpler and clearer as a result.

----

The remainder of this PR is test coverage for block and transaction validation.
We now have tests covering cut-through during early block and transaction validation and later during the block processing pipeline.

One area of complexity that arose during these tests is the difficulty in actually creating an invalid block without necessarily identifying it as invalid. We want to be able to create an invalid block (with cut-through) and have it accepted into the block processing pipeline. To do this it requires a valid block header.
As part of this we needed to introduce a "readonly header extension" to allow the `prev_root` on the block header to be set correctly, even in the presence of invalid inputs and outputs in the block body itself.

https://github.com/mimblewimble/grin/blob/6648904899bf88ca9dd07a30e17de91a8c53c442/chain/src/chain.rs#L692-L706

### DIFF
--- a/chain/tests/process_block_cut_through.rs
+++ b/chain/tests/process_block_cut_through.rs
@@ -1,0 +1,176 @@
+// Copyright 2020 The Grin Developers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod chain_test_helper;
+
+use grin_chain as chain;
+use grin_core as core;
+use grin_keychain as keychain;
+use grin_util as util;
+
+use self::chain_test_helper::{clean_output_dir, genesis_block, init_chain};
+use crate::chain::{pipe, Chain, Options};
+use crate::core::core::verifier_cache::LruVerifierCache;
+use crate::core::core::{block, transaction};
+use crate::core::core::{Block, KernelFeatures, Transaction, Weighting};
+use crate::core::libtx::{build, reward, ProofBuilder};
+use crate::core::{consensus, global, pow};
+use crate::keychain::{ExtKeychain, ExtKeychainPath, Keychain, SwitchCommitmentType};
+use crate::util::RwLock;
+use chrono::Duration;
+use std::sync::Arc;
+
+fn build_block<K>(
+	chain: &Chain,
+	keychain: &K,
+	txs: &[Transaction],
+	skip_roots: bool,
+) -> Result<Block, chain::Error>
+where
+	K: Keychain,
+{
+	let prev = chain.head_header().unwrap();
+	let next_height = prev.height + 1;
+	let next_header_info = consensus::next_difficulty(1, chain.difficulty_iter()?);
+	let fee = txs.iter().map(|x| x.fee()).sum();
+	let key_id = ExtKeychainPath::new(1, next_height as u32, 0, 0, 0).to_identifier();
+	let reward =
+		reward::output(keychain, &ProofBuilder::new(keychain), &key_id, fee, false).unwrap();
+
+	let mut block = Block::new(&prev, txs, next_header_info.clone().difficulty, reward)?;
+
+	block.header.timestamp = prev.timestamp + Duration::seconds(60);
+	block.header.pow.secondary_scaling = next_header_info.secondary_scaling;
+
+	// If we are skipping roots then just set the header prev_root and skip the other MMR roots.
+	// This allows us to build a header for an "invalid" block by ignoring outputs and kernels.
+	if skip_roots {
+		chain.set_prev_root_only(&mut block.header)?;
+	} else {
+		chain.set_txhashset_roots(&mut block)?;
+	}
+
+	let edge_bits = global::min_edge_bits();
+	block.header.pow.proof.edge_bits = edge_bits;
+	pow::pow_size(
+		&mut block.header,
+		next_header_info.difficulty,
+		global::proofsize(),
+		edge_bits,
+	)
+	.unwrap();
+
+	Ok(block)
+}
+
+#[test]
+fn process_block_cut_through() -> Result<(), chain::Error> {
+	let chain_dir = ".grin.cut_through";
+	global::set_local_chain_type(global::ChainTypes::AutomatedTesting);
+	util::init_test_logger();
+	clean_output_dir(chain_dir);
+
+	let keychain = ExtKeychain::from_random_seed(false)?;
+	let pb = ProofBuilder::new(&keychain);
+	let genesis = genesis_block(&keychain);
+	let chain = init_chain(chain_dir, genesis.clone());
+
+	// Mine a few empty blocks.
+	for _ in 1..6 {
+		let block = build_block(&chain, &keychain, &[], false)?;
+		chain.process_block(block, Options::MINE)?;
+	}
+
+	let key_id1 = ExtKeychainPath::new(1, 1, 0, 0, 0).to_identifier();
+	let key_id2 = ExtKeychainPath::new(1, 2, 0, 0, 0).to_identifier();
+	let key_id3 = ExtKeychainPath::new(1, 3, 0, 0, 0).to_identifier();
+
+	// Build a tx that spends a couple of early coinbase outputs and produces some new outputs.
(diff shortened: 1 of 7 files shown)


antiochp:
I'd like @tromp to give this a good look over before we merge this one.

tromp:
So currently a block orders inputs and outputs by hash. Would it make more sense to order them by commitment instead? This would be a consensus breaking change, that would simplify the code:
1) it would no longer need to compute hashes to check the order.
2) it would not need to sort again to check uniqueness.

core/tests/block.rs
```diff
@@ -784,3 +784,131 @@ fn validate_header_proof() {
...
+
+	// The block should fail validation due to cut-through.
+	assert_eq!(
+		block.validate(&BlindingFactor::zero(), verifier_cache()),
+		Err(Error::Transaction(transaction::Error::CutThrough))
+	);
+
+	// The block should fail lightweight "read" validation due to cut-through.
+	assert_eq!(
+		block.validate_read(),
```
tromp, line 829: how does validate_read call verify_cut_through ?

core/tests/block.rs
```diff
@@ -784,3 +784,131 @@ fn validate_header_proof() {
...
+	.expect("valid tx");
+
+	let prev = BlockHeader::default();
+	let key_id = ExtKeychain::derive_key_id(0, 0, 0, 0, 0);
+	let mut block = new_block(&[tx], &keychain, &builder, &prev, &key_id);
+
+	// The block should fail validation due to cut-through.
+	assert_eq!(
+		block.validate(&BlindingFactor::zero(), verifier_cache()),
+		Err(Error::Transaction(transaction::Error::CutThrough))
```
tromp, line 824: Does block.validate() by any chance also call block.validate_read() ?
For this test to be meaningful, it should not...

core/tests/block.rs
```diff
@@ -784,3 +784,131 @@ fn validate_header_proof() {
 	)
 	.is_err());
 }
+
+// Test coverage for verifying cut-through during block validation.
+// It is not valid for a block to spend an output and produce a new output with the same commitment.
+// This test covers the case where a plain output is spent, producing a plain output with the same commitment.
+#[test]
```
tromp, line 791: This test seems to duplicate the test in tests/transaction.rs ?!

core/tests/block.rs
```diff
@@ -784,3 +784,131 @@ fn validate_header_proof() {
...
+	// Block validates via lightweight "read" validation.
+	block.validate_read()?;
+
+	Ok(())
+}
+
+// Test coverage for verifying cut-through during block validation.
+// It is not valid for a block to spend an output and produce a new output with the same commitment.
+// This test covers the case where a coinbase output is spent, producing a plain output with the same commitment.
+#[test]
```
tromp, line 855: This test seems to duplicate the test in tests/transaction.rs ?!

core/tests/block.rs
antiochp: Starts off the same but tests block validity by building a block from the transaction.

chain/tests/process_block_cut_through.rs
```diff
@@ -0,0 +1,176 @@
...
+	// Transaction is invalid due to cut-through.
+	assert_eq!(
+		tx.validate(Weighting::AsTransaction, verifier_cache.clone()),
+		Err(transaction::Error::CutThrough),
+	);
+
+	// Transaction will not validate against the chain (utxo).
+	assert_eq!(
+		chain.validate_tx(&tx).map_err(|e| e.kind()),
+		Err(chain::ErrorKind::DuplicateCommitment(commit)),
```
tromp, line 136: This appears to be the only test running into a DuplicateCommitment.
It would be good to know the nested functions leading to this in current block processing: i.e.
pipe::process_block
pipe::validate_utxo (has wrong comment /// Validate block inputs against utxo.)
utxo_view::validate_block
utxo_view::validate_output

core/tests/block.rs
antiochp: Same as above. Starts off with a tx but then builds a block from it and validates the block itself.

chain/tests/process_block_cut_through.rs
antiochp: > pipe::validate_utxo (has wrong comment /// Validate block inputs against utxo.)
Fixed comment. Thanks for catching that.

core/tests/block.rs
tromp: ah, yes. would it make sense to avoid the redundancy by having both tests calling a single helper function to build this tx?

antiochp:
> So currently a block orders inputs and outputs by hash. Would it make more sense to order them by commitment instead? This would be a consensus breaking change, that would simplify the code:
> 
> 1. it would no longer need to compute hashes to check the order.
> 2. it would not need to sort again to check uniqueness.
>    Actually, that's not true. The block has inputs sorted and separately outputs sorted.
>    But that means a simple merge suffices to get all sorted.

If inputs and outputs were sorted consistently (by commitment as you suggest) we could use the existing "zero allocation, peekable iterator" approach.

I think it is something to definitely consider. 
The consensus breaking change aspect of this makes me hesitate, particularly as it affects both consensus via the MMR insertion order _and_ the p2p network message format simultaneously.

We should investigate this and explore further as a separate issue/PR. If we decide to do this then yes we can definitely simplify some of the implementation details here.

PR #3426, opened 2020-08-17
Update dockerfiles with rust 1.45 and Debian Buster

quentinlesceller:
Dockerfile was not working.
- Rust 1.35 -> 1.45
- Debian 9.4 -> 10


quentinlesceller:
Tested locally and works. Merging.

PR #3428, opened 2020-08-20
[DNM] Refactor "leaf_set"

antiochp:
**Do not merge prior to 5.0.0 (regardless of branch naming).**

----

We currently update the leaf_set file in realtime as blocks are accepted.
This is the _only_ file that we write to disk during block processing that is not append-only.

This PR is an exploration into removing this dependency and keeping the leafset bitmap purely in memory.
This would be backed by the existing output_pos index from the db.
The index is handled transactionally during block processing.

The leaf_set represents the list of output pos that make up the current UTXO.

Things we still need the leaf_set file for - 
* fast sync snapshot to provide a leaf_set to a peer
* fast sync receiving a leaf_set from a peer 

If the MMR files are strictly append-only then we can potentially - 
* never need to fsync them, yet
* always recover from "corruption" by simple truncation to earlier state


yeastplume:

PR #3431, opened 2020-09-02
Rename Floonet to Testnet

### LINKED ISSUE #3429: Consider: rename floonet -> testnet
`floonet` is a great name.

Its just not _necessarily_ a great name for `testnet` and it would amazing if we could repurpose the name for something more related to communication channels.

Right now `floonet` and `testnet` are synonymous and effectively interchangeable (and confusing).

Its going to take a little bit of investigation to understand what the fallout and impact would be around a name change like this in terms of how invasive it would be to the codebase, config files etc.

quentinlesceller:
Fixes #3429.

This should also be compatible with grin-wallet too.


quentinlesceller:
This works as is

antiochp:
Awesome!
I am planning to test this out over the next couple of days.

etc/Dockerfile.floonet
```diff
@@ -37,13 +37,13 @@ COPY --from=builder /usr/src/grin/target/release/grin /usr/local/bin/grin
 
 WORKDIR /root/.grin
 
-RUN grin --floonet server config && \
+RUN grin --testnet server config && \
```
antiochp, line 40: File itself needs renaming here I think as we now reference it via the new `testnet` suffix.

core/tests/consensus_floonet.rs
```diff
@@ -12,17 +12,17 @@
...
-	secondary_pow_ratio, valid_header_version, FLOONET_FIRST_HARD_FORK, FLOONET_SECOND_HARD_FORK,
-	FLOONET_THIRD_HARD_FORK, HARD_FORK_INTERVAL,
+	secondary_pow_ratio, valid_header_version, HARD_FORK_INTERVAL, TESTNET_FIRST_HARD_FORK,
+	TESTNET_SECOND_HARD_FORK, TESTNET_THIRD_HARD_FORK,
 };
 use grin_core::core::HeaderVersion;
 use grin_core::global;
 
 #[test]
 fn test_secondary_pow_ratio() {
```
antiochp, line 22: File needs renaming.

keychain/src/keychain.rs
```diff
@@ -45,8 +45,8 @@ impl ExtKeychain {
 }
 
 impl Keychain for ExtKeychain {
-	fn from_seed(seed: &[u8], is_floo: bool) -> Result<ExtKeychain, Error> {
-		let mut h = BIP32GrinHasher::new(is_floo);
+	fn from_seed(seed: &[u8], is_test: bool) -> Result<ExtKeychain, Error> {
```
antiochp, line 48: I have a WIP PR somewhere that attempts to cleanup passing around `is_floo` (now `is_test`). I'll see if I can resurrect it. I think we want to explicitly pass a "chain type" around rather than a simple bool like this.

antiochp (commented):
Couple of minor comments. Couple of files should be renamed.

I want to spend a bit of time today testing this locally to make sure we generate fresh config and read existing config correctly etc.

keychain/src/keychain.rs
quentinlesceller: Sure, I leave it like that for now.

quentinlesceller:
Thanks for the review. Addressed the file renaming issue.

> I want to spend a bit of time today testing this locally to make sure we generate fresh config and read existing config correctly etc.

Do you think it is worth it to write the migration code for it? Considering it's floonnet/testnet?

antiochp:
> Thanks for the review. Addressed the file renaming issue.
> 
> > I want to spend a bit of time today testing this locally to make sure we generate fresh config and read existing config correctly etc.
> 
> Do you think it is worth it to write the migration code for it? Considering it's floonnet/testnet?

No I think its fine if we just need to recreate config etc. I just want to make sure there are no surprises in there.

antiochp:
Regenerated `--testnet` config file and resyncing now. Everything looks good to me.

antiochp (approved):
:+1:

quentinlesceller:
Okay LGTM

PR #3433, opened 2020-09-07
Refactor p2p reader

jaspervdm:
This PR refactors the reader of the node p2p by introducing a codec that keeps track of the read state and emits the received p2p messages. The codec only interprets the message headers, the body stay as raw `Bytes`, leaving it to the `Protocol` struct to actually interpret them, as is the case now.

Some advantages of the refactored code:
- Clearer seperation between layers, we don't have to pass the tcp stream around between the reader and the part that actually handles the messages and interfaces with the chain (the `Protocol` struct)
- Preparation for a follow up PR: instead of each peer having a handle to the chain and updating the chain in place, introduce a dedicated thread to handle messages and a channel in between it and the peer readers. This has multiple advantages, one of them being we have less places to lock the chain, since updating the chain state is inherently sequential anyway. 
- Work towards async p2p: the codec will implement the `tokio_util::codec::Framed` trait, which simplifies reading and writing of async messages (future PR)
- Minimize allocations by re-using the same buffer

Unfortunately afaik we currently do not have a framework to test each p2p message and interaction individually, so this PR has been tested on mainnet by syncing from scratch and keeping it running for a while.


p2p/src/codec.rs
```diff
@@ -0,0 +1,137 @@
...
+
+	fn is_none(&self) -> bool {
+		match self {
+			State::None => true,
+			_ => false,
+		}
+	}
+}
+
+pub enum Output {
```
antiochp, line 34: This jumped out to me as confusing due to `Output` being heavily used elsewhere in grin codebase.
Maybe we can rename this to something less generic?

antiochp:
This looks great! :+1: on the overall approach. This makes a ton of sense.
I need to take some time to review and do some testing though.

antiochp:
Ran a full sync locally and everything went smoothly. :+1:

p2p/src/protocol.rs
```diff
@@ -51,31 +49,65 @@ impl Protocol {
...
 		// banned peers up correctly?
 		if adapter.is_banned(self.peer_info.addr) {
 			debug!(
-				"handler: consume: peer {:?} banned, received: {:?}, dropping.",
-				self.peer_info.addr, msg.header.msg_type,
+				"handler: consume: peer {:?} banned, received: {}, dropping.",
+				self.peer_info.addr, input,
 			);
-			return Ok(None);
+			return Ok(Consumed::Disconnect);
```
antiochp, line 63: I like how this is now explicit.

p2p/src/codec.rs
jaspervdm: You're right, I will rename it. Naming structs isn't my strong suit, I don't really like the `Consume` and `Consumed` names either but couldn't come up with better ones

core/src/ser.rs
```diff
@@ -561,6 +568,109 @@ impl<'a> Reader for StreamingReader<'a> {
 	}
 }
 
+/// Protocol version-aware wrapper around a `Buf` impl
```
antiochp, line 571: Does this actually need to be "protocol version aware"? I suspect we can get away with simplifying this.

jaspervdm:
Yes, `Consume` and `Output` are only subtly different. I'm trying to think if we can merge them somehow but it might make things a bit awkward.
And I think with some additional refactoring we can get rid of the `MsgHeaderWrapper`, as it is now only used during the handshake. I will have a look and update the PR.

core/src/ser.rs
jaspervdm: The reason for this is that the `Reader` trait has a method to access the protocol version. It is used in the protocol handler to convert from raw bytes to the structs, as deserializing these structs can be protocol version dependent.

antiochp: Ahh right. That was my fault... :+1:

quentinlesceller:
Could you merge upstream/master into your PR?
Had a quick look and everything looks fine. 
Also ran it from existing db and sync from scratch without any issues.

jaspervdm:
@quentinlesceller Yep will do! Thanks for testing.
@antiochp One other thing I was considering is this: should we even deal with deserializing in the `Protocol` handler? this part is concerned with reading/writing to the chain so maybe we should do the actual deserialization closer to the reader.

antiochp:
> we would have one variant for each type of message

So one for each known msg type and one unknown case?
I think that makes a lot of sense - might need to see what it looks like in practice to really know but :+1: on the idea.

jaspervdm:
> So one for each known msg type and one unknown case?

Yes, that is basically what I have done in the latest commit. It also allowed us to get rid of one of the enums.
We can also further simplify the code by getting rid of the `Type` and `MsgHeaderWrapper` enums and using a macro for `decode_message` but I think I will leave that to a future PR

antiochp:
Testing locally. Looks good so far.

core/src/core/block.rs
```diff
@@ -423,6 +423,22 @@ impl From<UntrustedBlockHeader> for BlockHeader {
 	}
 }
 
+/// List of block headers
+pub struct BlockHeaders(pub Vec<BlockHeader>);
+
+/// Deserialization of a list of block headers
+impl Readable for BlockHeaders {
```
antiochp, line 430: We used to "stream" headers in chunks of 32. 
`BlockHeaders` with a single `read` means this is no longer the case?

antiochp:
I'm planning to cut a `4.1.0` off master in the next day or so. This is so we have an official release with "commit only inputs".

Do you mind if we hold off on merging this PR until after that?

p2p/src/protocol.rs
```diff
@@ -50,223 +45,188 @@ impl Protocol {
...
-				let (count, bytes_read): (u16, _) = msg.streaming_read()?;
-				total_bytes_read += bytes_read;
-
+			Message::Headers(hs) => {
 				// Read chunks of headers off the stream and pass them off to the adapter.
-				let chunk_size = 32u16;
-				let mut headers = Vec::with_capacity(chunk_size as usize);
-				for i in 1..=count {
-					let (header, bytes_read) =
-						msg.streaming_read::<core::UntrustedBlockHeader>()?;
```
antiochp, line 235: Related to my previous comment - here we used to actually read headers of the bytes from the body in chunks of 32 headers.
We now (I think?) read and parse _all_ headers up front and _then_ process them in chunks of 32.

antiochp: Related - looks like our 32 header chunks are acting weird here also (first chunk is a single header...) - 

```
20200914 18:07:40.223 INFO grin_servers::common::adapters - Received 1 block headers from 47.90.208.51:3414
20200914 18:07:40.251 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.304 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.355 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.410 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.461 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.513 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.568 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.620 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.674 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.726 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.778 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.830 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.881 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.928 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:40.980 INFO grin_servers::common::adapters - Received 32 block headers from 47.90.208.51:3414
20200914 18:07:41.036 INFO grin_servers::common::adapters - Received 31 block headers from 47.90.208.51:3414
```

antiochp: Maybe we don't actually care about the "streaming" approach. 
I'm not entirely sure how much benefit it gave us for a non-trivial amount of additional complexity.

antiochp:
Actually I'm guessing we want to preserve the "streaming read" behavior for chunks of headers. 
This is likely to be useful generally once we have various PIBD related "large" messages (chunks of kernel MMR etc.) We are going to want to be able to start processing chunks of data before we have fully received the msg in these cases.

jaspervdm:
> Do you mind if we hold off on merging this PR until after that?

Sure, no problem.
> We now (I think?) read and parse all headers up front and then process them in chunks of 32.

Yes, correct.
> Actually I'm guessing we want to preserve the "streaming read" behavior for chunks of headers.

I agree. I will look into this. Since the codec is already holding state and we are decoding the messages in there, this shouldn't be too hard to do.

antiochp:
We know the overall size in bytes of the full 512 header payload.
But we do not know the size of each individual header.

The streaming reader allowed us to call `read_item()` repeatedly for each individual header, parsing them as soon as we had enough bytes to read.

Looking at this now it is conceptually close to an `Iterator`, just not defined in those terms.

You mention `Framed` here in the PR for async processing.
https://docs.rs/tokio-util/0.3.1/tokio_util/codec/struct.Framed.html

The `frame` here is something at a smaller granularity than the full msg body. A chunk of `n` headers in this specific case.

jaspervdm:
@antiochp Ok with the latest commit we now read the headers as they come in, and return them in batches of 32. This is done by overestimating the header size and performing deserialization when the buffer is filled. I also added some tests to prove that the header size calculation is actually correct.

There was a nasty bug in there that took me a long time to hunt down, but I managed to fix it. The problem was that if a read failed, the reserved part of the buffer was kept around. This messed up the next read attempt.

p2p/src/codec.rs
```diff
@@ -0,0 +1,244 @@
...
+	fn set_stream_timeout(&self) -> Result<(), Error> {
+		let timeout = match &self.state {
+			None => HEADER_IO_TIMEOUT,
+			_ => BODY_IO_TIMEOUT,
+		};
+		self.stream.set_read_timeout(Some(timeout))?;
+		Ok(())
+	}
+
+	fn read_inner(&mut self) -> Result<Message, Error> {
```
antiochp, line 101: Oh interesting. So `read_inner()` will actually return multiple `Message` instances here (each a batch of 32 headers) if we receive is a `Type::Headers`.
Whereas in most cases it only makes sense to call `read_inner()` once to read the full message?

p2p/src/codec.rs
```diff
@@ -0,0 +1,244 @@
...
+							let bytes_left = *bytes_left;
+							self.state = None;
+							if bytes_left > 0 {
+								return Err(Error::BadMessage);
+							}
+						}
+						return Ok(Message::Headers(h));
+					}
+				}
+				Attachment(left, meta, now) => {
```
antiochp, line 186: And in contrast with headers, reading an attachment returns a _single_ `Message`, basically a running total of the download? Each time with potentially more of the attachment read? 
Subsequent blocking calls to `read_inner()` may result in more of the attachment being read.

p2p/src/codec.rs
```diff
@@ -0,0 +1,244 @@
...
+					let update = AttachmentUpdate {
+						read: next_len,
+						left: *left,
+						meta: Arc::clone(meta),
+					};
+					if *left == 0 {
+						self.state = None;
+						debug!("attachment: DONE");
+					}
+					return Ok(Message::Attachment(update, Some(raw)));
```
antiochp, line 202: Does `raw` here represent _everything_ read for this attachment so far? Or just the part read in this particular call to `read_inner`?

p2p/src/codec.rs
```diff
@@ -0,0 +1,244 @@
...
+			None => MsgHeader::LEN,
+			Header(Known(h)) if h.msg_type == Type::Headers => {
+				// If we are receiving a list of headers, read off the item count first
+				min(h.msg_len as usize, 2)
+			}
+			Header(Known(header)) => header.msg_len as usize,
+			Header(Unknown(len, _)) => *len as usize,
+			BlockHeaders { bytes_left, .. } => {
+				// The header length varies with the number of edge bits. Therefore we overestimate
+				// its size and only actually read the bytes we need
```
antiochp, line 84: Does this mean we potentially _wait_ for (slightly) too many bytes to arrive before attempting to parse a batch of 32 headers? While still only subsequently reading the exact bytes required.
The last batch is handled by tracking `bytes_left`.

This makes a lot of sense - I think this is a far better way to decouple the blocking code.

p2p/src/codec.rs
```diff
@@ -0,0 +1,244 @@
...
+						// Return full message
+						let msg = decode_message(header, &mut raw, self.version);
+						self.state = None;
+						return msg;
+					}
+				}
+				Header(Unknown(_, msg_type)) => {
+					// Discard body and return
+					let msg_type = *msg_type;
+					self.buffer.advance(next_len);
```
antiochp, line 151: This just discards those bytes by advancing over them? :+1:

p2p/src/codec.rs
jaspervdm: Correct!

p2p/src/codec.rs
jaspervdm: An attachment will generate a lot of `Message::Attachment`s, one for every 48000 bytes that is read. This is the same size iirc as the buffer we were filling before.

p2p/src/codec.rs
jaspervdm: `raw` is a chunk of 48000 bytes. Storing the full attachment in a buffer would be very inefficient, as we would have to allocate a lot of memory.

p2p/src/codec.rs
jaspervdm: Yes we wait for a couple of bytes too many, but since we do it header by header the difference is not that big. On floonet/mainnet a header with C32 takes 415 bytes, while we wait for 578 bytes (the header size with C63, the maximum allowed graph size).

p2p/src/codec.rs
jaspervdm: Yes, `Bytes`/`BytesMut` has an internal cursor, this just increases it by `next_len` bytes. Since this is the only place that points to this part of the buffer (guaranteed because it is a `BytesMut`), that part gets marked as dropped.

Splitting the data off like we do when we read the body ([here](https://github.com/mimblewimble/grin/pull/3433/files#diff-194510b22fabb1d5607a8bca1200a847R130)) advances the cursor similarly, but also increases a refcount and returns a `Bytes` for that part of the buffer. We don't actually care about the content of the body, so we just advance the cursor which is a bit faster.

p2p/src/codec.rs
antiochp: I see it now. And then we write these bytes to the file itself in the main `reader_thread` loop.

antiochp (approved):
:+1:

Let's do it!

Issue #3430, opened 2020-08-30
P2P Protocol Version for HF4


antiochp:
Nodes are currently on p2p protocol version 2.

Version 1 "fixed size kernels".
Version 2: "variable sized kernels".

We are proposing bumping to version 3: "commit only inputs".
See #3419 for details on this.

https://github.com/mimblewimble/docs/wiki/P2P-Protocol#phasing-out-old-peers

In line with our protocol version major/minor versioning strategy we should consider bumping to version `1000` as part of our final scheduled HF, or potentially as part of 4.x.0 and a final bump to version `2000` at HF4.

It would be convenient if we could take advantage of the final scheduled HF to lock the supported protocol version in.
This would allow for legacy code to be cleaned up after HF4 with some confidence that old nodes will no longer need to be supported.

Post HF4 we will have no need to support either legacy "fixed size kernels" or legacy "features and commit inputs".

antiochp:
cc @DavidBurkett and @davidtavarez

DavidBurkett:
Our latest Grin++ versions are already doing variable-sized kernels, and I'm comfortable committing to supporting the new protocol version 3 as well.

It seems unnecessary to force such a rapid jump in protocol versions to kill old peers. I expected those kinds of major jumps to be saved for cases where there's an extraordinary amount of code for supporting the old versions or a noticeable decrease in network chattiness or performance increase as a result of an upgrade. If we do decide to increase the protocol version by 1000 anyway, we'll keep Grin++ consistent.

antiochp:
> It seems unnecessary to force such a rapid jump in protocol versions to kill old peers.

We could alternatively just implement an HF4 specific rule that peers are incompatible with older protocol versions, but it makes sense to take advantage of the existing rules I think.

PR #3434, opened 2020-09-07
bump protocol version to 1000 to facilitate version deprecation

antiochp:
Related #3430. 

This PR introduces a "major" protocol version bump to __1000__.

https://github.com/mimblewimble/docs/wiki/P2P-Protocol#phasing-out-old-peers

> Peers with a protocol version in the range [1000-1999] should be able to interact with any peer in the range [0-2999].

This maintains full support with all prior protocol versions.
In effect we deprecate (but still continue to support) versions __0-1000__ inclusive.

As part of HF4 the proposal is to introduce an additional "major" bump to __2000__.

> Peers with a protocol version in the range [2000-2999] should be able to interact with any peer in the range [1000-3999].

This would cleanly and intentionally remove support for old deprecated versions (<1000) allowing implementation internals to be cleaned up post HF4.

### DIFF
--- a/core/src/global.rs
+++ b/core/src/global.rs
@@ -45,7 +45,7 @@ use util::OneTime;
 /// Note: We also use a specific (possible different) protocol version
 /// for both the backend database and MMR data files.
 /// This defines the p2p layer protocol version for this node.
-pub const PROTOCOL_VERSION: ProtocolVersion = ProtocolVersion(3);
+pub const PROTOCOL_VERSION: ProtocolVersion = ProtocolVersion(1_000);
 
 /// Automated testing edge_bits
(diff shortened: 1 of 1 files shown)


quentinlesceller (commented):
I've lookup at the code and I don't see where this logic actually applies. Might have miss something. Can you point it out for me?

antiochp:
> I've lookup at the code and I don't see where this logic actually applies. Might have miss something. Can you point it out for me?

That's a good point - this code does not yet actually exist. 
We currently assume we support _all_ previous protocol versions.

antiochp:
I will open a separate PR to actually implement the deprecation logic...

PR #3435, opened 2020-09-07
maintain preferred peer connections

antiochp:
* If we specify `preferred_peers` in config then attempt to maintain these when cleaning excess peers.
* Pass `preferred_peers` into `clean_peers()` to facilitate this.
* General cleanup passing a slice of `PeerAddr` around rather than `Option<Vec<PeerAddr>>`.

Tested on `mainnet` and "preferred peers" works a lot more robustly now...


quentinlesceller (approved):
:+1: Looking good

PR #3436, opened 2020-09-08
Organize doc files

Paouky:
- Add `doc/translations` directory and move there all translation markdown files that were previously held in `doc`. Makes navigation easier on the eyes.
- Correct all translation links to suit the changes.
- Add Mimblewimble whitepaper to `doc` for future reference (this is the real motivation for the PR).
- Transfer [p2p_protocol](https://github.com/mimblewimble/docs/wiki/P2P-Protocol) and [validation_logic](https://github.com/mimblewimble/docs/wiki/Validation-logic) from regular docs wiki to here.


quentinlesceller (approved):
Looking good 👍 Much cleaner indeed. 
LGTM. Shall we remove P2P Protocol and Validation Logic from the wiki then?

Also what are we going to do with all these documents now that the docs are coming?

Paouky:
> LGTM. Shall we remove P2P Protocol and Validation Logic from the wiki then?

Yep we can remove the wiki documents, along with a bunch of others.

> Also what are we going to do with all these documents now that the docs are coming?

Leave them be, in my opinion. They're not really being maintained in any way, but still might be useful for future reference or perhaps for when some dev decides to do an overhaul. Some of them are in the new docs (in a slightly more polished form) and some were deemed too technical for it.

PR #3439, opened 2020-09-09
cleanup mining stats

bladedoyle:
Cleanup mining stats

1) make edge_bits dynamic (last size submitted)
2) dont show bogus stats when no miners are connected
3) move network hashrate calculation into stratum server and only call it when miners are connected, and pass the value through StratumStats object (like all the rest)
4) update a few comments

The two user visible (non-consensus) changes are:

Previously:  bogus default, or outdated stats would be shown when no miners are connected.
Now: status shows "n/a" for those stats when no miners are connected

Previously:  only stats for C31 are shown, which is no longer even mined
Now: stats are shown for the size of the last share submitted - with C32 by default (if no shares are submitted yet)

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -208,6 +208,10 @@ pub const AR_SCALE_DAMP_FACTOR: u64 = 13;
 pub fn graph_weight(height: u64, edge_bits: u8) -> u64 {
 	let mut xpr_edge_bits = edge_bits as u64;
 
+	if edge_bits < global::base_edge_bits() {
+		return 0;
+	}
+
 	let expiry_height = YEAR_HEIGHT;
 	if edge_bits == 31 && height >= expiry_height {
 		xpr_edge_bits = xpr_edge_bits.saturating_sub(1 + (height - expiry_height) / WEEK_HEIGHT);
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -208,10 +208,6 @@ pub const AR_SCALE_DAMP_FACTOR: u64 = 13;
 pub fn graph_weight(height: u64, edge_bits: u8) -> u64 {
 	let mut xpr_edge_bits = edge_bits as u64;
 
-	if edge_bits < global::base_edge_bits() {
-		return 0;
-	}
-
 	let expiry_height = YEAR_HEIGHT;
 	if edge_bits == 31 && height >= expiry_height {
 		xpr_edge_bits = xpr_edge_bits.saturating_sub(1 + (height - expiry_height) / WEEK_HEIGHT);
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -267,7 +267,7 @@ impl Default for StratumStats {
 			num_workers: 0,
 			block_height: 0,
 			network_difficulty: 0,
-			edge_bits: 32,
+			edge_bits: 0,
 			network_hashrate: 0.0,
 			worker_stats: Vec::new(),
 		}
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -127,7 +127,7 @@ pub struct StratumStats {
 	pub block_height: u64,
 	/// current network difficulty we're working on
 	pub network_difficulty: u64,
-	/// cuckoo size of last share sumitted
+	/// cuckoo size of last share submitted
 	pub edge_bits: u16,
 	/// current network Hashrate (for edge_bits)
 	pub network_hashrate: f64,
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -266,7 +266,7 @@ impl Default for StratumStats {
 			is_running: false,
 			num_workers: 0,
 			block_height: 0,
-			network_difficulty: 1000,
+			network_difficulty: 0,
 			edge_bits: 32,
 			network_hashrate: 0.0,
 			worker_stats: Vec::new(),
--- a/servers/src/common/stats.rs
+++ b/servers/src/common/stats.rs
@@ -19,7 +19,6 @@ use crate::util::RwLock;
 use std::sync::Arc;
 use std::time::SystemTime;
 
-use crate::core::consensus::graph_weight;
 use crate::core::core::hash::Hash;
 use crate::core::ser::ProtocolVersion;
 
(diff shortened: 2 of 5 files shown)


core/src/consensus.rs
```diff
@@ -208,6 +208,10 @@ pub const AR_SCALE_DAMP_FACTOR: u64 = 13;
 pub fn graph_weight(height: u64, edge_bits: u8) -> u64 {
 	let mut xpr_edge_bits = edge_bits as u64;
 
+	if edge_bits < global::base_edge_bits() {
+		return 0;
+	}
```
quentinlesceller, line 213: What is this for?

servers/src/mining/stratumserver.rs
```diff
@@ -574,16 +575,21 @@ impl Handler {
 					// set a new deadline for rebuilding with fresh transactions
 					deadline = Utc::now().timestamp() + config.attempt_time_per_block as i64;
 
-					self.workers.update_block_height(new_block.header.height);
-					self.workers
-						.update_network_difficulty(state.current_difficulty);
-
+					// Manage list of versions of this heights block
 					if clear_blocks {
```
quentinlesceller, line 579: A bit out of the scope but this would be probably be nicer to have:
```suggestion
					// If this is a new block, clear the current_block version history
					if current_hash != latest_hash {
```

servers/src/mining/stratumserver.rs
```diff
@@ -574,16 +575,21 @@ impl Handler {
...
 						state.current_block_versions.clear();
 					}
 					state.current_block_versions.push(new_block);
-					// Send this job to all connected workers
+
+					// Update the mining stats
+					self.workers
+						.update_block_height(state.current_block_versions[0].header.height);
```
quentinlesceller, line 586: Why not keep `self.workers.update_block_height(new_block.header.height);`?

servers/src/mining/stratumserver.rs
```diff
@@ -574,16 +575,21 @@ impl Handler {
...
+					let difficulty = state.current_block_versions[0].header.total_difficulty()
+						- head.total_difficulty;
+					self.workers.update_network_difficulty(difficulty.to_num());
```
quentinlesceller, line 589: Same here those two lines can be replaced by:
```suggestion
					self.workers
 						.update_network_difficulty(state.current_difficulty);
```

src/bin/tui/mining.rs
```diff
@@ -337,21 +337,32 @@ impl TUIStatusListener for TUIMiningView {
...
-			stratum_stats.network_difficulty
-		);
-		let stratum_edge_bits = format!("Cuckoo Size:           {}", stratum_stats.edge_bits);
+		let stratum_block_height = match stratum_stats.num_workers {
+			0 => "Solving Block Height:  n/a".to_string(),
+			_ => format!("Solving Block Height:  {}", stratum_stats.block_height),
+		};
+		let stratum_edge_bits = match stratum_stats.num_workers {
+			0 => "Mining POW:            n/a".to_string(),
+			_ => format!("Mining POW:            Cuckoo{}", stratum_stats.edge_bits),
```
quentinlesceller, line 350: Wouldn't it be clearer to have
```suggestion
			_ => format!("Latest Cuckoo Size:           {}", stratum_stats.edge_bits),
```
probably off in term of TUI spaces though.

servers/src/grin/server.rs
```diff
@@ -361,7 +361,6 @@ impl Server {
 
 	/// Start a minimal "stratum" mining service on a separate thread
 	pub fn start_stratum_server(&self, config: StratumServerConfig) {
-		let edge_bits = global::min_edge_bits();
```
quentinlesceller, line 364: I'd rather keep something like `let edge_bits = global::min_edge_bits() +1;` than having an hardcoded value later. What do you think?

quentinlesceller (changes requested):
Few comments. Also don't think it' super clean to have both the `edge_bits` and `network_hashrate` potentially oscillating for every share (i.e. mining c29 and c32 simultaneously). But we need to also consider that c29 has 4 months left.

core/src/consensus.rs
bladedoyle: It adds range checking.

servers/src/mining/stratumserver.rs
bladedoyle: At that point in the code "current_hash" has been updated (see line 571)
`current_hash = latest_hash;`
So the check needs to be done before that.

servers/src/mining/stratumserver.rs
bladedoyle: Sure.  That new_block was just pushed to the current_block_versions[0] so its the same thing.  I'll change it back.

servers/src/mining/stratumserver.rs
bladedoyle: > Few comments. Also don't think it' super clean to have both the `edge_bits` and `network_hashrate` potentially oscillating for every share (i.e. mining c29 and c32 simultaneously). But we need to also consider that c29 has 4 months left.

I think it is possible to mine C33(+?) so we still need some way to show multiple pow size and rate.  Showing data for the last submitted share (oscillating) is better than showing bogus data, IMO.

src/bin/tui/mining.rs
bladedoyle: idk, I dont have a preference (except for keeping TUI formatted).

servers/src/grin/server.rs
bladedoyle: Sure, good suggestion.

servers/src/mining/stratumserver.rs
bladedoyle: Now I recall why - new_block has been "moved" when it gets pushed onto the block version list.   But it seems reasonable to set the stats first then push it onto the list.

core/src/consensus.rs
bladedoyle: I can do the check in the caller if you prefer?

quentinlesceller: I think you can remove it altogether. Correct me if I'm wrong but, we don't need this check in the caller since it's only called for a `valid_share`.

bladedoyle: Its called when building a new block template.  That includes the first new block template before any shares are submitted.
Since we set a default size in run_loop() you are correct that the current code never calls with an invalid value - but maybe it would be best to have the check to futureproof a mistake?   I dont feel strongly so I'll remove the check as suggested.

servers/src/common/stats.rs
```diff
@@ -128,8 +127,10 @@ pub struct StratumStats {
 	pub block_height: u64,
 	/// current network difficulty we're working on
 	pub network_difficulty: u64,
-	/// cuckoo size used for mining
+	/// cuckoo size of last share sumitted
```
quentinlesceller, line 130:
```suggestion
	/// cuckoo size of last share submitted
```

quentinlesceller (approved):
Looking good for me 👍.
@antiochp can you take a look?

antiochp (commented):
This all looks ok to me.
I'm not overly familiar with the stratum related code though so not the best person to :+1: this.

PR #3451, opened 2020-09-25
refactor get_ser() to use get_with() internally

antiochp:
This PR refactors some of the lmdb batch internals.

* `get_with()` deserialization function now fallible
* `get_ser()` uses `get_with()` internally and passes in default deserialization strategy
* removed the awkward `get_ser_access()`
* `batch.exists()` now aware of current write tx and not simply a wrapper to `db.exists()`

Related #3450 (which prompted this refactor).

We use `get_ser()` heavily as this uses the default deserialization strategy. But we also occasionally want more flexible access (see #3450). This PR consolidates `get_with()` and `get_ser()`.

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -359,7 +359,9 @@ impl<'a> Batch<'a> {
 	fn get_legacy_input_bitmap(&self, bh: &Hash) -> Result<Bitmap, Error> {
 		option_to_not_found(
 			self.db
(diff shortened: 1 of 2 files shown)


antiochp:
Going to merge this. It's a refactor, no change to behavior. Tested locally and against mainnet.
The changes to `exists()` are limited as this is only called in a single location during block processing.

PR #3453, opened 2020-09-27
PMMR segment creation and validation

jaspervdm:
This PR implements a (P)MMR "segment", defined by [this WIP RFC](https://github.com/jaspervdm/grin-rfcs/blob/pibd_messages/text/0000-pibd-messages.md). In short, a segment is a set of `2**b` (with variable `b`) consecutive leaves along with the necessary data to reconstruct the subtree root and data to verify membership of the subtree in the original MMR (a Merkle proof). In case of a prunable MMR, the segment only contains the unpruned leaves in the segment range and also contains intermediary MMR hashes that are necessary to construct the segment root.

Concretely, a segment consist of the following elements:
- Segment identifier: `b`, the 2-log of the size of the number of leaves (before pruning) and a zero-based segment index `idx`
- List of intermediary hashes, sorted by MMR position in ascending order. Only for prunable MMRs, it contains the unpruned hashes in the segment range. They are necessary to reconstruct the segment root. As of this PR it contains _all_ unpruned hashes above the leaves, which possibly contains redundant data. This is something we can improve over time.
- List of unpruned leaves in the segment (leaf index range `[i*2**b, (i+1)*2**b)`), sorted by MMR position in ascending order.
- Segment merkle proof, required to reproduce the MMR root starting with the segment root, thereby proving membership.

Given that a segment contains a number of leaves that is a power of 2, a full segment forms a full subtree in the MMR and as such it has a single root. The final segment possibly has less than `2**b` elements. In this case the peaks in the segment are also peaks in the full MMR and we define its root as these peaks bagged together.

**Segment creation**
A segment is created by looping over all the MMR positions in the segment range. If the position is an unpruned leaf or its subling is an unpruned leaf, add it to the leaf data list. If not, check if its hash is unpruned and add it to the list of hashes. Next, generate the merkle proof by filling a list of hashes with:
1. the siblings along the path from the subtree root (final MMR position of the segment) up to its corresponding peak in the MMR
2. peaks to the right of our subtree root, bagged together to a single hash
3. peaks to the left (from right to left) with a position smaller than the first MMR position of the segment

Note that this procedure will also behave as expected for the partially filled final segment, since in that case step and 1 and 2 will not produce any hashes and step 3 will give us all the other peaks in the MMR.

**Segment verification**
Iterate over all the MMR positions in the segment range. If the position is a leaf: check (for prunable MMRs) to see if the element is expected to be there (based on presence of the element or its sibling in the bitmap) and if it is get it from the list of leaves, hash it and store the hash in a list of temporary hashes.
For all other positions: if both children are not present in the list of temporary hashes (i.e. they are both pruned), do nothing. If either or both hashes are present in the list, hash them together and store the new hash in the list of temporary hashes. If only one of the children was present, obtain the hash of the other child from the list of hashes in the segment instead.
After looping through all positions, we are either left with 1 entry (full segment) or multiple entries (partially filled, final segment) in the list of temporary hashes. If there are multiple, bag them together. We are left with the segment root.

Next, verify the proof by attempting to reproduce the MMR root hash: first, hash together with the siblings along the path from the subtree root to its peak in the MMR, then hash it together with the bagged peaks to the right and finally hash it together with the peaks on the left (going right to left). Verification passes if the calculated hash is equal to the MMR root hash.

**Open points before merging**
- ~~In case of a heavily pruned MMR, is this data always sufficient to reconstruct the full pruned MMR? Or would we miss any intermediary hashes?~~ Yes, now that we added support for fully pruned segments.
- ~~I think we need a function to extract the intermediary and proof hashes for purposes of storing them in the MMR we are building~~ TBD in a future PR
- ~~Would it be more natural to pass in a bitmap indicating the _spent_ or the _unspent_ positions?~~ Unspent
- ~~Is bitmap/leaf index 0-based (should be a quick check)~~ Yes, it is.


antiochp:
Quick question that occurred to me reading over the PR description (nice description by the way!) - 

> Segment identifier: b, the 2-log of the size of the number of leaves (before pruning) and a zero-based segment index idx

Does it potentially make more sense to provide a "starting leaf index" as segment identifier? The current "segment index" is heavily dependent on the "segment size". Clearly you can translate between these easily enough but it may make sense to identify these based on both number of leaves via `b` and leaf `idx`.

antiochp:
> List of unpruned leaves in the segment (leaf index range [i*2**b, (i+1)*2**b)), sorted by MMR position in ascending order.

Is the intention to have these segments be self-contained? For pruned MMR (outputs and rangeproofs) do we also provide the bitmap or is the assumption we already have the corresponding bitmap segment?

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+
+	/// Whether the segment is full (size == capacity)
+	#[inline]
+	fn full_segment(&self, last_pos: u64) -> bool {
+		self.segment_unpruned_size(last_pos) == self.segment_capacity()
+	}
+
+	/// Inclusive range of MMR positions for this segment
+	#[inline]
+	fn segment_pos_range(&self, last_pos: u64) -> (u64, u64) {
```
antiochp, line 115: Does it make sense to actually return range here?

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+			Writeable::write(data, writer)?;
+		}
+		Writeable::write(&self.proof, writer)?;
+		Ok(())
+	}
+}
+
+/// Merkle proof of a segment
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct SegmentProof {
```
antiochp, line 374: How much overlap do we have with the existing `MerkleProof` struct here?
Have you considered potentially combining/merging/reusing this?

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+		writer.write_u64(self.idx)
+	}
+}
+
+/// Segment of a PMMR: unpruned leaves and the necessary data to verify
+/// segment membership in the original MMR.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Segment<T> {
+	identifier: SegmentIdentifier,
+	hashes: Vec<(u64, Hash)>,
```
antiochp, line 80: I'd like to consider splitting this out into two separate vecs. One of pos and one of hashes.
On disk we store this data as `[H, H, H, H, ...]` but we need to convert this to `[P, H, P, H, P, H, ...]`.

For maximum flexibility it would be nice to keep this aligned with the on disk storage impl. 
We may want to consider alternative ser/deser strategies in the future where we simply read chunks of bytes from the memory mapped MMR files (both data and hash files) directly, rather than reading individual hashes.

So something along the lines of - 

```
hashes: Vec<Hash>,
hash_pos: Vec<u64>,
leaf_data: Vec<T>,
leaf_pos: Vec<u64>,
```

jaspervdm:
> Does it potentially make more sense to provide a "starting leaf index" as segment identifier? The current "segment index" is heavily dependent on the "segment size". Clearly you can translate between these easily enough but it may make sense to identify these based on both number of leaves via `b` and leaf `idx`.

I had considered this but decided against it, because it would make the proof more complicated. Right now it works under the assumption that a full segment contains elements that fall under a single common root `b` levels up. If we allow an arbitrary start leaf index, this is no longer the case. And if we would only allow `start_leaf_idx % 2**b == 0`, it would imo be more natural to define the index as `start_leaf_idx / 2**b` which is what we do now.

> Is the intention to have these segments be self-contained? For pruned MMR (outputs and rangeproofs) do we also provide the bitmap or is the assumption we already have the corresponding bitmap segment?

The first step in the sync process would be to download all bitmap segments and reconstruct the full bitmap and mmr. Once this is done, we can download the output and rangeproof segments and validate them using the bitmap.

core/src/core/pmmr/segment.rs
jaspervdm: In many of the calculations we actually need the first or last position, but as far as i can see there is no convenient way to obtain them from a [Range](https://doc.rust-lang.org/stable/std/ops/struct.Range.html)

core/src/core/pmmr/segment.rs
jaspervdm: There is some overlap but we have a few extra assumptions in here that change the calculation somewhat. I could try to see if we can somehow extract some common code without cluttering it.

antiochp:
> > Does it potentially make more sense to provide a "starting leaf index" as segment identifier? The current "segment index" is heavily dependent on the "segment size". Clearly you can translate between these easily enough but it may make sense to identify these based on both number of leaves via `b` and leaf `idx`.
> 
> I had considered this but decided against it, because it would make the proof more complicated. Right now it works under the assumption that a full segment contains elements that fall under a single common root `b` levels up. If we allow an arbitrary start leaf index, this is no longer the case. And if we would only allow `start_leaf_idx % 2**b == 0`, it would imo be more natural to define the index as `start_leaf_idx / 2**b` which is what we do now.

Yeah that makes sense. I wondered about the (hypothetical) scenario where we read a segment with a different size than earlier segments. but I guess if we are requesting them with consistent size then this is not relevant.

:+1:

core/src/core/pmmr/segment.rs
jaspervdm: That's fine with me, I will refactor the code accordingly.

jaspervdm:
> Yeah that makes sense. I wondered about the (hypothetical) scenario where we read a segment with a different size than earlier segments. but I guess if we are requesting them with consistent size then this is not relevant.

In these situations it would be fairly easy to calculate the changed index. I could add a `rescale` method to the identifier that does this.

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+			SegmentError::Mismatch => write!(f, "Root hash mismatch"),
+		}
+	}
+}
+
+/// Tuple that defines a segment of a given PMMR
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub struct SegmentIdentifier {
+	/// 2-Log of the size of a segment
+	pub log_size: u8,
```
antiochp, line 55: I think `log_size` is actually the same as our existing `height` terminology -

https://github.com/mimblewimble/grin/blob/cea546ceb8b5be0fc4462489e8a4761bc395fb18/core/src/core/pmmr/pmmr.rs#L540-L548

Would it make sense to use `height` here to avoid introducing the new `log_size` terminology?

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+
+impl<T> Segment<T> {
+	/// Maximum number of leaves in a segment, given by `2**b`
+	#[inline]
+	fn segment_capacity(&self) -> u64 {
+		1 << self.identifier.log_size
+	}
+
+	/// Offset (in leaf idx) of first leaf in the segment
+	#[inline]
```
antiochp, line 93: Do we need `#[inline]` everywhere? I'm not clear on guidelines for explicitly doing this.

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+	pub fn from_pmmr<U, B>(
+		segment_id: SegmentIdentifier,
+		pmmr: &ReadonlyPMMR<'_, U, B>,
+		prunable: bool,
+	) -> Result<Self, SegmentError>
+	where
+		U: PMMRable<E = T>,
+		B: Backend<U>,
+	{
+		let mut segment = Segment {
```
antiochp, line 151: Maybe a `Segment:empty()` here for readability?

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+		Ok(segment)
+	}
+}
+
+impl<T> Segment<T>
+where
+	T: PMMRIndexHashable,
+{
+	/// Calculate root hash of this segment
+	pub fn root(&self, last_pos: u64, bitmap: Option<&Bitmap>) -> Result<Hash, SegmentError> {
```
antiochp, line 199: This root calculation is good for now but it feels like longer term we want/need a more flexible (sub)PMMR interface and actually (re)build a (sub)PMMR up given the provided hashes and leaf data.

The segment is a subtree (potentially multiple smaller subtrees) in a larger PMMR. And the root is the root of this subtree.

If we could start at some leaf offset (rather than pos 1) and simply build a partial/sub PMMR from that point on then we effectively get the root for free.

We do assume we only ever add individual leaves to a PMMR currently. We would need to be able to add parent hashes directly, skipping pruned leaves, in some more flexible way to accomplish this.

I'm not suggesting we do this for this PR, just something for us to think about in the future.

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+						b.contains(idx_1 as u32) || b.contains(idx_2 as u32)
+					})
+					.unwrap_or(true)
+				{
+					// We require the data of this leaf if either the mmr is not prunable or if
+					// the bitmap indicates it (or its sibling) should be here
+					// TODO: possibly remove requirement on the sibling when we no longer support
+					//  syncing through the txhashset.zip method
+					let data = leaves
+						.find(|&&(p, _)| p == pos)
```
antiochp, line 225: Is this not always the next element in leaf_data? Do we need to find it here?
i.e. We just iterate over data as the leaves are already sorted?

antiochp: Similar to how we pop() twice for the left and right hashes below. We can do something similar here for left and right leaves (at least currently where we know both will exist due to pruning impl).

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+		for pos in segment_first_pos..=segment_last_pos {
+			if pmmr::is_leaf(pos) {
+				if let Some(data) = pmmr.get_data_from_file(pos) {
+					segment.leaf_data.push((pos, data));
+					continue;
+				} else if !prunable {
+					return Err(SegmentError::MissingLeaf(pos));
+				}
+			}
+			// TODO: optimize, no need to send every intermediary hash
```
antiochp, line 174: I think the list of hashes is the same as the list of pos that return true for `is_prune_root()`. i.e. We need the hash for the root of every pruned subtree.

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+
+impl Writeable for SegmentProof {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), Error> {
+		writer.write_u64(self.hashes.len() as u64)?;
+		for hash in &self.hashes {
+			Writeable::write(hash, writer)?;
+		}
+		Ok(())
+	}
+}
```
antiochp, line 519:
```suggestion
}

#[cfg(test)]
mod tests {
	    use croaring::Bitmap;

use crate::core::hash::{Hash, DefaultHashable};
	use crate::ser::{self, Writeable, Writer};
	use super::{Segment, SegmentIdentifier, SegmentProof};

	struct TestElem(pub [u32; 4]);

	impl DefaultHashable for TestElem {}

	impl Writeable for TestElem {
		fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
			writer.write_u32(self.0[0])?;
			writer.write_u32(self.0[1])?;
			writer.write_u32(self.0[2])?;
			writer.write_u32(self.0[3])
		}
	}

	//   3
	//  / \
	// 1   2
	//
	// Two leaves at pos 1 and 2.
	// Both leaves pruned.
	// Single root remaining at pos 3.
	//
	// No leaf_data (both leaves pruned).
	// The hash provided it the parent (root) hash at pos 3.
	//
	// The root of this segment should simply be the root hash provided for pos 3.
	#[test]
	fn test_single_pruned_root() {
		let hash = Hash::default();
		let segment: Segment<TestElem> = Segment {
			identifier: SegmentIdentifier { log_size: 1, idx: 0 },
			hashes: vec![(3, hash)],
			leaf_data: vec![],
			proof: SegmentProof::empty(),
		};
		assert_eq!(hash, segment.root(3, Some(&Bitmap::create())).unwrap());
	}
}
```

antiochp: `segment.root()` does not currently handle the case where we are provided with a root hash of a pruned subtree and neither child is present in the segment.

The test above is a minimal example of this with a manually constructed segment.
I may have constructed the segment incorrectly. 
Can you confirm this is what would be provided in this situation (single hash, no leaf_data etc.)?

antiochp: This stems from the `(None, None) => None,` after popping both children off the hashes vec.
There are two scenarios to handle here - 
* both children are `None` and we are also `None` because we deep in a pruned subtree.
* both children are `None` but we are a pruned root and the hash is provided in the segment directly.

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,519 @@
...
+		&self,
+		last_pos: u64,
+		mmr_root: Hash,
+		segment_first_pos: u64,
+		segment_last_pos: u64,
+		segment_root: Hash,
+	) -> Result<(), SegmentError> {
+		let root =
+			self.reconstruct_root(last_pos, segment_first_pos, segment_last_pos, segment_root)?;
+		if root == mmr_root {
```
antiochp, line 491: We commit to the `output_root`, `range_proof_root` and `kernel_root` in the block header.
We can pass the `range_proof_root` and `kernel_root` into `validate()` directly and confirm the root matches.

The `output_root` needs some thinking about as this is a merged root of both the output PMMR and the bitmap accumulator (they are hashed together).
So we cannot validate one is correct without providing the other.

If we receive a segment of the bitmap accumulator then I think we need to be provided with the corresponding output root to allow validation.
And conversely if we receive a segment of the output PMMR then we need the corresponding bitmap root.

Its not clear to me if this should be provided as part of the segment proof itself or provided alongside it.
There is some context specific logic that needs to occur there to know what kind of segment this is and what additional (if any) root hashes must be provided.

core/src/core/pmmr/segment.rs
jaspervdm: The leaves are guaranteed to be sorted, yes. However, my understanding is that if two sibling leaves are spent, but compaction has not been run, they still exist in the backend. During segment construction we include any leaf that still exists in the backend, without consulting a bitmap. That means the next leaf is not always the leaf we expect.

core/src/core/pmmr/segment.rs
jaspervdm: Ok, I think we should leave this for a future PR where we can check it properly, as there are already a lot of moving parts in this PR.

core/src/core/pmmr/segment.rs
jaspervdm: > The output_root needs some thinking about as this is a merged root of both the output PMMR and the bitmap accumulator (they are hashed together).
So we cannot validate one is correct without providing the other.

Yes, good point. So maybe it would make sense to introduce an additional p2p message to request the roots of the output PMMR and the bitmap accumulator. We can validate them by hashing them together to see if they reproduce `output_root` in the header. Then during the validation of segments we can use the roots directly. This sounds like the easiest solution to me, rather than having to alter the structure of segments for this specific situation.

core/src/core/pmmr/segment.rs
jaspervdm: My assumption was always that it should be added for "sufficiently small" functions to hint to the compiler it should be inlined, but I did some googling and apparently it only has an effect across crate boundaries. Since these functions are internal to the segment i will remove the inlining here.

core/src/core/pmmr/segment.rs
jaspervdm: I think this situation is already accounted for:
- Fully pruned segments cannot be created (see [here](https://github.com/mimblewimble/grin/pull/3453/files#diff-520aa054d83807bc7b454f4f7863aed94bc878ee393fa3d8df67039d47b568f8R189)).
- We do need to know the root of the segment if its sibling is not pruned, but the merkle proof of the sibling segment does contain it. That is why I still had this open point:
> I think we need a function to extract the intermediary and proof hashes for purposes of storing them in the MMR we are building

antiochp: It looks like `(None, None) => None` catches _any_ subtree in the segment in a similar way, not just the overall root? 
If both child nodes are `None` we do not always want to use `None` as the parent hash.

Does this happen somewhere else in the code?

core/src/core/pmmr/segment.rs
```diff
@@ -0,0 +1,538 @@
...
+			// TODO: optimize, no need to send every intermediary hash
+			if prunable {
+				if let Some(hash) = pmmr.get_from_file(pos) {
+					segment.hashes.push(hash);
+					segment.hash_pos.push(pos);
+				}
+			}
+		}
+
+		if segment.leaf_data.is_empty() {
```
antiochp, line 189: What are the implications of not being able to request a fully pruned segment?
If the first 1024 outputs are spent and the segment size is 1024, how do we proceed?

core/src/core/pmmr/segment.rs
jaspervdm: > Oh I get it now. We defer this to the next iteration.

Yes, correct.

core/src/core/pmmr/segment.rs
antiochp: Ok yeah this makes sense. So we are leveraging the iterator and just need to iterate once, but here we may skip over redundant leaves. :+1:

core/src/core/pmmr/segment.rs
jaspervdm: The first step in the new sync process will be to reconstruct the full bitmap (mmr) by downloading the segments. From that point on we know which segments will be empty, and just skip them.

antiochp: Yes of course, this makes sense if we have the bitmap (which itself is not pruned). :+1:
So in the pathological case of just a single unpruned leaf at the right hand side of the PMMR, this just results in a single segment being requested. We don't need the root of any of the fully pruned segments as this data is effectively included in the Merkle proof of that final segment.

jaspervdm: If the second segment is not fully spent, we download it, validate it and add the data to the PMMR we are building. This includes the root of the first segment since we obtained it from the merkle proof.

core/src/core/pmmr/segment.rs
antiochp: I'd personally prefer if these segment response msgs were self-contained and included this _other_ root. So not in the segment itself, but still included in the response msg.
Adding a whole new p2p msg for this seems overly complicated.
That said - I think either solution would work.

jaspervdm: Ok, I think that makes sense. The PR [here](https://github.com/mimblewimble/grin/pull/3470) defines the actual p2p messages (to be rebased after this one is merged). I will update it to add an extra field for the `OutputBitmapSegment` and `OutputSegment` messages.

jaspervdm:
Not at all! I think they highlight some of the subtleties in the PR so it is good that they are discussed explicitly.

core/src/core/pmmr/segment.rs
jaspervdm: I could use the `MerkleProof` struct instead, but we can't really use any of the functions already defined in it. We would have to add new functions to generate/validate the proof for the segment. So I think it would be more clean to keep them separate for now.

However if we possibly want to merge them in the future, we should maybe add the `mmr_size` field to `SegmentProof`, to keep it compatible with `MerkleProof`. Right now it is supplied as an argument to the validation function and assumed to be obtained from the block header.

antiochp:
Just posting here for reference - 

* We discussed "empty" segments. Proposal is to return a single root (found by recursing up the MMR from the empty segment). This provides all necessary hashes for reconstruction, even across a heavily pruned MMR, with pruning going beyond segment size. 

* There is an optimization in the above, if the requester can specify the segment size/height. Given the output bitmap they can determine which segments are empty and request _larger_ segments that cover these empty segments of the MMR.

jaspervdm:
Fixed an edge case: if there is an uneven number of leaves, and the final leaf is spent, we still require it to be present in the final segment. Previously we only checked the bitmap for it and its (non-existent) neighbour, which are both 0. This led us to assume they were pruned, but this is not the case.

This was actually caught by the pruning test in `store/tests/segment.rs`, except for the fact that there was a bug in the test itself. Fixing the bug in the test made the test fail as it should have.

jaspervdm:
@antiochp 
> * We discussed "empty" segments. Proposal is to return a single root (found by recursing up the MMR from the empty segment). This provides all necessary hashes for reconstruction, even across a heavily pruned MMR, with pruning going beyond segment size. The root position will exist outside the defined segment.

We now support pruned segments. The `root()` function now returns `Option<Hash>`, where a `None` indicates a full segment that is completely pruned. In this situation the root (or one of its parents) needs to be obtained from the list of hashes in the segment. This is done in `first_unpruned_parent()`.

In order to find the first parent up the path to the peak that is unpruned, the new `first_unpruned_parent()` function can be used. If the segment is not fully pruned, it will return `(hash, None)` where the hash is the root of the segment. If the segment is fully pruned, it will return `(hash, Some(pos))` where the hash is the hash of the first parent that isn't compacted away, and `pos` is the corresponding position.

I've also added a bunch of tests to make sure it behaves as expected for full segments and doesn't affect the partially filled final segment.

antiochp:
:+1: Sounds good - I'm planning to take a closer look at this today.

antiochp:
This looks good. Want to move it out of Draft status?

Just one minor point/question - 

> In order to find the first parent up the path to the peak that is unpruned, the new `first_unpruned_parent()` function can be used. If the segment is not fully pruned, it will return `(hash, None)` where the hash is the root of the segment. If the segment is fully pruned, it will return `(hash, Some(pos))` where the hash is the hash of the first parent that isn't compacted away, and `pos` is the corresponding position.

Do we need to have `None` vs `Some(pos)` here for this? 
I wonder if we could simply do `(hash, pos)` consistently for both scenarios? If its the "real" root of the subtree then its just the pos of the root. If its a fully pruned subtree then pos is just a higher up parent pos. Is there an advantage to only including an optional pos?

jaspervdm:
You are right, we don't really need to return a `(hash, Some(pos))`, will update it to `(hash, pos)`. For conceptual clarity I'd like to keep the `Option<Hash>` on the `root()` function though.

antiochp (approved):
I think we're looking good here with this. :+1:
What else is outstanding before we can merge?

jaspervdm:
I'm working on the deser of bitmap segments in this PR, but it probably makes more sense to do that in a separate one. I think we can merge this.

antiochp:
🎉

PR #3454, opened 2020-09-28
Refactor PMMR read methods into trait

jaspervdm:
`PMMR` and `ReadOnlyPMMR` have a few common functions that are essentially duplicate code. I also needed access to one of the `PMMR` functions on the `ReadOnlyPMMR`. This PR pulls these functions out of the specific impls and puts them in a common trait.

### DIFF
--- a/core/src/core/pmmr/pmmr.rs
+++ b/core/src/core/pmmr/pmmr.rs
@@ -26,6 +26,138 @@ use crate::ser::{PMMRIndexHashable, PMMRable};
 /// 64 bits all ones: 0b11111111...1
 const ALL_ONES: u64 = u64::MAX;
 
+/// Trait with common methods for reading from a PMMR
+pub trait ReadablePMMR {
+	/// Leaf type
+	type Item;
+
+	/// Get the hash at provided position in the MMR.
+	fn get_hash(&self, pos: u64) -> Option<Hash>;
+
+	/// Get the data element at provided position in the MMR.
+	fn get_data(&self, pos: u64) -> Option<Self::Item>;
+
+	/// Get the hash from the underlying MMR file (ignores the remove log).
+	fn get_from_file(&self, pos: u64) -> Option<Hash>;
+
+	/// Total size of the tree, including intermediary nodes and ignoring any pruning.
+	fn unpruned_size(&self) -> u64;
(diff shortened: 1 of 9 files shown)


antiochp (commented):
Looks good. I think this makes a lot of sense, yes.
The whole `ReadonlyPMMR` needs a rethink at some point - we still need `mut` on a bunch of internal state and its not _really_ readonly etc. (But that's for another PR).

antiochp:
Testing this locally now.

antiochp:
> I also needed access to one of the PMMR functions on the ReadOnlyPMMR.

What did you need here?

antiochp (approved):
:+1:

core/src/core/pmmr/pmmr.rs
```diff
@@ -396,6 +372,63 @@ where
...
 }
 
+impl<'a, T, B> ReadablePMMR for PMMR<'a, T, B>
+where
+	T: PMMRable,
+	B: 'a + Backend<T>,
+{
+	type Item = T::E;
+
+	fn get_hash(&self, pos: u64) -> Option<Hash> {
```
antiochp, line 382: Actually - can `get_hash()` and `get_data()` (and similar) not also be moved to the trait if we had a `backend()` fn?

jaspervdm:
> What did you need here?

`peak_path` - the hashes of the "other" peaks - is very useful for the merkle proofs in the chunks

core/src/core/pmmr/pmmr.rs
jaspervdm: Yes, but that does mean we give access to the backend outside of the module. It might give use the temptation to accidentally use the backend directly.

jaspervdm: Hm on further thought, we could maybe add the `backend()` function to a separate trait that isn't visible outside of the module..

jaspervdm: Turns out this is not really possible, as it would expose the private trait:
```
error[E0445]: restricted trait `core::pmmr::pmmr::PMMRBackend<B>` in public interface
```
So I'd rather keep it like it is, with some duplicate code.

PR #3455, opened 2020-09-30
Cleanup usage of legacy bitmaps.

antiochp:
The legacy bitmaps are no longer required and have not been in use for a while.

We migrated to using the "spent index" via `BLOCK_SPENT_PREFIX` prior to `4.0.0`.


jaspervdm (approved):
Yes, can be safely removed.

quentinlesceller (approved):
LGTM

PR #3456, opened 2020-09-30
cleanup validate() no need to return sums

antiochp:
`block.validate()` used to return the kernel sums via `Result<Commitment, Error>`.

This is not actually used anywhere and we can simplify this to `Result<(), Error>`.


antiochp:
Merging. This is a small refactor/cleanup and does not affect any behavior. 
The return type was unused.

PR #3457, opened 2020-10-01
use choose_multiple to clean get_peers() up

antiochp:
Small refactor to cleanup `find_peers()`.
We can use `choose_multiple()` to conveniently find `count` random entries, rather than shuffling them up manually.

### DIFF
--- a/p2p/src/store.rs
+++ b/p2p/src/store.rs
@@ -16,8 +16,7 @@
 
 use chrono::Utc;
 use num::FromPrimitive;
-use rand::seq::SliceRandom;
-use rand::thread_rng;
+use rand::prelude::*;
 
 use crate::core::ser::{self, Readable, Reader, Writeable, Writer};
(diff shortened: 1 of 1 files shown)


p2p/src/store.rs
```diff
@@ -146,31 +145,28 @@ impl PeerStore {
...
+		let peers = self
 			.db
-			.iter::<PeerData>(&to_key(PEER_PREFIX, ""))?
+			.iter::<PeerData>(&key)?
 			.map(|(_, v)| v)
 			.filter(|p| p.flags == state && p.capabilities.contains(cap))
-			.collect::<Vec<_>>();
-		peers[..].shuffle(&mut thread_rng());
-		Ok(peers.iter().take(count).cloned().collect())
+			.choose_multiple(&mut thread_rng(), count);
```
quentinlesceller, line 161: TIL

quentinlesceller (approved):
LGTM 👍

PR #3458, opened 2020-10-01
Flexible peer filtering

antiochp:
Takes inspiration from https://github.com/mimblewimble/grin/pull/3021 replacing custom cursor impl with `impl Iterator`.

As part of the PIBD work we will want to run nodes on the network that support various different sync strategies. some nodes may support PIBD while others only support "legacy" txhashset downloads.
So we need more flexible peer management and filtering of connected peers. This PR is a step toward this.

----

Introduce `PeersIter` so we can return iterator via "newtype iterator delegation".

```
pub fn peers_iter(&self) -> PeersIter<impl Iterator<Item = Arc<Peer>>> {
    ...
}
```

Then expose various filters and operators on `PeersIter` so we can chain them in various flexible ways -

```
self
    .peers_iter()
    .outbound()
    .connected()
    .count()
```

----

* moved the random sorting/shuffle out of the api and into a simpler `choose_random()` filter on `PeersIter`
* flexible `with_difficulty()` filtering to filter based on minimum difficulty
* flexible `with_capabilities()` filtering that can be combined with other chain-able filters

The following fns have all been replaced with the flexible `with_difficulty()` filter - 

* `more_work_peers()`
* `more_or_same_work_peers()`
* `more_work_peer()`
* `most_work_peers()`
* `most_work_peer()`

----

Note: As part of this PR the sync process has been modified slightly. 
During sync (both header_sync and state_sync) we now sync explicitly against _outbound_ peers. 
This makes the sync process more consistent across "public" and "private" nodes. We no longer attempt to sync against inbound nodes regardless of whether we accept incoming connections or not.

### DIFF
--- a/api/src/handlers/server_api.rs
+++ b/api/src/handlers/server_api.rs
@@ -56,7 +56,7 @@ impl StatusHandler {
 		Ok(Status::from_tip_and_peers(
 			head,
 			w(&self.peers)?
-				.peers_iter()
+				.iter()
 				.connected()
 				.count()
 				.try_into()
--- a/api/src/handlers/server_api.rs
+++ b/api/src/handlers/server_api.rs
@@ -56,7 +56,8 @@ impl StatusHandler {
 		Ok(Status::from_tip_and_peers(
 			head,
 			w(&self.peers)?
-				.connected_peers()
+				.peers_iter()
+				.connected()
 				.count()
 				.try_into()
 				.unwrap(),
--- a/api/src/handlers/server_api.rs
+++ b/api/src/handlers/server_api.rs
@@ -21,6 +21,7 @@ use crate::types::*;
 use crate::web::*;
 use hyper::{Body, Request};
 use serde_json::json;
+use std::convert::TryInto;
 use std::sync::Weak;
 
 // RESTful index of available api endpoints
@@ -54,7 +55,11 @@ impl StatusHandler {
 		let (api_sync_status, api_sync_info) = sync_status_to_api(sync_status);
 		Ok(Status::from_tip_and_peers(
 			head,
-			w(&self.peers)?.peer_count(),
+			w(&self.peers)?
+				.connected_peers()
+				.count()
+				.try_into()
+				.unwrap(),
 			api_sync_status,
 			api_sync_info,
 		))
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -18,8 +18,7 @@ use std::fs::File;
 use std::path::PathBuf;
 use std::sync::Arc;
 
-use rand::seq::SliceRandom;
-use rand::thread_rng;
+use rand::prelude::*;
 
 use crate::chain;
 use crate::core::core;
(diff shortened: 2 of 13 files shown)


p2p/src/peers.rs
```diff
@@ -105,139 +104,32 @@ impl Peers {
...
 
-	/// Get vec of peers we are currently connected to.
-	pub fn connected_peers(&self) -> Vec<Arc<Peer>> {
+	/// Iterator over our current peers.
+	/// This allows us to hide try_read_for() behind a cleaner interface.
+	/// PeersIter lets us chain various adaptors for convenience.
+	pub fn peers_iter(&self) -> PeersIter<impl Iterator<Item = Arc<Peer>>> {
 		let peers = match self.peers.try_read_for(LOCK_TIMEOUT) {
-			Some(peers) => peers,
+			Some(peers) => peers.values().cloned().collect(),
```
jaspervdm, line 112: We are potentially cloning a lot of `Arc`s here while we might only be interested in a few of them. Would it be possible to refactor `PeersIter` to borrow them instead, and only `clone` them whenever we use the `IntoIterator ` trait?

p2p/src/peers.rs
```diff
@@ -353,7 +245,7 @@ impl Peers {
 	/// Ping all our connected peers. Always automatically expects a pong back
 	/// or disconnects. This acts as a liveness test.
 	pub fn check_all(&self, total_difficulty: Difficulty, height: u64) {
-		for p in self.connected_peers().iter() {
+		for p in self.peers_iter().connected() {
```
jaspervdm, line 248: Given that these functions are defined on the `Peer` struct, would it make sense to remove the "peers" word from them? So `iter()` instead of `peers_iter()`

api/src/handlers/server_api.rs
```diff
@@ -54,7 +55,12 @@ impl StatusHandler {
 		let (api_sync_status, api_sync_info) = sync_status_to_api(sync_status);
 		Ok(Status::from_tip_and_peers(
 			head,
-			w(&self.peers)?.peer_count(),
+			w(&self.peers)?
+				.peers_iter()
+				.connected()
+				.count()
+				.try_into()
+				.unwrap(),
```
jaspervdm, line 63: Unwraps always make me uneasy, unless we can prove they never reach the bad path. In this case we can't, so could we replace it with an `unwrap_or(0)`?

jaspervdm (commented):
Apologies for the delay in review. I really like the approach here 👍 Just have a few small comments

p2p/src/peers.rs
antiochp: I don't think this is really possible because peers are behind a `RwLock`. The cloning effectively takes a snapshot of the current state of our list of peers at that point in time. Otherwise we need to keep a reference to the lock and things get messy.

Let me take another look at this though.

antiochp: I think our peer management needs a larger overhaul to get rid of this cloning. I'm not convinced having this outer `RwLock` wrapping the peers is the right approach long term. 
Lets revisit this and tackle as a separate PR.

jaspervdm: Ok, agreed :+1:

jaspervdm (approved):

PR #3460, opened 2020-10-02
CI: no directories in checksum file

quentinlesceller:
Equivalent of https://github.com/mimblewimble/grin-wallet/pull/360


quentinlesceller:
Merging since already tested in https://github.com/mimblewimble/grin-wallet/pull/360

PR #3463, opened 2020-10-06
Get the hash of a PMMR leaf ignoring the leafset

jaspervdm:
When a leaf is marked for removal (using the leafset), its hash is not actually removed from the backend until its sibling is also removed. The `get_hash` function on the PMMR will return a `None` for hashes that are marked for removal but not actually pruned yet. However in [this PR](https://github.com/mimblewimble/grin/pull/3453) we need access to this hash. So here we introduce a new function on the PMMR and Backend that will return a leaf hash, ignoring the leafset.

### DIFF
--- a/core/src/core/pmmr/backend.rs
+++ b/core/src/core/pmmr/backend.rs
@@ -39,6 +39,9 @@ pub trait Backend<T: PMMRable> {
 	/// Get a Hash by insertion position.
 	fn get_hash(&self, position: u64) -> Option<Hash>;
 
+	/// Get a leaf hash by insertion position, ignoring the leaf set
+	fn get_leaf_hash(&self, position: u64) -> Option<Hash>;
+
 	/// Get underlying data by insertion position.
 	fn get_data(&self, position: u64) -> Option<T::E>;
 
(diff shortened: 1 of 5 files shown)


core/src/core/pmmr/backend.rs
```diff
@@ -39,6 +39,9 @@ pub trait Backend<T: PMMRable> {
 	/// Get a Hash by insertion position.
 	fn get_hash(&self, position: u64) -> Option<Hash>;
 
+	/// Get a leaf hash by insertion position, ignoring the leaf set
```
antiochp, line 42: This should not be necessary - we already have `get_from_file()` below, which does exactly this.

core/src/core/pmmr/pmmr.rs
```diff
@@ -34,6 +34,9 @@ pub trait ReadablePMMR {
 	/// Get the hash at provided position in the MMR.
 	fn get_hash(&self, pos: u64) -> Option<Hash>;
 
+	/// Get the hash of a leaf, ignoring the leaf set
+	fn get_leaf_hash(&self, pos: u64) -> Option<Hash>;
```
antiochp, line 38: Same - we already have `get_from_file()` below.

antiochp (commented):
I don't think we need to do this. We already have `get_from_file()` on both `PMMR` and `ReadonlyPMMR`.

jaspervdm:
Yep, my mistake.

PR #3464, opened 2020-10-06
Define get_data_from_file function on ReadablePMMR trait

jaspervdm:
This function is currently only defined on the backend but it would be very convenient to have it on the PMMR as well.


antiochp (approved):
Yeah makes sense. :+1:

PR #3466, opened 2020-10-08
[4.1.x] fix v2 conversion to ensure we provide blocks in correct format (#3465)

antiochp:
Porting #3465 across to `4.1.x` branch for a bugfix release.

`4.1.0` introduced a subtle issue that can cause problems if a `4.0.0` node attempts to sync full blocks from a `4.1.0` node.


antiochp:
Merging. Same changes as #3465.

PR #3476, opened 2020-10-15
PMMR backend support for append_pruned_root()

### LINKED ISSUE #3475: Support rebuilding PMMR from pruned subtrees
Related #3453 

We need a way to rebuild a PMMR given a number of PIBD "segments".
Currently the PMMR api supports `append()` for new leaf nodes and associated hashes.

https://github.com/mimblewimble/grin/blob/cea546ceb8b5be0fc4462489e8a4761bc395fb18/core/src/core/pmmr/backend.rs#L30

We do not currently have a way to skip leaves and append a pruned subtree (root) directly.

We would need to do something along these lines - 

* append the single hash to the hash file
* append an entry to the PMMR `prune_list`
  * update the "shift chaches" in the `prune_list` to reflect the new entry

Exposing this `append_pruned()` on the PMMR backend _should_ (at least in principle) allow us to easily rebuild a PMMR.
For a given segment we would simply iterate over the pos, appending leaves and pruned subtrees as necessary.

antiochp:
~Do not merge until after HF4.~
Need to revisit this now that #3573 has been merged and the prune list internals simplified.

----

Resolves #3475.
Related #3453.

The previous impl of `append()` assumed we _always_ append individual leaves and associated hash(es). This makes perfect sense with append-only semantics during regular usage, adding a new block to the chain etc.

This PR relaxes this assumption and allows us to easily append pruned "subtree roots" directly.

The prune list stored in disk consists solely of pruned subtree roots so we can now reuse the same logic across node startup (reading the prune list off disk and initializing the caches) _and_ rebuilding a prune list from PIBD segments.

We take advantage of "append-only" semantics where any given subtree root will roll up pruned-ness of _previous_ contained subtrees and leaves.

Note that during chain compaction we can now simply rewrite the prune_list by converting it to a bitmap, merging it (bitmap `OR`) with the set of new pruned leaves and reopening the prune list with the updated bitmap. The logic for opening and rebuilding is the same.

Prune list "shift cache" and "leaf shift cache" no longer need to be rebuilt as we can efficiently build them as we populate the prune list (for both leaves and pruned subtrees). We take advantage of "append-only" semantics here. A new prune list entry will only ever shift subsequent positions.


jaspervdm:
Are we ok to remove the 5.0.0 milestone on this? As far as I can see we will really only need this improvement once we are actually processing segments to build the PMMRs.

antiochp:
> Are we ok to remove the 5.0.0 milestone on this? As far as I can see we will really only need this improvement once we are actually processing segments to build the PMMRs.

Yes. Probably makes sense. I wanted to get this in but you are right - we don't technically need it for 5.0.0 and it just adds risk to the HF.

phyro:
is this one waiting for PIBD to be continued? would it make sense to create a label PIBD or something to make it clear in the list of PRs?

yeastplume:
Closing and replacing with #3659

Issue #3473, opened 2020-10-14
Tracking issue: Fix DAA RFC


lehnberg:
_placeholder for https://github.com/mimblewimble/grin-rfcs/pull/61_

antiochp:
Done, #3477 is merged.

PR #3477, opened 2020-10-17
fixDAA; implement wtema

tromp:
---
name: [WIP] fixDAA
about: implement simpler wtema difficulty adjustment algorithm
title: fixDAA
labels: consensus breaking
assignees: antiochp, quentinlesceller, joltz

---
addresses https://github.com/mimblewimble/grin-rfcs/pull/61


antiochp:
Posted a question related to wtema and FlyClient in the DAA RFC - 
https://github.com/mimblewimble/grin-rfcs/pull/61#issuecomment-711152543

core/src/core/block.rs
```diff
@@ -432,10 +432,9 @@ pub struct UntrustedBlockHeader(BlockHeader);
 impl Readable for UntrustedBlockHeader {
 	fn read<R: Reader>(reader: &mut R) -> Result<UntrustedBlockHeader, ser::Error> {
 		let header = read_block_header(reader)?;
-		if header.timestamp
-			> Utc::now() + Duration::seconds(12 * (consensus::BLOCK_TIME_SEC as i64))
-		{
-			// refuse blocks more than 12 blocks intervals in future (as in bitcoin)
+		if header.timestamp > Utc::now() + Duration::seconds(5 * 60) {
```
antiochp, line 435: @tromp this is what you were asking in keybase about?

We do not have access to config directly here. But we could potentially expose it via `lazy_static!` and `thread_local!` similar to what we do for `CHAIN_TYPE` and `NRD_FEATURE_ENABLED`.

That way we could have something like `global::get_future_time_limit()` following the same pattern as `global::get_chain_type()`.

The reason for introducing the `thread_local!` complexity is to make this configurable in tests. Otherwise we have a single global config value that ends up causing havoc across multiple tests.

core/src/core/block.rs
```diff
@@ -432,10 +432,9 @@ pub struct UntrustedBlockHeader(BlockHeader);
...
+			// refuse blocks more than 5 minutes in the future
+			// ideally this Future Time Limit (FTL) is specified in grin-server.toml
 			// TODO add warning in p2p code if local time is too different from peers
 			error!(
 				"block header {} validation error: block time is more than 12 blocks in future",
```
antiochp, line 440: We should update this log msg if "12 blocks" is no longer applicable.

core/src/core/block.rs
tromp: I'm not sure this is something we would want to test anytime soon.
Is implementation much simpler if we forego testing different settings?

antiochp: Not really. We're better off just following the existing pattern I think.
Or just leaving it as hardcoded for now. Maybe just move the constant to `global.rs` and reference it from there.

src/bin/grin.rs
```diff
@@ -148,7 +148,7 @@ fn real_main() -> i32 {
 
 	// Initialize our global chain_type and feature flags (NRD kernel support currently).
 	// These are read via global and not read from config beyond this point.
-	global::init_global_chain_type(config.members.unwrap().server.chain_type);
+	global::init_global_chain_type(config.members.as_mut().unwrap().server.chain_type);
```
antiochp, line 151: `as_mut()` not required here?

core/src/global.rs
```diff
@@ -179,19 +193,43 @@ pub fn get_chain_type() -> ChainTypes {
...
+pub fn get_future_time_limit() -> u64 {
+	FUTURE_TIME_LIMIT.with(|ftl| match ftl.get() {
+		None => {
+			if GLOBAL_FUTURE_TIME_LIMIT.is_init() {
+				let ftl = GLOBAL_FUTURE_TIME_LIMIT.borrow();
+				set_local_future_time_limit(ftl);
+				ftl
+			} else {
+				// Global config unset, default to 5 minutes
+				5 * 60
```
antiochp, line 219: We should call `set_local_future_time_limit()` either way here so we just read local config next time.
This way we are simply lazily initializing the local config.

servers/src/common/types.rs
```diff
@@ -214,6 +217,7 @@ impl Default for ServerConfig {
 			dandelion_config: pool::DandelionConfig::default(),
 			stratum_mining_config: Some(StratumServerConfig::default()),
 			chain_type: ChainTypes::default(),
+			future_time_limit: 5 * 60,
```
antiochp, line 220: We should make `5 * 60` a `const` in `global.rs` and reuse it in `get_future_time_limit()` above.

src/bin/grin.rs
tromp: my understanding is that all but the last access to config.members needs to have as_mut(),
as suggested by lines 127 and 128.

servers/src/common/types.rs
tromp: Sounds good; but then we should also use config::API_SECRET_FILE_NAME rather than ".api_secret" just above?

src/bin/grin.rs
antiochp: Ugh.

antiochp:
Cam across this and trying to wrap my head around some of it - 
https://github.com/zawy12/difficulty-algorithms/issues/30

> POW consensus must:
> ...
> 2. Have a future timestamp limit that is > 2x "expected" node time error & << difficulty averaging window (strictly speaking, << block time).

Is (2) here relevant to our decision to use a 5 minute FTL?

src/bin/grin.rs
antiochp: Yeah - this whole init of the config needs rethinking in `grin.rs`. It looks like we have repeatedly just hacked this until it worked...

We should not need to have `mut` config here as we do not want to inadvertently write to it. But it needs to be `mut` because something else higher up in the file has modified a default via a `mut` on some _other_ piece of nested config.

:+1: on getting this working for now - but we should revisit at some point.

tromp: i see that grin-wallet uses

```
        global::init_global_chain_type(
                config
                        .members
                        .as_ref()
                        .unwrap()
                        .wallet
                        .chain_type
                        .as_ref()
                        .unwrap()
                        .clone(),
        );
```
perhaps that's better than as_mut() ?

core/src/global.rs
tromp: Does that apply to is_nrd_enabled() as well?

src/bin/grin.rs
tromp: do you agree with adding a global::init_global_accept_fee_base() method and associated
global::get_accept_fee_base() for the benefit of grin-wallet?
It would call global::init_global_accept_fee_base() after parsing the grin-wallet.toml file,
and then pass global::get_accept_fee_base() as 4th argument whenever calling txlib::mod::tx_fee

tromp:
> Is (2) here relevant to our decision to use a 5 minute FTL?

zawy argues that an FTL shorter than the blocktime is the best guard against timestamp manipulation. Which is undeniably true; but other people think that doing so carries too much risk of having the local clocks move out of sync beyond the FTL. So our choice of 5 minutes is a compromise that allows some but not much timestamp manipulation, while keeping the risk of out of sync local time very small (the idea being that a clock off by 5 minutes is quite human noticeable, while 1 minute is not so much).

src/bin/grin.rs
tromp: I decided to have an accept_fee_base  parameter in BOTH grin-server.toml and grin-wallet.toml,
and both using global::get_accept_fee_base() to query its value. It's even possible (but not particularly useful) to set then to different values in simultaneously running server and wallet.

As a result, we can simply libtx::mod::tx_fee to no longer have an optional fee base argument. If you want to deviate from the default, you can just change it locally, like I do in some tests.

servers/src/common/types.rs
tromp: default now defined as global::DEFAULT_FUTURE_TIME_LIMIT

core/src/global.rs
tromp: set_local_future_time_limit() always called now

core/src/core/block.rs
tromp: Following existing pattern now.

core/tests/consensus_automated.rs
```diff
@@ -148,7 +215,7 @@ fn repeat(interval: u64, diff: HeaderInfo, len: u64, cur_time: Option<u64>) -> V
 		.collect::<Vec<_>>()
 }
 
-fn repeat_offs(from: u64, interval: u64, diff: u64, len: u64) -> Vec<HeaderInfo> {
+fn repeat_offs(interval: u64, diff: u64, len: u64, from: u64) -> Vec<HeaderInfo> {
```
antiochp, line 218: Glad this is not called many times and we don't need to review those params getting juggled around... 😅

antiochp (approved):
Looks good to me.
I did not review the actual `next_wtema_difficulty()` logic.

lehnberg:
fixes: #3473

antiochp:
@quentinlesceller @j01tz I added you as reviewers (@tromp has you listed in the PR description).

core/src/consensus.rs
```diff
@@ -187,18 +184,21 @@ pub fn valid_header_version(height: u64, version: HeaderVersion) -> bool {
 	height < 4 * HARD_FORK_INTERVAL && version == header_version(height)
 }
 
-/// Number of blocks used to calculate difficulty adjustments
-pub const DIFFICULTY_ADJUST_WINDOW: u64 = HOUR_HEIGHT;
+/// Number of blocks used to calculate difficulty adjustment by Damped Moving Average
+pub const DMA_WINDOW: u64 = HOUR_HEIGHT;
+
+/// Difficulty adjustment half life is 4 hours
+pub const WTEMA_HALF_LIFE: u64 = 4 * HOUR_SEC;
```
j01tz, line 191: How was `WTEMA_HALF_LIFE` parameter selection determined?

doc/pow/pow.md
```diff
@@ -1,43 +1,34 @@
-# Grin's Proof-of-Work
```
jaspervdm, line 1: I think the changes in this doc are not relevant to the difficulty adjustment algorithm. Maybe we should have them in a separate PR instead?

core/src/consensus.rs
```diff
@@ -345,37 +358,66 @@ where
...
+	// last two headers
+	let last_header = last_headers.next().unwrap();
+	let prev_header = last_headers.next().unwrap();
+
+	let last_block_time: u64 = last_header.timestamp - prev_header.timestamp;
+
+	let last_diff = last_header.difficulty.to_num();
+
+	// wtema difficulty update
+	let next_diff =
```
j01tz, line 388: I'm struggling a bit verifying this algorithm against the details in the RFC and WTEMA specification. It seems straightforward enough here, but I think I need to better wrap my head around the `WTEMA_HALF_LIFE` value. Maybe a line or two of documentation could be helpful here.

core/src/consensus.rs
```diff
@@ -345,37 +358,66 @@ where
...
+		last_diff * WTEMA_HALF_LIFE / (WTEMA_HALF_LIFE - BLOCK_TIME_SEC + last_block_time);
```
jaspervdm, line 389: Took me a bit to process so putting a note here: in terms of the RFC, `WTEMA_HALF_LIFE` corresponds to `W * T`. So in this PR we have a `W` of 240, meaning the difficulty can change by a factor `~e` every 240 blocks (4 hours, hence the name of `WTEMA_HALF_LIFE`) 👍

jaspervdm (approved):
Looks good! 👍

core/src/consensus.rs
tromp: it's a tradeoff between response speed and stability. the 240 blocks closely matches BTH's choice of the equivalent of 288 blocks.

doc/pow/pow.md
tromp: the doc fixes don't need review so this seemed the easiest way to get then applied.

core/src/consensus.rs
tromp: correct. the name HALF_LIFE is a bit misleading in that sense, since it corresponds not to a doubling.

j01tz (approved):

Issue #3472, opened 2020-10-14
Bump header version + any pow changes


lehnberg:

antiochp:
Done in #3478.

PR #3478, opened 2020-10-18
HardFork4

tromp:
---
name: HardFork4
about: make header version 5 valid forever at 2*YEAR_HEIGHT
title: Hard Fork 4
labels: consensus-breaking, must-have, critical
assignees: antiochp, quentinlesceller, yeastplume

---

Still needs proper setting of TESTNET_FOURTH_HARD_FORK;
waiting for testnet to progress at predictable pace...

### DIFF
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -136,8 +136,8 @@ pub const TESTNET_SECOND_HARD_FORK: u64 = 298_080;
 /// Testnet second hard fork height, set to happen around 2020-06-20
 pub const TESTNET_THIRD_HARD_FORK: u64 = 552_960;
 
-/// Testnet second hard fork height, set to happen around 2020-12-??
-pub const TESTNET_FOURTH_HARD_FORK: u64 = 720_000;
+/// Testnet second hard fork height, set to happen around 2020-12-8
+pub const TESTNET_FOURTH_HARD_FORK: u64 = 649_440;
 
 /// Fork every 3 blocks
 pub const TESTING_HARD_FORK_INTERVAL: u64 = 3;
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -137,7 +137,7 @@ pub const TESTNET_SECOND_HARD_FORK: u64 = 298_080;
 pub const TESTNET_THIRD_HARD_FORK: u64 = 552_960;
 
 /// Testnet second hard fork height, set to happen around 2020-12-8
-pub const TESTNET_FOURTH_HARD_FORK: u64 = 649_440;
+pub const TESTNET_FOURTH_HARD_FORK: u64 = 642_240;
 
 /// Fork every 3 blocks
 pub const TESTING_HARD_FORK_INTERVAL: u64 = 3;
--- a/core/src/consensus.rs
+++ b/core/src/consensus.rs
@@ -136,22 +136,22 @@ pub const TESTNET_SECOND_HARD_FORK: u64 = 298_080;
 /// Testnet second hard fork height, set to happen around 2020-06-20
 pub const TESTNET_THIRD_HARD_FORK: u64 = 552_960;
 
-/// AutomatedTesting and UserTesting HF1 height.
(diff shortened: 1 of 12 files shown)


core/src/consensus.rs
```diff
@@ -136,43 +136,30 @@ pub const TESTNET_SECOND_HARD_FORK: u64 = 298_080;
...
+pub const TESTING_HARD_FORK_INTERVAL: u64 = 3;
 
 /// Compute possible block version at a given height, implements
 /// 6 months interval scheduled hard forks for the first 2 years.
 pub fn header_version(height: u64) -> HeaderVersion {
-	let chain_type = global::get_chain_type();
 	let hf_interval = (1 + height / HARD_FORK_INTERVAL) as u16;
-	match chain_type {
-		global::ChainTypes::Mainnet => HeaderVersion(hf_interval),
+	let testing_hf_interval = (1 + height / TESTING_HARD_FORK_INTERVAL) as u16;
```
antiochp, line 149: This can be moved into the match arm relevant for testing. We don't use it anywhere outside of there.

```
global::ChainTypes::AutomatedTesting | global::ChainTypes::UserTesting => {
	let testing_hf_interval = (1 + height / TESTING_HARD_FORK_INTERVAL) as u16;
	HeaderVersion(min(5, testing_hf_interval))
}
```

core/src/global.rs
```diff
@@ -224,35 +222,22 @@ pub fn create_pow_context<T>(
...
-			new_cuckaroom_ctx(edge_bits, proof_size)
-		}
-		ChainTypes::Testnet if valid_header_version(height, HeaderVersion(2)) => {
-			new_cuckarood_ctx(edge_bits, proof_size)
+	if chain_type == ChainTypes::Mainnet || chain_type == ChainTypes::Testnet {
+		// Mainnet and Testnet have Cuckatoo31+ for AF and Cuckaroo{,d,m,z}29 for AR
+		if edge_bits > 29 {
+			new_cuckatoo_ctx(edge_bits, proof_size, max_sols)
+		} else {
+			match header_version(height) {
```
antiochp, line 230: This reads so much better than `valid_header_version()`. :+1:

core/src/global.rs
```diff
@@ -224,35 +222,22 @@ pub fn create_pow_context<T>(
...
-			new_cuckarooz_ctx(edge_bits, proof_size)
-		}
-		ChainTypes::Testnet if valid_header_version(height, HeaderVersion(3)) => {
-			new_cuckaroom_ctx(edge_bits, proof_size)
-		}
-		ChainTypes::Testnet if valid_header_version(height, HeaderVersion(2)) => {
-			new_cuckarood_ctx(edge_bits, proof_size)
+	if chain_type == ChainTypes::Mainnet || chain_type == ChainTypes::Testnet {
+		// Mainnet and Testnet have Cuckatoo31+ for AF and Cuckaroo{,d,m,z}29 for AR
+		if edge_bits > 29 {
```
antiochp, line 227: Consider making this explicit on `header_version()` also?

```
if header_version(height) >= 5 || edge_bits > 29 {
   ...
}
```

Would we then want to explicitly fail if we actually had `edge_bits <= 29` after HF4?

antiochp (approved):
This looks good to me.
Couple of minor comments.

:+1:

lehnberg:
fixes: #3472

core/src/global.rs
tromp: thanks!

core/src/global.rs
tromp: Asking for a 29 context past HF4 should result in a clear error such as no_cuckaroo_ctx() rather than a dubious CuckAToo29 context, which is not an outright error, but otherwise quite invalid.

antiochp: Yes absolutely - just wondering if there is a more explicit way of implementing this, failing earlier in the code with a `no_cuckatoo_ctx()`.

Issue #3479, opened 2020-10-19
FlyClient support via a modified "Variable Difficulty MMR"


antiochp:
FlyClient (revised Aug2020): https://eprint.iacr.org/2019/226.pdf

> To enable handling difficulty-based sampling, we need to make two adjustments. We need a data-structure
which efficiently and securely enables the verifier to sample blocks based on their relative difficulty positions,
rather than their absolute positions as in the standard MMR. Second, Assumption 2, which states that the
adversary’s forks have only a fraction of the honest chain’s weight, requires that all difficulty transitions
are correct. In fact, as described in Section 3.1, the assumption is broken if the adversary can arbitrarily
manipulate block difficulties. We show how an adapted variable-difficulty MMR which aggregates information
about the total difficulty as well as difficulty transitions can resolve both issues at once.

> Definition 7 (Difficulty MMR). A difficulty MMR is a variant of the MMR with identical properties, but
such that each node contains additional data. Every node can be written as h, w, t, Dstart, Dnext, n, data, where
h is a λ bit output of a hash function, w is an integer representing a total difficulty, t is a timestamp, D
is a difficulty target, n is an integer representing the size of the subtree rooted by the node and data is an
arbitrary string. In a block, i.e., a leaf node, n = 1 and t is the time it took to mine the current block (the
blocks time stamp minus the previous block’s timestamp). w, Dstart is the current difficulty targets and Dnext
is the difficulty of the next block computed using the difficulty adjustment function defined Definition 1.
Each non-leaf node is defined as {H(lc, rc), lc.w + rc.w, lc.t + rc.t, lc.Dstart, rc.Dnext, lc.n + rc.n, ⊥}, where
lc = LeftChild and rc = RightChild.

> Difficulty MMR Verification. We need to modify the MMR verification algorithm in several ways. The
prover algorithm will be the same generating a proof consisting of a Merkle path. In general the verifier will
check the Merkle path and that the targets assigned to each node are feasible. For simplicity, we assume
that the epoch length m and the total number of leafs n are powers of 2. Given a left child (lc), a right child
(rc) and a parent node (p), the verifier performs the following checks:
> 1. Compute p using lc and rc following Definition 7.
> 2. Verify that lc.Dnext = rc.Dstart.
> 3. For both lc and rc verify that they are internally consistent. That is, ensure that there is a possible
set of legal difficulty transitions given the aggregate timing and difficulty parameters of these nodes:
[...]

PR #3480, opened 2020-10-22
[DNM] hash_with_index() rework to support flexible hash entries

antiochp:
[WIP]

Something along these lines would allow us to support additional data at parent nodes in the MMR.
This is similar in concept to our early "sumtree" idea.

This could be used to support the "variable difficulty MMR" for FlyClient if this indeed turns out to be necessary.

Related: #3479 

Most MMR structures simply use `Hash` as the `HashEntry`.
The header MMR would use a `HeaderHashEntry` which could be extended to include various additional pieces of aggregate data, summarizing the header leaf nodes contained in the subtree.

The way this is implemented in this PR we can include additional data in each `HashEntry` and _exclude_ it when hashing or _include_ it when hashing. So this would allow us to maintain additional arbitrary data in an MMR without affecting the MMR root _or_ do something along the lines of the "variable difficulty MMR" and commit to this additional data when hashing and producing the root.
I'm pretty sure our early "sumtree" concept excluded the commitment sum from the hash (and did not commit to it directly).

### DIFF
--- a/api/src/types.rs
+++ b/api/src/types.rs
@@ -15,7 +15,7 @@
 use crate::chain;
 use crate::core::core::hash::Hashed;
 use crate::core::core::merkle_proof::MerkleProof;
-use crate::core::core::{KernelFeatures, TxKernel};
+use crate::core::core::{KernelFeatures, OutputIdentifier, TxKernel};
 use crate::core::{core, ser};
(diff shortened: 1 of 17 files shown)


antiochp:
Closing. Flyclient investigation to be continued.

PR #3481, opened 2020-10-24
Fixfees

tromp:
---
name: fixfees
about: implement fixfees RFC
title: Fix Fees
labels: consensus-breaking, must-have

---


core/src/core/transaction.rs
```diff
@@ -36,6 +39,168 @@ use util::static_secp_instance;
...
+			}
+		}
+
+		deserializer.deserialize_any(FeeFieldsVisitor)
+	}
+}
+
+/// Conversion from a valid fee to a FeeFields with 0 fee_shift
+/// The valid fee range is 1..FEE_MASK
+impl TryFrom<u64> for FeeFields {
```
antiochp, line 113: Would it be beneficial to also have a `From<u32> for FeeFields` that never fails?
I'm not sure if this simplifies any use of this but it would potentially get rid of some uses of `unwrap()`.
We'd probably need to convert to `u64 -> u32` safely though, so maybe it just moves the problem around.

tromp: You can only get rid of unwrap by limiting fees to 32 bits, but if that was a good idea, then the RFC should have used 32 bits instead of 40 bits. This would come back to bite us in the distant future if blocks fill up and people do want to pay 5 Grin in fees, but either their wallet won't let them, or worse, will fail inexplicably.

antiochp: No I mean you can always fit a 32 bit fee safely in the 40 available bits. Not to limit fees to 32 bits.
So have the ability to turn a u32 safely into a FeeFields via `into()` (for convenience) and a u64 via `try_into()`.

tromp: So this would only be used in tests? I thought you were suggesting a new method that wallets could use.
For literal constant fees, it would be fine. But personally, I don't mind the unwrap()s.

core/src/global.rs
```diff
@@ -154,6 +162,9 @@ thread_local! {
 	/// Mainnet|Testnet|UserTesting|AutomatedTesting
 	pub static CHAIN_TYPE: Cell<Option<ChainTypes>> = Cell::new(None);
 
+		/// minimum transaction fee per unit of transaction weight for mempool acceptance
```
antiochp, line 165: whitespace/indentation

pool/src/pool.rs
```diff
@@ -413,12 +413,12 @@ where
 						weighting,
 						self.verifier_cache.clone(),
 					) {
-						if new_bucket.fee_to_weight >= bucket.fee_to_weight {
-							// Only aggregate if it would not reduce the fee_to_weight ratio.
+						if new_bucket.fee_rate >= bucket.fee_rate {
+							// Only aggregate if it would not reduce the fee_rate ratio.
 							tx_buckets[pos] = new_bucket;
```
antiochp, line 418: If we now support tx aggregation with combination of overpaying fees and suitable `fee_shift` values - does this allow us to relax this aggregation rule?

From the RFC -

> For instance, suppose two transactions have the same minfee. If one pays 3 times the minfee with a fee shift of 1, while the other pays exactly minfee with a zero fee shift, then both can be aggregated into one that pays twice the joint minfee (which is double the old one).

Currently this aggregation rule allows later txs to aggregate (during bucketing) with earlier txs iff the `fee_rate` does not decrease.

Txs `A` and `B` here being added to the _same_ bucket because `B` spends an output from `A`.

antiochp (commented):
:+1: Looks good to me.

Couple of really minor formatting comments.

One question about our bucketing/aggregation logic and how `fee_shift` and `fee_rate` interacts with this. I'm not clear how this _should_ work when bucketing dependent transactions together.

core/src/core/transaction.rs
```diff
@@ -866,53 +1060,59 @@ impl TransactionBody {
...
-	pub fn body_weight(&self) -> u64 {
-		TransactionBody::weight(
-			self.inputs.len() as u64,
-			self.outputs.len() as u64,
-			self.kernels.len() as u64,
-		)
+	/// Shifted fee for a TransactionBody is the sum of fees shifted right by the maximum fee_shift
+	/// this is used to determine whether a tx can be relayed or accepted in a mempool
+	/// where transactions can specify a higher block-inclusion priority as a positive shift up to 15
+	/// but are required to overpaythe minimum required fees by a factor of 2^priority
```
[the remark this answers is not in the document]
tromp: fixed

antiochp:
Do we need to handle `FeeFields` differently per and post HF4 (via header version or height)? 
I don't see anything conditional on version or height.

core/src/global.rs
tromp: fixed

pool/src/pool.rs
tromp: Yes, we should relax the aggregation rules to allow any aggregation that preserves shifted_fee compliance.
But I don't want to make this logic height dependent, so we can make that (non-consensus) change later.

tromp:
> Do we need to handle `FeeFields` differently per and post HF4 (via header version or height)?
> I don't see anything conditional on version or height.

Yes, FeeFields::fee() should be version dependent. I need to ponder how best to provide it with the header version.

pool/src/pool.rs
antiochp: Yes. That makes sense. :+1:

antiochp:
> Yes, FeeFields::fee() should be version dependent. I need to ponder how best to provide it with the header version.

I'm thinking about this as well.

core/src/core/transaction.rs
```diff
@@ -106,21 +279,24 @@ pub enum KernelFeatures {
 	/// Plain kernel (the default for Grin txs).
 	Plain {
 		/// Plain kernels have fees.
-		fee: u64,
+		#[serde(serialize_with = "fee_fields_as_int")]
+		fee_fields: FeeFields,
```
jaspervdm, line 283: To fully preserve backwards (de)serialization compatibility, we should name this field `fee` instead of `fee_fields

tromp: renamed in latest commit

antiochp (approved):
I'm :+1:. 
I'd like another pair of eyes on this before we merge though.

antiochp (approved):
Passing `height` into `accept_fee()` looks :+1:

j01tz (approved):
Looks good overall but wasn't able to spend too much time thinking through further possible corner cases and did not get a chance to manually test. Will be good to test this change extensively with the beta release. :+1:

core/src/core/transaction.rs
```diff
@@ -36,6 +40,184 @@ use util::static_secp_instance;
...
+impl Writeable for FeeFields {
+	fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ser::Error> {
+		writer.write_u64(self.0)
+	}
+}
+
+impl Readable for FeeFields {
+	fn read<R: Reader>(reader: &mut R) -> Result<Self, ser::Error> {
+		let fee_fields = reader.read_u64()?;
+		Ok(Self(fee_fields))
```
jaspervdm, line 58: Should we use `try_from` here? Currently this impl allows us to construct FeeFields with an illegal fee value.

jaspervdm (commented):
Looks good, I just have one question.

core/src/core/transaction.rs
tromp: the height oblivious reader has no notion of illegal

jaspervdm: Ok, cleared up after discussion on keybase.

jaspervdm (approved):
👍 to merge assuming CI completes succesfully.

quentinlesceller (approved):

quentinlesceller (commented):
Questions about displaying the fees below. Other than that nothing stood up.

PR #3482, opened 2020-10-28
add segmenter for generating segments from txhashset with consistent rewind

antiochp:
Related #3453.

* Cache a `segmenter` instance based on current archive header (every 12 hours)
* The cached `segmenter` contains a "snapshot" of the bitmap accumulator
  * This can be read directly without needing to rewind (based on archive header)
* Fast reads of output, rangeproof and kernel MMR data via `ReadonlyPMMR::at()` (no rewind required).

The caller (p2p handling peer PIBD segment requests) is responsible for verifying the provided header is valid with respect to the current archive header. 

We only support generation of PIBD segments at the current local archive header (single cached `segmenter`).
Note: As discussed below, PMMR data is apppend-only (bitmap is an exception) so this is not as limiting as it appears.

This PR is going to have a bunch of placeholder code in it until #3453 is merged. 
This is the bridge between actual segment creation and current chain state.

### DIFF
--- a/chain/src/txhashset/segmenter.rs
+++ b/chain/src/txhashset/segmenter.rs
@@ -1,0 +1,66 @@
+// Copyright 2020 The Grin Developers
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Generation of the various necessary segments requested during PIBD.
+
+use crate::core::core::BlockHeader;
+use crate::error::Error;
+use crate::txhashset::{self, TxHashSet};
+
+/// TODO - Replace this with SegmentIdentifier from segment PR.
+pub struct SegmentIdentifier {
+	height: u8,
+	idx: u64,
+}
+
+/// Segmenter for generating PIBS segments.
+pub struct Segmenter<'a> {
+	txhashset: &'a TxHashSet,
+	header: BlockHeader,
+}
+
+impl<'a> Segmenter<'a> {
+	/// Create a new segmenter based on the provided txhashset.
+	pub fn new(txhashset: &'a TxHashSet, header: BlockHeader) -> Segmenter<'a> {
+		Segmenter { txhashset, header }
+	}
+
+	/// Create a kernel segment.
+	pub fn kernel_segment(&self, _id: SegmentIdentifier) -> Result<&str, Error> {
+		txhashset::rewindable_kernel_view(&self.txhashset, |view, batch| {
+			view.rewind(&self.header)?;
+
+			// now build a segment from the kernel MMR in our view.
+
+			Ok(())
+		})?;
+
+		Ok("this will return a segment")
+	}
+
+	/// Note: we need to rewind both the bitmap and the output MMR as we need both roots here.
+	pub fn bitmap_segment(&self, _id: SegmentIdentifier) -> Result<&str, Error> {
+		unimplemented!()
+	}
+
+	/// Note: we need to rewind both the bitmap and the output MMR as we need both roots here.
+	pub fn output_segment(&self, _id: SegmentIdentifier) -> Result<&str, Error> {
+		unimplemented!()
+	}
+
+	/// Create a rangeproof segment.
+	pub fn rangeproof_segment(&self, _id: SegmentIdentifier) -> Result<&str, Error> {
+		unimplemented!()
+	}
+}
--- a/chain/src/txhashset/segmenter.rs
+++ b/chain/src/txhashset/segmenter.rs
@@ -14,13 +14,15 @@
 
 //! Generation of the various necessary segments requested during PIBD.
 
-use std::sync::Arc;
+use std::{marker, sync::Arc};
 
 use crate::core::core::hash::Hash;
-use crate::core::core::pmmr::ReadablePMMR;
-use crate::core::core::BlockHeader;
+use crate::core::core::pmmr::{Backend, ReadablePMMR, ReadonlyPMMR};
+use crate::core::core::{BlockHeader, OutputIdentifier, TxKernel};
+use crate::core::ser::{PMMRable, Readable, Writeable};
 use crate::error::{Error, ErrorKind};
-use crate::txhashset::{self, PMMRHandle, TxHashSet};
+use crate::txhashset::{self, BitmapChunk, PMMRHandle, TxHashSet};
+use crate::util::secp::pedersen::RangeProof;
 use crate::util::RwLock;
 
 /// TODO - Replace this with SegmentIdentifier from segment PR.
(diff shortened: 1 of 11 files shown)


jaspervdm:
This looks good. So in https://github.com/mimblewimble/grin/pull/3470, responding to a request would be as simple as getting the segmenter from the chain, comparing the request block header hash to the one in the segmenter and getting the appropriate segment.

antiochp:
> It would be bad if a node is syncing from a specific point and the nodes its syncing from switch to the newer block. It would mean all the segments have to be downloaded from scratch.

Yeah I wondered about that. This is 60 secs every 12 hours, so its going to be rare. But that said it is undoubtedly going to happen. Maybe the simplest thing to do would be to support this for the last _two_ recent header heights.

antiochp:
Been doing some playing around with this and the rewinds are too slow currently. We need to think about how to avoid doing all this work, specifically rewinding to rebuild the utxo bitmap each time.

antiochp:
@jaspervdm 

It may be possible to avoid needing to support PIBS for multiple blocks (cross boundary between 2 archive periods).
The _only_ PIBD data that is not append-only is the bitmap.

Ignoring the bitmap for a second - if the client (syncing) keeps track of the block being downloaded for, it can safely switch to requesting the later period. This assumes it receives the latest header via broadcast in a timely manner.
The old segments (outputs, rangeproofs, kernels) will have Merkle proofs against the earlier header, but the segments themselves will be identical. Later segments will switch to having Merkle proofs against the later header.

This puts the responsibility on the client to correctly reconstruct the PMMR based on segments potentially split across two headers. Segments are fine, we just need to account for the varying roots of the Merkle proofs.
This all works because the data is append-only. Segments for earlier header are simply a truncation of data for the later header.

tl;dr Client can reconstruct (output, rangeproof, kernel) PMMR from segments regardless of the root they are under. All data up to and including earlier PMMR size is identical and append-only. 
Data beyond that, up to and including the later PMMR size, is strictly append-only.

----

So the issue is the bitmap. This _will_ be different across the two headers for the two sync time periods.
This is effectively "all or nothing" for a given header.
If the client successfully receives all segments, consistently, then the bitmap can be constructed.
If the sync period switches _during_ bitmap segment download then the client can restart the sync process. but _only_ for the bitmap segments.

----

Still need to think this through more but I suspect we can avoid requiring nodes to maintain and serve multiple versions of cached data to support this.

jaspervdm:
@antiochp I think that makes a lot of sense and reduces the amount of data nodes need to keep around. Just so I make sure I fully understand your proposal, if a node is syncing outputs/rproofs/kernels at a specific block and a new archive period starts, the following should happen:
- Get latest headers (presumably we already keep up to date with the header chain during current sync)
- Wait for all pending segments to come in, verify and process them against the header
- Get new output bitmap segments, rebuild the new output bitmap chunk MMR
- Update partially built output/rproof PMMRs and prune anything that is spent
- Continue with sync by requesting more segments against the new header

antiochp:
> We just have to do the pruning step all the way at the end, once we built the full trees.

Yes agreed. We may be reconstructing the PMMRs from segments against different headers. We only really know what can be pruned once we have all the data locally and we have verified all PMMRs are consistent.

antiochp:
> Get new output bitmap segments, rebuild the new output bitmap chunk MMR

I think we only need to do this _if_ we have not received a full set of bitmap segments against the original header.
If we can reconstruct the full bitmap MMR then we know we can reconstruct the other PMMRs against the original header (we can always simply truncate data appended beyond that).

antiochp:
This is now exercising the segmenter and the various segment variants during `chain.init()`

```
20201118 10:46:23.770 DEBUG grin_chain::chain - init_segmenter: initializing new segmenter for 0003d63dd89f at 961200
20201118 10:46:23.771 DEBUG grin_chain::txhashset::txhashset - Rewind extension to 0003d63dd89f at 961200 from 0003a83b0650 at 964452
20201118 10:46:24.737 DEBUG grin_chain::txhashset::bitmap_accumulator - applied 4194 chunks from idx 71 to idx 4264 (100ms)
20201118 10:46:24.739 DEBUG grin_chain::chain - init_segmenter: done, took 968ms
20201118 10:46:24.748 DEBUG grin_chain::txhashset::segmenter - kernel_segment: id: (9, 0), leaves: 512, hashes: 0, proof hashes: 13, took 8ms
20201118 10:46:24.748 DEBUG grin_chain::txhashset::segmenter - bitmap_segment: id: (9, 0), leaves: 512, hashes: 0, proof hashes: 4, took 0ms
20201118 10:46:24.754 DEBUG grin_chain::txhashset::segmenter - output_segment: id: (11, 0), leaves: 100, hashes: 469, proof hashes: 12, took 5ms
20201118 10:46:24.760 DEBUG grin_chain::txhashset::segmenter - rangeproof_segment: id: (7, 0), leaves: 10, hashes: 43, proof hashes: 16, took 5ms
```

chain/src/txhashset/segmenter.rs
```diff
@@ -0,0 +1,149 @@
...
+use crate::util::secp::pedersen::RangeProof;
+use crate::util::RwLock;
+
+/// Segmenter for generating PIBD segments.
+#[derive(Clone)]
+pub struct Segmenter {
+	txhashset: Arc<RwLock<TxHashSet>>,
+	bitmap_snapshot: BitmapAccumulator,
+	header: BlockHeader,
+}
```
jaspervdm, line 33: Because `Chain::segmenter()` clones the `Segmenter` I think we should put the `BitmapAccumulator` in an `Arc`. Cloning the full accumulator each time would be quite expensive

antiochp: Oh I see what you mean here. I thought you were originally talking about having a _single_ instance of the accumulator.
I was confused because we need to be able to have multiple snapshots over time.
But you are talking about a single instance of a given _snapshot_. 

Yes :+1: we don't need to clone this snapshot each time. I'll update this.

jaspervdm:
Coming back to the discussion about switching archive periods during sync: how would we be able to verify the segments of prunable MMRs after the switch, without first downloading the new bitmap? The new segments possibly contain less leaves (because they have been spent and pruned in between the archive periods), but we still require them to be present according to our old bitmap.

antiochp:
> Coming back to the discussion about switching archive periods during sync: how would we be able to verify the segments of prunable MMRs after the switch, without first downloading the new bitmap? The new segments possibly contain less leaves (because they have been spent and pruned in between the archive periods), but we still require them to be present according to our old bitmap.

So technically yes we would need to download the full _later_ bitmap to be able to verify prunable segments.
But only if pruning were to occur in the in the interim period.

In practical terms we only prune data after 7 days. and we are talking about two archive periods in the most recent 12 hours.
So nothing will be pruned between the current archive period and the previous archive period.

jaspervdm:
Okay, makes sense 👍. So after a prolonged shutdown we'd still need to redownload the bitmap segments regardless, but we can keep around the validated segments of all the other MMRs

antiochp:
> Okay, makes sense 👍. So after a prolonged shutdown we'd still need to redownload the bitmap segments regardless, but we can keep around the validated segments of all the other MMRs

Yes I think that makes sense. Bitmap segments will no longer be valid beyond our pruning/compaction horizon.
And syncing the bitmap segments is going to be fast relative to the rest of the data.

----

Edit: This got me thinking about something. Once these new p2p msgs are available we will always have the option to go ask a connected peer for a "recent" bitmap. So even in a relatively catastrophic data corruption situation locally we should be able to leverage this to give us a relatively quick way to fix our local data and resync without needing to clear everything out.
So we should be able to provide more robustness for stopping/starting with relatively little extra work here.

Currently there is no way to rebuild our local bitmap from a peer without requesting the full txhashset archive.

core/src/core/pmmr/segment.rs
```diff
@@ -132,6 +132,7 @@ impl<T> Segment<T> {
 		(first, last)
 	}
 
+	/// TODO - binary_search_by_key() here (can we assume these are sorted by pos?)
```
jaspervdm, line 135: We can't use `binary_search_by_key` because we only have an iterator here, not a slice

jaspervdm (approved):

PR #3484, opened 2020-11-05
default capabilities cleanup

antiochp:
We previously set node "capabilities" explicitly in the config file.
The capabilities are then advertised to peers via hand/shake protocol. 

```
bits = 15
```

This makes it hard to introduce new capabilities (for example PIBD support) as every node has these effectively hard-coded in their existing config files.
And every node needs to update the config to explicitly enable new capabilities.

This PR removes this configuration and simply defines "default" capabilities. This "default" can now change over time without explicit modification of the config file.

----

One (surprising and probably unexpected) thing we used node capabilities for was to express an "affinity" toward other peers. When requesting a peer address list we asked for peers that exhibited the same capabilities as our local node.
I don't think this ever makes sense, particularly in an environment where peers have heterogenous capabilities.
Nodes should advertise their local capabilities and it should be up to other nodes to decide, based on these capabilities, how to interact.  A node wants to know about a wide range of other nodes, not just those that appear similar.

This PR modifies the "peer address list" logic to request lists of peers that themselves are also capable of providing lists of peers. No other capabilities are required. Previously we only asked for `FULL_NODE` (i.e. full set of capabilities) peers.

----

Introduce a new `PIBD_HIST` capabilities flag, currently unused and currently _not_ included in default capabilities.

----

This PR should make it significantly easier to roll PIBD out across nodes incrementally. Only a subset of nodes may support PIBD initially.

We may want to introduce support for _disabling_ some capabilities over time. This is outside the scope of this PR.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -267,7 +267,7 @@ fn comments() -> HashMap<String, String> {
 	);
 
 	retval.insert(
-		"[server.p2p_config.capabilities]".to_string(),
+		"[server.pool_config]".to_string(),
 		"#If the seeding type is List, the list of peers to connect to can
 #be specified as follows:
 #seeds = [\"192.168.0.1:3414\",\"192.168.0.2:3414\"]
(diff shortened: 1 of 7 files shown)


antiochp:
Planning to enable the new `PIBD_HIST` capability in a separate PR as a final step once we are comfortable with #3470 and it is merged.

phyro:
Out of curiosity, would a 'nimble node' also be a possible node capability value?

antiochp:
> Out of curiosity, would a 'nimble node' also be a possible node capability value?

I'm imagining a "nimble" node as having a limited or reduced set of capabilities.
A nimble node may not support `PIBD_HIST` for example.

jaspervdm (approved):
Looks good 👍 Ran it locally for a while without any issues

PR #3486, opened 2020-11-09
on_block_accepted logging and webhooks all the time

### LINKED ISSUE #3485: on_block_accepted hooks should fire regardless of sync
If we successfully process and accept a new block we call `block_accepted` on the adapter.
This call to `block_accepted` performs several operations - 

* triggers all registered `on_block_accepted` hooks (webhooks and logging)
* broadcast the new block (header first) to our peers
* reconcile the mempool
* reconcile the mempool related "reorg cache"

The first two operations are suppressed if we are processing blocks in "sync" mode.
We do not want to broadcast blocks to our peers if we are syncing.

I believe the logic here is incorrect for the first operation.
I think we _do_ want to log (and fire webhooks) for all blocks accepted, regardless of sync/broadcast.

Note the "gap" here between blocks at height `949737` and `949748` -

```
20201108 02:47:03.845 DEBUG grin_servers::common::hooks - block_accepted (head+): 0002385f0a23 at 949735 (prev: 000004f47538 at 949734)
20201108 02:47:15.616 DEBUG grin_servers::common::hooks - block_accepted (head+): 00011eb0abc8 at 949736 (prev: 0002385f0a23 at 949735)
20201108 02:48:03.850 DEBUG grin_servers::common::hooks - block_accepted (head+): 0001a30458c9 at 949737 (prev: 00011eb0abc8 at 949736)
20201108 02:49:46.899 DEBUG grin_servers::common::hooks - block_accepted (head+): 0001aa69e5bd at 949748 (prev: 000064b9880a at 949747)
20201108 02:50:22.396 DEBUG grin_servers::common::hooks - block_accepted (head+): 0001c69874d5 at 949749 (prev: 0001aa69e5bd at 949748)
20201108 02:51:05.126 DEBUG grin_servers::common::hooks - block_accepted (head+): 0000e91560eb at 949750 (prev: 0001c69874d5 at 949749)
```

This was from a recent large reorg that triggered a "catch up" sync on this node (node was >5 blocks diff behind head).
But the "REORG!" `block_accepted` log msg as we were syncing.

----

https://github.com/mimblewimble/grin/blob/25fcefada2e290ea4322ec8e4483e31ef168f91b/servers/src/common/adapters.rs#L739-L744

----

Proposed:

* Suppress block broadcast if SYNC (retain current behavior).
* Trigger hooks regardless of SYNC (new behavior).

antiochp:
Resolves #3485

We want to fire the registered `on_block_accepted` hooks regardless of `SYNC`.
We do still want to suppress block broadcast if `SYNC`.

Pretty sure this is fine to do even if we have webhooks registered. We use `futures` in the webhook code (hyper client) so these are not blocking and should be fine to run even during sync.

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -737,11 +737,13 @@ where
 	V: VerifierCache + 'static,
 {
 	fn block_accepted(&self, b: &core::Block, status: BlockStatus, opts: Options) {
-		// not broadcasting blocks received through sync
+		// Trigger all registered "on_block_accepted" hooks (logging and webhooks).
+		for hook in &self.hooks {
+			hook.on_block_accepted(b, status);
+		}
+
+		// Suppress broadcast of new blocks received during sync.
 		if !opts.contains(chain::Options::SYNC) {
-			for hook in &self.hooks {
-				hook.on_block_accepted(b, status);
-			}
 			// If we mined the block then we want to broadcast the compact block.
 			// If we received the block from another node then broadcast "header first"
 			// to minimize network traffic.


quentinlesceller (approved):
👍

Issue #3483, opened 2020-11-02
API thread hangs attempting to get (old) block data from LMDB


bladedoyle:
**Describe the bug**
API thread hangs attempting to get some (old) block data from LMDB.  It appears to never recover and requires a restart of the node process to regain API response. 

**To Reproduce**
`curl -ugrin:<secret> localhost:3413/v1/blocks/29`
The query is against an archive node.  
Blocks before and after this work fine.
More than one block causes this issue (its hard to test because a node restart is needed to try again)

**Relevant Information**
No error messages in the node log or on the screen.
The stack trace for this thread shows:
`
(gdb) thread 7
[Switching to thread 7 (Thread 0x7ef85b01c700 (LWP 723))]
#0  0x00005646f8d788b3 in mdb_page_search_root (mc=mc@entry=0x7ef85b0068c0, key=key@entry=0x7ef85b006d38, flags=flags@entry=0) at mdb/libraries/liblmdb/mdb.c:5318
5318    mdb/libraries/liblmdb/mdb.c: No such file or directory.
(gdb) where
#0  0x00005646f8d788b3 in mdb_page_search_root (mc=mc@entry=0x7ef85b0068c0, key=key@entry=0x7ef85b006d38, flags=flags@entry=0) at mdb/libraries/liblmdb/mdb.c:5318
#1  0x00005646f8d78b92 in mdb_page_search (mc=mc@entry=0x7ef85b0068c0, key=key@entry=0x7ef85b006d38, flags=flags@entry=0) at mdb/libraries/liblmdb/mdb.c:5446
#2  0x00005646f8d79285 in mdb_cursor_set (mc=mc@entry=0x7ef85b0068c0, key=key@entry=0x7ef85b006d38, data=data@entry=0x7ef85b006d48, op=op@entry=MDB_SET, exactp=exactp@entry=0x7ef85b0068bc)
    at mdb/libraries/liblmdb/mdb.c:5905
#3  0x00005646f8d79880 in mdb_get (txn=<optimized out>, dbi=<optimized out>, key=0x7ef85b006d38, data=0x7ef85b006d48) at mdb/libraries/liblmdb/mdb.c:5575
#4  0x00005646f8cbb1c5 in lmdb_zero::tx::ConstAccessor::get (self=0x7ef85b007678, db=0x5646fafba8c0, key=...)
[...]

PR #3487, opened 2020-11-09
Disable merkle proofs for v2 get_block

antiochp:
This PR defaults to `include_merkle_proof = false` in the v2 api `get_block` jsonrpc endpoint.
These Merkle proofs are not actively used by anyone and were originally included (I believe) as a "proof of concept".
They are expensive to generate for historical blocks (specifically against a full archival node).

Related #3483.

----

Tested via - 

```
curl -X POST \
     -H 'Content-Type: application/json' \
     -d '{"jsonrpc":"2.0","id":1,"method":"get_block","params":[952019, null, null]}' \
         http://localhost:3413/v2/foreign
```

----

This is equivalent behavior to the optional `?no_merkle_proof` url param in the original v1 api endpoint - 

```
curl "localhost:3413/v1/blocks/952022?no_merkle_proof"
```


api/src/foreign_rpc.rs
```diff
@@ -130,116 +130,125 @@ pub trait ForeignRpc: Sync + Send {
...
-				453337210,
-				459401597,
-				496068509,
-				511300624
+
+	curl -X POST \
+		 -H 'Content-Type: application/json' \
+		 -d '{"jsonrpc":"2.0","id":1,"method":"get_block","params":[952019, null, null]}' \
+		 http://localhost:3413/v2/foreign
+
```
quentinlesceller, line 142: What's that? The doctests (currently disabled) are just below.

api/src/foreign_rpc.rs
```diff
@@ -130,116 +130,125 @@ pub trait ForeignRpc: Sync + Send {
...
+
+
+		```
+		# grin_api::doctest_helper_json_rpc_foreign_assert_response!(
+		# r#"
+		{
+			"jsonrpc": "2.0",
+			"method": "get_block",
+			"params": [374274, null, null],
```
quentinlesceller, line 152: This is the value you want to change here I guess

quentinlesceller (changes requested):

quentinlesceller:
Other than that everything looks good to me.

api/src/foreign_rpc.rs
antiochp: lol 😭

quentinlesceller (approved):

PR #3490, opened 2020-11-16
add peer capabilities to tui peers screen

antiochp:
Related #3484. 

Add a `Capab` column showig a simple u32 value for peer advertised capabilities.
`15` is the default (`FULL_NODE`) currently.
We plan to introduce a new value for PIBD support.

Screen Shot 2020-11-16 at 2 51 10 PM


jaspervdm (approved):
Tested locally, works as expected

PR #3491, opened 2020-11-17
optimize images

Th0rgal:
Hey there,

This PR optimizes the included images by about 40%.

Sincerely,
Thomas


quentinlesceller:
Thanks @Th0rgal for the PR. I think we will not update the docs anymore and keep them as an archive since the real docs are now on the docs repo.

PR #3492, opened 2020-11-20
More efficient serialization for bitmap segments

jaspervdm:
This PR introduces a new type `BitmapSegment` that can be converted back and forth into a `Segment<BitmapChunk>`. It provides a more efficient (de)serialization than the `Segment` by choosing different strategies depending on the occupancy of the bitmap in the range. This is achieved by splitting the segment up in blocks of 2^16 bits (`BitmapBlock`s). If a block has less than 4096 positive indices, we serialize a list of positive indices as unsigned 16bit integers. If a block instead has less than 4096 negative indices, we serialize a list of negative indices as unsigned 16bit integers. In other cases, we serialize the full bitmap bytes directly.

TODO:
- ~We can speed up the conversion between the two types by manipulating the storage of the internal bitvecs directly, using an `unsafe` block. Is this something we want to do?~ Can be done in a future PR
- ~Check whether the current tests are sufficient~

### DIFF
--- a/chain/src/txhashset/bitmap_accumulator.rs
+++ b/chain/src/txhashset/bitmap_accumulator.rs
@@ -292,11 +292,10 @@ impl From<Segment<BitmapChunk>> for BitmapSegment {
 		let (identifier, _, _, _, leaf_data, proof) = segment.parts();
 
 		let mut chunks_left = leaf_data.len();
-		let mut blocks = Vec::with_capacity(
-			(chunks_left + BitmapBlock::CHUNKS_PER_BLOCK - 1) / BitmapBlock::CHUNKS_PER_BLOCK,
-		);
+		let mut blocks =
+			Vec::with_capacity((chunks_left + BitmapBlock::NCHUNKS - 1) / BitmapBlock::NCHUNKS);
 		while chunks_left > 0 {
-			let n_chunks = min(BitmapBlock::CHUNKS_PER_BLOCK, chunks_left);
+			let n_chunks = min(BitmapBlock::NCHUNKS, chunks_left);
 			chunks_left = chunks_left.saturating_sub(n_chunks);
 			blocks.push(BitmapBlock::new(n_chunks));
 		}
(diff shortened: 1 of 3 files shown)


antiochp:
:eyes:

antiochp:
Bitmap chunks are 1024 bits long.
Segments are variable length (based on height).
We flip serialization strategies at 4096 bits.
So does this mean that for some segment sizes/heights we only ever use the positive/negative indices and never fallback to the bitmap representation?

jaspervdm:
In theory yes, but such segments are currently not allowed. Per [the RFC](https://github.com/jaspervdm/grin-rfcs/blob/pibd_messages/text/0000-pibd-messages.md), the minimum height of a bitmap segment is 9.

antiochp:
> * We can speed up the conversion between the two types by manipulating the storage of the internal bitvecs directly, using an `unsafe` block. Is this something we want to do?

We can always revisit this later - I would not optimize like this yet.

chain/src/txhashset/bitmap_accumulator.rs
```diff
@@ -239,3 +246,283 @@ impl Readable for BitmapChunk {
...
+		assert!(length <= Self::NBITS as usize);
+		assert_eq!(length % BitmapChunk::LEN_BITS, 0);
+		writer.write_u8((length / BitmapChunk::LEN_BITS) as u8)?;
+
+		let count_pos = self.inner.iter().filter(|&v| v).count() as u32;
+		let count_neg = Self::NBITS - count_pos;
+		let threshold = Self::NBITS / 16;
+		if count_pos < threshold {
+			// Write positive indices
+			writer.write_u8(1)?;
```
antiochp, line 400: We should pull these `0, 1, 2` magic numbers out into const values or maybe an enum?

chain/src/txhashset/bitmap_accumulator.rs
```diff
@@ -239,3 +246,283 @@ impl Readable for BitmapChunk {
...
+
+impl Readable for BitmapBlock {
+	fn read<R: Reader>(reader: &mut R) -> Result<Self, ser::Error> {
+		let n_chunks = reader.read_u8()?;
+		if n_chunks as usize > BitmapBlock::NCHUNKS {
+			return Err(ser::Error::TooLargeReadErr);
+		}
+		let n_bits = n_chunks as usize * BitmapChunk::LEN_BITS;
+
+		let mode = reader.read_u8()?;
```
antiochp, line 432: If these magic numbers are an enum then we can read the enum here based on the `u8` value.
And then match on the enum below.

antiochp (approved):
This looks good. 
Minor comment about the positive/negative/raw_bytes magic numbers.

chain/src/txhashset/bitmap_accumulator.rs
jaspervdm: Makes sense, I've updated the PR.

Issue #3489, opened 2020-11-10
Increase "reorg cache" timer limit? 


deevope:
We should increase for every node the reorg_cache timer limit which is set by default to 30 minutes if it affects the mempool, so in case of future large reorg, nodes will rebroadcast automatically all unconfirmed transactions which were confirmed in the stale blocks.

If it only affects our local txs pool, we should consider to add the mempool in the reorg cache for the reason mentioned above.

antiochp:
> If it only affects our local txs pool, we should consider to add the mempool in the reorg cache for the reason mentioned above.

I don't understand what you mean here by "add the mempool in the reorg cache". Can you clarify?

deevope:
> I don't understand what you mean here by "add the mempool in the reorg cache". Can you clarify?

Sorry if this was unclear. I meant to add each transaction from the mempool in the reorg cache as they become confirmed on the blockchain. We will "age them out" after x minutes (x : timer limit). And in case of large reorg with empty block, nodes will rebroadcast them automatically if the cached txs were not contained in the reorg.

antiochp:
> I meant to add each transaction from the mempool in the reorg cache as they become confirmed on the blockchain.

We actually add txs to the reorg cache at the same time as we add them to the mempool.
If a tx is received and accepted by the mempool, it goes in the reorg cache as well. Its effectively just a record of every tx that we successfully added to the mempool (capped at rolling recent 30 mins currently).
So there should be no entries in the mempool that are not _also_ in the reorg cache.

A fork (or competing forks) will cause the mempool to be reconciled (every tx in the mempool is valid w.r.t current chain state). But txs remain in the reorg cache, independent of current chain state. This is by design, invalid txs may become valid, precisely so we can handle reorg like this.

----

> nodes will rebroadcast them automatically

We should check the current impl. I'm not entirely confident this happens today.
We definitely use the reorg cache to repopulate the local mempool in a reorg scenario. I'm just not sure if the node _also_ rebroadcasts these re-accepted txs.

Edit: Just checked and it does look like we simply add txs back to the mempool directly when reconciling the reorg cache. And this does _not_ cause the node to (re)broadcast these txs to peers.

This is probably a gap in our logic. For every tx in the reorg cache that is successfully added back to the mempool during reconciliation should _also_ be (re)broadcast out to our peers. We should treat these txs as "new" in this scenario.

antiochp:
I think a 24 hour period had been suggested in keybase for the reorg cache retention period.
[...]

PR #3495, opened 2020-11-22
reorg cache fix

deevope:
Related to #3489 

- **reorg cache period configurable** :  Allow everyone to configure themselves the reorg cache period, so they could increase it easily in function of their constraint. To repopulate their local mempool.

- ~~**treat cached txs as "new" during reconciliation** :  Allow our nodes to broadcast txs to their peers.~~
>  ~~_For every tx in the reorg cache that is successfully added back to the mempool during reconciliation should also be (re)broadcast out to our peers. We should treat these txs as "new" in this scenario._~~

- ~~**increase max pool size to 300K txs** : To prevent for the future, in case of large reorg with empty blocks, against stale full blocks, nodes mempool should be able to handle 6 hours of transactions of full blocks.~~

- ~~**Evict duplicate txs in reorg cache** : Given we add transactions back transaction to the pool, and in the pool we add them back in the reorg cache, we remove them from the reorg cache before they are sent in the pool.~~

- ~~**Evict only txs accepted to pool and sourced from reorg cache when rec…** : Fix of commit [3d8df3a](https://github.com/mimblewimble/grin/pull/3495/commits/3d8df3a7c82ecd8c0f05c25c87b05a7013957ee4), It will evict _only_ from reorg cache, transactions sourced from reorg cache, accepted by the pool to be broadcasted and added to the reorg_cache again. As we want to keep every transactions that were not accepted by the pool in the reorg_cache in the case where a reorg with larger depth happen again.~~

**test**: 
- ~~reorg on testnet made, node rebroadcasted successfully the transactions to other peers~~

- ~~I plan also to make a stress test on testnet with 50k transactions - not done yet.~~

### DIFF
--- a/pool/src/types.rs
+++ b/pool/src/types.rs
@@ -107,6 +107,11 @@ pub struct PoolConfig {
 	#[serde(default = "default_accept_fee_base")]
 	pub accept_fee_base: u64,
 
+	// Reorg cache retention period in minute.
+	// The reorg cache repopulates nodes mempool in a reorg scenario.
+	#[serde(default = "default_reorg_cache_period")]
+	pub reorg_cache_period: i64,
+
 	/// Maximum capacity of the pool in number of transactions
 	#[serde(default = "default_max_pool_size")]
 	pub max_pool_size: usize,
@@ -126,6 +131,7 @@ impl Default for PoolConfig {
 	fn default() -> PoolConfig {
 		PoolConfig {
 			accept_fee_base: default_accept_fee_base(),
+			reorg_cache_period: default_reorg_cache_period(),
 			max_pool_size: default_max_pool_size(),
 			max_stempool_size: default_max_stempool_size(),
 			mineable_max_weight: default_mineable_max_weight(),
@@ -135,6 +141,9 @@ impl Default for PoolConfig {
 
 fn default_accept_fee_base() -> u64 {
 	consensus::MILLI_GRIN
+}
+fn default_reorg_cache_period() -> i64 {
+	30
 }
 fn default_max_pool_size() -> usize {
 	50_000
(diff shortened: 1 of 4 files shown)


deevope:
@antiochp There's one problem on my PR, I'm working to solve it. I should remove from reorg_cache only txs that are sourced from reorg cache and accepted by the pool to be broadcasted again, and as well added back to the reorg cache. Currently it removes every txs in reorg cache once a reorg happen. As we want to keep every transactions that were not accepted by the pool in the reorg_cache for the case where a reorg with larger depth happen again.

deevope:
@antiochp Everything should be fine now and it's been tested on my side. 👍

pool/src/types.rs
```diff
@@ -107,6 +107,11 @@ pub struct PoolConfig {
 	#[serde(default = "default_accept_fee_base")]
 	pub accept_fee_base: u64,
 
+	// Reorg cache retention period in minute.
+	// The reorg cache repopulates local mempool in a reorg scenario.
+	#[serde(default = "default_reorg_cache_period")]
+	pub reorg_cache_period: i64,
```
antiochp, line 113: Lets make this `u64` as it does not make sense to support `-ve` values in the config.

antiochp (commented):
Yes this looks good.
We should investigate the various re-broadcast strategies that have been discussed recently. But exposing this config now makes a lot of sense.

Minor comment on `i64` vs `u64`.

:+1:

pool/src/types.rs
deevope: I can't do `u64`, as [minutes()](https://docs.rs/chrono/0.4.11/chrono/struct.Duration.html#method.minutes) requires `i64`. Either we set `u32` and turn it to `i64` in minutes() or we just let `i64` but no error will be thrown if users put a negative value.

antiochp: `u32` works then, we're not going to need huge values in here.

antiochp (commented):
:+1: Looks good.

antiochp (approved):

antiochp:
Going to merge this for 5.0.0 beta.

PR #3496, opened 2020-11-22
PIBD segment p2p messages

jaspervdm:
This PR introduces the new p2p messages as defined in [this RFC](https://github.com/mimblewimble/grin-rfcs/pull/68). Only merge after https://github.com/mimblewimble/grin/pull/3482 and https://github.com/mimblewimble/grin/pull/3492. 
Until then we use a placeholder `Segmenter` that just panics on usage.

- ~Enforce segment height limits~
- ~Rebase once 3482 is merged~
- ~Switch to `BitmapSegment` in place of `Segment<BitmapChunk>` in `OutputBitmapSegment` responses~

### DIFF
--- a/p2p/src/protocol.rs
+++ b/p2p/src/protocol.rs
@@ -287,6 +287,16 @@ impl MessageHandler for Protocol {
 
 				Consumed::Attachment(Arc::new(meta), file)
 			}
+
+			Message::GetOutputBitmapSegment(_)
+			| Message::OutputBitmapSegment(_)
+			| Message::GetOutputSegment(_)
+			| Message::OutputSegment(_)
+			| Message::GetRangeProofSegment(_)
+			| Message::RangeProofSegment(_)
+			| Message::GetKernelSegment(_)
+			| Message::KernelSegment(_) => Consumed::None,
+
 			Message::Unknown(_) => Consumed::None,
 		};
 		Ok(consumed)
--- a/p2p/src/protocol.rs
+++ b/p2p/src/protocol.rs
@@ -17,8 +17,8 @@ use crate::conn::MessageHandler;
 use crate::core::core::{hash::Hashed, CompactBlock};
 
 use crate::msg::{
-	Consumed, Headers, Message, Msg, PeerAddrs, Pong, SegmentRequest, SegmentResponse,
-	SegmentResponseWithRoot, TxHashSetArchive, Type,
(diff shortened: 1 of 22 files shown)


antiochp:
#3482 is now merged.

antiochp:
:eyes:

p2p/src/protocol.rs
```diff
@@ -287,6 +290,92 @@ impl MessageHandler for Protocol {
...
+					)?)
+				} else {
+					Consumed::None
+				}
+			}
+			Message::OutputBitmapSegment(_)
+			| Message::OutputSegment(_)
+			| Message::RangeProofSegment(_)
+			| Message::KernelSegment(_) => Consumed::None,
+
```
antiochp, line 378: This is just a `/dev/null` placeholder for now?

servers/src/common/adapters.rs
```diff
@@ -477,6 +482,66 @@ where
 	fn get_tmpfile_pathname(&self, tmpfile_name: String) -> PathBuf {
 		self.chain().get_tmpfile_pathname(tmpfile_name)
 	}
+
+	fn get_kernel_segment(
+		&self,
+		hash: Hash,
+		id: SegmentIdentifier,
+	) -> Result<Segment<TxKernel>, chain::Error> {
+		if id.height < 9 || id.height > 13 {
```
antiochp, line 491: We should pull these min and max heights out into `const` so its easier to see them all together.

antiochp (approved):
Looks good.
Minor comment about the min/max values.

p2p/src/protocol.rs
jaspervdm: Yes, nodes receiving segments just ignore them.

servers/src/common/adapters.rs
jaspervdm: Done

servers/src/common/adapters.rs
```diff
@@ -46,6 +46,12 @@ use crate::util::OneTime;
 use chrono::prelude::*;
 use chrono::Duration;
 use rand::prelude::*;
+use std::ops::Range;
+
+const KERNEL_SEGMENT_HEIGHT_RANGE: Range<u8> = 9..14;
```
antiochp, line 51: Nice. Didn't think of using ranges here. :+1:

PR #3497, opened 2020-11-23
use extend_from_slice when appending multiple hashes to hash file

antiochp:
This PR cleans up the `Backend` trait, passing a slice of hashes to `append()`.

Pass `&[Hash]` into `append()` rather than `Vec<Hash>`.
This lets us call `extend_from_slice()` in `VecBackend` without cloning the vec of hashes.
In PMMR backend we can use the slice directly and append multiple hashes.


jaspervdm (approved):
Looks like a good improvement :+1:

PR #3498, opened 2020-11-23
try and call refreshenv before running tests on windows

antiochp:
Try and call `refreshenv` as part of the script when running tests on windows (env is not shared across tasks).

### DIFF
--- a/.ci/test.yml
+++ b/.ci/test.yml
@@ -1,5 +1,7 @@
 steps:
-  - script: 'cargo test --all'
+  - script: |
+      refreshenv
+      cargo test --all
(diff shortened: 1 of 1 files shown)


antiochp:
This might look like it worked on CI but it did not actually run the tests on windows...

Still investigating.

antiochp:
Ok i think we got past it this time - 

```
   Compiling croaring-sys-mw v0.4.5
   Compiling hyper v0.13.6
```

antiochp:
Going to merge. This is a single line change to the windows test pipeline.

PR #3499, opened 2020-11-23
we have no patch file and nothing to patch

antiochp:
Remove legacy code in the azure pipeline config that runs `git apply .ci/win.patch` as this file no longer exists.


quentinlesceller (approved):

quentinlesceller:
Indeed forgot about that. LGTM.

PR #3501, opened 2020-11-25
fix calls to next_diff() in tests to use next height correctly

antiochp:
Related #3477 

We had some test failures related to bad (legacy) test setup code.
We need to use the correct next height when determining the next difficulty, when building blocks in the tests.

Also had some tests just using `Difficulty::min_dma()` but this does not work beyond HF4.
So reworked those tests to use `next_header_info.difficulty`.


tromp (approved):
good fixes; would be better still to have Block::new accept both difficulty and secondary_scale

PR #3502, opened 2020-11-25
make min_wtema difficulty chain type dependent

tromp:
---
name: fix min_wtema
about: Make minimum wtema difficulty chain type dependent, to prevent slow test mining
title: ''
labels: ''
assignees: @antiochp 

---


antiochp (approved):
Looks good to me. :+1:

Testing locally then will merge.

PR #3503, opened 2020-11-25
Disable API v1

quentinlesceller:
As scheduled, the Node v1 API is now disabled.
A subsequent cleanup should be done but this is the minimal to do without causing too much disruption.


antiochp (commented):
Testing locally but looks ok so far.

antiochp (approved):
Just did some quick "does v2 still work" sanity checks.
All looks good.

:+1:

PR #3504, opened 2020-11-25
fix, no more FULL_NODE

antiochp:
Not really sure how this happened but this PR cleans up a reference to `FULL_NODE` which no longer exists.

Edit: It was added in #3490 and should not have been as `FULL_NODE` no longer exists, we should be using `Capabilities::default()`.


antiochp:
Merging. This is a 1 line fix to fix tests on master.

PR #3507, opened 2020-11-26
refreshenv as part of windows release script

antiochp:
We need to run `refreshenv` after installing `llvm` as part of the windows release script.
Related #3498.


antiochp:
Waiting for CI to complete before merging. This code is not actually exercised unless we run the release process. 
But I'd rather not accidentally break anything else right now.

PR #3521, opened 2020-12-11
prefer outbound peers when syncing

### LINKED ISSUE #3518: no peers available, disabling sync (finally explained, inbound vs outbound inconsistency)
Thanks to @bladedoyle I think we tracked down one scenario where we see the following regularly in the logs - 

```
20201211 16:04:08.807 WARN grin_servers::grin::sync::syncer - sync: no peers available, disabling sync
```

When syncing we find "max diff" from _all_ our peers (inbound _and_ outbound) then we randomly select an _outbound_ peer with that diff to sync from.
But we can be in a situation where "max diff" is from an inbound peer without any outbound peers currently being at that diff (say a new block was very recently found) and our peers had not yet updated local diff via ping/pong (every 5-10secs).

----

We only want to sync from peers we know we reached out to.

Need to think through how best we can approach this.

antiochp:
Resolves #3518 

This PR fixes an edge case that we appear to be hitting on `testnet` due to low number of nodes and overall network health.

There was an inconsistency during sync where we would see a `max_diff` on an _inbound_ peer that would trigger sync but then we would attempt to sync against a randomly chosen _outbound_ peer. But it is possible that no outbound peers have sufficient difficulty to handle the sync request.

When checking for known `max_diff` we need to ensure we check against a set of peers that is consistent with those that we will actually sync from.

We want to prioritize syncing against _outbound_ peers whenever possible.
But we we also want to handle the case where an _inbound_ peer advertises more work than any other connected peers - in this case we will sync against that _inbound_ peer, but only if no outbound peers are known to have the same total work.

Outbound peers are considered more reliable as they are more likely to be long lived (and likely with faster connections and more bandwidth).
Related discussion - https://github.com/bitcoin/bitcoin/issues/5315#issuecomment-63675901

Rules are as follows - 
* find `max_diff` across _all_ connected peers (inbound and outbound)
* randomly choose an _outbound_ peer with `diff >= max_diff`
* if no _outbound_ peer then randomly choose _inbound_ peer with `diff >= max_diff`

This means we prefer outbound peers when syncing. And we sync against inbound peers only when necessary.

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -26,6 +26,7 @@ use std::sync::{mpsc, Arc};
 use std::{cmp, str, thread, time};
 
 use crate::core::global;
+use crate::core::pow::Difficulty;
 use crate::p2p;
 use crate::p2p::types::PeerAddr;
 use crate::p2p::ChainAdapter;
@@ -172,15 +173,10 @@ fn monitor_peers(
 		total_count += 1;
 	}
 
-	let peers_count = peers.iter().connected().count();
-
-	let max_diff = peers.max_peer_difficulty();
-	let most_work_count = peers
-		.iter()
-		.outbound()
-		.with_difficulty(|x| x >= max_diff)
-		.connected()
-		.count();
+	let peers_iter = || peers.iter().connected();
+	let peers_count = peers_iter().count();
+	let max_diff = peers_iter().max_difficulty().unwrap_or(Difficulty::zero());
+	let most_work_count = peers_iter().with_difficulty(|x| x >= max_diff).count();
 
 	debug!(
 		"monitor_peers: on {}:{}, {} connected ({} most_work). \
--- a/servers/src/grin/sync/body_sync.rs
+++ b/servers/src/grin/sync/body_sync.rs
@@ -93,14 +93,25 @@ impl BodySync {
 		let head = self.chain.head()?;
 
 		// Find connected peers with strictly greater difficulty than us.
-		let peers: Vec<_> = self
-			.peers
-			.iter()
-			.outbound()
-			.with_difficulty(|x| x > head.total_difficulty)
-			.connected()
-			.into_iter()
-			.collect();
+		let peers_iter = || {
+			self.peers
+				.iter()
+				.with_difficulty(|x| x > head.total_difficulty)
+				.connected()
+		};
+
+		// We prefer outbound peers with greater difficulty.
+		let mut peers: Vec<_> = peers_iter().outbound().into_iter().collect();
+		if peers.is_empty() {
+			debug!("no outbound peers with more work, considering inbound");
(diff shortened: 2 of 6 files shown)


antiochp:
Related https://github.com/bitcoin/bitcoin/issues/5315
Specifically https://github.com/bitcoin/bitcoin/issues/5315#issuecomment-63675901

> It's not a strong guarantee, but typically large block chain fetches come from incoming and not from outgoing connections.

servers/src/grin/seed.rs
```diff
@@ -172,15 +173,10 @@ fn monitor_peers(
...
-
-	let max_diff = peers.max_peer_difficulty();
-	let most_work_count = peers
-		.iter()
-		.outbound()
-		.with_difficulty(|x| x >= max_diff)
-		.connected()
-		.count();
+	let peers_iter = || peers.iter().connected();
+	let peers_count = peers_iter().count();
```
j01tz, line 177: The change to using empty closures here and in other files when iterating peers is just a rust convenience thing?

j01tz (approved):
:+1: on the concept and initial review. Tested for a quick sync but not likely under the troublesome conditions. As mentioned elsewhere, we probably want to revisit this area after the HF to ensure we have a clear big-picture robustness with good edge case security wrt rules around incoming/outgoing peer connections.

servers/src/grin/seed.rs
antiochp: We need to build a couple of instances of the same (or similar) iterator. 
And the closure just makes this a lot easier to keep them consistent.

servers/src/grin/sync/body_sync.rs
```diff
@@ -93,14 +93,25 @@ impl BodySync {
...
+		// We prefer outbound peers with greater difficulty.
+		let mut peers: Vec<_> = peers_iter().outbound().into_iter().collect();
+		if peers.is_empty() {
+			warn!("no outbound peers with more work, considering inbound");
+			peers = peers_iter().inbound().into_iter().collect();
+		}
+
+		// If we have no peers (outbound or inbound) then we are done for now.
+		if peers.is_empty() {
+			warn!("no peers (inbound or outbound) with more work");
```
quentinlesceller, line 112: Nitpicking but why `warn` here. This is going to happen all the time no?

quentinlesceller (approved):
Looking good. Minor comment regarding the logs. Just synced with it.

jaspervdm (approved):
Looks good. Don't have anything to add that wasn't mentioned before.

servers/src/grin/sync/body_sync.rs
antiochp: Let me think about this a bit - it only happens once per actual transition out of "sync mode" as we only run body sync when we are actively syncing.

PR #3523, opened 2020-12-14
[DNM] Remove v1 node api code

quentinlesceller:
This is large cleanup of the unused v1 API code following v5.0.0 removal.
I've also splitted the `rest.rs` file to:
- `error.rs` that contains the error types.
- `api.rs` that contains basic things to make the API works.


quentinlesceller:
Might want to put that one on hold until HF.

antiochp:
Yeah can we do this on master post HF?

PR #3524, opened 2020-12-14
Update Client to v2 API

### LINKED ISSUE #3522: Client use v1 API
The current `grin client` uses the V1 API which is now removed.
Convert it to v2. https://github.com/mimblewimble/grin/blob/v5.0.0-beta.2/src/bin/cmd/client.rs#L159v

quentinlesceller:
As titled. Fixes #3522.
JSON-RPC code is taken from https://github.com/mimblewimble/grin-wallet/pull/329 (json-rpc.rs). Might be good to use the same code for both in the future.


antiochp:
Looks like there are build issues related to the `json!` macro.

jaspervdm (approved):
👍 Tested locally and behaves as expected.

antiochp (commented):
Changes look fine.
I'm testing this locally just to sanity check the various `client` api calls.

antiochp (approved):
:+1:

antiochp:
Merging this. My understanding is this is the last PR outstanding for `5.0.x`.

PR #3525, opened 2020-12-14
better naming of APIs file and variables

deevope:
Related to PR https://github.com/mimblewimble/grin-wallet/pull/544
- Changed name of  `.foreign_api_secret` to `.node_foreign_api_secret` and `.api_secret` to `.node_owner_api_secret`
- Better naming of node `owner_api` variables


jaspervdm:
I like the new names however we should consider how it breaks current setups and think if/how we can have some type of backwards compatibility.

antiochp:
I believe we are using a limited fix from https://github.com/mimblewimble/grin-wallet/pull/546 for `5.0.x`.
We can revisit the naming post HF.

deevope:
@quentinlesceller feel free to review when you have some time 👍

PR #3514, opened 2020-12-08
Update tokio, hyper versions

jaspervdm:
Update tokio and hyper

PR #3526, opened 2020-12-15
Merge current/5.0.x on master

quentinlesceller:
Applies:
https://github.com/mimblewimble/grin/pull/3514
https://github.com/mimblewimble/grin/pull/3520.


quentinlesceller:
I think we want to go over each commit/pr since the split instead.

PR #3532, opened 2020-12-23
[5.0.x] Fix PIBD segments

jaspervdm:
I performed a bunch of tests on testnet using other 5.0.x nodes and found a couple of bugs in the code that generates/validates PIBD segments
- When generating the bitmap segments, we incorrectly assumed that the segments were always full, resulting in a panic
- We forgot a negation in the range proof segment height check
- During validation of bitmap segments, we need to hash the `output|bitmap` root with the output MMR size, not the bitmap MMR size, so we need to add an argument to the `validate_with` functions. This does not affect current nodes since they don't do any validation but we might as well fix it now.


antiochp (approved):
:+1: looks good. Thanks for catching these!

PR #3539, opened 2021-01-06
override old accept_fee_base setting of 1000000 to new default 500000

tromp:
override old accept_fee_base setting of 1000000 to new default 500000
so node operators needn't update their grin-server.toml in order to respect the new fee regime.


antiochp (approved):
:+1:
As discussed - its not elegant but it avoids needing to update config network wide.
To be revisited once we have a robust solution for writing/rewriting config during node upgrades.

PR #3544, opened 2021-01-12
[5.0.x] Do not use serde private export

quentinlesceller:
Fixes https://github.com/mimblewimble/grin/issues/3543


antiochp (approved):
Makes sense. :+1: if everything is still working.

quentinlesceller:
This still makes sense for a bug fix release. LGTM.

PR #3549, opened 2021-01-20
[5.0.x] Safety: check length during deserialization

jaspervdm:
Investigation into https://github.com/mimblewimble/grin-wallet/issues/576 led me to the `secp_ser` code, where we attempt to take a byte slice of length 32. If the original vec is shorter, attempting to take a slice of that length will result in a panic. To prevent this we should first check the length of the vector and return an error if its too short.

We do something similar in 2 other places so I decided to fix those too.


quentinlesceller (approved):
Makes sense. Can you port it to master too?

Issue #3516, opened 2020-12-11
Server Cannot Run Because of NotFoundErr


garyyu:
Tried multiple times by clean the whole `chain_data` folder, and eventually the `testnet` synced with latest 5.0.0-beta.2 release. but after a quit and it can not run again because of a not found error:

```
20201211 10:16:29.485 INFO grin_servers::grin::server - Starting server, genesis block: edc758c1370d
20201211 10:16:31.892 ERROR grin_util::logger -
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Chain(Error { inner: NotFoundErr("Block with hash: 0e29ee5b51f1")

Store Error: NotFoundErr("Block with hash: 0e29ee5b51f1"), reason: DB Not Found Error: Block with hash: 0e29ee5b51f1 })': src/bin/cmd/server.rs:58   0: grin_util::logger::send_panic_to_log::{{closure}}
   1: std::panicking::rust_panic_with_hook
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:573
   2: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:476
   3: std::sys_common::backtrace::__rust_end_short_backtrace
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/sys_common/backtrace.rs:153
   4: rust_begin_unwind
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/std/src/panicking.rs:475
   5: core::panicking::panic_fmt
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/panicking.rs:85
   6: core::option::expect_none_failed
             at /rustc/18bf6b4f01a6feaf7259ba7cdae58031af1b7b39/library/core/src/option.rs:1221
   7: grin::cmd::server::start_server_tui
   8: grin::cmd::server::server_command
   9: grin::real_main
  10: grin::main
```

tromp:
That block should be available. I can see it here at pre-fork height 641520:

https://floonet.grinscan.net/block/641520

garyyu:
Yes, reproducible.

[...]

PR #3555, opened 2021-02-03
fix for missing block under certain startup conditions

antiochp:
This PR fixes a "cannot find block" error that can occur during node startup.
We have some code that runs during node initialization that exercises the PIBD segmenter as we wanted to keep this code "in use".
There is an edge case during startup that we did not account for - if the node is shutdown while it is in the process of syncing then on next startup the chain can find itself in a state where traversing back to the "archive header" (our 720 block archival period) can fail as we do not have the full set of blocks for this time period yet.
This PR simply ensures the code that touches the segmenter does not fail in this situation (during startup).

This is related to a jump from one 720 block archive period to the next due to chain height, during sync. It is possible for the jump to result in "missing" blocks as we are still syncing. This is not a problem with data integrity or missing data - its just that we were making a bad assumption during node initialization without accounting for this scenario.

Resolves https://github.com/mimblewimble/grin/issues/3516

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -236,8 +236,7 @@ impl Chain {
 		// In reality we will do this based on PIBD segment requests.
 		// Initialization (once per 12 hour period) will not be this slow once lmdb and PMMRs
 		// are warmed up.
-		{
-			let segmenter = chain.segmenter()?;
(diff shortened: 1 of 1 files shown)


antiochp:
Going to merge this. Its a small change to node init code, making it more failure tolerant.

PR #3557, opened 2021-02-03
[5.0.x] fix for missing block under certain startup conditions

antiochp:
Porting https://github.com/mimblewimble/grin/pull/3555 across from `master`.
This is worth including if we release a new version on `5.0.x` branch.


antiochp:
Merging. Same as #3555 on `master`.

PR #3558, opened 2021-02-03
Block sync hash traversal perf

### LINKED ISSUE #3554: Performance issue when syncing historical blocks
We identified a performance issue during sync, specifically when we are requesting full blocks with a large difference between the height of the current header and the height of the full block.
Specifically this is exacerbated during an experimental "full archive sync" when header height is `1_073_555` and block height is `1`. We iterate back over the headers to identify which full blocks are missing, so we know which to request. This is _slow_ when we iterate over a million headers.

Related https://github.com/mimblewimble/grin/issues/3552

----

We know the height of the current chain head and we know all headers beyond this (full blocks not yet sync'd).
It should be fairly easy to rework this logic to iterate _forwards_ from the current chain head (last full block), taking advantage of `get_header_by_height()` to do this.

While this is particularly noticeable when syncing historical blocks I suspect the performance issue is still present during a regular "fast sync" for the most recent 48 hours of full blocks. There is no need to continually keep iterating back over all the headers in this "recent history" time period.

Planning to fix this issue as a standalone task.

antiochp:
Resolves #3554.

When we transition from "header sync" to "body sync" we identify missing blocks based on the delta between the chain of headers and the current chain of full blocks. i.e. We sync headers and then we go request the full blocks for each of those headers.
There is some additional complexity where we support a fork between the header chain and the full block chain as one is not always a subset of the other. And to do this we identify the "fork point" which is the last common header between header chain and full block chain before they begin to diverge.

----

Fork example - 

Headers: `A -> B -> C' -> D -> E`

Full blocks: `A -> B -> C`

Header sync gave us a chain of headers building on `C'` while our local node was previously aware of `C`.
We need to "rewind" to the fork point `B` immediately prior to `C` and `C'` and request missing blocks `[C', D, E]`.

----

The existing implementation rewinds back from the head of the header chain to identify both the fork point and a set of missing block hashes. This is fine if the delta between header chain and full block chain is small but is not suitable if the header chain must be rewound a significant number of blocks.

It fails badly in the "full archval node sync" scenario where we must rewind from height ~1,500,00 back to 1.

----

The PR introduces a more efficient approach -

1. identify height of full block chain
2. increase this height by the number of blocks we will request in parallel
3. lookup this "max" header directly on header chain via `get_header_by_height()`
4. Iterate back from this "max" header to the fork point.

This places an upper bound on the iteration required as we can lookup headers directly based on height.
This greatly reduces the amount of iteration required when identifying missing blocks to sync.

----

As part of this rework `check_txhashset_needed()` was refactored, splitting out `get_fork_point()` and `check_txhashset_needed()` into separate fns.
This allowed the implementation to be simplified significantly.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -15,7 +15,7 @@
 //! Facade and handler for the rest of the blockchain implementation
 //! and mostly the chain pipeline.
 
-use crate::core::core::hash::{Hash, Hashed, ZERO_HASH};
+use crate::core::core::hash::{Hash, Hashed};
 use crate::core::core::merkle_proof::MerkleProof;
 use crate::core::core::verifier_cache::VerifierCache;
 use crate::core::core::{
@@ -38,7 +38,6 @@ use crate::util::RwLock;
 use crate::ChainStore;
 use grin_core::ser;
 use grin_store::Error::NotFoundErr;
-use std::cmp::min;
 use std::fs::{self, File};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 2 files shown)


antiochp:
Going to merge this to `master`. 
We need to make sure we test sync thoroughly as part of the next release.

Issue #3553, opened 2021-02-02
Rate limit outbound p2p messages


antiochp:
Exploration/investigation into full block archive sync identified an issue relating to banning abusive peers.

Related https://github.com/mimblewimble/grin/issues/3552

We ban a peer if we receive (or send) excessive numbers of messages to/from that peer within a 60s time period.

The limit is 500 messages within 60s. If we see more than this then the peer is flagged as abusive and eventually banned.

It makes sense to monitor this for inbound received messages but I'm not convinced this is the right thing to do when we initiate requests by sending messages to peers. 
Specifically this is an issue if we request large numbers of blocks from a small number of peers (say those that support full archive mode) as we can easily hit the 500 msg limit when requesting historical blocks.

Proposal is to limit the abusive peer handling to _inbound_ msgs only. If we engage in abusive behavior when sending messages to a peer then they will react accordingly. We do not need to ban them for something we are responsible for.

We would need to monitor (and rate limit) our outbound msg rates to ensure we did not exceed peer limits.
We can do this by inspecting the `Tracker` (msg rates) associated with each peer and potentially rate limit by introducing a small delay across large numbers of outbound "get_block" requests.

PR #3560, opened 2021-02-10
add rate limiting to outbound p2p msg sending

antiochp:
Resolves https://github.com/mimblewimble/grin/issues/3553

We treat peers as abusive if inbound or outbound msg rates exceed 500/min.

This PR modifies this to only consider inbound msg rate. 
We do not want to treat a peer as abusive if _we_ are responsible for sending lots of p2p messages.

Add basic "rate limiting" to `write_message()` by inspecting the previous timestamp tracked by the tracker.
We ensure 150ms has elapsed based on previous sent message before attempting to send another.
In normal operation this is rarely the case. We occasionally send a ping and a block header close together and these are now separated by 150ms.
The scenario where this _does_ come into play is during "fast sync" when requesting large numbers of full blocks from a small number of peers. Specifically "block archival sync" against archival nodes.
We no longer flood the peer connection with hundreds of block requests and these are now spaced out to avoid hitting the abusive peer limit on the receiving end.

----

As part of this PR cleaned up the readonly usage of `tracker`.
The following fns have been removed, replaced with exposing a simple `tracker()` accessor - 
* `last_min_sent_bytes`
* `last_min_received_bytes`
* `last_min_message_counts`

For example - 

```
    let received = peer.tracker().received_bytes.read().count_per_min();
```

This makes the read locks more explicit and removes a case where we take multiple read locks in one go.

----

Introduced `elapsed_since_last_msg()` to `RateCounter` for convenience. 
This simplifies the implementation of the rate limiting logic.

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -34,10 +34,10 @@ use crate::types::{
 use crate::util::secp::pedersen::RangeProof;
 use bytes::Bytes;
 use num::FromPrimitive;
-use std::fmt;
 use std::fs::File;
 use std::io::{Read, Write};
 use std::sync::Arc;
+use std::{fmt, thread, time::Duration};
 
 /// Grin's user agent with current version
 pub const USER_AGENT: &str = concat!("MW/Grin ", env!("CARGO_PKG_VERSION"));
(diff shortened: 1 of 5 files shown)


antiochp:
Going to merge.
Inbound "abusive" peer checking is unchanged. This only applies to _outbound_ traffic, which we should not be flagging as abusive.

PR #3563, opened 2021-02-15
BLOCK_HIST capability enabled if archive_mode=true

### LINKED ISSUE #3562: Add BLOCK_HIST capability and enable it if archive_mode=true
We want to advertise an additional `BLOCK_HIST` capability if we are running with `archive_mode=true`.
This will enable other peers to request historical blocks from us along with other peers that advertise support for this.

Related #3552.

antiochp:
Resolves #3562

### DIFF
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -391,6 +391,8 @@ bitflags! {
 		const TX_KERNEL_HASH = 0b0000_1000;
 		/// Can provide PIBD segments during initial byte download (fast sync).
 		const PIBD_HIST = 0b0001_0000;
+		/// Can provide historical blocks for archival sync.
+		const BLOCK_HIST = 0b0010_0000;
 	}
 }
 
--- a/servers/src/grin/server.rs
+++ b/servers/src/grin/server.rs
@@ -214,9 +214,13 @@ impl Server {
 			init_net_hooks(&config),
 		));
 
-		// Use our default capabilities here.
-		// We will advertize these to our peers during hand/shake.
-		let capabilities = Capabilities::default();
+		// Initialize our capabilities.
+		// Currently either "default" or with optional "archive_mode" (block history) support enabled.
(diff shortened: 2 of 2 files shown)


p2p/tests/ser_deser.rs
```diff
@@ -81,6 +81,6 @@ fn test_capabilities() {
 fn test_pibd_capabilities() {
 	assert_eq!(
 		p2p::types::Capabilities::default() | p2p::types::Capabilities::PIBD_HIST,
-		p2p::types::Capabilities::from_bits_truncate(0b11111111 as u32),
+		p2p::types::Capabilities::from_bits_truncate(0b00011111 as u32),
```
phyro, line 84: If I understand it correctly, the above value does an OR of default+PIBD bitmask. If I'm reading it correctly, the default already includes PIBD_HIST https://github.com/mimblewimble/grin/blob/010730e6daaa268a8316cc702c51fcb408fa0a6d/p2p/src/types.rs#L406
So maybe we can remove the `| p2p::types::Capabilities::PIBD_HIST` and correct the comment above the test which tells us that it is not in the defaults.

antiochp: Oh good catch. 👍 
I'll update in a separate PR (unrelated to this).

antiochp: https://github.com/mimblewimble/grin/pull/3565

antiochp: I'll clean this PR up once https://github.com/mimblewimble/grin/pull/3565 is merged.

phyro:
would it make sense to add an assertion that it does not contain the BLOCK_HIST capability here?https://github.com/mimblewimble/grin/blob/master/p2p/tests/capabilities.rs#L44
or maybe in ser_deser.rs. Anyway, this looks good to me and don't want to stall it

phyro (approved):

DavidBurkett (approved):

Issue #3561, opened 2021-02-11
Node peers list requests, and handling of them


bladedoyle:
When my grin node starts it gets initial peers from dns and attempts to connect to each:
```
20210211 13:21:12.468 DEBUG grin_servers::grin::seed - Retrieving addresses from dns mainnet.seed.grin.icu:3414
20210211 13:21:12.516 DEBUG grin_servers::grin::seed - Retrieving addresses from dns mainnet.seed.713.mw:3414
...
20210211 13:21:13.080 TRACE grin_p2p::serv - connect_peer: on 0.0.0.0:3414. connecting to 157.230.142.11:3414
```
25 initial peer addresses are returned.

My node then adds the ones it can connect to as peers:
```
20210211 13:21:13.100 DEBUG grin_p2p::peer - connect: handshaking with Ok(157.230.142.11:3414)
20210211 13:21:13.117 DEBUG grin_p2p::store - save_peer: PeerAddr(157.230.142.11:3414) marked Healthy
```

All good so far.

For each of those peers my node then requests a peers list:
```
20210211 13:21:13.118 TRACE grin_p2p::peer - Asking 157.230.142.11:3414 for more peers PEER_LIST
```
Each peer returns a peer address list containing 256 entries.
My node then adds all of them to my peers database without checking to see if they are any good.
see: https://github.com/mimblewimble/grin/blob/298accb6f382ab9e96c73ea3fae70ab88d2372a9/p2p/src/peers.rs#L668

This could result in as many as 25 * 256 = 6400 peers added.
Each peer has to be added to the lmbd, then later almost all of them are found to be bad and removed.  Thats possibly many thousands of lmdb write lock requests within a second or two.
This flood of write lock requests overloads the lmdb and on a slow disk some peers db lock requests exceed the timeout and report "failed to get peers lock".

In practice some of the addresses in the peers lists are duplicates.
A quick test shows that 2741 (unique) peers are added to my peers database within the first second of starting a new node.
Of those 2741 peers added, only 57 accept a connection.
Within the first 30 seconds, 8357 (unique) peers are processed through my peers database.

DavidBurkett:
[...]

PR #3566, opened 2021-02-16
[5.0.x] inefficient locking on recv of peers lists can result in failure to get peers lock

bladedoyle:
see: https://github.com/mimblewimble/grin/issues/3561

This change batches the write of PEER_LIST results into the peers lmdb to increase efficiency and reduce lock timeouts.

20210215 11:28:20.196 ERROR grin_p2p::peers - add_connected: failed to get peers lock

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -295,6 +295,11 @@ impl Peers {
 		self.store.save_peer(p).map_err(From::from)
 	}
 
+	/// Saves updated information about mulitple peers in batch
+	pub fn save_peers(&self, p: Vec<PeerData>) -> Result<(), Error> {
+		self.store.save_peers(p).map_err(From::from)
+	}
+
 	/// Updates the state of a peer in store
 	pub fn update_state(&self, peer_addr: PeerAddr, new_state: State) -> Result<(), Error> {
 		self.store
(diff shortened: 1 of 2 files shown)


DavidBurkett (approved):

p2p/src/store.rs
```diff
@@ -127,6 +127,15 @@ impl PeerStore {
 		batch.commit()
 	}
 
+	pub fn save_peers(&self, p: Vec<PeerData>) -> Result<(), Error> {
```
phyro, line 130: would it make sense to also redefine the `save_peer` above to delegate the call to `save_peers` for a vector of size 1?

p2p/src/peers.rs
```diff
@@ -682,9 +688,10 @@ impl NetAdapter for Peers {
 				ban_reason: ReasonForBan::None,
 				last_connected: Utc::now().timestamp(),
 			};
-			if let Err(e) = self.save_peer(&peer) {
-				error!("Could not save received peer address: {:?}", e);
-			}
+			to_save.push(peer);
+		}
+		if let Err(e) = self.save_peers(to_save) {
```
phyro, line 693: Is there anything wrong if we end up saving no peers because this one failed?

DavidBurkett: It's not ideal, but how would you know which one failed?

DavidBurkett (approved):

p2p/src/store.rs
bladedoyle: That would require every call to save_peer to create a Vec and add the single item.  Im not convinced thats better than having a few lines of similar code.

DavidBurkett: I think the suggestion is to make save_peer just a wrapper around save_peers, instead of duplicating the batch DB code.

bladedoyle:
@antiochp    I am happy with this change.  Please merge or reject when you get a chance.

phyro (approved):

antiochp (approved):
This looks good to me. :+1:
I'm planning to test this locally and will merge once done.

PR #3568, opened 2021-02-18
rework prune rewrite with iterators

antiochp:
This PR refactors the handling of `pos_to_rm` when pruning during compaction.
We used to pass a slice `&[u64]` around and this has been reworked to accept an `IntoIterator<Item = u64>`.
This allows us to avoid `collect()` in various places with intermediate memory allocation.

Split `save_prune()` into two functions for flexibility so we can pass an iterator into the earlier `write_tmp_pruned()` without needing `&mut self` (as the iterator passed in has a reference to immutable self).

### DIFF
--- a/store/src/lib.rs
+++ b/store/src/lib.rs
@@ -25,7 +25,6 @@ extern crate log;
 use failure;
 #[macro_use]
 extern crate failure_derive;
-#[macro_use]
 extern crate grin_core as core;
 extern crate grin_util as util;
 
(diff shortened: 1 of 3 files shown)


antiochp:
This is purely a refactor. We have decent test coverage. Testing locally to ensure we can still prune/compact successfully.
Will merge if local testing is good.

antiochp:
```
20210219 11:37:56.842 DEBUG grin_chain::txhashset::txhashset - txhashset: starting compaction...
20210219 11:38:23.665 DEBUG grin_chain::txhashset::txhashset - txhashset: check_compact output mmr backend...
20210219 11:38:26.931 DEBUG grin_chain::txhashset::txhashset - txhashset: check_compact rangeproof mmr backend...
20210219 11:38:31.452 DEBUG grin_chain::txhashset::txhashset - txhashset: ... compaction finished
```

PR #3573, opened 2021-02-23
Retire pruned cache

antiochp:
Our "prune list" implementation maintains an internal "prune cache".
* The prune list consists of roots of fully pruned subtrees.
* The cache contains all pos beneath these roots.
* We check if a pos was pruned by looking in the cache.

We can improve this without needing to maintain an internal cache.

To check if a pos is pruned we can look for the "next" pruned subtree and check if pos is beneath it.
We only need to consider the "next" subtree due to the append-only PMMR semantics.
If the next subtree does not include pos then we know that no other subtree will either.

### DIFF
--- a/core/src/core/pmmr/pmmr.rs
+++ b/core/src/core/pmmr/pmmr.rs
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::u64;
-use std::{marker, ops::RangeInclusive};
+use std::{marker, ops::Range, u64};
 
 use croaring::Bitmap;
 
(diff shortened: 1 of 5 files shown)


core/src/core/pmmr/pmmr.rs
```diff
@@ -541,10 +540,18 @@ pub fn peak_map_height(mut pos: u64) -> (u64, u64) {
 /// index. This function is the base on which all others, as well as the MMR,
 /// are built.
 pub fn bintree_postorder_height(num: u64) -> u64 {
-	if num == 0 {
+	let mut pos = num.saturating_sub(1);
+	if pos == 0 {
 		return 0;
 	}
-	peak_map_height(num - 1).1
```
antiochp, line 547: This is a small optimization inlining only the logic required to get the height. 
We don't need to calculate peaks and then throw them away.

PR #3579, opened 2021-03-02
Enable block archival sync

antiochp:
Archival Peers (as of 20210315) - 

* `78.47.168.92:3414`
* `46.101.247.184:3414`
* `3.81.230.109:3414`
* `23.124.76.209:3414`

```
# configure the above as peers_preferred to help archive node connectivity
peers_preferred = ["78.47.168.92:3414", "46.101.247.184:3414", "3.81.230.109:3414", "23.124.76.209:3414"]
```

----

Two relatively minor changes that allow a node to run in "block archival mode" and sync fully from other archival peers.
This works in conjunction with various related PRs recently merged.

1. Simply return false for `check_txhashset_needed()` if running with `archive_mode` enabled
2. No longer skip compaction during sync (compact periodically during full archival sync).

----

The local `chain_data` dir should be cleared out prior to a full archival sync - the node will sync from scratch.

A node will only be able to sync historical blocks from other archival nodes.
The best way of configuring the local node to peer with archival nodes is via the following config.

```
#
# with [server] section -
#

archive_mode = true


#
# within [server.p2p_config] section -
# 

# These are known archival nodes. Sync works best if these are both "preferred" and "allow".
peers_preferred = ["85.10.201.143:3414", "49.12.108.173:3414", "176.9.86.219:3414", "116.203.154.223:3414", "213.239.217.14:3414", "46.101.247.184:3414"]
peers_allow = ["85.10.201.143:3414", "49.12.108.173:3414", "176.9.86.219:3414", "116.203.154.223:3414", "213.239.217.14:3414", "46.101.247.184:3414"]
```

----

Once the node has completed an archival sync we can update the config to allow other nodes to connect.

```
# No explicit peers_allow list, so all peers implicitly allowed.
# peers_allow = ["85.10.201.143:3414", "49.12.108.173:3414", "176.9.86.219:3414", "116.203.154.223:3414", "213.239.217.14:3414", "46.101.247.184:3414"]
```

----

Full archival sync takes several hours to complete, probably close to 24 hours.
On completion the local node will have full block history from height 1 onwards.

Disk usage is approx 9.5GB for a full archival node.

----

It is anticipated that `peers_preferred` will be a required config for archival nodes, to ensure we good good connectivity.

The additional `peers_allow` config is a **temporary solution** while we test this out prior to having nodes correctly advertising archival capabilities.

----

### TODO

- [x] uncomment the BLOCK_HIST capability filtering (we only sync from other archival nodes)
- [x] collect a list of known good ip addresses for existing archival nodes so we can publish a list of `preferred` peers for bootstrapping
- [ ] release this as part of `5.1.0`

### DIFF
--- a/servers/src/grin/sync/body_sync.rs
+++ b/servers/src/grin/sync/body_sync.rs
@@ -100,6 +100,8 @@ impl BodySync {
 			peers = peers_iter().inbound().into_iter().collect();
 		}
 
+		debug!("*** body_sync: peers: {}", peers.len());
+
 		// If we have no peers (outbound or inbound) then we are done for now.
 		if peers.is_empty() {
 			debug!("no peers (inbound or outbound) with more work");
--- a/servers/src/grin/sync/body_sync.rs
+++ b/servers/src/grin/sync/body_sync.rs
@@ -72,6 +72,11 @@ impl BodySync {
 		Ok(false)
 	}
 
+	/// Is our local node running in archive_mode?
+	fn archive_mode(&self) -> bool {
+		self.chain.archive_mode()
+	}
+
 	/// Return true if txhashset download is needed (when requested block is under the horizon).
 	/// Otherwise go request some missing blocks and return false.
 	fn body_sync(&mut self) -> Result<bool, chain::Error> {
@@ -89,16 +94,13 @@ impl BodySync {
 		let peers = {
 			// Find connected peers with strictly greater difficulty than us.
 			let peers_iter = || {
-				// TODO - Once we have "critical mass" of archival nodes on network
-				// we can uncomment this so archive nodes only sync from archive nodes.
-				//
-				// let cap = if self.chain.archive_mode {
-				// 	Capabilities::BLOCK_HIST
-				// } else {
-				// 	Capabilities::UNKNOWN
-				// };
-
-				let cap = Capabilities::UNKNOWN;
+				// If we are running with archive mode enabled we only want to sync
+				// from other archive nodes.
+				let cap = if self.archive_mode() {
+					Capabilities::BLOCK_HIST
+				} else {
+					Capabilities::UNKNOWN
+				};
 
 				self.peers
 					.iter()
--- a/servers/src/grin/sync/body_sync.rs
+++ b/servers/src/grin/sync/body_sync.rs
@@ -116,6 +116,7 @@ impl BodySync {
 				peers = peers_iter().inbound().into_iter().collect();
 			}
 
+<<<<<<< HEAD
 			// If we have no peers (outbound or inbound) then we are done for now.
 			if peers.is_empty() {
 				debug!("no peers (inbound or outbound) with more work");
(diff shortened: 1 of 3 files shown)


servers/src/grin/sync/body_sync.rs
```diff
@@ -85,26 +86,42 @@ impl BodySync {
...
-		let peers_iter = || {
-			self.peers
-				.iter()
-				.with_difficulty(|x| x > head.total_difficulty)
-				.connected()
-		};
+		let peers = {
+			// Find connected peers with strictly greater difficulty than us.
+			let peers_iter = || {
+				// TODO - Once we have "critical mass" of archival nodes on network
```
antiochp, line 92: This is commented out currently so we can "bootstrap" a new archive node from nodes known to be archive nodes, but not necessarily advertising as such.
With this change uncommented we would _only_ be able to archival sync (historical blocks) from nodes explicitly advertising via the `BLOCK_HIST` capability flag.

We want to move to this being explicit (capability based) but we can only safely do this if we have enough nodes advertising "archive mode" via the capability flag.

antiochp:
If we have consensus that this is something we want to roll out to `mainnet` then I propose we consider a `5.1.0` release in the next couple of weeks. This would include this change along with other changes currently on `master`.

The "bootstrapping" change (discussed above) will ensure network traffic and peering is unaffected for non-archive nodes (archival sync restricted to known archive peers).

antiochp:
I understand @quentinlesceller is experiencing _really_ slow archive sync.
This appears limited to some local nodes as we are not seeing this across every archive node. Possibly slow disk or some other local environmental factor.
Solving that is outside the scope of this PR - we can (and will) investigate and optimize in a subsequent PR.

bladedoyle:
I completed a sync on bare metal i9 cpu + ssd in just over 38 hours.
ending disk use: 7.9 GB
Everything looks good to me.

antiochp:
For reference a $5/month Digital Ocean instance sync'd in about 20 hours.
So maybe bandwidth is a bigger factor than disk speed.

bladedoyle:
My test sync was done with 1 GB fiber...
None of the peers data xfer rate ever got about about 12K/sec, including a local full node peer.

quentinlesceller (approved):

antiochp:
> My test sync was done with 1 GB fiber...
> None of the peers data xfer rate ever got above about 12K/sec, including a local full node peer.
> 
> ftr: the sync started with 3 peers, gained peers over the first few hours, and ended with (currently has) 7 peers.
> 
> one more note: my test sync did not allow inbound connections, all outbound

Is this with `allow` and `preferred` lists configured against known archival nodes?

servers/src/grin/sync/body_sync.rs
antiochp: This is now uncommented and we filter peers by `BLOCK_HIST` capability when choosing peers to sync blocks from if running in archive mode.

antiochp:
Going to merge this to `master`.
Anybody wanting to run this code with `archive_mode=true` enabled is advised to specify an explicit set of known archive nodes in their `peers_preferred` config.

PR #3584, opened 2021-03-03
Mining tui updates

bladedoyle:
* Add column ordering to Mining:Mining Server Status view
* Set "Connected" column default sort with connected miners on top in Mining:Mining Server Status view
* Remove "Type" and "Sec. Scaling" columns from Mining::Difficulty view 
* Adjust size of columns for Mining views
* Fix "Pow Difficulty" value in Mining:Mining Server Status view


bladedoyle:
image


1. Removed "Latest POW submitted" and added "Blocks Found" (total for all workers)
2. Shorter column header text and resized columns so the last column would fit better
3. Added ordering to Mining Workers table and set default sort so that connected workers will be sorted to the top
*. Changed "Number of workers" to "Active workers" 
6. Fixed share difficulty value
--
4. Removed "Type" 
5. and "Sec. Scaling" columns from "Divviculty" sub menu
*. Resized columns so that the last column would fit better

quentinlesceller (approved):
Code looks good ! Tested a bit this morning. Overall that looks much better and updated to the current post HF conditions. Thanks @bladedoyle! LGTM 🚀

PR #3585, opened 2021-03-04
Update and do not use serve private export

quentinlesceller:
Exactly like #3544 but including update of the corresponding crates:
Serde and serde derive to 1.0.123
Syn to 1.0.60

Those are bug fix release so it should be safe to merge.


antiochp (approved):
Looks good to me.
This is going to be pretty obvious if it breaks something I guess.
:+1:

PR #3590, opened 2021-03-08
Remove imports and cleanups

quentinlesceller:
Mostly removal of unused imports.
Three additional things:
1. Consolidation of `if else` structure with the same return type
2. Removal of unnecessary clone
3. Do not call `string()` in Display function. This can lead to infinite recursion.

Overall should be consensual.


antiochp (approved):
Looks good!

antiochp:
Were the if/else changes caught by clippy or just manual?

quentinlesceller:
All Clippy (https://rust-lang.github.io/rust-clippy/master/index.html#if_same_then_else). Those were "Deny" type so blocking any further clippy processing.

PR #3591, opened 2021-03-08
[DNM] Hash file simplification

antiochp:
Deprecate the (P)MMR data file (and size file) and store (leaf) data in the database.
This PR reworks the PMMR backend data strcutures to _only_ store the hash file on disk.
Everything else is pushed into the db (indexed by MMR pos).

Very much still an experimental WIP.

Outputs, rangeproofs and kernels -
* store entries in the db indexed by MMR pos (per leaf) `pos -> output`
* existing index still maps output commitment to output pos etc. `commitment -> pos`
* `commitment -> pos -> output`

Headers - 
* already store headers in the db indexed by hash `hash -> header`
* store new index mapping `height -> hash`
* `height -> hash -> header`

----

TODO

- [ ] get chain crate tests passing
- [ ] get other tests passing
- [ ] cleanup all the commented out code
- [ ] lots of missing docs on new fns etc.
- [ ] migration process for existing node 
  - [ ] populate the db on startup if missing if required
- [ ] handle receiving txhashset state sync
  - [ ] given a raw "data file" and a `prune_list` we can sequentially read leaf data for all unpruned leaf pos
- [ ] handle providing txhashset state sync (need to recreate data files from db etc.) 
  - [ ] iterate over unpruned leaf data in db and write bytes to "data file" when constructing `txhashset.zip`
  - [ ] `txhashset.zip` is cached every 720 blocks so we can take advantage of this to mitigate extra cost here 
  - [ ] consider not supporting state sync initially (no `TXHASHSET_HIST` capabilities)

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -990,9 +990,8 @@ impl Chain {
 		{
 			self.validate_kernel_history(&header, &txhashset)?;
 
-			let header_pmmr = self.header_pmmr.read();
 			let batch = self.store.batch()?;
-			txhashset.verify_kernel_pos_index(&genesis, &header_pmmr, &batch)?;
+			txhashset.verify_kernel_pos_index(&genesis, &batch)?;
 		}
 
 		// all good, prepare a new batch and update all the required records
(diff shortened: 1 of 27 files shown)


yeastplume:

PR #3592, opened 2021-03-09
Update copyright year to 2021

quentinlesceller:
As titled.


antiochp (approved):
We should just get rid of these years.
:+1:

quentinlesceller:
Can we?

PR #3597, opened 2021-03-18
Removing verifier caches

DavidBurkett:
Same fix on `current/5.0.x` - https://github.com/mimblewimble/grin/pull/3598


antiochp (approved):
:+1:

PR #3600, opened 2021-03-18
[5.0.x] Rewind bad block

antiochp:
* If we see the "bad block" on the current chain then rewind back to the previous block.
* Make sure we remove all subsequent blocks from our local db, back to and including the bad block.
* Update chain head to reflect the previous good block.


quentinlesceller (approved):
Looking good.

PR #3602, opened 2021-03-18
[DNM] Rewind bad block (#3600)

antiochp:
**Do not merge.**
----

We don't actually want this on `master`.
We want to decide on the correct long-term approach and implement that here.

* Currently the rangeproof verification cache is disabled on both master and 5.0.x branches.
  *  We need to decide if we want to reintroduce this fully and what the cache key impl needs to be.
* "rewind bad block" should not be required on `master` as it was simply a hot fix to get past the "bad fork"
* We do want to propagate the `InvalidBlockProof` error (to ban peers correctly)
  * the whole error/bool/ban logic needs a rethink and cleanup as its hard to reason about currently
* Related #3603 (on `5.0.x` branch) "rewind headers and ban"
  * "rewind headers" also a hot fix and not required on `master`
* but we _do_ need to handle the "good header, bad block" scenario robustly somehow and this needs impl on `master`
  * one option is explicit list of "bad headers" as we have in that PR but ideally this is dynamic

I'll open a tracking issue for all of above and we can close this PR once its tracked elsewhere.

---- 

Cherry-picked #3600 from `current/5.0.x`

* rewind prior to known bad block on chain init
* rewind prior to bad block if found on current chain
* pass InvalidBlockProof through

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -206,6 +206,10 @@ impl Chain {
 			genesis: genesis.header,
 		};
 
+		// If known bad block exists on "current chain" then rewind prior to this.
+		// Suppress any errors here in case we cannot find
+		chain.rewind_bad_block()?;
+
 		chain.log_heads()?;
 
 		// Temporarily exercising the initialization process.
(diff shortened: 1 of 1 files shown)


antiochp:
Tracking issue - https://github.com/mimblewimble/grin/issues/3605

Closing this PR.

PR #3603, opened 2021-03-18
rewind bad headers, ban on explicit bad headers

antiochp:
* validate against explicit list of bad headers
* rewind both `header_head` and `head` if bad block on current chain


MCM-Mike:
Will this PR be released before 3am CET or next days?

quentinlesceller (approved):

DavidBurkett (approved):

PR #3611, opened 2021-03-22
[DNM] Revert "Removing verifier caches (#3597)"

trevyn:
This reverts commit 87ff219d37b59c8e3cd8ab7f8629af4b9fd12dd1.

Per discussion with @antiochp. To be followed by a separate PR that fixes the cache keys.

Issue #3606, opened 2021-03-19
tracking issue - enable rangeproof verifier cache


antiochp:
* Related #3605 
* Related #3597

I think we want to do the following initially -
* fix the cache key impl for rangeproofs
  * @jaspervdm suggested implementing `DefaultHashable` for the full output (identifier + rangeproof) 
* revert #3597 ensuring we use the correct cache key impl 

----

Additional possible improvements - 

We get most of the benefit from the verifier cache internally in the mempool. We do a *lot* of tx verification in various mempool operations. For example after aggregating txs together we verify the aggregated tx. Each verification leverages the rangeproof verifier cache as we do not want to verify the rangeproofs repeatedly each time.

But we may want to consider disabling the cache when we actually verify a new block.
Blocks are consensus critical, transactions are not.
We would need to take care to handle edge cases where a potentially invalid tx entered the mempool. 
This makes me hesitate a little.

antiochp:
After some investigation and discussion we are removing the "verifier cache" entirely (for now) - 
https://github.com/mimblewimble/grin/pull/3628

Future work may involve reintroducing some form of cache.

PR #3613, opened 2021-03-22
[DNM] Fix verifier cache keys by hashing full `Output`

trevyn:
Per #3606 and discussion with @antiochp, this includes:

- Revert "Removing verifier caches (#3597)" from #3611
- Impl `DefaultHashable` for `Output`, and hash the full `Output` including the `OutputIdentifier`, instead of just the `RangeProof`. This looks to be provided via the existing `Writeable` impl on `Output`.

Outstanding questions:

- Is there any other specific test coverage required?
- Address comments from #3606:
>We get most of the benefit from the verifier cache internally in the mempool. We do a lot of tx verification in various mempool operations. For example after aggregating txs together we verify the aggregated tx. Each verification leverages the rangeproof verifier cache as we do not want to verify the rangeproofs repeatedly each time.
>
>But we may want to consider disabling the cache when we actually verify a new block.
Blocks are consensus critical, transactions are not.
We would need to take care to handle edge cases where a potentially invalid tx entered the mempool.
This makes me hesitate a little.


core/src/core/verifier_cache.rs
```diff
@@ -28,20 +28,20 @@ pub trait VerifierCache: Sync + Send {
...
 	fn filter_kernel_sig_unverified(&mut self, kernels: &[TxKernel]) -> Vec<TxKernel>;
 	/// Takes a vec of tx outputs and returns those outputs
-	/// that have not yet had their rangeproofs verified.
-	fn filter_rangeproof_unverified(&mut self, outputs: &[Output]) -> Vec<Output>;
+	/// that have not yet been verified.
+	fn filter_output_unverified(&mut self, outputs: &[Output]) -> Vec<Output>;
 	/// Adds a vec of tx kernels to the cache (used in conjunction with the the filter above).
 	fn add_kernel_sig_verified(&mut self, kernels: Vec<TxKernel>);
 	/// Adds a vec of outputs to the cache (used in conjunction with the the filter above).
-	fn add_rangeproof_verified(&mut self, outputs: Vec<Output>);
```
DavidBurkett, line 36: nit: The names and comments of these 2 functions were more accurate before. There's no need to rebrand all of this as "output verification." It's still rangeproof verification imho. It just includes the output commitment as part of the hash.

core/tests/verifier_cache.rs
```diff
@@ -42,20 +42,20 @@ fn test_verifier_cache_rangeproofs() {
...
 		let mut cache = cache.write();
-		cache.add_rangeproof_verified(vec![out]);
+		cache.add_output_verified(vec![out]);
 	}
 
 	// Check it shows as verified according to the cache.
 	{
 		let mut cache = cache.write();
-		let unverified = cache.filter_rangeproof_unverified(&[out]);
+		let unverified = cache.filter_output_unverified(&[out]);
```
DavidBurkett, line 58: Please add a test to prove that the inflation issue is not reintroduced. That is, use the same rangeproof with a different output commitment and make sure that the cache does not consider it already verified.

.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
+name: test
```
DavidBurkett, line 1: We already have CI setup. We use Azure pipelines

trevyn: Yep, am not intending this to be merged. Still figuring out the workflow so I can run tests on my own forks.

trevyn:
Thanks @DavidBurkett ! All comments addressed.

antiochp:
It would be good to add a test here that exercises the full block processing pipeline resulting in an `InvalidBlockProof`.
Something along the lines of the test in `nrd_validation_rules.rs`

Ideally we can reproduce the inflation bug running the test with `5.0.1` codebase.
Then confirm the inflation bug is no longer present in `5.0.4`.
And finally that reintroducing the verifier cache logic does not also reintroduce the inflation bug.

We will have to jump through some hoops to build an "invalid" block (presumably we need to swap out a valid rangeproof for an invalid one).
This is worth spending some time on as we definitely have less test coverage (and less ability to add them) around "invalid blocks" like this.

trevyn:
> It would be good to add a test here that exercises the full block processing pipeline resulting in an `InvalidBlockProof`.
> Something along the lines of the test in `nrd_validation_rules.rs`

Got it. I looked the relevant code and I think I'll be able to make progress on this. Will ping when ready for a review, or with any questions that arise. Thanks!

antiochp:
We should pull the test from #3621 into this PR and consolidate the work.
@snape479 @trevyn

trevyn:
I definitely support removing the cache. Do we have a sense of the actual, measured performance impact?

antiochp:
> I definitely support removing the cache. Do we have a sense of the actual, measured performance impact?

For a handful of txs per block like we have currently and a txpool that handles those few txs the performance impact is going to be close to zero.
The verifier cache was originally put in place to handle future increased tx volume.

antiochp:
PR to remove the verifier cache - https://github.com/mimblewimble/grin/pull/3628

trevyn:
> PR to remove the verifier cache - #3628

👍🏼. Will sort out the tests once that is merged.

trevyn:
Closing in favor of https://github.com/mimblewimble/grin/pull/3628.

The only new test in this PR requires direct access to the verifier cache, which no longer exists. Tests from https://github.com/mimblewimble/grin/pull/3621 should still work, I'll look at porting them over.

PR #3614, opened 2021-03-23
Allow rest api to shutdown

i1skn:
I think we can re-enable this functionality, `grin-wallet` also uses this crate and it's good to have an ability to shutdown the api without killing the process.

### DIFF
--- a/api/src/rest.rs
+++ b/api/src/rest.rs
@@ -30,6 +30,7 @@ use rustls::internal::pemfile;
 use std::convert::Infallible;
 use std::fmt::{self, Display};
 use std::fs::File;
+use std::mem;
 use std::net::SocketAddr;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


quentinlesceller:
Looking good. Do you know why we can reenable this now?

i1skn:
As far as I got after conversation with @hashmap there was a temporary bug in hyper version, which is now fixed. I've tested it as a part of Ironbelly and it shutdowns the API properly.

quentinlesceller (approved):
Tested and looking good. LGTM 👍

quentinlesceller:
@i1skn seems like the API does not start with this change.

i1skn:
Hi, is there any specific reason or case when it does not start?

quentinlesceller:
Basically any call on the API will just fails. The server is not listening.

PR #3617, opened 2021-03-23
Update zip dependency and fix deprecation warnings

GeneFerneau:
Update zip crate to latest version, and fix deprecation warnings using functionally equivalent APIs

### DIFF
--- a/util/src/zip.rs
+++ b/util/src/zip.rs
@@ -21,6 +21,21 @@ use std::thread;
 use self::zip_rs::write::FileOptions;
 use zip as zip_rs;
 
+// Sanitize file path for normal components, excluding '/', '..', and '.'
+// From private function in zip crate
+fn path_to_string(path: &std::path::Path) -> String {
+    let mut path_str = String::new();
+    for component in path.components() {
+        if let std::path::Component::Normal(os_str) = component {
+            if !path_str.is_empty() {
+                path_str.push('/');
+            }
+            path_str.push_str(&*os_str.to_string_lossy());
+        }
+    }
+    path_str
+}
+
 /// Create a zip archive from source dir and list of relative file paths.
 /// Permissions are set to 644 by default.
 pub fn create_zip(dst_file: &File, src_dir: &Path, files: Vec<PathBuf>) -> io::Result<()> {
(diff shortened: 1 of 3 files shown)


util/src/zip.rs
```diff
@@ -57,7 +72,7 @@ pub fn extract_files(from_archive: File, dest: &Path, files: Vec<PathBuf>) -> io
 		let mut archive = zip_rs::ZipArchive::new(from_archive).expect("archive file exists");
 		for x in files {
 			if let Ok(file) = archive.by_name(x.to_str().expect("valid path")) {
-				let path = dest.join(file.sanitized_name());
+				let path = dest.join(file.mangled_name());
```
trevyn, line 75: Is `enclosed_name` a better option here, based on the [doc comments](https://github.com/zip-rs/zip/blob/26710ee8a5aba9fff72725a81db9519a4dcd7778/src/read.rs#L689..L730)?

util/src/zip.rs
```diff
@@ -37,7 +52,7 @@ pub fn create_zip(dst_file: &File, src_dir: &Path, files: Vec<PathBuf>) -> io::R
 		let file_path = src_dir.join(x);
 		if let Ok(file) = File::open(file_path.clone()) {
 			info!("compress: {:?} -> {:?}", file_path, x);
-			writer.get_mut().start_file_from_path(x, options)?;
+			writer.get_mut().start_file(path_to_string(x), options)?;
```
trevyn, line 55: Would just `x.to_str()` work here? We generate the path, so it shouldn't need to be sanitized with the `path_to_string` function, right?

phyro:
@GeneFerneau check the comments when you find the time, let's get this one in if it is easy to resolve those :+1:

util/src/zip.rs
GeneFerneau: I looked at that too, but we actually want the side-effects of `mangled_name`.

util/src/zip.rs
GeneFerneau: `path_to_string` performs extra checks that `x.to_str()` does not, like normalizing the path.

GeneFerneau:
> @GeneFerneau check the comments when you find the time, let's get this one in if it is easy to resolve those +1

Addressed the comments. Let me know if that answers your questions.

util/src/zip.rs
antiochp: I'm :+1: on replacing `sanitized_name` with `mangled_name` based on the deprecated `sanitized_name` simply delegating to `mangled_name`.

But it might be worth revisiting with a separate PR to consider replacing with `enclosed_name`.

From the docs referenced by @trevyn - 
```
    /// This will read well-formed ZIP files correctly, and is resistant
    /// to path-based exploits. It is recommended over
    /// [`ZipFile::mangled_name`].
```

util/src/zip.rs
antiochp: :+1:

antiochp (approved):
:+1:

PR #3618, opened 2021-03-24
client invalidateheader and resetchainhead

### LINKED ISSUE #3607: tracking issue - good header, bad block
* Related #3605 

Some rough notes on a potential approach to this.

We sync headers from peers that advertise more work than our local node.
We then try and sync blocks based on our local header chain.

The "good header, bad block" issue means we see a header locally that as far as we know is valid (valid pow etc.)
We then request the corresponding blocks.
But there is a possible scenario where the block _body_ is invalid based on current consensus rules, yet the corresponding header itself is valid. So its a "good" header, but a "bad" block.

We can ban the peer sending us the bad block. But our local node is simply going to continue trying (unsuccessfully) to sync the missing blocks along that same header chain.

We need to flag that header as "bad" in some way, based on the block being invalid.

----

Proposal:

I think there is a relatively simple way to do this.
On receiving an invalid block we want to do the following - 

* immediately reset the `header_head` to the previous good header (with corresponding good block)
* ban the peer that sent us the invalid block as normal

We will repeatedly swap between header_sync and body_sync for a time as our local node will attempt to sync blocks and fail.
Eventually we will start to sync headers along the "good " header chain. And we can then progress.

This still needs investigating but I think the "reset header_head immediately" is the cleanest way to implement this.
We don't need to track "bad" headers this way as we can effectively do this on the fly.
Its not necessarily the most efficient way to do this but we will ban peers based on the block itself.
If no peer can provide the block in question we will eventually switch to an alternative header chain (I think).

----

Modified Proposal:

Do this "manually" for now. Expose api endpoints allowing the local node to selectively choose a "good" chain.

The following PR exposes an owner api endpoint (and client cmd) to manually `resetchainhead` on local node - 
https://github.com/mimblewimble/grin/pull/3618

If we add a companion `invalidateheader` api endpoint then a node operator will have the tool necessary to invalidate "bad" forks and reset the chain state to a previous header allowing the node to sync against a known "good" chain.

If these prove robust and practical in use then we can consider automating this process based on header and block validation etc.

antiochp:
Resolves #3607

----

Experimenting with a different approach here to resolving the various issues around sync and how it interacts with a chain fork.

This PR adds an endpoint and handler for `resetchainhead` via the owner api.

This will take a block hash and will do something roughly along the following lines -

* check the header being reset to is indeed present in the local db
* reset `header_head` to the existing `head` via "rewind and reapply" on the header extension
* then reset `head` (and `header_head` consistently) via "rewind and reapply" on txhashset extension (full blocks)

The _should_ result in the ability to arbitrarily "rewind" or "reset" to an alternative block header.

Rather than trying to resolve all edge cases related to sync I'm proposing it makes more sense to have additional flexibility and allow a node operator to "manually" work around a fork via this api endpoint.

If we can get this working robustly then there is an additional api endpoint that could be added - 
* `invalidateheader`

These could be used together to invalidate one chain fork and then reset the local chain to a header on an alternative fork.
I believe these would allow most fork scenarios to be recovered from reasonably gracefully albeit manually.

If this approach works _then_ we may consider extending this and doing the "reset" in a more automated fashion without api involvement.

----

Non-archival node should be able to "reset" to prior to the horizon by -
* remove recent full blocks from db (effectively all blocks, horizon onwards)
* reset `head` (and `tail`)
* truncate the MMR structures
* force a state sync

This would potentially be a convenient way of retaining headers (up to reset point) locally while still resyncing state.

----

TODO

- [x] error checking and basic validation (make sure its actually a hash)
- [x] user feedback - success/failure messages (jsonrpc issues?)
- [x] "can we rewind beyond horizon"
  - [x] For a full archival node - need to confirm but technically possible?
  - [x] For non-archival node see above (for now this is prevented and full resync is required)
- [x] cleanup

### DIFF
--- a/api/src/owner_rpc.rs
+++ b/api/src/owner_rpc.rs
@@ -133,7 +133,7 @@ pub trait OwnerRpc: Sync + Send {
 	 */
 	fn compact_chain(&self) -> Result<(), ErrorKind>;
 
-	fn reset_chain_head(&self, hash: Hash) -> Result<(), ErrorKind>;
+	fn reset_chain_head(&self, hash: String) -> Result<(), ErrorKind>;
 
 	/**
 	Networked version of [Owner::get_peers](struct.Owner.html#method.get_peers).
@@ -366,6 +366,10 @@ impl OwnerRpc for Owner {
 		Owner::validate_chain(self).map_err(|e| e.kind().clone())
 	}
 
+	fn reset_chain_head(&self, hash: String) -> Result<(), ErrorKind> {
+		Owner::reset_chain_head(self, hash).map_err(|e| e.kind().clone())
+	}
+
 	fn compact_chain(&self) -> Result<(), ErrorKind> {
 		Owner::compact_chain(self).map_err(|e| e.kind().clone())
 	}
@@ -384,10 +388,6 @@ impl OwnerRpc for Owner {
 
 	fn unban_peer(&self, addr: SocketAddr) -> Result<(), ErrorKind> {
 		Owner::unban_peer(self, addr).map_err(|e| e.kind().clone())
-	}
-
-	fn reset_chain_head(&self, hash: Hash) -> Result<(), ErrorKind> {
-		Owner::reset_chain_head(self, hash).map_err(|e| e.kind().clone())
 	}
 }
 
--- a/api/src/owner_rpc.rs
+++ b/api/src/owner_rpc.rs
@@ -14,7 +14,6 @@
 
 //! JSON-RPC Stub generation for the Owner API
 
-use crate::core::core::hash::Hash;
 use crate::owner::Owner;
 use crate::p2p::types::PeerInfoDisplay;
 use crate::p2p::PeerData;
--- a/api/src/owner_rpc.rs
+++ b/api/src/owner_rpc.rs
@@ -14,6 +14,7 @@
 
 //! JSON-RPC Stub generation for the Owner API
 
+use crate::core::core::hash::Hash;
 use crate::owner::Owner;
 use crate::p2p::types::PeerInfoDisplay;
 use crate::p2p::PeerData;
@@ -131,6 +132,8 @@ pub trait OwnerRpc: Sync + Send {
 	```
 	 */
 	fn compact_chain(&self) -> Result<(), ErrorKind>;
+
+	fn reset_chain_head(&self, hash: Hash) -> Result<(), ErrorKind>;
 
 	/**
 	Networked version of [Owner::get_peers](struct.Owner.html#method.get_peers).
@@ -382,6 +385,10 @@ impl OwnerRpc for Owner {
 	fn unban_peer(&self, addr: SocketAddr) -> Result<(), ErrorKind> {
 		Owner::unban_peer(self, addr).map_err(|e| e.kind().clone())
 	}
+
+	fn reset_chain_head(&self, hash: Hash) -> Result<(), ErrorKind> {
+		Owner::reset_chain_head(self, hash).map_err(|e| e.kind().clone())
+	}
 }
 
 #[doc(hidden)]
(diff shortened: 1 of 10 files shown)


tromp:
> check the header being reset to is indeed present in the local db

the header is specified by its (cycle)hash?

> reset header_head to the existing head via "rewind and reapply" on the header extension

what is a header extension? what does reapplying a header mean? is this updating a constant amount of state?

> then reset head (and header_head consistently) via "rewind and reapply" on txhashset extension (full blocks)

can this rewind beyond the horizon, if the node saw preceding blocks?

antiochp:
No guarantees to this working 100% correctly and it may brick you local `chain_data`.

That said, the following works for me - 

```
../target/release/grin client resetchainhead 0001df6e93e6462b754d4616a51a0b65917c9492a259175cccbf74af2a990ddf
Successfully reset chain head 0001df6e93e6462b754d4616a51a0b65917c9492a259175cccbf74af2a990ddf
```

antiochp:
> the header is specified by its (cycle)hash?
> 

Yes by hash.

> what is a header extension? what does reapplying a header mean?

We interact with the header MMR via a "header extension". This allows us to rewind/truncate and append new data to the header MMR.

> can this rewind beyond the horizon, if the node saw preceding blocks?

No sure yet. I have only experimented with rewinding within the horizon.

antiochp:
```
curl -X POST \
     -H 'Content-Type: application/json' \
     -d '{"jsonrpc":"2.0","id":1,"method":"reset_chain_head","params":["0002574861b6931d24533549418541a86ae4f8efadc6e2dd23016a518ddee32d"]}' \
         http://localhost:3413/v2/owner

{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "Ok": null
  }
}%
```

antiochp:
```
../target/release/grin client resetchainhead 0002574861b6931d24533549418541a86ae4f8efadc6e2dd23016a518ddee32d

Successfully reset chain head 0002574861b6931d24533549418541a86ae4f8efadc6e2dd23016a518ddee32d
```

antiochp:
We can no do the following - 

* `grin client invalidateheader 0002691b87125bb45aa76e54543d9375f2047b10ca1ee674dea22a890447e262`
* `grin client resetchainhead 0000b9e51d5733d778a0a13f2c85e843863db0ea4ea6a4b0b358873636dc5db3`

We maintain a local set of invalid headers (denylist). 
This does not persist across node restart to avoid getting a node into a persistent bad state.

We can simply invalidate a particular chain fork by identifying a header on the fork and invalidating it via `invalidateheader`.
This header(s) will be treated as invalid and will result in a peer ban.

By resetting chain state to an earlier header via `resetchainhead` we can then force the local node onto an alternative fork.
This assumes at least one of our peers is advertising the alternative fork that does not involve an invalid header.
Other peers that advertise a chain head on a fork with an invalid header will be banned over time.

phyro:
is this one ready for assignees?

api/src/owner_rpc.rs
```diff
@@ -363,6 +367,14 @@ impl OwnerRpc for Owner {
 		Owner::validate_chain(self).map_err(|e| e.kind().clone())
 	}
 
+	fn reset_chain_head(&self, hash: String) -> Result<(), ErrorKind> {
```
GeneFerneau, line 370: Do calls to this API endpoint require a password / extra authentication?

The attack vector would be for a malicious caller to send a hash to their chain head.

chain/src/chain.rs
```diff
@@ -347,14 +399,14 @@ impl Chain {
...
-		// Check if we already know about this block.
-		self.is_known(&b.header)?;
-
 		// Process the header first.
 		// If invalid then fail early.
 		// If valid then continue with block processing with header_head committed to db etc.
 		self.process_block_header(&b.header, opts)?;
 
+		// Check if we already know about this full block.
+		self.is_known(&b.header)?;
```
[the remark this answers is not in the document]
antiochp: Actually no, `is_known()` is checking for existence of the _full block_ corresponding to the header hash, not the header itself.

api/src/owner_rpc.rs
antiochp: Yes this endpoint  is exposed via the "owner" api alongside ban/unban peer endpoints etc.

antiochp:
Going to merge this shortly to `master`.

GeneFerneau (approved):

Issue #3620, opened 2021-03-26
VerifiedCache should be re-enabled


snape479:
Due to the recent CVE, the VerifierCache was disabled as an emergency measure. It should be enabled with the appropriate fix. PR will be forthcoming with this fix.

antiochp:
Duplicate of https://github.com/mimblewimble/grin/issues/3606

phyro:
I suggest we close this one

snape479:
Added a perf test to https://github.com/mimblewimble/grin/pull/3621. It shows a significant performance impact from disabling the verifier_cache. Here is the output:

# No cache
% ./target/release/perf
[2021-04-02 08:02:54]: Starting perf test
[2021-04-02 08:02:55]: building tx1
[2021-04-02 08:03:42]: building tx2
[2021-04-02 08:04:28]: building tx3
[2021-04-02 08:05:14]: building tx4
[2021-04-02 08:06:01]: building tx5
[2021-04-02 08:06:47]: start validate tx
[2021-04-02 08:06:51]: tx val time elapsed = 3599ms
[2021-04-02 08:06:51]: start process block
[2021-04-02 08:06:52]: block val time elapsed = 909ms

# With cache
% ./target/release/perf
[2021-04-02 07:58:08]: Starting perf test
[2021-04-02 07:58:09]: building tx1
[2021-04-02 07:58:55]: building tx2
[2021-04-02 07:59:42]: building tx3
[2021-04-02 08:00:28]: building tx4
[2021-04-02 08:01:15]: building tx5
[2021-04-02 08:02:01]: start validate tx
[2021-04-02 08:02:02]: tx val time elapsed = 878ms
[2021-04-02 08:02:02]: start process block
[2021-04-02 08:02:02]: block val time elapsed = 231ms

I think you should reconsider closing out this issue and merge pull request #3621.

antiochp:
@snape479 Thanks for digging into this. Let me take a look at the test results today.
I'm surprised we are seeing this large a difference.

snape479:
Yes, the test uses a pathological case for the tx validation, as only one of the outputs is changed. Keep in mind an attacker may use the fact that they can make transactions that look almost the same that are actually different and use the pathological case because nodes would propagate these 5 transactions separately through the transaction pools whereas if they were exactly the same, they would only propagate as a single transaction. In terms of the block though, I think all the transactions will generally be cached and thus I think that test shows the actual picture.

phyro:
[...]

PR #3621, opened 2021-03-26
[DNM] https://github.com/mimblewimble/grin/issues/3620

snape479:
This pull request re-enables the VerifierCache and addresses issue: https://github.com/mimblewimble/grin/issues/3620
The added test can be used against the previous version of the code to verify that it fails without the associated fix in this PR.


trevyn:
Do you know of a way to exercise the full block processing pipeline resulting in an `InvalidBlockProof` against the 5.0.1 code, something along the lines of the test in `nrd_validation_rules.rs`? :)

snape479:
> Do you know of a way to exercise the full block processing pipeline resulting in an `InvalidBlockProof`, something along the lines of the test in `nrd_validation_rules.rs`? :)

Not familiar with that. Wouldn't hurt to add a test at the transaction.rs level though.

phyro:
@antiochp I remember you were discussing an option where a block validation would not use any cache, did you guys agree on what the next steps would be? I think having block validation that ignores cache might be a good idea. We can optimize this later when blocks are actually heavy and we have a lot more eyes on the code.

trevyn:
@snape479 If you're interested in trying, I can (personally) offer a bounty of US$1,000 paid in BTC for a test at the block/chain level mentioned above that exercises the inflation bug without reaching into the verifier cache directly.

snape479:
Ok, I will take a stab at it.

antiochp:
This is effectively a duplicate of the changes proposed in https://github.com/mimblewimble/grin/pull/3613 right?
I think leveraging `DefaultHashable` is the most appropriate way of achieving what we want here.

snape479:
@trevyn I checked in a test that demonstrates an invalid range proof at the block level. What it doesn't do is show the inflation bug directly. To do that you need to create a negative output to offset a positive output. That would require a change to the grin_secp256k1 library c code, as the function currently accepts a uint64_t type. I don't suggest doing that just for this test. If you want the test as it stands, I can do a separate pull request for that since it sounds like there's another solution for the verifier_cache itself.

trevyn:
@snape479 Nice. Would it be possible to make the `grin_secp256k1` modifications locally and produce a short chain that shows the bug, and hardcode that chain data into a test?

snape479:
That might work. I will take a look at that and see what I can do.

chain/tests/mine_simple_chain.rs
```diff
@@ -817,6 +817,89 @@ fn output_header_mappings() {
...
+				build::output(consensus::REWARD - 20001, key_id31.clone()),
+				build::output(1, key_id32.clone()),
+			],
+			&kc,
+			&pb,
+		)
+		.unwrap();
+
+		// overwrite our rangeproof with a rangeproof from last block
+		tx2.body.outputs[1].proof = tx1.body.outputs[0].proof;
```
antiochp, line 890: We probably want to ensure we are explicit in _which_ rangeproof is being replaced and which one we are replacing it with. The order (`outputs[1]` etc.) will not necessarily be consistent and we may accidentally pick one that would cause the validation to fail for unrelated reasons (swapping out a coinbase one for example).

chain/tests/mine_simple_chain.rs
```diff
@@ -817,6 +817,89 @@ fn output_header_mappings() {
 	clean_output_dir(".grin_header_for_output");
 }
 
+/// Test the duplicate rangeproof bug
+#[test]
+fn test_duplicate_rangeproof() {
```
antiochp, line 822: 🚀

chain/tests/mine_simple_chain.rs
```diff
@@ -817,6 +817,89 @@ fn output_header_mappings() {
...
+		.unwrap();
+
+		// overwrite our rangeproof with a rangeproof from last block
+		tx2.body.outputs[1].proof = tx1.body.outputs[0].proof;
+
+		let next = prepare_block_tx(&kc, &prev_main, &chain, 8, &[tx2.clone()]);
+		// process_block fails with verifier_cache disabled or with correct verifier_cache
+		// implementations
+		assert_eq!(
+			chain.process_block(next, chain::Options::SKIP_POW).is_ok(),
```
antiochp, line 896: We just merged #3616 to `master` and we now propagate up a more useful error on failure here.
We should test for the error explicitly as part of this test - it should be an `InvalidBlockProof`. 
We don't want to accidentally have this test passing because of some unrelated error occurring during block processing.

antiochp:
> @trevyn I checked in a test that demonstrates an invalid range proof at the block level. What it doesn't do is show the inflation bug directly. To do that you need to create a negative output to offset a positive output. That would require a change to the grin_secp256k1 library c code, as the function currently accepts a uint64_t type. I don't suggest doing that just for this test. If you want the test as it stands, I can do a separate pull request for that since it sounds like there's another solution for the verifier_cache itself.

One thing we could do here is leverage `secp.commit_sum()` to build a negative commitment (we can "subtract" a large positive value from a smaller positive value).
At least I think we could use this.

Might be a lot of hoops to jump through to set the offending block up but this would be the perfect opportunity to put this in place.

snape479:
Yes, commit_sum worked. I was able to invert the output and generate 1 million grin without modifying the secp library. The test is working for me. In checkins before disabling the verifier_cache, it fails and since that checkin it passes. As I mention in the TODO there's probably a better way to check error, but I notice that errors changed in master, so when I merge it over to the other PR, I'll see if I can improve that somehow.

trevyn:
@snape479 Awesome! Let me know where to send the bounty award, I’m @trevyn on keybase if you prefer a private channel.

snape479:
Ok, sent you a message on keybase. There's a proof that the keybase account is connected to this github account.

trevyn:
> Ok, sent you a message on keybase. There's a proof that the keybase account is connected to this github account.

👍🏼. Bounty awarded & payment sent.

@antiochp wants to remove the verifier cache altogether for now in https://github.com/mimblewimble/grin/pull/3628, and we can definitely pull these tests in on top of that.

antiochp:
ok https://github.com/mimblewimble/grin/pull/3628 is merged.
Feel free to either rework this PR to cover just the test or close and open a new PR.

antiochp:
Closing. Tests included as part of #3630.

PR #3628, opened 2021-03-31
Remove verifier cache entirely

antiochp:
Drastic different approach - just get rid of the verifier cache entirely.
We can revisit tx validation performance _internal_ to the mempool in the future when required.

This is a lot of lines of changes for a relatively small functional change.
This just highlights the impact of threading it through the code everywhere.


phyro:
I support removing the cache. Having no cache logic means we remove any possibility of validation/invalidation bugs for the next 5-10 years at least while also removing the need to maintain and reason about this layer. This makes me believe it's a good move for now.

antiochp:
Reintroducing _some_ form of caching is definitely in the plan here - but it may not follow the exact same form as we see currently. Ideally it would be more limited in scope and less invasive to the codebase.

It would be crazy not to take advantage of _many_ nodes seeing _many_ transactions before validating full blocks if there was a way to robustly/safely/cleanly cache validation of rangeproofs and kernel signatures.

antiochp:
Planning to merge this later on today unless anybody has any serious objections.

antiochp:
ok merged. I'm going to open an issue to track the future cache work - got some thoughts around that.

PR #3629, opened 2021-04-01
implement fix past fees RFC with fee shift since genesis

tromp:
---
name: fixpastfees
about: remove dependence of fee() and fee_shift() functions on height
title: 'fix past fees'
labels: ''
assignees: ''

---
This PR implements the "fix past fees" RFC at https://github.com/mimblewimble/grin-rfcs/blob/master/text/0021-fix-prior-fees.md

It is consensus breaking in the weakest possible sense, in that older clients can disagree only if a many months deep reorg occurs that includes a > 40 bit fee (roughly 1100 Grin) prior to HF4.

This code has been tested on a sync-from-scratch.
No new tests are added, and only a few were removed that used to test that> 40 bit fees were allowed prior to HF4.


core/tests/core.rs
```diff
@@ -247,14 +244,14 @@ fn transaction_cut_through() {
 	let tx1 = tx1i2o();
 	let tx2 = tx2i1o();
 
-	let height = 42; // arbitrary
-	assert!(tx1.validate(Weighting::AsTransaction, height).is_ok());
-	assert!(tx2.validate(Weighting::AsTransaction, height).is_ok());
+
```
DavidBurkett, line 247: nit: remove superfluous blank line

DavidBurkett (approved):

phyro:
We should block the merge of this one until https://github.com/mimblewimble/grin-rfcs/pull/77 is accepted

tromp:
Yes, that's what it says in top description:-)

phyro:
Woops, missed it. But I did mean a literal merge block by making the PR draft (WIP, DNM or similar)

core/src/core/transaction.rs
```diff
@@ -1470,16 +1459,8 @@ impl Transaction {
 	}
 
 	/// Transaction minimum acceptable fee
-	pub fn accept_fee(&self, height: u64) -> u64 {
```
antiochp, line 1473: This is a different rule change than that discussed in the RFC right?
The RFC is related to the 40 bit fee limit.
This specific change involves the minimum acceptable fee.

Or does the RFC also cover minimum `accept_fee` rules?

tromp: The RFC covers all of 
* the 40-bit limit
*  the 4-bit fee_shift
* use of uniform weights
* new minimum accept fee base to make output cost about a grin cent

antiochp:
#3634 updates the config file -

~`accept_fee_base = 1000000`~  
`accept_fee_base = 500000`

as part of updating the config file to "version 2".

So we should be safe to include cleanup around the "rewriting" of this config item at startup in this PR.

phyro:
@tromp what's the state of this one?

tromp:
Ready to be merged upon approval.

antiochp (approved):
:+1:

PR #3632, opened 2021-04-05
Fix Windows CI `cargo test` silently passing

trevyn:
I don't really know why, but the changes I made in the CI scripts for #3596 were causing the Windows CI tests to silently pass without doing anything. Changing the way the env variables are set seems to fix this, the full tests now run. `grin-wallet` counterpart in a moment.

```
Starting: Windows Cargo Test
==============================================================================
Task         : Command line
Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
Version      : 2.182.0
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
==============================================================================
Generating script.
========================== Starting Command Output ===========================
"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\a4073cdf-5e02-40b8-8632-2b6fd7d6a3c8.cmd""
Refreshing environment variables from registry for cmd.exe. Please wait...Finished..
Finishing: Windows Cargo Test
```

### DIFF
--- a/.ci/windows-release.yml
+++ b/.ci/windows-release.yml
@@ -1,19 +1,16 @@
 steps:
-  - script: |
-      refreshenv
-      LIBCLANG_PATH=C:\Program Files\LLVM\lib
-      LLVM_CONFIG_PATH=C:\Program Files\LLVM\bin\llvm-config
-      ROARING_ARCH=x86-64-v2
-      cargo test --all
+  - script: 'refreshenv && cargo test --all'
+    env:
+      LIBCLANG_PATH: C:\Program Files\LLVM\lib
+      LLVM_CONFIG_PATH: C:\Program Files\LLVM\bin\llvm-config
+      ROARING_ARCH: x86-64-v2
     displayName: Cargo Test All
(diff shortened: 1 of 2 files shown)


antiochp:
Good catch.
I think something similar has actually bitten us in the past. I think it involves the first command returning 0 and treated as success or something similar.

.ci/windows-release.yml
```diff
@@ -1,19 +1,16 @@
...
-      cargo test --all
+  - script: 'refreshenv && cargo test --all'
+    env:
+      LIBCLANG_PATH: C:\Program Files\LLVM\lib
+      LLVM_CONFIG_PATH: C:\Program Files\LLVM\bin\llvm-config
+      ROARING_ARCH: x86-64-v2
     displayName: Cargo Test All
     condition:  and(succeeded(), contains(variables['Build.SourceBranch'], 'refs/tags/'), eq(variables['CI_JOB'], 'release' ))
-  - script: |
-      cargo clean
```
antiochp, line 11: Do we need `cargo clean` here?

trevyn: Fixed, thanks!

PR #3634, opened 2021-04-15
Add migration to `config_file_version = 2`

### LINKED ISSUE #3540: Need support for rewriting grin-server.toml file
The fixfees RFC lowered the accept_fee_base default from 1_000_000 to 500_000, but most nodes
have the 1_000_000 setting in the grin-server.toml configuration file, which would incorrectly override the new 500_000 default.

This resulted in the ugly fight-override fix at https://github.com/mimblewimble/grin/pull/3539

To get rid of that exception code, we would like to be able to rewrite the grin-server.toml file
in order to update obsolete defaults.

trevyn:
This adds an automatic migration of `grin-server.toml` to `config_file_version = 2`, performed as such:

- Add `config_file_version = 2` line after header comments
- If `server.pool_config.accept_fee_base` is 1000000, change it to 500000
- Remove the comment `#a setting to 1000000 will be overridden to 500000 to respect the fixfees RFC`, since it no longer will once `config_file_version` >= 2
- Verify that the updated toml parses to the expected result
- Write the updated `grin-server.toml` file to disk

This migration is only applied when the `config_file_version` key is not present in the existing `grin-server.toml` file.

For a fresh install, the migration is performed automatically when reading the initial default config.

Error handling can be improved upon request if this general approach is OK.

Closes #3540.

### DIFF
--- a/config/src/comments.rs
+++ b/config/src/comments.rs
@@ -19,7 +19,7 @@ use std::collections::HashMap;
 fn comments() -> HashMap<String, String> {
 	let mut retval = HashMap::new();
 	retval.insert(
-		"[server]".to_string(),
+		"config_file_version".to_string(),
 		"
 # Generated Server Configuration File for Grin
 #
@@ -31,7 +31,13 @@ fn comments() -> HashMap<String, String> {
 # -[user home]/.grin
 #
 
-#########################################
+"
+		.to_string(),
+	);
+
+	retval.insert(
+		"[server]".to_string(),
+		"#########################################
 ### SERVER CONFIGURATION              ###
 #########################################
 
@@ -323,7 +329,6 @@ fn comments() -> HashMap<String, String> {
 		"accept_fee_base".to_string(),
 		"
 #base fee that's accepted into the pool
-#a setting to 1000000 will be overridden to 500000 to respect the fixfees RFC
 "
 		.to_string(),
 	);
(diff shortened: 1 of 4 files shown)


phyro:
Awesome work, I have a couple of questions around the changes and what they mean
1. Should we define a fresh install as config version 2 with 500000 as default since this would allow us to remove this migration later on? or did I understand it wrong and we do that already?
2. If we are adding versioning to the node configuration, should we consider doing similar on the wallet side?

trevyn:
> Should we define a fresh install as config version 2 with 500000 as default

Functionally, it doesn't really matter. Leaving the version 1 defaults provides some implicit testing of the migration mechanism, which should probably get an explicit test if the default is changed. I see your point though, I'll think on it.

> since this would allow us to remove this migration later on?

I don't think it'd be a good idea to ever remove the migration — users could have arbitrarily old configs, so the code should always be able to migrate from a day-1 config file all the way up to whatever the current version is.

> If we are adding versioning to the node configuration, should we consider doing similar on the wallet side?

I took a quick look, and didn't see anything that would currently benefit from a new config version. If I missed something, happy to add that.

I'm also happy to spend the time on a more generalized approach (dreams of a general `toml_migration` crate), but I thought that might be falling into the over-engineering trap, and would make for a more difficult code review. :)

antiochp (commented):
This looks like a good approach here. :+1:
Planning to do some local testing of this over the next day or so.

antiochp:
We still include the offending comment via `comments.rs` when auto generating the config file -

https://github.com/mimblewimble/grin/blob/1b8acee72e7a4236cdf8561a7af5f894bfe11985/config/src/comments.rs#L322-L329

antiochp:
So I guess we should also update the generated file (if we generate one) to include the latest version number.

PR #3637, opened 2021-05-01
make the job_id of every task in stratum server unique to reduce iPollo G1 rejected shares.

xiaojay:
Problem to fix

https://forum.grin.mw/t/ipollo-g1-rejected-shares/8432

I work with iPollo G1 miner's developer and found that the reason of ipollo miner rejected shares is 
grin stratum server sometimes send job with same job_id but is not same pre_work, which make miner failed to work on the new job then submit the wrong share, which to be rejected.

In this pr, we use (current_state.current_block_versions.len() as u64) * bh.height as unique job_id. 

we have tested it in goblinpool.com, and reduce iPollo G1 rejected shares from ~5% -> ~0.2%

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -343,9 +343,10 @@ impl Handler {
 		}
 		let pre_pow = header_buf.to_hex();
 		let current_state = self.current_state.read();
+		let job_id = (current_state.current_block_versions.len() as u64) * bh.height;
 		let job_template = JobTemplate {
 			height: bh.height,
-			job_id: (current_state.current_block_versions.len() - 1) as u64,
+			job_id: job_id,
 			difficulty: current_state.minimum_share_difficulty,
 			pre_pow,
 		};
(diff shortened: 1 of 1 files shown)


tromp:
I'm no longer in physical posession of the G1-mini miner. But I recall that the testing I did showed that the rejected shares were not due to job id confusion.

I don't see how your patch makes a difference.

You're changing the outgoing job template from (height, job_id) to (height, (job_id+1)*height),
and then then undoing this transformation when it comes back from the miner.

Presuming that the miner preserves templates, this results in exactly the same state.current_block_versions lookups.

What am I missing?

bladedoyle:
Can you please detail the steps leading up to the bad share?

The stratum server sends out a job height x job_id 0
The G1 starts mining that job
The stratum server send out a new job height y job_id 0
The G1 continues to mine the original job because it looks only at job_id not height and the job_id didnt change so it ignores the new job
The G1 finds a solution and submits it 

But this wont result in an invalid share, it would result in a stale share.....unless the G1 uses the new block height when submitting?  That would be bad.

So the G1 must accept the new job data, continue to mine the old job, and then put the old jobs solution into the new job template and submit that?

phyro:
@xiaojay could you check the comments above when you find some free time? 🙏

xiaojay:
I close this pr and the new g1 miner software fix this on their side.

PR #3650, opened 2021-07-12
Update build.md

noobvie:
Add cargo package (and recommended packages) to be installed before excecuting cargo command.


yeastplume:
I've just installed a fresh dev system on debian 10 and didn't need to install the additional packages specified in the PR. Is anyone else able to verify whether these are needed on any particular system?

quentinlesceller:
Yeah I don't these packages are necessaries for building grin. We can close this one I think.

PR #3655, opened 2021-08-19
Update digest, hmac, ripemd160, sha2 and pbkdf2 crates

quentinlesceller:
As titled. Update:

- `digest`
- `hmac`
-  `ripemd160`
- `sha2`
- `pbkdf2`

crates. Get rid of an issue where publishing crates fails based on a yanked version of `crypto-mac`.


yeastplume (approved):
All looks good, thanks!

PR #3659, opened 2021-11-04
PMMR Backend Support for append_pruned_root (Continued)

yeastplume:
This is a continuation/completion of #3476

All of the comments and discussion from the previous PR still apply. I've merged this with with the most recent code, and reviewed/tested, particularly with respect to the simplification of the prune list cache in #3573.

As of now, this all appears to be working, and at least doesn't appear to break the basic existing case of expanding the pmmr by adding individual leaf nodes. Fuller testing of appending pruned segments to the pmmr will be performed in later tests/PRs.

### DIFF
--- a/core/src/core/pmmr/pmmr.rs
+++ b/core/src/core/pmmr/pmmr.rs
@@ -495,6 +495,24 @@ pub fn insertion_to_pmmr_index(mut sz: u64) -> u64 {
 	2 * sz - sz.count_ones() as u64 + 1
 }
 
+/// Returns the insertion index of the given leaf index
+pub fn pmmr_leaf_to_insertion_index(pos1: u64) -> Option<u64> {
+	if !is_leaf(pos1) {
+		return None;
+	}
+	let has_odd_leaf = n_leaves(pos1 - 1) % 2;
+	let cur_peaks = if has_odd_leaf == 1 {
+		peaks(pos1 - 1)
+	} else {
+		peaks(pos1)
+	};
+	Some(
+		cur_peaks.iter().fold(has_odd_leaf, |sum, n| {
+			sum + u64::pow(2, (bintree_postorder_height(*n) as u32).into()) + 1 / 2
+		}) - 1,
+	)
+}
+
 /// sizes of peaks and height of next node in mmr of given size
 /// Example: on input 5 returns ([3,1], 1) as mmr state before adding 5 was
 ///    2
@@ -672,7 +690,7 @@ pub fn bintree_rightmost(pos: u64) -> u64 {
 	pos - bintree_postorder_height(pos)
 }
 
-/// Gets the position of the rightmost node (i.e. leaf) beneath the provided subtree root.
+/// Gets the position of the leftmost node (i.e. leaf) beneath the provided subtree root.
 pub fn bintree_leftmost(pos: u64) -> u64 {
 	let height = bintree_postorder_height(pos);
 	pos + 2 - (2 << height)
@@ -680,9 +698,9 @@ pub fn bintree_leftmost(pos: u64) -> u64 {
 
 /// Iterator over all leaf pos beneath the provided subtree root (including the root itself).
 pub fn bintree_leaf_pos_iter(pos: u64) -> impl Iterator<Item = u64> {
-	let leaf_start = max(1, bintree_leftmost(pos as u64));
-	let leaf_end = bintree_rightmost(pos as u64);
-	(leaf_start..=leaf_end).filter(|x| is_leaf(*x))
+	let leaf_start = pmmr_leaf_to_insertion_index(bintree_leftmost(pos)).unwrap();
+	let leaf_end = pmmr_leaf_to_insertion_index(bintree_rightmost(pos)).unwrap();
+	(leaf_start..=leaf_end).map(|n| insertion_to_pmmr_index(n))
 }
 
 /// Iterator over all pos beneath the provided subtree root (including the root itself).
--- a/core/src/core/pmmr/pmmr.rs
+++ b/core/src/core/pmmr/pmmr.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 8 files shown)


core/src/core/pmmr/pmmr.rs
```diff
@@ -665,14 +668,27 @@ pub fn family_branch(pos: u64, last_pos: u64) -> Vec<(u64, u64)> {
...
+pub fn bintree_leftmost(pos: u64) -> u64 {
+	let height = bintree_postorder_height(pos);
+	pos + 2 - (2 << height)
+}
+
+/// Iterator over all leaf pos beneath the provided subtree root (including the root itself).
+pub fn bintree_leaf_pos_iter(pos: u64) -> impl Iterator<Item = u64> {
+	let leaf_start = max(1, bintree_leftmost(pos as u64));
+	let leaf_end = bintree_rightmost(pos as u64);
+	(leaf_start..=leaf_end).filter(|x| is_leaf(*x))
```
tromp, line 685: It would be more efficient to iterate over leaf_indices, and map those into node_indices with the very cheap
map( \leaf_index -> 2 * leaf_index - count_ones(leaf_index) )

core/src/core/pmmr/pmmr.rs
```diff
@@ -665,14 +668,27 @@ pub fn family_branch(pos: u64, last_pos: u64) -> Vec<(u64, u64)> {
...
+}
+
+/// Iterator over all pos beneath the provided subtree root (including the root itself).
+pub fn bintree_pos_iter(pos: u64) -> impl Iterator<Item = u64> {
+	let leaf_start = max(1, bintree_leftmost(pos as u64));
```
tromp, line 690: why the max(1, ?

store/src/prune_list.rs
```diff
@@ -54,16 +55,21 @@ pub struct PruneList {
...
+	pub fn new(path: Option<PathBuf>, bitmap: Bitmap) -> PruneList {
+		let mut prune_list = PruneList {
 			path,
-			bitmap,
+			bitmap: Bitmap::create(),
 			shift_cache: vec![],
 			leaf_shift_cache: vec![],
+		};
+
+		for pos in bitmap.iter().filter(|x| *x > 0) {
```
tromp, line 67: the non-zero test is either wrong (for 0-based pos) or redundant (for 1-based pos).

store/src/prune_list.rs
```diff
@@ -179,6 +180,18 @@ impl PruneList {
 		}
 	}
 
+	// Calculate the next shift based on provided pos and the previous shift.
+	fn calculate_next_shift(&self, pos: u64) -> u64 {
+		let prev_shift = self.get_shift(pos.saturating_sub(1));
```
tromp, line 185: pos.saturating_sub(1) should be simply pos-1 since pos is 1-based and cannot be 0.

tromp: I suggest we change all occurrences of pos to either pos0 or pos1 depending on whether they're 0-based or 1-based.

store/src/prune_list.rs
```diff
@@ -179,6 +180,18 @@ impl PruneList {
...
+		let shift = if self.is_pruned_root(pos) {
+			let height = bintree_postorder_height(pos);
+			2 * ((1 << height) - 1)
```
tromp, line 188: Why is shift 1 less than a subtree size?

core/src/core/pmmr/pmmr.rs
yeastplume: Is there a way to iterate over the leaf indices of a subtree without reading them from an already-stored bitmap, (which is implementation-specific and not available in these helpers which I assume are supposed to be 'pure' MMR operations)?

tromp: i don't see what a bitmap has to do with this routine which is only about indices.
i'm suggesting to convert both leaf_start and leaf_end into leaf indices (which is as expensive as an is_leaf test), using a node_to_leaf_index method, then construct a range from those, and map each element back to a node index.

yeastplume: thanks, that makes it clearer, will implement

tromp: the max(1, part is also redundant here.

PS: I just made a PR rewriting that MMR doc to stick with 0-based positions.

yeastplume: added a `pmmr_leaf_to_insertion_index` function along with some tests and modified `bintree_leaf_pos_iter` to use it. (Note this introduces error handling that wasn't previously needed, so there's still an awkward unwrap in there to deal with)

tromp (changes requested):
pmmr_leaf_to_insertion_index is way too complicated.
    let (insertidx, height) = peak_map_height(pos1 - 1);
    if (height == 0) then Some (insertidx) else None;
should do it...

core/src/core/pmmr/pmmr.rs
yeastplume: it's redundant in theory however the tests are constructed with the assumption that `bintree_pos_iter(0)` returns an empty iterator which is produced by this, so I don't want to break this assumption ATM.

store/src/prune_list.rs
yeastplume: agreed in theory, but removing this test breaks existing assumptions about the underlying data

store/src/prune_list.rs
yeastplume: In this case yes, we can just change the saturating sub since it's only called in one place at the moment that will fail with assertion beforehand is pos is 0

yeastplume: I'm going to do a bit of a 'clean as you go' on changing pos to pos0 or pos1 at the moment, but if we do a complete pass on all instances it should be within a separate PR.

tromp: Let me know when you're done making pmmr changes so I can do my own cleanup...

tromp (approved):
So far so good.
Needs more work on 1-based indices but will work on that in separate PR...

PR #3660, opened 2021-11-05
don't confuse with different numbering bases

tromp:
Minor change in MMR doc, rewriting a few paragraphs that for no good reason used 1-based positions.

### DIFF
--- a/doc/mmr.md
+++ b/doc/mmr.md
@@ -65,46 +65,20 @@ so we need a method to compute one (otherwise it would defeat the purpose of
 using a hash tree). This process is called "bagging the peaks" for reasons
 described in [1].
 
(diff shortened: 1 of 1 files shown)


phyro:
Unrelated to this change, but I do wonder if the hashing in the document is correct e.g. `P = Blake2b(N | Blake2b(N | Node(p3) | Node(p2)) | Node(p1))` is correct. I don't understand why we'd hash this way.

tromp:
> Unrelated to this change, but I do wonder if the hashing in the document is correct e.g. `P = Blake2b(N | Blake2b(N | Node(p3) | Node(p2)) | Node(p1))`. I don't understand why we'd hash this way.

Thanks for spotting that mistake. Fixed now.

yeastplume (approved):

PR #3663, opened 2021-11-10
refactor pmmr functions, improve efficiency; panic on 0 pos1

tromp:
---
name: fixmmr_part1
about: streamline MMR code
title: Fix MMR part 1
labels: 
assignees: Yeastplume

---
This PR refactors pmmr.rs, improving efficiency and clarifying use of 0-based vs 1-based positions, introducing panics when passing 0 as a 1-based position, which makes no sense in normal use. The latter did require removal of corresponding test-cases.
A followup PR is planned to make all methods use 0-based positions, which are preferable for two reasons:
1) they're baked into the consensus model through hash_with_index
2) they simplify some computations, such as converting leaf to node index:
     node_index = 2 * leaf_index - count_ones(leaf_index)
There appears to be no advantage to 1-based positions. The APIs use them though, so the API implementation is where all the conversion should take place.

### DIFF
--- a/core/src/core/pmmr/pmmr.rs
+++ b/core/src/core/pmmr/pmmr.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::{cmp::max, iter, marker, ops::Range, u64};
+use std::{iter, marker, ops::Range, u64};
 
 use croaring::Bitmap;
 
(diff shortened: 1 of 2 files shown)


yeastplume:
All looks good from my perspective. The panics should be replaced with proper error types, but moving all of the internals to be 0 based should remove the need from them. The API will remain 1-based, so we'll need to make sure all the error propagation is at that level.

yeastplume (approved):

PR #3665, opened 2021-11-15
[PIBD] Chain Segmenter Validation Test + Block Archive Horizon Change

yeastplume:
Review Version:

* Adds an integration test that acts upon small test instances of chain data, both compacted and uncompacted. The test reads all 4 datasets from the PIBD segmenter in the order which it might happen during a live sync, then validates all of the segments. It also reconstructs the raw `croaring::Bitmap` from the read bitmap segments for use in output/rangeproof segment validation.
* Adds the sample test chains (as output by the new wallet test https://github.com/mimblewimble/grin-wallet/pull/628) as binary data. Not ideal, but much better than what we have now to test this (i.e. nothing). It's anticipated this test set will be updated and used in further testing.
* Adds an #[ignored] test that will allow users to manually run the validation test against another instance of the chain. I've run this test against a copy of the recent live chain, with segments validating successfully.
* Removed the temporary initialisation at startup of the chain segmenter (it will need to be invoked and updated lazily once messaging is in place)
* IMPORTANT TO REVIEW - During chain compaction on test data, the chain discards blocks that the segmenter expects to be in place. Added a manual check to adjust the horizon at which blocks are discarded if the calculated cutoff in `chain::remove_historical_blocks` is greater than the horizon expected by the segmenter
* Adds a function to BitmapChunk to return an iterator over the set of integers represented by that chunk (with an offset added). Used as a helper to reconstruct a raw `croaring::Bitmap` for segment validation.
* Adds functions to `SegmentIdentifier` to determine the number of segments that would need to be requested given the target pmmr size and the desired segment height.

---- Old Version of this ----

Snapshot of current explorations into PIBD, note this PR in its current state is more an attempt at understanding and starting to put together tests to exercise the the PIBD segmenter and segment validation. The current goal is to get an entire compressed/pruned hashset 'copied' locally via PIBD functions, and once this is in place (with sufficient unit tests exercising all of the PIBD PMMR work done to date,) we can start thinking about the actual sync process. 

Thus far:

- Introduction of some static test data into the `chain` crate, a small but 'real' transaction hash set with real transaction data generated via a wallet test (will commit that test separately to grin_wallet). (Yes, I'm adding static binary data to the repository for the purposes of testing PIBD/PMMR segmentation. Never ideal but much better than the nothing we currently have).
- WIP 'megatest' that reads a test chain, copies headers into a new chain instance then attempts to read and validate segments from the original. Note this will likely evolve into series of tests exercising the segmenter and segment validation (as none exist at the moment), which will likely be what this PR ends up being.

Next up:
- Figure out why output root validation isn't working
- Read multiple segments from the source hashset
- Addition of chain/txhashset functions to rebuild a new chain based on received segments (likely the focus of the next PR)

Don't forget:
- NRD Kernels need to be considered in the validation (to be discussed)


yeastplume:
Top comment updated, ready for review. This mostly adds segmenter tests against live data (developed both because they weren't there and for my own understanding,) which should also serve as a reference for how clients would call and validate data from other clients. Next up would be functionality + further tests that recreate hash set data from read segments.

chain/tests/test_pibd_validation.rs
```diff
@@ -0,0 +1,235 @@
...
+			let h = src_chain.get_header_by_height(h).unwrap();
+			dest_chain.process_block_header(&h, options).unwrap();
+		}*/
+
+		// Init segmenter, (note this still has to be lazy init somewhere on a peer)
+		// This is going to use the same block as horizon_header
+		let segmenter = src_chain.segmenter().unwrap();
+
+		// BITMAP - Read + Validate, Also recreate bitmap accumulator for target tx hash set
+		// Predict number of leaves (chunks) in the bitmap MMR from the number of outputs
```
tromp, line 84: This is compute rather than predict. No uncertainty involved:-)

chain/tests/test_pibd_validation.rs
```diff
@@ -0,0 +1,235 @@
...
+			.clone();
+		println!("BITMAP PMMR SIZE: {}", bitmap_pmmr_size);
+		println!(
+			"Bitmap Segments required: {}",
+			SegmentIdentifier::count_segments_required(bitmap_pmmr_size, target_segment_height)
+		);
+		// TODO: This can probably be derived from the PMMR we'll eventually be building
+		// (check if total size is equal to total size at horizon header)
+		let identifier_iter =
+			SegmentIdentifier::traversal_iter(bitmap_pmmr_size, target_segment_height);
```
tromp, line 102: Since you just computed count_segments_required, you might as well check that it equals the identifier_iter size?!

core/src/core/pmmr/segment.rs
```diff
@@ -69,6 +69,28 @@ impl Writeable for SegmentIdentifier {
...
+				height: segment_height,
+				idx: idx as u64,
+			},
+		)
+	}
+
+	/// Returns number of segments required that would needed in order to read a
+	/// pmmr of size `target_mmr_size` in segments of height `segment_height`
+	pub fn count_segments_required(target_mmr_size: u64, segment_height: u8) -> usize {
+		pmmr::n_leaves(target_mmr_size) as usize / (1 << segment_height as usize) + 1
```
tromp, line 90: why + 1 ?
Do you add an empty segment if mmr size is an exact multiple of 2^segment_height ?

tromp (approved):
All looks reasonable. See my few in-line comments for minor issues.

core/src/core/pmmr/segment.rs
yeastplume: Left over from earlier hack-test, thanks for spotting

chain/tests/test_pibd_validation.rs
yeastplume: It's prediction with 100% certainty!

chain/tests/test_pibd_validation.rs
yeastplume: Yes, I was doing that at one point but removed it (led to redundant code since `count()` on an iterator consumes it, I'm sure there are other ways but didn't go further). Will add something similar into future tests.

yeastplume:
Merged, thanks for review!

PR #3667, opened 2021-11-22
[PIBD] PMMR Desegmenter Structure (Pt. 1)

yeastplume:
This is the first part of several anticipated PRs to enable and test the reconstruction of the txhashset from PMMR segments. Note that none of the code introduced here is called from anywhere but the main test `test_pibd_copy`. Merging this a bit earlier than I'd planned, but breaking up the work like this keeps the review process a bit more manageable, and the included test is needed to help track down a potential serious performance issue in header sync 😄 

Included in this PR:

* Introduce a `desegmenter` type containing functions to validate and append PMMR segments.
* Add optional instance of desegmenter to chain, with lazy init method
* Function in `BitmapAccumulator` to return the accumulator as a raw bitmap for use in desegmenter validation
* Add functions in txhashset extension to apply outputs, rangeproofs and kernels within an extension
* Add test to copy sample dataset from previous PR #3665 into new txhashset.
* Fix for SegmentIdentifier returning wrong number of segments in `count_segments_required`
* change `last_pos` to `mmr_size` in Segment to be consistent with changes #3666 

Note the current copy tests do work, however they haven't been tested or verified against actual pruned data.


chain/src/chain.rs
```diff
@@ -863,6 +865,43 @@ impl Chain {
 		))
 	}
 
+	/// instantiate desegmenter (in same lazy fashion as segmenter, though this should be be
```
tromp, line 868: to be be or not to be be...

chain/src/txhashset/desegmenter.rs
```diff
@@ -0,0 +1,243 @@
...
+		)?;
+		Ok(())
+	}
+
+	// Calculate and store number of leaves and positions in the bitmap mmr given the number of
+	// outputs specified in the header. Should be called whenever the header changes
+	fn calc_bitmap_mmr_sizes(&mut self) {
+		// Number of leaves (BitmapChunks)
+		self.bitmap_mmr_leaf_count =
+			(pmmr::n_leaves(self.archive_header.output_mmr_size) as f64 / 1024f64).ceil() as u64;
```
tromp, line 126: Don't use floating point!
(n+1023) / 1024 computes rounded up division with plain integers.

chain/src/txhashset/desegmenter.rs
```diff
@@ -0,0 +1,243 @@
...
+			.unwrap_or(&pmmr::insertion_to_pmmr_index(self.bitmap_mmr_leaf_count))
+			.clone();
+		debug!(
+			"pibd_desgmenter - expected size of bitmap MMR: {}",
+			self.bitmap_mmr_size
+		);
+	}
+
+	/// Adds and validates a bitmap chunk
+	/// TODO: Still experimenting, this expects chunks received to be in order
```
tromp, line 143: Why do you need them in order?

core/src/core/pmmr/segment.rs
```diff
@@ -87,7 +87,8 @@ impl SegmentIdentifier {
 	/// Returns number of segments required that would needed in order to read a
 	/// pmmr of size `target_mmr_size` in segments of height `segment_height`
 	pub fn count_segments_required(target_mmr_size: u64, segment_height: u8) -> usize {
-		pmmr::n_leaves(target_mmr_size) as usize / (1 << segment_height as usize)
+		(pmmr::n_leaves(target_mmr_size) as f64 / (1 << segment_height as usize) as f64).ceil()
```
tromp, line 90: No floating point please! See other comment.

chain/src/txhashset/txhashset.rs
```diff
@@ -1232,6 +1239,33 @@ impl<'a> Extension<'a> {
...
+			self.output_pmmr
+				.push(&output_identifier)
+				.map_err(&ErrorKind::TxHashSetErr)?;
+		}
+		Ok(())
+	}
+
+	/// Apply a rangeproof segment to the output PMMR. must be called in order
+	/// TODO: Not complete
+	pub fn apply_rangeproof_segment(&mut self, segment: Segment<RangeProof>) -> Result<(), Error> {
```
tromp, line 1259: why feed in a whole segment when you only use one component?

chain/src/txhashset/txhashset.rs
```diff
@@ -1232,6 +1239,33 @@ impl<'a> Extension<'a> {
...
+		let (_sid, _hash_pos, _hashes, _leaf_pos, leaf_data, _proof) = segment.parts();
+		for proof in leaf_data {
```
tromp, line 1261: too much copy-paste?!
shouldn't you use the proof component here?

chain/tests/test_pibd_copy.rs
```diff
@@ -0,0 +1,233 @@
...
+	clean_output_dir(&dest_root_dir);
+}
+
+#[test]
+#[ignore]
+// As above, but run on a real instance of a chain pointed where you like
+fn test_pibd_copy_real() {
+	util::init_test_logger();
+	// if testing against a real chain, insert location here
+	let src_root_dir = format!("/Users/yeastplume/Projects/grin_project/server/chain_data");
```
tromp, line 228: This is not gonna work on my machine:-(

chain/tests/test_pibd_validation.rs
```diff
@@ -87,7 +87,7 @@ fn test_pibd_chain_validation_impl(is_test_chain: bool, src_root_dir: &str) {
 		println!("BITMAP PMMR NUM_LEAVES: {}", bitmap_mmr_num_leaves);
 
 		// And total size of the bitmap PMMR
-		let bitmap_pmmr_size = pmmr::peaks(bitmap_mmr_num_leaves + 1)
+		let bitmap_pmmr_size = pmmr::peaks(bitmap_mmr_num_leaves)
```
tromp, line 90: I assume this is because the former was written before my 1-based -> 0-based conversion?

tromp (approved):
Not a particularly through review, but found some minor points to be addressed...

yeastplume:
Thank you, will address bits before merging this. It's perhaps a bit early for review and there are still a lot of things here that are going to change, but I'd just like to get everything I have to date merged before continuing with this.

chain/tests/test_pibd_copy.rs
yeastplume: it's not meant to, it's meant for running a manual test against a live chain instance (so while testing you modify the path).. I'll make this more clear in comments

core/src/core/pmmr/segment.rs
yeastplume: 👍

chain/src/txhashset/desegmenter.rs
yeastplume: :+1:

chain/tests/test_pibd_validation.rs
yeastplume: yes

yeastplume:
Addressed a few issues, will keep the rest in consideration during future work

PR #3670, opened 2021-11-30
[SYNC PERFORMANCE] Replace header proof serialisation with more efficient algorithm

yeastplume:
In performing test for PIBD work, it's become evident that header validation is not as efficient as it could be. There are a few reasons for this (with more to be addressed in coming PRs), but much of it comes down to the number of calls to the difficulty iterator in `store::DifficultyIter`, which deserialises several entire headers from the DB each time it's called. This iterator next method is called 60 times on each header validation to get the block's expected difficulty, which uncovered a large performance issue around the inefficient 'BitVec' struct.

A test (ignored in CI, meant to be run manually against live chain data) is included that copies and validates 100k headers from one chain to another. Before this change, this test in release mode (Mac, 2.3 Ghz 8-Core Intel i9) took about 76 seconds. With this change, the test takes 35 seconds. Flamegraphs are attached demonstrating the before and after (note the huge amount of time spent in `DifficultyIter::next()`)

Before:
image

After:
image

Note there's still more to do here performance wise, but was an obvious first candidate for optimisation. Also investigated the use of a bitpacking lib https://github.com/quickwit-inc/bitpacking that actually uses SSE and AVX2 instructions to optimise further if available, but it unfortunately only supports packing up to u32.

Exact Changes: 
* Remove BitVec
* Add `pack_bits` function, which packs an array of u64s to a specified bit length more efficiently than the previous bitvec implementation
* Add `Proof::pack_nonces` function, which packs an array of u64s to a specified bit length more efficiently
* Add `Proof::pack_len` helper function and use instead of previous bitvec

### DIFF
--- a/core/src/pow/types.rs
+++ b/core/src/pow/types.rs
@@ -21,7 +21,7 @@ use rand::{thread_rng, Rng};
 use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
 /// Types for a Cuck(at)oo proof of work and its encapsulation as a fully usable
 /// proof of work within a block header.
-use std::cmp::{max, min};
+use std::cmp::{max, min, Ordering};
 use std::ops::{Add, Div, Mul, Sub};
 use std::u64;
 use std::{fmt, iter};
@@ -405,11 +405,20 @@ impl Proof {
 	/// Pack the nonces of the proof to their exact bit size as described above
 	pub fn pack_nonces(&self) -> Vec<u8> {
 		let mut compressed = vec![0u8; Proof::pack_len(self.edge_bits)];
+		pack_bits_old(
+			self.edge_bits,
+			&self.nonces[0..self.nonces.len()],
+			&mut compressed,
+		);
+		println!("PACK OLD: {:?}", compressed);
+
+		let mut compressed = vec![0u8; Proof::pack_len(self.edge_bits)];
 		pack_bits(
 			self.edge_bits,
 			&self.nonces[0..self.nonces.len()],
 			&mut compressed,
 		);
+		println!("PACK NEW: {:?}", compressed);
 		compressed
 	}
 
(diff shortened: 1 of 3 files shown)


yeastplume:
Also, suggestions to speed up `pack_bits` welcome!

core/src/pow/types.rs
```diff
@@ -448,8 +511,7 @@ impl Readable for Proof {
 		// prepare nonces and read the right number of bytes
 		let mut nonces = Vec::with_capacity(global::proofsize());
 		let nonce_bits = edge_bits as usize;
-		let bits_len = nonce_bits * global::proofsize();
-		let bytes_len = BitVec::bytes_len(bits_len);
+		let bytes_len = (nonce_bits * global::proofsize() + 7) / 8;
```
tromp, line 514: Should use pack_len here?!

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
 }
 
+/// Pack an array of u64s into `compressed` at the specified bit width. Caller
+/// must ensure `compressed` is the right size
+fn pack_bits(bit_width: u8, uncompressed: &[u64], mut compressed: &mut [u8]) {
+	// We will use a `u64` as a mini buffer of 64 bits.
+	// We accumulate bits in it until capacity, at which point we just copy this
+	// mini buffer to compressed.
+	let mut mini_buffer: u64 = 0u64;
+	let mut cursor = 0; //< number of bits written in the mini_buffer.
```
tromp, line 430: i'd use only remaining, no cursor.

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+	let mut pack_bytes_remaining = compressed.len();
+	for el in uncompressed {
+		let remaining = 64 - cursor;
```
tromp, line 433: this becomes redundant

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+		match bit_width.cmp(&remaining) {
```
tromp, line 434: use a plain comparison bit_width.cmp <= remaining, since

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+			Ordering::Less => {
+				// Plenty of room remaining in our mini buffer.
+				mini_buffer |= el << cursor;
```
tromp, line 437: this needs to happen in both Less and Equal

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+				cursor += bit_width;
```
tromp, line 438: this becomes remaining -= bit_width

which is harmless in case of Equal

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+			}
+			Ordering::Equal => {
+				mini_buffer |= el << cursor;
+				// We have completed our minibuffer exactly.
+				// Let's write it to `compressed`.
+				compressed[..8].copy_from_slice(&mini_buffer.to_le_bytes());
```
tromp, line 444: these remaining lines can be in a nested if bit_width.cmp == remaining

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+				compressed = &mut compressed[8..];
+				pack_bytes_remaining -= 8;
+				mini_buffer = 0u64;
+				cursor = 0;
```
tromp, line 448: this becomes remaining = 64

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+	for el in uncompressed {
+		let remaining = 64 - cursor;
+		match bit_width.cmp(&remaining) {
+			Ordering::Less => {
+				// Plenty of room remaining in our mini buffer.
+				mini_buffer |= el << cursor;
+				cursor += bit_width;
+			}
+			Ordering::Equal => {
+				mini_buffer |= el << cursor;
```
tromp, line 441: this becomes << (64 - remaining)

core/src/pow/types.rs
tromp: mini_buffer |= el << (64 - remaining)
happens in all 3 cases, and so should be done here before the comparison.

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+				// Let's write it to `compressed`.
+				compressed[..8].copy_from_slice(&mini_buffer.to_le_bytes());
+				compressed = &mut compressed[8..];
+				pack_bytes_remaining -= 8;
+				mini_buffer = 0u64;
+				cursor = 0;
+			}
+			Ordering::Greater => {
+				mini_buffer |= el << cursor;
+				// We have completed our minibuffer.
```
tromp, line 452: i'd say overflowed:-)

core/src/pow/types.rs
tromp, line 434: There's too much overlap between cases Equal and Greater to merit separating.

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
 
+/// Pack an array of u64s into `compressed` at the specified bit width. Caller
+/// must ensure `compressed` is the right size
+fn pack_bits(bit_width: u8, uncompressed: &[u64], mut compressed: &mut [u8]) {
+	// We will use a `u64` as a mini buffer of 64 bits.
+	// We accumulate bits in it until capacity, at which point we just copy this
+	// mini buffer to compressed.
+	let mut mini_buffer: u64 = 0u64;
+	let mut cursor = 0; //< number of bits written in the mini_buffer.
+	let mut pack_bytes_remaining = compressed.len();
```
tromp, line 431: no need to keep track of this.

core/src/pow/types.rs
```diff
@@ -396,13 +402,70 @@ impl Proof {
...
+				// with the remaining bits.
+				compressed[..8].copy_from_slice(&mini_buffer.to_le_bytes());
+				compressed = &mut compressed[8..];
+				pack_bytes_remaining -= 8;
+				cursor = bit_width - remaining;
+				mini_buffer = el >> remaining;
+			}
+		}
+	}
+	if pack_bytes_remaining > 0 {
```
tromp, line 463: pack_bytes_remaining is simply compressed.len() % 8.

tromp (approved):
looking good; left some suggestions for minor improvement.

tromp (approved):
See the comments with suggestions for minor improvement.

We should similarly rewrite the repeated calls to read_number in Proof::read
into a single unpack function.

yeastplume:
Thank you, the compression function has been greatly compressed with suggestions above

PR #3674, opened 2021-12-12
Make fn rewind_hash pub

deevope:
Following phyro suggestion on the [grin-wallet #630](https://github.com/mimblewimble/grin-wallet/pull/630#discussion_r767196139). 
We need to make the rewind_hash function in ViewKey public in order to use it directly in the get_rewind_hash function on the wallet.

### DIFF
--- a/keychain/src/view_key.rs
+++ b/keychain/src/view_key.rs
@@ -94,7 +94,7 @@ impl ViewKey {
 		})
 	}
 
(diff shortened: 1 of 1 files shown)


yeastplume (approved):
All good, no reason to not do this.

Anynomouss:
@deevope Can you point me to any information on using the rewind hash argument? 
I have run the command and got a hash and now want to to use the hash to scan an archive node for historically associate outputs? The aim is on the long term to add the function to partially restore a wallets transaction history by showing historically created and destroyed outputs.

Issue #3649, opened 2021-07-10
ApiServer should shutdown gracefully


snape479:
Previously, 'with_graceful_shutdown' was added to allow the ApiServer to shutdown gracefully here:

https://github.com/mimblewimble/grin/commit/4a09fed36ca8f2b9f3ccd2e1c518e777a4c1da6a

However, it had to be reverted due to the fact that the 'rx' was being dropped at the end of the code block and a 'cancel' error occurred right away. The solution to this problem is to store the (tx, rx) on the heap using Box and moving them to the startup routines in src/bin/grin.rs. Additionally, a monitor thread is needed to call stop when the StopState indicates the server stopped.

I have implemented this in bitcoinmw here:

https://github.com/bitcoinmw/bitcoinmw/commit/2f25ec2a5d2c6b7cf646e25f4f18858e68702473

I was going to do a pull request to Grin, but my repo had some merge issues (due to verifier_cache). If someone could copy these changes from BMW over to Grin that would be great. Opening this issue to track.

phyro:
I believe this is now resolved. Thanks for the issue and the fix.

PR #3677, opened 2021-12-27
ApiServer gracefully shutdown

deevope:
copy/past from @snape479 https://github.com/mimblewimble/grin/issues/3649


yeastplume (approved):
Looks good to me, thanks @snape479 for providing the original fix.

Issue #3609, opened 2021-03-20
[feature] Introduce command 'verify-chain' to verify the whole chain state


phyro:
**Is your feature request related to a problem? Please describe.**
There is no way to verify the chain state from the command line. We assume the UTXO set, the set of kernels and the offset produce a valid state. In general this should be true, but as was seen in the recent events, it is possible to produce an invalid state through block processing due to the use of various optimization techniques.

**Describe the solution you'd like**

As a user, I would like to run `./grin verify-chain` which would verify the whole state of the blockchain at that point in time.

Verification processing includes:
1. Verification of block headers
2. Verification of kernels
3. Verification of the UTXO set
4. Validity of the chain by verifying that the `sum(kernels) = sum(utxo_set) + height*H + offset*G` (this formula is not precise but you get the point)

It is possible I missed some elements of the chain that would need to be verified.

**Describe alternatives you've considered**
I think the alternative right now is to resync the node - although I'm not sure if this does such verification until the horizon point and then block validation for the last week which may include the optimizations that I would like to avoid.

**Additional context**
This feature may be useful for catching live processing bugs due to optimizations or other reasons e.g. an inflation bug could have been caught earlier if someone was running this command periodically for their local chain.

Note: Since this command requires a lot of computation, it would probably mean that the node would need to stop processing and propagating blocks to avoid changing the state while we verify it. An example of what we don't want is to prune the rangeproofs or outputs that are in the UTXO set at the time we started verifying the chain state. Stopping processing seems conceptually simpler and safer than labeling merkle entries as removed until the computation is finished and later running a garbage collector.

trevyn:
This could be a useful command for an “extended test” CI job as well.

trevyn:
There is a [`validate_chain`](https://docs.rs/grin_api/5.0.1/grin_api/struct.Owner.html#method.validate_chain) Owner API call, I wonder what that does and how it differs from the desired behavior.

phyro:
> There is a [`validate_chain`](https://docs.rs/grin_api/5.0.1/grin_api/struct.Owner.html#method.validate_chain) Owner API call, I wonder what that does and how it differs from the desired behavior.

[...]

PR #3678, opened 2021-12-27
`verify-chain` node client arg

deevope:
Following this feature suggestion https://github.com/mimblewimble/grin/issues/3609
- Add the parameter on the api `fast_validation`
- Add verify-chain node client argurment

Usage:
Full chain validation
```grin client verify-chain```

Partial chain validation
```grin client verify-chain --fast```

PS: This PR need the API to gracefully shutdown, need to merge https://github.com/mimblewimble/grin/pull/3677

### DIFF
--- a/api/src/client.rs
+++ b/api/src/client.rs
@@ -35,7 +35,7 @@ pub fn get<T>(url: &str, api_secret: Option<String>) -> Result<T, Error>
 where
 	for<'de> T: Deserialize<'de>,
 {
-	handle_request(build_request(url, "GET", api_secret, None)?)
+	handle_request(build_request(url, "GET", api_secret, None)?, None)
 }
 
 /// Helper function to easily issue an async HTTP GET request against a given
@@ -52,7 +52,7 @@ where
 /// on a given URL that returns nothing. Handles request
 /// building and response code checking.
 pub fn get_no_ret(url: &str, api_secret: Option<String>) -> Result<(), Error> {
-	send_request(build_request(url, "GET", api_secret, None)?)?;
+	send_request(build_request(url, "GET", api_secret, None)?, None)?;
 	Ok(())
 }
 
@@ -60,13 +60,18 @@ pub fn get_no_ret(url: &str, api_secret: Option<String>) -> Result<(), Error> {
 /// object as body on a given URL that returns a JSON object. Handles request
 /// building, JSON serialization and deserialization, and response code
 /// checking.
-pub fn post<IN, OUT>(url: &str, api_secret: Option<String>, input: &IN) -> Result<OUT, Error>
+pub fn post<IN, OUT>(
+	url: &str,
+	api_secret: Option<String>,
+	input: &IN,
+	read_timeout: Option<u64>,
+) -> Result<OUT, Error>
 where
 	IN: Serialize,
 	for<'de> OUT: Deserialize<'de>,
 {
 	let req = create_post_request(url, api_secret, input)?;
-	handle_request(req)
+	handle_request(req, read_timeout)
 }
 
 /// Helper function to easily issue an async HTTP POST request with the
@@ -94,7 +99,7 @@ pub fn post_no_ret<IN>(url: &str, api_secret: Option<String>, input: &IN) -> Res
 where
 	IN: Serialize,
 {
-	send_request(create_post_request(url, api_secret, input)?)?;
+	send_request(create_post_request(url, api_secret, input)?, None)?;
 	Ok(())
 }
 
@@ -110,7 +115,7 @@ pub async fn post_no_ret_async<IN>(
 where
 	IN: Serialize,
 {
-	send_request_async(create_post_request(url, api_secret, input)?).await?;
+	send_request_async(create_post_request(url, api_secret, input)?, None).await?;
 	Ok(())
 }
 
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: api/src/client.rs [hunk 1]
```diff
@@ -52,21 +52,26 @@ where
...
 /// object as body on a given URL that returns a JSON object. Handles request
 /// building, JSON serialization and deserialization, and response code
 /// checking.
-pub fn post<IN, OUT>(url: &str, api_secret: Option<String>, input: &IN) -> Result<OUT, Error>
+pub fn post<IN, OUT>(
+	url: &str,
+	api_secret: Option<String>,
+	input: &IN,
+	read_timeout: Option<u64>,
+) -> Result<OUT, Error>
```
tromp, line 68: Instead of adding Option parameters to multiple methods, isn't it preferable to add a new timeout member,
and a method to change it from its default value of 20secs?

api/src/owner.rs
```diff
@@ -80,17 +80,20 @@ impl Owner {
 
 	/// Trigger a validation of the chain state.
 	///
+	/// # Arguments
+	/// * `fast_validation` - if false verify the rangeproof associated with each unspent output and all kernels
```
tromp, line 84: i'd prefer this is called something like assume_valid_rangeproofs_kernels,
to reflect its importance and potential for harm.

src/bin/cmd/client.rs
```diff
@@ -149,6 +159,21 @@ impl HTTPNodeClient {
 		e.reset().unwrap();
 	}
 
+	pub fn validate_chain(&self, fast_validation: bool) {
```
tromp, line 162: how come this is a new method, rather than a changed method?

src/bin/cmd/client.rs
```diff
@@ -149,6 +159,21 @@ impl HTTPNodeClient {
...
+		let mut e = term::stdout().unwrap();
+		let params = json!([fast_validation]);
+		writeln!(
+			e,
+			"Checking the state of the chain. This might take time..."
+		)
+		.unwrap();
+		match self.send_json_request::<()>("validate_chain", &params) {
+			Ok(_) => writeln!(e, "Successfully validated chain.").unwrap(),
```
tromp, line 171: This output should reflect whether rangeproofs and kernels were validated.

src/bin/cmd/client.rs
```diff
@@ -191,6 +216,10 @@ pub fn client_command(client_args: &ArgMatches<'_>, global_config: GlobalConfig)
 			let hash = args.value_of("hash").unwrap();
 			node_client.invalidate_header(hash.to_string());
 		}
+		("verify-chain", Some(args)) => {
+			let fast_validation = if args.is_present("fast") { false } else { true };
```
tromp, line 220: This is simply let fast_validation = !args.is_present("fast")

tromp (commented):
See in-line comments.

src/bin/cmd/client.rs
deevope: I don't really get what you mean here by a changed method?

tromp: I mean that it looks like you're adding a validate_chain method to HTTPNodeClient,
whereas elsewhere in this commit, there are changes to existing validate_chain methods.

deevope: ah ok, will call it, `verify_chain` to make it clear it's a new method.

tromp: I don't object to the name. If it behaves identical to those others, then the same name is appropriate. But I just wondered, since I'm not very familiar with this code, why HTTPNodeClient didn't need such a method before and does now.

api/src/client.rs
deevope: I did it a bit more cleaner but just adding a method in api/client.rs and trigger a change of timeout in function of the api method used is a bit painful in the form it's built. If you have better idea, i'm open.
However these method are almost only used for cmd/client.rs

api/src/handlers/chain_api.rs
```diff
@@ -53,9 +53,9 @@ pub struct ChainValidationHandler {
 }
 
 impl ChainValidationHandler {
-	pub fn validate_chain(&self) -> Result<(), Error> {
+	pub fn validate_chain(&self, fast_validation: bool) -> Result<(), Error> {
```
tromp, line 56: why not use assume_valid_rangeproofs_kernels here as well?

src/bin/cmd/client.rs
```diff
@@ -149,6 +158,27 @@ impl HTTPNodeClient {
 		e.reset().unwrap();
 	}
 
+	pub fn verify_chain(&self, fast_validation: bool) {
```
tromp, line 161: why not use the name assume_valid_rangeproofs_kernels here as well?

tromp (approved):
I recommend more use of the name assume_valid_rangeproofs_kernels
as indicated with in-line comments.

src/bin/cmd/client.rs
deevope: Done but i let `--fast` as the option for the arg verify-chain.

yeastplume:
Reviewed requested changes, all good with me. Thanks for this PR at @deevope and for the detailed review @tromp !

yeastplume (approved):

Issue #3681, opened 2021-12-29
Grin sync from scratch broken on master


phyro:
**Describe the bug**
If you're syncing from scratch, the process panics soon after launch.

**To Reproduce**
```
git checkout origin/master
cargo clean
cargo build
cs target/debug
./grin clean
./grin
```

**Relevant Information**
Some logs: https://gist.github.com/phyro/9cbb0f2ad2311f333309fec936a5fcd3

It seems like the issue was introduced in this PR https://github.com/mimblewimble/grin/pull/3671/files

**Desktop (please complete the following information):**
 - OS: Linux

PR #3683, opened 2021-12-30
Handle attempts to get block headers at invalid heights

yeastplume:
Twofold 'belt and suspender' fix for #3681:

* First, add an error check into `get_header_hash_by_height` in `txhashset` to return an error if the header height requested is too large for `pmmr::insertion_to_pmmr_index` to handle. Note there's no error handling in the core pmmr functions (which all assume arguments are correct, rightly or wrongly), so this check is placed as close to it as possible.
* Second, manually handle negative heights from the difficulty iterator in `get_server_stats` code. Note the call to `get_header_by_height` is now needed here due to earlier performance optimizations on DifficultyIterator that removed the deserialization of proof nonces and left iterator entries unable to calculate their hash.

Either fix on its own would work, but `get_header_by_height` should be returning an error for invalid heights, while the call to it from `get_server_stats` shouldn't be attempting to provide invalid heights via an incorrect cast. (Note that casting from i64 to u64 is allowed in rust regardless of value, meaning that -59i64 to u64 results in 18446744073709551557).

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -50,6 +50,8 @@ const KERNEL_SUBDIR: &str = "kernel";
 
 const TXHASHSET_ZIP: &str = "txhashset_snapshot";
 
+const HALF_U64_MAX: u64 = u64::MAX / 2;
+
 /// Convenience wrapper around a single prunable MMR backend.
 pub struct PMMRHandle<T: PMMRable> {
 	/// The backend storage for the MMR.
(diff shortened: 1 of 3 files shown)


chain/src/txhashset/txhashset.rs
```diff
@@ -110,6 +112,9 @@ impl PMMRHandle<BlockHeader> {
 
 	/// Get the header hash at the specified height based on the current header MMR state.
 	pub fn get_header_hash_by_height(&self, height: u64) -> Result<Hash, Error> {
+		if height >= HALF_U64_MAX {
```
tromp, line 115: Can't we compare height with the header MMR size?

tromp (commented):
see single in-line comment

phyro:
I have run the initial sync that broke before and it no longer does. Apart from the minor comment from tromp, the changes look good :+1: I was not able to sync the node, but I now suspect the problem is on my end with the disk.

Hecim1984 (approved):
🤑

PR #3684, opened 2022-01-04
TUI Freeze Fix - Add manually maintained hash to difficulty iterator

yeastplume:
Due to recent performance optimisations to the difficulty iterator, a call to the chain's `get_header_by_height` needed to be added. Unfortunately, this function relies on a read lock on the txhashset. A write lock is held unyielded for a very long time by validation code after the txhashset is downloaded, leading to a TUI freeze when calling this method.

Fix is to avoid that call by re-adding a hash to the `HeaderDifficultyInfo` struct, populated by maintaining hashes as provided in the `prev_header` field in the Headers themselves. This optionally keeps hashes available in the `HeaderDifficultyInfo` struct while still avoiding the need to deserialize the entire header to derive the hash.

Note this point in the TUI code is the only place where a block's hash is needed from the DifficultyIterator, so impact should be minimal.

### DIFF
--- a/chain/src/store.rs
+++ b/chain/src/store.rs
@@ -441,6 +441,7 @@ pub struct DifficultyIter<'a> {
 	// toward the genesis block (while maintaining current state)
 	header: Option<BlockHeader>,
 	prev_header: Option<BlockHeader>,
+	prev_header_hash: Option<Hash>,
 }
 
 impl<'a> DifficultyIter<'a> {
@@ -453,6 +454,7 @@ impl<'a> DifficultyIter<'a> {
 			batch: None,
 			header: None,
 			prev_header: None,
+			prev_header_hash: None,
 		}
 	}
 
@@ -465,6 +467,7 @@ impl<'a> DifficultyIter<'a> {
 			batch: Some(batch),
 			header: None,
 			prev_header: None,
+			prev_header_hash: None,
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


chain/src/store.rs
```diff
@@ -479,18 +482,26 @@ impl<'a> Iterator for DifficultyIter<'a> {
 		// Items returned by this iterator cannot be expected to correctly
 		// calculate their own hash - This iterator is purely for iterating through
 		// difficulty information
-		self.header = if self.header.is_none() {
+		let (cur_header, cur_header_hash) = if self.header.is_none() {
 			if let Some(ref batch) = self.batch {
-				batch.get_block_header_skip_proof(&self.start).ok()
+				(
+					batch.get_block_header_skip_proof(&self.start).ok(),
+					Some(self.start.hash()),
```
phyro, line 489: I may be missing something obvious due to not being familiar with Rust. Looking at types `start` seems to be of type `Hash`. Is calling `hash()` on it different than just `Some(self.start)`?
(same question for the `else if` branch below)

servers/src/grin/server.rs
```diff
@@ -451,18 +451,10 @@ impl Server {
...
 					height += 1;
 
-					// We need to query again for the actual block hash, as
-					// the difficulty iterator doesn't contain enough info to
-					// create a hash
-					// The diff iterator returns 59 block headers 'before' 0 to give callers
-					// enough detail to calculate initial block difficulties. Ignore these.
-					let block_hash = if height < 0 {
-						ZERO_HASH
+					let block_hash = if let Some(nh) = next.hash {
```
phyro, line 454: syntactic sugar nitpicking: I think we could get rid of this branching by doing something like `block_hash: next.hash.unwrap_or(ZERO_HASH)` (might be a bit different in rust)

phyro (approved):
I can confirm this resolved my TUI freezing issue. Unfortunatelly, I'm still having issues syncing which seems unrelated to this change. Peers drop and everything starts working slowly for me. It might be due to some disk issues I'm experiencing that are unrelated to Grin, but will be able to tell only after I try on a different disk.

chain/src/store.rs
yeastplume: Had just noticed that the latest block in the TUI 'difficulty' tab was showing up incorrectly, this would be why (hash-of-hash instead of hash). good catch!

servers/src/grin/server.rs
yeastplume: :+1:

PR #3687, opened 2022-01-12
Update to cursive 0.17

gyscos:


yeastplume:
If someone has the time, it might be worth just checking out this PR and confirming no obvious issues in the TUI with these changes? (Thanks @gyscos for providing the update)

PR #3688, opened 2022-01-18
[PIBD_IMPL] PIBD Desegmenter State

yeastplume:
* Add state functions to Desegmenter to allow it to request the next segments it needs
* Initial attempt at adding state to the network adapter side to keep track of what segments have been requested from others.
* Addition of higher-level SegmentTypeIdentifier to allow sync code to uniquely identify a segment for a particular archive header by SegmentIdentifier+Type
* Move some segment calculation logic out of `Segment` and into `SegmentIdentifier` to allow functions to be used on identifiers.
* Change chain's `desegmenter()` function to return an RwLock
* Add field to sync status to track segment types that have been requested from a peer (TODO: when to remove this as failed)
* Re-add buffer to store received bitmap segments


yeastplume:
Going to stop this PR here and move on to the next one. With these changes, the sync process identifies the correct bitmap segments it needs, requests them from a peer with the correct capabilities, then adds them to the desegmenter buffer. Next steps are to piece together the bitmap again, and move on to requesting and applying output segments from peers.

yeastplume:
Note pibd copy test will fail with this PR due to desegmenter cache changes, will fix in next PR.

PR #3689, opened 2022-01-21
[PIBD_IMPL] Bitmap accumulator reconstruction + TxHashset set reconstruction

yeastplume:
Still WIP, Goal of this PR is to have all mmrs in the txhashset reconstructed via p2p messaging, (with no kernel or rangeproof validation just yet)

* Reconstruct and `finalize` the desegmenter's BitmapAccumulator once all segments have been received
* Add ability to find and request the next output mmr segments in `Desegmenter::next_desired_segments`
* `apply_next_segments` function, which applies the next set of downloaded segments to the underlying pmmrs as and when they become available. (Note this is currently being called once per sync loop iteration and 'pops' off the next segments on each call, this is not the final timing strategy)
* Add remaining peer and p2p functions

### DIFF
--- a/chain/src/txhashset/desegmenter.rs
+++ b/chain/src/txhashset/desegmenter.rs
@@ -386,10 +386,20 @@ impl Desegmenter {
 			let txhashset = self.txhashset.read();
 			local_output_mmr_size = txhashset.output_mmr_size();
 		}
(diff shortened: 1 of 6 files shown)


yeastplume:
Applying to move on to modifying chain copy test cases to ensure segment application is accurate

PR #3705, opened 2022-04-13
[PIBD_IMPL] BitmapAccumulator Serialization Fix

yeastplume:
Fixes a crucial issue in which peers serializing BitmapAccumulator blocks could potentially either select the incorrect serialization mode for the block or write an incorrect length for the number of negative entries written. (This, in turn could potentially cause a panic on deserialization causing PIBD to fail).

When sending an block that isn't full (i.e. contains less than 2^16 bits), the previous version of the code was failing to remove the unused bits from calculations concerning the number of entries, causing the read length to be incorrectly written (or the wrong serialization mode to be chosen). This fix adjusts the calculation with the difference between the max block size and the actual.

Also adds another capability, `PIBD_HIST_1`, which peers will now need to respond with before another peer can request a PIBD segment from them, to ensure they have this fix.

Also adds unit tests to exercise the negative failure case for non-full blocks, (as well as the positive instance).

### DIFF
--- a/chain/src/txhashset/bitmap_accumulator.rs
+++ b/chain/src/txhashset/bitmap_accumulator.rs
@@ -421,6 +421,8 @@ impl Writeable for BitmapBlock {
 		let count_pos = self.inner.iter().filter(|&v| v).count() as u32;
 		let count_neg = Self::NBITS - count_pos;
 
+		// Negative count needs to be adjusted if the block is not full,
+		// which affects the choice of serialization mode and size written
 		let count_neg_adjust = Self::NBITS - length as u32;
 		let count_neg = count_neg - count_neg_adjust;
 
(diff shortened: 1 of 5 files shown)


chain/src/txhashset/bitmap_accumulator.rs
```diff
@@ -419,6 +419,12 @@ impl Writeable for BitmapBlock {
 
 		let count_pos = self.inner.iter().filter(|&v| v).count() as u32;
 		let count_neg = Self::NBITS - count_pos;
+
+		// Negative count needs to be adjusted if the block is not full,
+		// which affects the choice of serialization mode and size written
+		let count_neg_adjust = Self::NBITS - length as u32;
+		let count_neg = count_neg - count_neg_adjust;
```
tromp, line 426: This is a needlessly complicated way to
  let count_neg = length - count_pos;

PR #3707, opened 2022-05-12
Visibility DNSSeed scope tweaks to aid seed test utilities

yeastplume:
Just a few minor scope tweaks to DNSSeed constants and functions to assist with the seed check utility service I'm putting together at:

https://github.com/yeastplume/grin-history-tools/blob/main/crates/seedcheck/src/service/seed.rs

### DIFF
--- a/p2p/src/lib.rs
+++ b/p2p/src/lib.rs
@@ -44,7 +44,7 @@ mod peer;
 mod peers;
 mod protocol;
 mod serv;
-mod store;
+pub mod store;
 pub mod types;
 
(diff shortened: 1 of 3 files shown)


phyro (approved):
I don't know if store being private was a design decision or just nothing was needed until now. I'll assume the latter.
Looks good to me :+1:

yeastplume:
> I don't know if store being private was a design decision or just nothing was needed until now. I'll assume the latter. Looks good to me 👍

In our case it's mostly 'keep it private until a need arises for it not to be.' Thanks both!

PR #3710, opened 2022-05-24
Update Seed nodes

JulsKawa:
hello,
i update seed node with https://forum.grin.mw/t/call-for-new-dns-seed-nodes-mainnet-and-testnet/9790 and https://grintools.revcore.net/seedcheck/


MCM-Mike:
We will keep `mainnet-seed.grinnode.live` online which will redirect to different GRIN nodes we are running on mainnet. 
Contact e-mails did not change.

PR #3633, opened 2021-04-10
[WIP] Replace deprecated `failure` crate with `thiserror`

trevyn:
Will write detailed review notes soon.

To do:
- [ ] Consider how best to handle conversion from `.context()`

PR #3728, opened 2022-07-04
Thiserror changeover

yeastplume:
This is an update of the change to `thiserror` from the deprecated `failure` crate originally proposed by by @trevyn in #3633 (with many thanks for doing this).

As noted in that PR, we should be updating from the deprecated crate, and I'd like to make sure the error handling is updated and consistent between the node, wallet and gui code. Will be updating the wallet project similarly.


phyro:
There are some occurrences in the lock file e.g. https://github.com/mimblewimble/grin/blob/master/core/fuzz/Cargo.lock#L312-L330
Should we do something about these?

yeastplume:
> There are some occurrences in the lock files e.g. https://github.com/mimblewimble/grin/blob/master/core/fuzz/Cargo.lock#L312-L330 Should we do something about these?

I've updated the fuzz tests (which I have never looked at) to latest versions of everything, this should remove those references, as well as clean up most of the dependencybot warnings we have against the code.

phyro:
Successfully synced a node and tried a basic error. I noticed some RPC calls got a different return type e.g. `get_header` went from `-> Result<BlockHeaderPrintable, ErrorKind>;` to `-> Result<BlockHeaderPrintable, Error>;`. I tried a simple request for a header at height that didn't exist at the time on both master and this branch and they returned the same response
```
> curl -XPOST -s -w "%{http_code}" -ugrin:(cat ~/.grin/main/.foreign_api_secret) -d '{"jsonrpc": "2.0","method": "get_header","params": [1817338,null,null],"id": 1}' http://127.0.0.1:3413/v2/foreign
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "Err": "NotFound"
  }
}200⏎ 
```

Do we care if all the error responses are exactly as they were before?

yeastplume:
> Successfully synced a node and tried a basic error. I noticed some RPC calls got a different return type e.g. `get_header` went from `-> Result<BlockHeaderPrintable, ErrorKind>;` to `-> Result<BlockHeaderPrintable, Error>;`. I tried a simple request for a header at height that didn't exist at the time on both master and this branch and they returned the same response
> 
> ```
> > curl -XPOST -s -w "%{http_code}" -ugrin:(cat ~/.grin/main/.foreign_api_secret) -d '{"jsonrpc": "2.0","method": "get_header","params": [1817338,null,null],"id": 1}' http://127.0.0.1:3413/v2/foreign
> {
>   "id": 1,
>   "jsonrpc": "2.0",
>   "result": {
>     "Err": "NotFound"
>   }
> }200⏎ 
> ```
> 
> Do we care if all the error responses are exactly as they were before?

They can't be the same as before, because `thiserror` gets rid of `ErrorKind`. As you demonstrate there, the printed output is okay. Anyone linking directly may need to modify some error handling code, but it's straightforward (will include this in the release notes)

core/fuzz/fuzz_targets/block_read_v2.rs
```diff
@@ -1,12 +1,13 @@
...
 
 use grin_core::core::UntrustedBlock;
-use grin_core::ser;
+use grin_core::ser::{self, DeserializationMode};
 
 fuzz_target!(|data: &[u8]| {
 	let mut d = data.clone();
-	let _t: Result<UntrustedBlock, ser::Error> = ser::deserialize(&mut d, ser::ProtocolVersion(2));
+	let _t: Result<UntrustedBlock, ser::Error> =
+		ser::deserialize(&mut d, ser::ProtocolVersion(2), DeserializationMode::Full);
```
phyro, line 12: ah, we missed these in one of the previous refactorings didn't we? Good this was found and fixed :+1:
Do you know if we run any fuzz testing when making a release or on CI?

chain/src/pipe.rs
```diff
@@ -456,9 +458,7 @@ fn validate_header(header: &BlockHeader, ctx: &mut BlockContext<'_>) -> Result<(
 
 fn validate_block(block: &Block, ctx: &mut BlockContext<'_>) -> Result<(), Error> {
 	let prev = ctx.batch.get_previous_header(&block.header)?;
-	block
-		.validate(&prev.total_kernel_offset)
-		.map_err(ErrorKind::InvalidBlockProof)?;
+	block.validate(&prev.total_kernel_offset)?;
```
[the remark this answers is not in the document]
yeastplume: no, it's just implicitly done here now https://github.com/yeastplume/grin/blob/ba7eb9739987fb2687c8b72c3049de0eee3ab183/chain/src/error.rs#L54

core/fuzz/fuzz_targets/block_read_v2.rs
yeastplume: No, fuzz testing is very manual and I don't think anyone's looked at it in years. I updated all of it because they're small and only took a few minutes to update, so someone can have a look later if they like. If it had been a non-trivial task I would have just removed or disabled them.

yeastplume:
Think I'm happy with the state of review now, so merging. We'll have a while to shake issues out on the master branch before release anyhow.

PR #3734, opened 2022-08-31
The foreign_rpc and owner_rpc auto generated helper modules should be made public.

flomang:
I've got a project that I want to build using the easy_jsonrpc_mw generated modules. In order to access those modules I need public access to the foreign_rpc and owner_rpc modules. This change shouldn't affect or require any tests nor pose any security issues.

Here is an example of client side usage that requires public access.
https://github.com/flomang/grin-foreign-api-client

I also noticed the gin-wallet api is also in need of this change. I can submit a pull request for that as well if this gets approved.


yeastplume:
Yes, that makes sense, I'm coming across a lot of things that 'should have been public' while doing GUI work. Will merge, and thanks for the PR.

yeastplume (approved):

PR #3737, opened 2022-10-03
Override global one times.

flomang:
The grin-gui needs to be able to start the embedded node either in Testnet or Mainnet. If the user selects a Testnet wallet while the embedded node is running on Mainnet the grin-gui shall shutdown the Mainnet node and restart a Testnet node. Unfortunately, since the init_global_... calls need to be called again with different parameters (i.e. change chain_type, fee, etc) the OneTime panics. 

These functions shall permit the globals to be set again so the grin-gui can dynamically change its global context for Mainnet or Testnet.


core/src/core/pmmr/readonly_pmmr.rs
```diff
@@ -73,7 +73,9 @@ where
 			Some(p) => p,
 			None => self.size,
 		};
-		let mut pmmr_index = pmmr_index1 - 1;
+		// default index to 0 if arg index1 is 0
+		let mut pmmr_index = if pmmr_index1 > 1 { pmmr_index1 - 1 } else { 0 };
```
yeastplume, line 77: The endless saga of 1 vs 0 indexed pmmrs.

This is fine as a belt+suspenders check, though I think the idea was that this would always be called assuming a 1 index, as exposed by all of the APIs. A lot of work was done to use 0 indices internally, but this confusion still abounds.

In any case, all fine, but you can just use `pmmr_index = pmmr_index1.saturating_sub(1);' instead of if statements

flomang:
Yes, these functions are only really needed for the gui atm. I was thinking maybe a feature?

yeastplume:
> Yes, these functions are only really needed for the gui atm. I was thinking maybe a feature?

Think we're fine as is, no need to complicate with features since there are no downsides to including this.

yeastplume (approved):

PR #3740, opened 2022-12-14
refactor: simplify some exact instances of Option/Result combinators

huitseeker:
This replaces some pattern-matches that are exact reimplementation of stdlib combinators.

This, as well as running clippy on the last commit involved changes that assume a more recent version of Rust, e.g.:
- captured identifiers (Rust 1.58)
- boolean combinators (`then`, ` then_some`, since Rust 1.62)

Opinions on this loosely held, feel free to suggest changes.
First commit tool-aided by [comby-rust](https://github.com/huitseeker/comby-rust)


huitseeker:
Cleaning my PR queue, feel free to re-open if interested.

PR #3743, opened 2023-04-13
Add nix/flake support

chives101:
# Motivation

This PR enables power users to easily and reliably build the project from source code.

# Expectation

A user with the nix package manager installed can build the project with the following command:

`nix build github:mimblewimble/grin`

and expect the resulting binaries in `./result/bin`.

# Testing

A test run for this PR can be achieved using my forked repo with the following command:

`nix build github:chives101/grin/flake` and expect a binary at `./result/bin/grin`.

# Maintenance

Each time `cargo.lock` is updated, the hash of the combined dependencies inside that file needs to be
updated in `etc/default.nix`. It can be done by simply giving a dummy hash, and observe the error
generated from the build log.

# Dev notes

This PR is the result of a simple adaptation of the code from this blog post:

https://www.tweag.io/blog/2022-09-22-rust-nix/


quentinlesceller:
@phyro iirc this might be something related with libncurses.

chives101:
@phyro Is it tested with the latest repo? I did a force push that updated the nixpkgs dependency which was pretty outdated.

I just tested the repo, works fine for me. BTW, you can do a local test by entering the project folder and do `nix build` after applying the PR.

chives101:
@phyro I did a fresh build on debian bullseye, I couldn't reproduce your issue though.

pkariz:
on nixos 22.11 it worked for me although it's weird that @phyro is getting these characters. I recall having the same problem but i've fixed it by installing newer ncurses or some other package. Will try to reproduce the problem.

chives101:
@pkariz It is weird, because flake is supposed to be hermetic. Guess it's not that airtight after all.

chives101:
@phyro It's fine, it's nothing. I'm pretty sure there's more I can do :D

Issue #3746, opened 2023-05-01
Config parsing error causes panic


ardocrat:
After adding `migrate_config_file_version_none_to_2` function by this PR https://github.com/mimblewimble/grin/pull/3634, `read_config` method is not handling parsing error anymore, panic occurs at `toml::from_str` function when config file is in wrong format.

PR #3747, opened 2023-05-01
Handle parsing error at config

ardocrat:
Prevent panic from `toml::from_str` at `migrate_config_file_version_none_to_2` function when config is in wrong format.
Fixes https://github.com/mimblewimble/grin/issues/3746


ardocrat:
Hardcoded fees setup will be refactored at future, so this PR is pointless.

PR #3750, opened 2023-06-01
'Turn on' PIBD on main-net, in advance of 5.2.0 release

yeastplume:
Removes the check that restricts PIBD requests to testnet only.

### DIFF
--- a/servers/src/grin/sync/state_sync.rs
+++ b/servers/src/grin/sync/state_sync.rs
@@ -100,13 +100,7 @@ impl StateSync {
 			} else if self.pibd_aborted {
 				false
 			} else {
-				// Only on testing chains for now
-				if global::get_chain_type() != global::ChainTypes::Mainnet {
-					true
-				//false
-				} else {
-					false
-				}
+				true
 			};
 
 		// Check whether we've errored and should restart pibd


servers/src/grin/sync/state_sync.rs
```diff
@@ -100,13 +100,7 @@ impl StateSync {
...
 				false
 			} else {
-				// Only on testing chains for now
-				if global::get_chain_type() != global::ChainTypes::Mainnet {
-					true
-				//false
-				} else {
-					false
-				}
+				true
```
tromp, line 103: Lines 102 - 104 can be rewritten as

                        } else {
                                !self.pibd_aborted
                        };

Even better would be having some accessor for the aborted field in SyncStatus::TxHashsetPibd (is there?) so that we could write

let using_pibd = ! self.sync_state.status().aborted() && ! self.pibd_aborted;

Btw, why do we have these two different bits of data to record what seems to be the same fact??

tromp (changes requested):
See my question about different bits of data recording aborted status...

servers/src/grin/sync/state_sync.rs
yeastplume: Good catch, no idea why there's a redundant flag in there, vestigial logic left over from previous mutations I guess.

I've removed that, think it reads fine now. You need to pattern match somehow to access the values in `SyncStatus::TxHashsetPibd`, so creating an accessor for that one value would be likely more trouble than it's worth.

tromp: It is better rewritten as

                let using_pibd = matches!(self.sync_state.status(), SyncStatus::TxHashsetPibd { aborted: true, .. });

yeastplume: `aborted: false` but otherwise good

tromp: No, changing aborted value is not the same as negating matches; we need


let using_pibd = !matches!(self.sync_state.status(), SyncStatus::TxHashsetPibd { aborted: true, .. });

yeastplume: Coming to that realisation. Okay, last time is the charm.

tromp (approved):

PR #3756, opened 2023-06-16
Optimize desegmenter check progress

ardocrat:
Save `latest_block_height` at `Desegmenter::check_progress` to request less data from db at `PMMRHandle<BlockHeader>::get_first_header_with` (`from_height` parameter) with each iteration.


yeastplume:
All looks good, and tested, thanks!

yeastplume (approved):

Issue #3754, opened 2023-06-15
TxHashsetDownload is starting in parallel with TxHashsetPibd


ardocrat:
**Describe the bug**
After some time at `TxHashsetPibd` step synchronization is switching to non-PIBD `TxHashsetDownload` step in parallel, then coming back to `TxHashsetPibd`, sometimes it can lead to hang of synchronization.

**Screenshots**
![Screenshot_20230615_144833](https://github.com/mimblewimble/grin/assets/127264394/66bfad5d-2c3c-4b50-a157-c0e23c435985)

At logs I am seeing:
![Screenshot_20230615_144709](https://github.com/mimblewimble/grin/assets/127264394/b1b85c4f-6792-4fc4-9426-ba7ea8adf563)

**Desktop (please complete the following information):**
 - OS: Linux
 - Version 5.2.0-beta

Anynomouss:
My guess would be that this happens when a new horizon header is defined. The horizon header is 48 hours back in time, but is updated every 12 hours. So it would be important to look at the time, probably it coincides with this 12 hourly horizon update.

Just a wild guess, but it happens at block header height 2414175 according to the log.

2414175/(1260) = 3353.02083333, so 0.0212*60 = 14.4. Your screenshot is from 14 minutes after a horizon update. Can be coincidence, most likely it is not. 
Probably the node gets into some conflict state since it starts to sync using the txHAshsetDownload since it updated all blocks except the last 48 hours using PIDB. Then, before the node has finished downloading the next 12 hours using the TxHashsetDownlaod method, the horizon is updated, resulting in a switch back to PIDB to download the remaining blocks to the new horizon header.

yeastplume:
I'm looking into this as well, would be good to narrow down how to reproduce it, let me know if you determine anything further.

cekickafa:
I did several interruptions, it worked and resumed until 2/7 state.
![pibd2](https://github.com/mimblewimble/grin/assets/74456317/8813cdca-17c5-4e47-a076-2af5a5e9c8f3)

 **Desktop**
    - OS: **Windows 10**
    - Version 5.2.0-beta

cekickafa:
Desktop
OS: **Windows 10**
Version 5.2.0-beta

![pibd2](https://github.com/mimblewimble/grin/assets/74456317/1586cc25-c8bf-422d-be29-b287c4f17801)

PR #3757, opened 2023-06-16
PIBD - Revert earlier pibd_aborted flag in sync logic

yeastplume:
Reverts StateSync's `pibd_aborted` flag, to test whether it fixes PIBD appearing to resume after having fallen back to txhashset.zip download.

Also includes changes #3756 (will merge that separately if all tests go well here)

### DIFF
--- a/servers/src/grin/sync/state_sync.rs
+++ b/servers/src/grin/sync/state_sync.rs
@@ -37,6 +37,7 @@ pub struct StateSync {
 	prev_state_sync: Option<DateTime<Utc>>,
 	state_sync_peer: Option<Arc<Peer>>,
 
+	pibd_aborted: bool,
 	earliest_zero_pibd_peer_time: Option<DateTime<Utc>>,
 }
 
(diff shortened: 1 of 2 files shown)


yeastplume:
Have tested now, this fixes #3754. The SyncStatus struct essentially keeps track of a large state machine which is executed every tick (essentially getting reset) in order to determine what the state of sync should be. A separate `pibd_aborted` flag, which persists between SyncStatus invocations, is required in state sync struct to record whether we should stop attempting PIBD altogether and revert to the earlier txhashset.zip method.

servers/src/grin/sync/state_sync.rs
```diff
@@ -62,6 +64,12 @@ impl StateSync {
 		self.earliest_zero_pibd_peer_time = t;
 	}
 
+	/// Flag to abort PIBD process within StateSync, intentionally separate from `sync_state`,
+	/// which can be reset between calls
+	pub fn set_pibd_aborted(&mut self) {
```
tromp, line 69: abort_pid() seems like a more straightforward name

tromp (approved):

yeastplume:
Thanks for reviewing!

PR #3759, opened 2023-06-20
chrono deprecations warning cleanup 

yeastplume:
Cleans up warnings related to specific errors, mostly `from_timestamp` moving to a `from_timestamp_opt` function that returns an options.

Note `from_timestamp_opt` returns `None` in the following circumstances:

```
Returns None on the out-of-range number of seconds (more than 262 000 years away from common era) and/or invalid nanosecond (2 seconds or more).
```

* Where timestamps are hardcoded on initialization (such as being hardcoded with 0s), I've just called unwrap as there's no reason for the function to return `None`
* Where timestamps are initialized dynamically in critical code check for `None` and return appropriate error
* where timestamps are less crucial, (tui, mining) just `unwrap_or_default`

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -231,7 +231,10 @@ impl Default for BlockHeader {
 		BlockHeader {
 			version: HeaderVersion(1),
 			height: 0,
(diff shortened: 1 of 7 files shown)


yeastplume:
6 Month re-review: think all of these are fine. PIBD is released and master set to an alpha so good time to integrate these.

PR #3770, opened 2023-09-22
Update croaring to 0.5.2

ardocrat:
Previous `croaring` crate version `0.4.6` uses `bindgen` crate version `0.56.0` that leads to build errors, such as: https://github.com/rust-lang/rust-bindgen/issues/2500 and https://github.com/rust-lang/rust-bindgen/issues/2312 on recent C compilers (e.g. GCC `13.2.1` under Arch Linux), more recent `croaring` crate version `0.5.2` uses `bindgen` crate version `0.59.0` and has no such issue (tested under Arch Linux with GCC `13.2.1` and glibc `2.38`).


yeastplume:
Yep, getting all sorts of errors on grin wallet tests. Looks good, thanks for looking into the cause.

PR #3773, opened 2023-10-03
`get_blocks` API call - set tail height to include genesis

yeastplume:
When calling the `get_blocks` function with a start height of 0, the start height is adjusted to the 'tail', or the earliest stored block. The tail function appears to be returning 1 in the case of an archive node, causing the call to skip block 0. This adjusts the start height in that single case to make the API more consistent.


pkariz:
`get_blocks` is returning as expected now, thanks!

PR #3774, opened 2023-10-18
Handle invalid MMR root to prevent sync thread panic

ardocrat:
Return `Result<TxHashSetRoots, Error>` at `TxHashSet::roots()` method to prevent thread panic on MMR root check.


yeastplume:
merging to apply #3775 on top

PR #3775, opened 2023-10-26
[PIBD] Fix to restart of PIBD download on corrupted data or mismatched roots

yeastplume:
In tandem with, #3774 this fixes a case where PIBD fails to restart properly if the downloads txhashset state does not match with expected roots.

The previous logic attempted to manually roll back the output PMMR states to genesis, which could not work given rollback relies on blocks data existing in the DB (which is not guaranteed beyond the horizon block). This adds a `reset_chain_head_to_genesis` function to `chain.rs`, which re-initializes the body PMMRs to the genesis block while leaving the header MMR intact, allowing PIBD to start again from the genesis block.

Once we've tested this, I'll merge #3774  first and then merge into this one to maintain @ardocrat's credit where credit is due.

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -271,10 +271,9 @@ impl Chain {
 	}
 
 	/// wipes the chain head down to genesis, without attempting to rewind
-	/// Used upon PIBD failure, where you want to keep the header chain but
-	/// restart the MMRs from scratch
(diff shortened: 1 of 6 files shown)


ardocrat:
Thanks, its working, in the end of `SyncStatus::TxHashsetPibd` it restarts PIBD synchronization from 0 if root is missing.

PR #3778, opened 2024-01-29
Dependency upgrade: Cursive + Chrono warnings

yeastplume:
* Upgrades cursive to 0.20 as previous version was so out of date as to be causing conflicts updating Cargo.lock
* Further chrono warning updates with `DateTime::<Utc>::from_utc` becoming `DateTime::from_naive_utc_and_offset`
* `timestamp::nanos()` => `.timestamp_nanos_opt().unwrap_or_default()` (only in TUI so not critical)

Supercedes #3687


yeastplume:
Merging, need to investigate windows illegal instruction failure separately

PR #3779, opened 2024-01-30
croaring-rs upgrade 0.5.2 -> 1.0.1

yeastplume:
Attempt to upgrade croaring-rs to 1.0.1, from 0.5.2, in an attempt to address odd instruction set issues such as found in https://github.com/mimblewimble/grin-gui/issues/73

Note this has not yet been tested, and this is here as a first pass, mostly to determine whether it does indeed address issues. A windows test was failing on the CI server at `pow::lean::test::lean_miner`. If this no longer fails after the update, this is a good sign.

```
pub fn trim(&mut self) {
        // trimming successively
        while self.edges.cardinality() > (7 * (self.params.num_edges >> 8) / 8) as u64 {
            self.count_and_kill();
        }
    }
 ```

Issues that need to be looked into:

* Several `Bitmap` functions that take a range are required to be u32 as opposed to u64 in croaring 0.5.6. I've just cast types for now, we need to confirm this is okay.
* croaring 1.0.1 now has several serialization formats, and I can't find a clear statement in the documentation saying which one corresponds to the serialization format in 0.5.2. I'm assuming this is `Portable` (and seems to work with existing chain data, though I could still be wrong).
* Tests in `store` appear to be failing following this update, and we need to trace why:

```
---- pmmr_reload stdout ----
env logger already initialized: SetLoggerError(())
idx: 1, shift_cache.len(): 0
thread 'pmmr_reload' panicked at store\src\prune_list.rs:156:26:
attempt to subtract with overflow


failures:
    compact_twice
    pmmr_compact_entire_peak
    pmmr_compact_horizon
    pmmr_compact_single_leaves
    pmmr_reload
    pmmr_rewind
```


yeastplume:
* Subtle change `bitmap.remove_range_closed` became `remove_range` and accepts an inclusive range as an argument, took a bit to track down. Caught by store pmmr tests.

yeastplume:
Also note this reverts the CI testing environment to the `windows-2019` image as per compiler bug found in https://github.com/RoaringBitmap/croaring-rs/pull/128

tromp (approved):
Looks ok, but this code needs to be made future proof in the next few years so it can support MMR indices beyond 2^32.

PR #3783, opened 2024-03-10
doc: clarify why blinding factor range proof is needed

starius:
This place in the intro was very hard to me to understand. I think now it is easier to understand the logic of why range proofs of blinding factors are needed. Hopefully it helps future readers!

### DIFF
--- a/doc/intro.md
+++ b/doc/intro.md
@@ -252,7 +252,9 @@ which can be signed by the attacker because Carol's blinding factor cancels out 
 
 This output (`(113 + 99)*G + 2*H`) requires that both the numbers 113 and 99 are known in order to be spent; the attacker
 would thus have successfully locked Carol's UTXO. The requirement for a range proof for the blinding factor prevents this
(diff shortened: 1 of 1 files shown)


IDRozenblad:
True

Anynomouss (changes requested):
I agree that the description could be made a bit clearer, but the proposed changes are also confusing. I would suggest to change the text as follows:

> This output (`(113 + 99)*G + 2*H`) requires that both the numbers 113 and 99 are known in order to be spent; the attacker would thus have successfully locked Carol's UTXO. The requirement for a range proof for the blinding factor prevents this because the attacker doesn't know the number 113 and thus neither (113 + 99). In other words, without knowing the private key (blinding factor), the attacker would not know the value in the output and would not be able to produce a valid range proof for it.   A more detailed description of range proofs is further detailed in the [range proof paper](https://eprint.iacr.org/2017/1066.pdf).

wiesche89:
PR #3840 includes @Anynomouss review.

Issue #3791, opened 2024-06-05
CPU consumption and DoS on calling get_outputs rpc


aglkm:
1. Call `get_block` rpc to ensure it can produce the expected output in time.
`curl -u grin:secret --data '{"jsonrpc":"2.0","method":"get_block","params":[2824162, null, null],"id":1}' 127.0.0.1:3413/v2/foreign`
2. Call `get_outputs` rpc.
`curl -u grin:secret --data '{"jsonrpc":"2.0","method":"get_outputs","params":[["08b7e57c448db5ef25aa119dde2312c64d7ff1b890c416c6dda5ec73cbfed2edea"], null, null, true, true],"id":1}' 127.0.0.1:3413/v2/foreign
`
3. Check CPU consumption.
4. Call `get_block` rpc again and notice long time response until `get_outputs` finishes its job.
`curl -u grin:secret --data '{"jsonrpc":"2.0","method":"get_block","params":[2824162, null, null],"id":1}' 127.0.0.1:3413/v2/foreign`

yeastplume:
Thanks for reporting, just a quick video below of me running these commands against a docker image using monitoring utility I've been putting together:

https://github.com/mimblewimble/grin/assets/7074070/2945ae56-a86a-45af-b738-cced17bd19b7

As you can see, the call to `get_outputs` is taking a couple of seconds (which definitely could be looked into to see how to speed it up), but it doesn't seem anywhere close to DoS territory from a single request.

Is the node in the middle of syncing when you're performing this call? Also, if you could give me as much detail about the machine the node is running on, including as much detail about processor specs, ram, disk etc.

aglkm:
If you want to re-create a longer period of time waiting from the `get_outputs`, then send the rpc to grinnode public node API. I was able to re-create the same long waiting time against my archival node. My guess, this depends on whether a node is archival or not, causing deep searching over the outputs comparing to non archival node.

Also, even with shorter waiting time periods from `get_outputs`, it still causing the issue by locking(?) the db, making node struggling to sync or handle other rpc requests, e.g. `get_block`. So, if someone constantly sending you `get_outputs` one after another (not even parallel requests are needed), your node is not able to be in sync with the network or handle certain other requests, obviously this is way worse for the archival node.

Some other observations regarding the issue:

this is blocking the db:

[...]

PR #3792, opened 2024-06-17
Remove Merkle Proof Generation from Foreign API `get_outputs`

yeastplume:
As evidenced by #3791, Merkle proof generation is an experimental feature that currently isn't being used but also has the potential to grind nodes to a halt, as a request for a proof for an early output requires the chain to rewind to that position. This is an obvious problem for archival nodes, but also an issue on non-archival nodes due to the ability to request the genesis coinbase output and force a rewind of the entire chain.

This functionality had been turned off for the `get_block` function, but for some reason remained within the `get_outputs` function.

https://github.com/mimblewimble/grin/pull/3487

Note there's no known use for these Merkle proofs as present (PIBD uses a form of them, but has its own code for proof generation and never requires a rewind due to it only operating on tree states at the horizon). This PR modifies the foreign API call to ignore the `include_merkle_proof` field in the `get_outputs` foreign api, while leaving the parameter in place to avoid breaking existing installations. Instructions to turn it back on are embedded.

Once merged into master, will cherry-pick to 5.3.x branch and rebuild.


tromp (approved):
Looks fine to me.

IDRozenblad:
seems fine

PR #3796, opened 2024-08-21
Rust 1.80+ fixes & accumulated warning cleanup

yeastplume:
Fix for current build issues on rust 1.8+, as well as fixes for all warnings and future deprecation warnings. Touches core code (though not in any complex way) so needs another level of review.

* Update cursive (and related `time` deps) to 0.21
* (Core) Remove unread `edge_bits` field in `CuckooParams` struct 
* (grin) Remove unused string arg from `RPCError` 
* Rust 1.80 now giving #[cfg] warnings if config values don't exist: https://blog.rust-lang.org/2024/05/06/check-cfg.html
* `NaiveDateTime::from_timestamp_opt` replaced with `DateTime::from_timestamp ... .naive_utc()`
* `NaiveDateTime::timestamp()` replaced with `.and_utc().timestamp()`
* `timestamp_nanos()` -> `timestamp_nanos_opt()`
* `clean_output_dir()` in `chain/tests/common` made conditional on test env
* update `built` crate (build-only dependency) to 0.7
* modify grin `build.rs` according to `built` crate updates
* update `log4rs` crate to 1.3, modify `util::logger` accordingly
* add `anyhow` crate to `util` to satisfy log4rs update dependency
* allow dead code in `chain_test_helper` test (incorrect unused code warning reported otherwise)
* rename `.cargo/config` -> `.cargo/config.toml`


chain/tests/chain_test_helper.rs
```diff
@@ -11,6 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+#![allow(dead_code)]
```
tromp, line 14: What is that needed for?

yeastplume: there's some cargo oddness that's causing a 'not used' warning for the `clean_output_dir` function below during compilation, even though it's configured as test only. I spent some time trying to track it down and couldn't find a reason for it. I could possibly spend more time on it

tromp: Maybe ask around on IRC #rust or other rust forums if other people have run into that issue. I see that googling for
   rust 'not used' warning for test code
also brings up discussions  on stack overflow, reddit, and the Rust programming language forum...

tromp (approved):
Looks fine to me. Just would prefer not to see ignore dead code directive...

yeastplume:
> Looks fine to me. Just would prefer not to see ignore dead code directive...

Just moved the directive to be function-scope rather than module scope, which it should have been in the first place.

PR #3804, opened 2025-03-02
Upgrade Tokio to v1.x and Hyper to v0.14

yeastplume:
This PR updates several long-overdue dependencies:

* Tokio: upgraded to version 1.x
* Hyper: upgraded to version 0.14
* Hyper-rustls: upgrade to 0.23
* Tokio-util: upgrade version 0.7
* Rustls: Added as an explicit dependency at version 0.20

* HTTP Connector Fix: Modified the HTTP connector to allow both HTTP and HTTPS connections for testing purposes.

This will likely require extensive testing once reviewed and merged into master. I'll likely move onto upgrading hyper once these changes are sufficiently verified.


api/src/rest.rs
```diff
@@ -214,13 +218,26 @@ impl ApiServer {
...
 				let server = async move {
-					let mut listener = TcpListener::bind(&addr).await.expect("failed to bind");
-					let listener = listener
-						.incoming()
-						.and_then(move |s| acceptor.accept(s))
-						.filter(|r| r.is_ok());
+					let listener = TcpListener::bind(&addr).await.expect("failed to bind");
+
+					let tls_stream = async_stream::stream! {
+						loop {
```
tromp, line 224: Why does this need a loop that was not needed before?

servers/src/mining/stratumserver.rs
```diff
@@ -602,64 +602,72 @@ impl Handler {
...
+				match listener.accept().await {
+					Ok((socket, _)) => yield socket,
+					Err(e) => {
+						error!("accept error = {:?}", e);
+						continue;
+					}
+				}
+			}
+		}
+		.for_each(move |socket| {
```
tromp, line 619: I can't believe how bad git is at diffing this change. It looks like it matches perfectly from the old line 611 vs this line 619 down, but git manages to mismatch it all...

tromp (approved):
It all looks perfectly acceptable. Just wondering about the introduction of the two loops...

api/src/rest.rs
yeastplume: Without the loop it would handle a single connection before stopping, this is the common pattern in this particular version of hyper (and beyond, I think, though we'll get to updating hyper to the latest next. An example here https://github.com/hs-CN/async-rustls-stream/blob/ec4298493353d3fae6f86e106d07543b9f1bea56/examples/server_client.rs#L38

PR #3806, opened 2025-03-31
Refactor: Improve Inbound Capabilities Counting

martinthonon:
Improve correctness by incrementing the inbound capabilities by one in the 'peers and sync' menu. With the default value for inbound set to 128, it was a bit misleading to see 127. Non-programmers start counting at 1, not 0.

Before :

```
Address [  ]  |  . . . .  |  Capab [  ]  |  User Agent [  ]
x.x.x.x:3414                 127            MW/Grin 5.4.0-Alpha
```

After :

```
Address [  ]  |  . . . .  |  Capab [  ]  |  User Agent [  ]
x.x.x.x:3414                 128            MW/Grin 5.4.0-Alpha
```


yeastplume:
Sorry, no. That's a field representing bitmask values, and it's value is 127, not 128.

PR #3807, opened 2025-04-13
[Do not Merge] Exploratory fix for PMMR header crash

yeastplume:
(This is here for testing and diagnosis only, for now.)

This is tryiong track down a capacity overflow panic reported in the keybase dev channel that can happen when a node receives PMMR segment data from a peer during PIBD sync. The code was blindly trusting the segment size announced by the peer and trying to allocate huge amounts of memory if that size was bogus (either intentionally malicious or just corrupted).

To fix / diagnose this, we're trying to:

* introduce checks in Segment::read within core/src/core/pmmr/segment.rs. 
* set a max limit based on the largest segment height currently configured in chain/src/pibd_params.rs 
* Improve error handling and log message when the issue occurs

This should help pinpoint where the error is coming from (whether it's an intentionally malicious peer or an error in the client code somewhere)

### DIFF
--- a/core/src/core/pmmr/segment.rs
+++ b/core/src/core/pmmr/segment.rs
@@ -14,15 +14,26 @@
 
 //! Segment of a PMMR.
 
-use crate::core::hash::Hash;
+use crate::core::hash::{Hash, Hashed};
 use crate::core::pmmr::{self, Backend, ReadablePMMR, ReadonlyPMMR};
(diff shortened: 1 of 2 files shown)


yeastplume:
Interestingly, that change triffered a test failure in the `ser_round_trip` test indicating that a nonsensical value was being read. After some tracking, it appears there was a mismatch between how a segment was being written and read, which may explain the original issue. Running a few tests to confirm we can still sync after the fix.

PR #3809, opened 2025-08-15
Update copyright year to 2025

martinthonon:
Reminder for next year:
`find . -type f -exec sed -i 's/Copyright 2025/Copyright 2026/g' {} +`

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2021 The Grin Developers
+# Copyright 2025 The Grin Developers
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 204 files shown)


IRsoctierDT (approved):
approve

IRsoctierDT (approved):
everything looks good

ardocrat:
This is unusual practice, header is updating only on first file creation, example:
https://github.com/cocos/google-game-sdk/blob/master/GameActivity/src/main/java/com/google/androidgamesdk/GameActivity.java

martinthonon:
It's a follow up to this PR, but as Antiochp suggested, we should just get rid of these years.
https://github.com/mimblewimble/grin/pull/3592

wiesche89:
I agree with Ardocrat here. Updating copyright years across the whole repository every year feels outdated and mostly creates unnecessary diff noise.

Issue #2825, opened 2019-05-14
4x limit on p2p msg lengths


antiochp:
https://github.com/mimblewimble/grin/blob/e56cd55980ea53022d7481c3168630c4cd9ef546/p2p/src/msg.rs#L278-L281

We currently allow p2p msg to be 4x the max allowed message length.
This was originally to allow for some breathing room as message formats stabilized.

But - it also put us in a place where we are no longer 100% sure we got the (1x) limits correct as we did not need to keep these limits updated over time.

Proposed approach: Add some logging around this to get an idea of where each msg is with respect to its currently defined max length.

Once we are happy we can consider introducing stricter limits on msg lengths.

PR #3811, opened 2025-09-15
p2p: log message sizes; warn on >1x; demote Ping/Pong to debug (#2825) also fixes PIBD Sync bug

syntaxjak:
Summary
Adds logging to surface P2P message sizes versus configured limits, so we can validate the 1x limits before changing enforcement (ref #2825). Also raises the stale Header size cap to 420 B (was 365 B) since headers on mainnet are consistently ~415 B.

What changed
	•	Inserted size logging in p2p/src/codec.rs after header decode:
	•	info! (or debug! for Ping/Pong) line: p2p(codec): <Type> size=<N> (1x=<A>, 4x=<B>)
	•	warn! when msg_len > 1x (still under the existing 4× cap).
	•	Demoted Ping/Pong logs to debug! to avoid info-level spam in normal operation.
	•	Bumped Header 1× limit from 365 B → 420 B, and updated Headers batch accordingly.
	•	No wire format, consensus, or API changes; logging + cap maintenance only.

Why
Historically we allowed up to 4× the nominal limits for flexibility. With this patch we surface real sizes so limits can be tuned. Logs revealed that all headers today are ~415 B, making the old 365 B limit stale. Raising it to 420 B removes noisy warnings while keeping 4× safety margin unchanged.

**How I tested**
- Built from source and ran a mainnet node.
- Observed lines like:
  - `p2p(codec): Ping size=16 (1x=16, 4x=64)` (at DEBUG)
  - `p2p(codec): CompactBlock size=1268 (1x=134803, 4x=539212)` (at INFO)
  - `p2p(codec): Header exceeds 1x limit (415 > 365)` (at WARN)
- Verified `warn!` only triggers when >1x, and that Ping/Pong no longer spam INFO.

**Risks / Perf**
- Slightly more logging; minimal runtime cost. No protocol/consensus impact.

**Notes**
- Builds on the intent of #2825 (originally opened by @antiochp), but implements logging in the codec so it covers all message types uniformly. Happy to tweak format/severity based on maintainer feedback.

### DIFF
--- a/servers/src/grin/sync/state_sync.rs
+++ b/servers/src/grin/sync/state_sync.rs
@@ -167,6 +167,8 @@ impl StateSync {
 		if sync_need_restart || header_head.height == highest_height {
 			if using_pibd {
 				if sync_need_restart {
+					self.state_sync_reset();
+					self.sync_state.clear_sync_error();
 					return true;
 				}
 				let (launch, _download_timeout) = self.state_sync_due();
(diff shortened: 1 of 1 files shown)


Anynomouss:
I will test this when I have time. 
Sounds reasonable. All those Ping messages on the INFO level are also useful since it indicates we might have more ping messages than needed. I still want to test if syncing would speed up when decreasing the amount of ping messages to for example only ones per minute. I would prefer to reduce the amount of ping messages above demoting it do DEBUG level though.

PR #3823, opened 2026-03-30
PIBD peers fix

ardocrat:
- choose peers based on minimal height (minimal 2 blocks behind max tip)
- temporary block peers for stale segments disconnecting only outbound: 
     - 1st time for 1 min.,
     - 2nd time for 3 min.,
     - 3rd time for 10min.
- force request for output and rangeproof segments to avoid stuck
- do not check for max cached segments on selecting next desired segment for request (stuck happened when peer got disconnected and we reached max segments cache size, so not tried to make another request).

### DIFF
--- a/chain/src/pibd_params.rs
+++ b/chain/src/pibd_params.rs
@@ -43,6 +43,9 @@ pub const SEGMENT_REQUEST_TIMEOUT_SECS: i64 = 20;
 /// will always be requested first)
 pub const SEGMENT_REQUEST_COUNT: usize = 15;
 
+/// How many blocks behind the tip a PIBD peer may be and still be considered usable.
+pub const PIBD_PEER_HEIGHT_SLACK_BLOCKS: u64 = 2;
+
 /// If the syncer hasn't seen a max work peer that supports PIBD in this number of seconds
 /// give up and revert back to the txhashset.zip download method
 pub const TXHASHSET_ZIP_FALLBACK_TIME_SECS: i64 = 60;
--- a/chain/src/pibd_params.rs
+++ b/chain/src/pibd_params.rs
@@ -48,4 +48,4 @@ pub const PIBD_PEER_HEIGHT_SLACK_BLOCKS: u64 = 2;
 
 /// If the syncer hasn't seen a max work peer that supports PIBD in this number of seconds
 /// give up and revert back to the txhashset.zip download method
-pub const TXHASHSET_ZIP_FALLBACK_TIME_SECS: i64 = 60;
+pub const TXHASHSET_ZIP_FALLBACK_TIME_SECS: i64 = 660;
--- a/chain/src/txhashset/desegmenter.rs
+++ b/chain/src/txhashset/desegmenter.rs
@@ -497,102 +497,129 @@ impl Desegmenter {
 					}
 				}
 			}
-		} else {
-			// We have all required bitmap segments and have recreated our local
-			// bitmap, now continue with other segments, evenly spreading requests
-			// among MMRs
-			let local_output_mmr_size;
-			let local_kernel_mmr_size;
-			let local_rangeproof_mmr_size;
-			{
-				let txhashset = self.txhashset.read();
-				local_output_mmr_size = txhashset.output_mmr_size();
-				local_kernel_mmr_size = txhashset.kernel_mmr_size();
-				local_rangeproof_mmr_size = txhashset.rangeproof_mmr_size();
-			}
-			let total_output_segments = SegmentIdentifier::count_segments_required(
-				self.archive_header.output_mmr_size,
-				self.default_output_segment_height,
-			);
-			let mut elems_added = 0;
-			if let Some(mut next_output_idx) = self.next_required_output_segment_index() {
-				while (next_output_idx as usize) < total_output_segments {
-					if self.output_segment_cache.len() >= self.max_cached_segments {
-						break;
-					}
-					if elems_added == max_elements / 3 {
-						break;
-					}
-					let output_id = SegmentIdentifier {
-						height: self.default_output_segment_height,
-						idx: next_output_idx,
-					};
-					let (_first, last) =
-						output_id.segment_pos_range(self.archive_header.output_mmr_size);
-					if last > local_output_mmr_size && !self.has_output_segment_with_id(output_id) {
-						return_vec.push(SegmentTypeIdentifier::new(SegmentType::Output, output_id));
-						elems_added += 1;
-					}
-					next_output_idx += 1;
-				}
-			}
+			// Bitmap is not finished yet, continue at next iteration when it will be ready.
+			return return_vec;
+		}
 
-			let total_rangeproof_segments = SegmentIdentifier::count_segments_required(
-				self.archive_header.output_mmr_size,
-				self.default_rangeproof_segment_height,
-			);
-			elems_added = 0;
-			if let Some(mut next_rp_idx) = self.next_required_rangeproof_segment_index() {
-				while (next_rp_idx as usize) < total_rangeproof_segments {
-					if self.rangeproof_segment_cache.len() >= self.max_cached_segments {
-						break;
-					}
-					if elems_added == max_elements / 3 {
-						break;
-					}
-					let rp_id = SegmentIdentifier {
-						height: self.default_rangeproof_segment_height,
-						idx: next_rp_idx,
-					};
-					let (_first, last) =
-						rp_id.segment_pos_range(self.archive_header.output_mmr_size);
-					if last > local_rangeproof_mmr_size
-						&& !self.has_rangeproof_segment_with_id(rp_id)
-					{
-						return_vec.push(SegmentTypeIdentifier::new(SegmentType::RangeProof, rp_id));
-						elems_added += 1;
-					}
-					next_rp_idx += 1;
+		// We have all required bitmap segments and have recreated our local
+		// bitmap, now continue with other segments, evenly spreading requests
+		// among MMRs
+		let local_output_mmr_size;
+		let local_kernel_mmr_size;
+		let local_rangeproof_mmr_size;
+		{
+			let txhashset = self.txhashset.read();
+			local_output_mmr_size = txhashset.output_mmr_size();
+			local_kernel_mmr_size = txhashset.kernel_mmr_size();
+			local_rangeproof_mmr_size = txhashset.rangeproof_mmr_size();
+		}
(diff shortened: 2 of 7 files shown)


syntaxjak:
IMG_5687

chain/src/txhashset/desegmenter.rs
```diff
@@ -497,102 +497,126 @@ impl Desegmenter {
...
-					let (_first, last) =
-						k_id.segment_pos_range(self.archive_header.kernel_mmr_size);
-					if last > local_kernel_mmr_size && !self.has_kernel_segment_with_id(k_id) {
-						return_vec.push(SegmentTypeIdentifier::new(SegmentType::Kernel, k_id));
-						elems_added += 1;
-					}
-					next_kernel_idx += 1;
+		// Explicitly add segment identifier to request if not exists.
+		let mut maybe_add_to_request = |seg_id: SegmentTypeIdentifier| {
+			if return_vec.iter().any(|i| i == &seg_id) {
```
wiesche89: Should this condition be inverted to: !return_vec.iter().any(|i| i == &seg_id)? 
It looks like the helper is supposed to add the segment when it is not already in return_vec, but currently it only does so when the segment is already present.

p2p/src/peers.rs
```diff
@@ -434,6 +434,51 @@ impl Peers {
...
+	/// Unblock blocked peers.
+	pub fn unblock_peers(&self) -> Result<(), Error> {
+		let peers = self.peers.try_read_for(LOCK_TIMEOUT).ok_or_else(|| {
+			error!("unblock_peers: failed to get peers lock");
+			Error::PeerException
+		})?;
+		let peers = peers.iter().into_iter();
+		let _ = peers
+			.filter(|(_, peer)| peer.is_blocked())
+			.map(|(_, peer)| peer.unblock());
```
wiesche89: I think this iterator is never consumed, so peer.unblock() will not actually run. Should this use for_each or a simple loop instead?

servers/src/grin/sync/state_sync.rs
```diff
@@ -256,6 +257,46 @@ impl StateSync {
...
+							.iter()
+							.outbound()
+							.by_addr(PeerAddr(peer_addr.clone()))
+							.is_some()
+					};
+					if is_outbound {
+						debug!("state_sync: disconnecting peer {}", peer_addr);
+						if let Err(e) = self
+							.peers
+							.disconnect_peer(PeerAddr(*peer_addr), "PIBD segment timeout")
```
wiesche89, line 281: For outbound peers, we block the peer and then immediately remove it from the peer map via disconnect_peer 
Doesn't that drop the in-memory state? If the peer reconnects later, it looks like the block counter starts from 1 again!?
The question is whether we actually want this behavior?

wiesche89 (commented):
Thanks for working on this. I found three points that I think should be checked before merging.

servers/src/grin/sync/state_sync.rs
ardocrat: Yes, I think to rework it to store blocked list in memory, updating with every iteration at peers monitor, so we can reuse this functionality for headers sync, for example.

servers/src/grin/sync/state_sync.rs
```diff
@@ -318,11 +359,25 @@ impl StateSync {
 					.connected()
 			};
 
+			// Get peers with reasonable height for pibd.
+			let height_slack = pibd_params::PIBD_PEER_HEIGHT_SLACK_BLOCKS;
+			let max_pibd_height = peers_iter_pibd()
```
wiesche89, line 364: This situation does not feel ideal overall, although it already existed before, since the height selection can be skewed by a single outlier peer.

In theory, it should be possible to validate against the required height instead, right?

That said, this could also introduce other risks.

Should we open a separate issue for this?

servers/src/grin/sync/state_sync.rs
```diff
@@ -318,11 +359,25 @@ impl StateSync {
...
+					p.info.height().saturating_add(height_slack) >= max_pibd_height
+				})
+			};
+
 			// If there are no suitable PIBD-Enabled peers, AND there hasn't been one for a minute,
 			// abort PIBD and fall back to txhashset download
 			// Waiting a minute helps ensures that the cancellation isn't simply due to a single non-PIBD enabled
 			// peer having the max difficulty
-			if peers_iter_pibd().count() == 0 {
+			if available_pibd_peers().count() == 0 {
```
wiesche89: For the fallback decision, we should only count peers that are actually eligible for requests, meaning height-eligible and not blocked.
Otherwise, this could also lead to the sync getting stuck.

servers/src/grin/sync/state_sync.rs
```diff
@@ -339,72 +394,95 @@ impl StateSync {
...
 				.or_else(|| {
-					peers_iter_pibd()
+					available_pibd_peers()
 						.inbound()
+						.with_filter(|p| !peers.is_blocked(p.info.addr))
 						.exclude(excluded_peer)
 						.choose_random()
+						.or_else(|| {
+							// Select from blocked if we have no peers (could be network issue).
+							available_pibd_peers()
```
wiesche89: If no unblocked peers are found, the selection falls back to blocked peers again. Could this potentially lead to an endless loop?

chain/src/types.rs
```diff
@@ -316,6 +316,16 @@ impl SyncState {
 		removed_segments
 	}
 
+	/// Drop all tracked PIBD requests, returning how many entries were removed.
+	pub fn clear_pibd_requests(&self) -> usize {
```
wiesche89, line 320: Currently, pending PIBD requests are cleared, which means the node loses the mapping of which segments are currently expected from which peers.

I have effectively fixed this in my branch: pending PIBD requests are no longer cleared globally. Instead, retryable segments are reconsidered via retryable_pibd_segments(), and when they are requested again, the existing request state is updated via refresh_pibd_segment().

So the question is whether we should wait for my change here, or handle this separately in a new PR?

chain/src/txhashset/desegmenter.rs
```diff
@@ -497,102 +497,126 @@ impl Desegmenter {
...
+			local_rangeproof_mmr_size = txhashset.rangeproof_mmr_size();
+		}
+		let total_output_segments = SegmentIdentifier::count_segments_required(
+			self.archive_header.output_mmr_size,
+			self.default_output_segment_height,
+		);
+		let mut elems_added = 0;
+		if let Some(mut next_output_idx) = self.next_required_output_segment_index() {
+			while (next_output_idx as usize) < total_output_segments {
+				if elems_added == max_elements / 3 {
```
wiesche89, line 523: I intentionally kept the max_cached_segments checks in place. This prevents creating additional normal requests for output, rangeproof, and kernel segments once the respective segment cache is full.
That avoids requesting too many extra segments under bad peer ordering or malicious peers, which would otherwise increase retry pressure.
Only the next required kernel is still explicitly forced, to avoid the known PIBD stall case.

servers/src/grin/sync/state_sync.rs
ardocrat: as I remember we ban peers with wrong height (in headers sync)

chain/src/types.rs
ardocrat: lets wait your change here I guess

servers/src/grin/sync/state_sync.rs
ardocrat: it means we have no PIBD peers available at all and need to switch to .zip download

servers/src/grin/sync/state_sync.rs
ardocrat: OK, it should help

servers/src/grin/sync/state_sync.rs
```diff
@@ -256,6 +257,46 @@ impl StateSync {
 			.sync_state
 			.remove_stale_pibd_requests(pibd_params::SEGMENT_REQUEST_TIMEOUT_SECS);
 
+		if !stale_segments.is_empty() {
+			for (seg_id, peer_addr) in stale_segments.iter() {
```
DavidBurkett, line 261: A single network stall could cause up to 15 segment failures for a single peer, resulting in block_peer being called all 15 times for a single incident. Doing a simple dedup by peer_addr would be better:

```
let stale_peers: HashSet<_> = stale_segments
    .iter()
    .filter_map(|(_, peer_addr)| peer_addr.map(PeerAddr))
    .collect();

for peer_addr in stale_peers {
    let _ = self.peers.block_peer(peer_addr, "PIBD segment timeout");
    // disconnect if outbound, as today
}
```

chain/src/pibd_params.rs
```diff
@@ -43,6 +43,9 @@ pub const SEGMENT_REQUEST_TIMEOUT_SECS: i64 = 20;
 /// will always be requested first)
 pub const SEGMENT_REQUEST_COUNT: usize = 15;
 
+/// How many blocks behind the tip a PIBD peer may be and still be considered usable.
+pub const PIBD_PEER_HEIGHT_SLACK_BLOCKS: u64 = 2;
+
 /// If the syncer hasn't seen a max work peer that supports PIBD in this number of seconds
 /// give up and revert back to the txhashset.zip download method
```
DavidBurkett, line 50: This PR makes temporarily blocked PIBD peers indistinguishable from having no PIBD peers at all. Since the fallback timeout is 60s but the new block durations can be 180s/600s, PIBD can abort before a temporarily blocked peer is eligible again. We should increase this to be longer than the maximum temporary block duration (600s). So, maybe change this constant to 660s (11 minutes)?

DavidBurkett:
The description says “outbound resetting counter after reconnect”... Is that intended? I don’t see reconnect clearing the `blocked` entry, so outbound peers appear to keep their timeout count until `unblock_peers()` clears all PIBD blocks.

ardocrat:
> The description says “outbound resetting counter after reconnect”... Is that intended? I don’t see reconnect clearing the `blocked` entry, so outbound peers appear to keep their timeout count until `unblock_peers()` clears all PIBD blocks.

fixed at last commits, fixed description

chain/src/pibd_params.rs
ardocrat: updated `TXHASHSET_ZIP_FALLBACK_TIME_SECS` to 660s

servers/src/grin/sync/state_sync.rs
ardocrat: good catch, thank you

Issue #3461, opened 2020-10-05
Use multiple LMDB dbs (tables)


quentinlesceller:
Meta issue to track the implementation of having multiple LMDB tables instead of `chain` and `peers`.
More context here https://github.com/mimblewimble/grin/pull/3320.

ardocrat:
Fixed at https://github.com/mimblewimble/grin/pull/3825

PR #3825, opened 2026-04-16
Migrate from lmdb-zero to heed

ardocrat:
Migrate from non-maintained [lmdb-zero](https://docs.rs/lmdb-zero/) to [heed](https://docs.rs/heed).
- Speed up of sync as result
- No more stuck at parallel commands while syncing
- Single environment for `peers` and `chain` data, migrate existing dbs outside default `lmdb` environment (solves https://github.com/mimblewimble/grin/issues/3461)
- Ability to use multiple shared environments (per path).
- Use separate databases instead of prefixes, use default database for values without prefixes, migrate old environment (https://github.com/mimblewimble/grin/pull/3320)


aglkm:
Compiled master branch with this PR and got the following core dump:
```
(gdb) bt
#0  __memcpy_avx_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:265
#1  0x000055b7e12b2d8c in mdb_txn_renew0 (txn=0x55b7f38018f0)
    at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lmdb-master-sys-0.2.6/lmdb/libraries/liblmdb/mdb.c:3176
#2  0x000055b7e12b341e in mdb_txn_begin (env=0x55b7f36f4520, parent=0x0, flags=131072, ret=0x7ffe09ade890)
    at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lmdb-master-sys-0.2.6/lmdb/libraries/liblmdb/mdb.c:3349
#3  0x000055b7e128df1d in heed::txn::RoTxn<heed::txn::WithoutTls>::new<heed::txn::WithoutTls> (env=0x55b7f36f2270)
    at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/heed-0.22.1/src/txn.rs:67
#4  0x000055b7e128f047 in heed::envs::env::Env<heed::txn::WithoutTls>::read_txn<heed::txn::WithoutTls> (self=0x55b7f36f2270)
    at /home/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/heed-0.22.1/src/envs/env.rs:427
#5  0x000055b7e1213e17 in grin_store::lmdb::Store::get_ser<grin_chain::types::Tip> (self=0x55b7f36f2240, key=..., deser_mode=...)
    at store/src/lmdb.rs:309
#6  0x000055b7e116d270 in grin_chain::store::ChainStore::head (self=0x55b7f36f2240) at chain/src/store.rs:65
#7  0x000055b7e1257799 in grin_chain::chain::Chain::head (self=0x55b7f36f9ce0) at chain/src/chain.rs:1457
#8  0x000055b7e124ae78 in grin_chain::chain::Chain::difficulty_iter (self=0x55b7f36f9ce0) at chain/src/chain.rs:1669
#9  0x000055b7e096ec81 in grin_servers::grin::server::Server::get_server_stats (self=0x7ffe09ae2b80) at servers/src/grin/server.rs:433
#10 0x000055b7e05f1b48 in grin::tui::ui::Controller::run (self=0x7ffe09ae25a0, server=...) at src/bin/tui/ui.rs:212
#11 0x000055b7e062cb62 in grin::cmd::server::start_server_tui::{closure#0} (serv=..., logs_rx=...) at src/bin/cmd/server.rs:60
#12 0x000055b7e06295e0 in grin_servers::grin::server::Server::start<grin::cmd::server::start_server_tui::{closure_env#0}> (
    config=..., logs_rx=..., info_callback=..., stop_state=..., api_chan=0x55b7f36f2180) at servers/src/grin/server.rs:121
#13 0x000055b7e064748a in grin::cmd::server::start_server_tui (config=..., logs_rx=..., api_chan=0x55b7f36f2180)
    at src/bin/cmd/server.rs:53
#14 0x000055b7e0645d31 in grin::cmd::server::start_server (config=..., logs_rx=..., api_chan=0x55b7f36f2180)
    at src/bin/cmd/server.rs:40
#15 0x000055b7e064670b in grin::cmd::server::server_command (server_args=..., global_config=..., logs_rx=..., api_chan=0x55b7f36f2180)
    at src/bin/cmd/server.rs:146
#16 0x000055b7e056f83e in grin::real_main () at src/bin/grin.rs:247
#17 0x000055b7e056d206 in grin::main () at src/bin/grin.rs:81
```
Stopped at: `Sync step 1/7: Downloading headers: 13%`

Not sure if the above stack is the place where the crash occurred, or just Linux defaulted to show thread 1 for `bt` command.
So here is the output from `thread apply full bt` command:
[thread_apply_full_bt.txt](https://github.com/user-attachments/files/27169994/thread_apply_full_bt.txt)

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
 	}
 
-	/// Increments the database size by as many ALLOC_CHUNK_SIZES
-	/// to give a minimum threshold of free space
-	pub fn do_resize(&self) -> Result<(), Error> {
-		let env_info = self.env.info()?;
-		let stat = self.env.stat()?;
-		let size_used = stat.psize as usize * env_info.last_pgno;
+	/// Resize database environment if needed.
+	fn maybe_resize(&self) {
```
wiesche89, line 445: Between checking open_txs_count == 0 and setting resizing = true, a new transaction can still start. However, heed::Env::resize() requires that absolutely no transaction is active at the time of resizing.

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+		self.wait_for_resize();
 
-		let new_mapsize = if env_info.mapsize < self.alloc_chunk_size {
-			self.alloc_chunk_size
+		// Check only one resize requirement per time to avoid multiple resizes.
+		if self.resize_checking() {
```
wiesche89, line 449: resize_checking() and set_resize_checking(true) are separate operations (race window).
Would it make sense to introduce a lock here to make the check-and-set operation atomic?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
-			self.env.set_mapsize(new_mapsize)?;
+		let (resize, new_size) = needs_resize(&self.env, self.alloc_chunk_size);
+		if resize {
+			let env_path = self.env_path.clone();
+			let env = self.env.clone();
+
+			// Resize immediately or at another thread to not interrupt current
+			// transaction waiting all open transactions to be closed.
+			if self.open_txs_count() != 0 {
+				thread::spawn(move || {
```
wiesche89, line 470: The resize thread runs in a tight loop without any pause as long as transactions are still open.
Do we need a sleep here (CPU)?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
-				res.env.clone(),
-				Some(&res.name),
-				&lmdb::DatabaseOptions::new(lmdb::db::CREATE),
-			)?));
+		// Migrate to default environment if needed.
+		let env_name = env_name.unwrap_or(DEFAULT_ENV_NAME);
+		if env_name != DEFAULT_MULTI_DB_ENV_NAME {
+			let migrate_from = Path::new(root_path).join(env_name);
+			if migrate_from.exists() {
+				match s.migrate_to_default_env(db_name, &migrate_from) {
```
wiesche89: If the migration fails, it is only logged and the store is still returned. FIs that too risky? Because the node may continue running with an empty or only partially migrated database?!

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+		let db_from = {
+			let mut write = from_env.write_txn()?;
+			let db: Database<Bytes, Bytes> = from_env.create_database(&mut write, from_name)?;
+			write.commit()?;
+			db
+		};
+		let mut write_to = self.env.write_txn()?;
+		let read_from = from_env.read_txn()?;
+		let mut count = 0;
+		for kv in db_from.iter(&read_from)? {
```
wiesche89: Errors while reading individual records during migration are currently ignored.

example idea:
let (k, v) = kv?; inside the loop and abort the migration immediately on any record read error.

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
-				&lmdb::DatabaseOptions::new(lmdb::db::CREATE),
-			)?));
+		// Migrate to default environment if needed.
+		let env_name = env_name.unwrap_or(DEFAULT_ENV_NAME);
+		if env_name != DEFAULT_MULTI_DB_ENV_NAME {
+			let migrate_from = Path::new(root_path).join(env_name);
+			if migrate_from.exists() {
+				match s.migrate_to_default_env(db_name, &migrate_from) {
+					Ok(_) => {
+						let _ = fs::remove_dir_all(&migrate_from);
```
wiesche89: The old database should only be deleted after successful validation, or initially just renamed into a backup directory?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+			let db: Database<Bytes, Bytes> = from_env.create_database(&mut write, from_name)?;
+			write.commit()?;
+			db
+		};
+		let mut write_to = self.env.write_txn()?;
+		let read_from = from_env.read_txn()?;
+		let mut count = 0;
+		for kv in db_from.iter(&read_from)? {
+			if let Ok((k, v)) = kv {
+				if k.contains(&PREFIX_KEY_SEPARATOR) {
```
wiesche89: Would it be safer to only treat keys in the exact format [prefix, ':', rest...] as legacy-prefixed keys?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+					let db_name = k.split_at(1).0;
+					let db = self.pre_dbs.get(&db_name[0]).unwrap();
```
wiesche89: unwrap() can panic here if a legacy key contains an unknown prefix. During migration, unexpected data should not cause a panic, but instead return a clear error and preserve the old database.

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+			let env_options = options.map_size(self.alloc_chunk_size).max_dbs(24);
+			env_options.open(from_path)?
 		};
-		Store {
-			env: self.env.clone(),
-			db: self.db.clone(),
-			name: self.name.clone(),
-			version,
-			alloc_chunk_size,
+		let (resize, new_size) = needs_resize(&from_env, self.alloc_chunk_size);
```
wiesche89, line 334: It may be safer to calculate the target size separately based on the target environment instead of reusing the source size?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+		if resize {
+			unsafe {
+				from_env.resize(new_size)?;
+				self.env.resize(new_size)?;
```
wiesche89, line 339: During migration, self.env.resize() is called directly without the same safeguards used in the runtime resize path.

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+		read: RoTxn<'static, WithoutTls>,
 		deserialize: F,
-	) -> PrefixIterator<F, T> {
-		PrefixIterator {
-			tx,
-			cursor,
-			seek: false,
-			prefix: prefix.to_vec(),
+	) -> DatabaseIterator<F, T> {
+		let keys = if let Ok(iter) = db.iter(&read) {
```
wiesche89: The new iterator loads all keys into a Vec<Vec<u8>> upfront. For chain and peer databases, this creates a scalability risk. Would it be better to keep a cursor-based or chunked iterator approach instead?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
-		};
-		kv.ok()
-			.filter(|(k, _)| k.starts_with(self.prefix.as_slice()))
-			.map(|(k, v)| match (self.deserialize)(k, v) {
-				Ok(v) => Some(v),
-				Err(_) => None,
-			})
-			.flatten()
+		if let Some(k) = self.keys.iter().skip(self.skip).next() {
+			let v = self.db.get(&self.read, k).unwrap_or(None);
```
wiesche89: Errors could silently get swallowed here. Would it be safer to change the iterator to Item = Result<T, Error> instead?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+			if let Some(v) = v {
+				return match (self.deserialize)(k, v) {
+					Ok(v) => {
+						self.skip += 1;
+						Some(v)
+					}
+					Err(_) => None,
```
wiesche89: Make deserialization errors visible here. Should we log them explicitly, or preferably return a result?

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+	pre_dbs: Arc<HashMap<u8, Database<Bytes, Bytes>>>,
+	def_db: Database<Bytes, Bytes>,
 	version: ProtocolVersion,
 	alloc_chunk_size: usize,
 }
 
+impl Drop for Store {
+	fn drop(&mut self) {
+		{
+			let mut w_map = ENV_MAP.get().unwrap().write();
```
wiesche89, line 121: ENV_MAP is removed as soon as no store is open anymore, but there may still be active iterators, TxCounters, or similar

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
+
+impl Drop for TxCounter {
+	fn drop(&mut self) {
+		Self::on_change_tx_count(&self.env_path, false);
+	}
+}
+
+impl TxCounter {
+	/// Increment or decrement active transactions count for current environment.
+	fn on_change_tx_count(env_path: &String, inc: bool) {
```
wiesche89, line 682: The transaction counter is managed manually as a plain u32 without additional safeguards. If the accounting ever gets out of sync, it could underflow or lead to incorrect resize decisions.

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
-		let name = match env_name {
-			Some(n) => n.to_owned(),
-			None => "lmdb".to_owned(),
-		};
-		let db_name = match db_name {
-			Some(n) => n.to_owned(),
-			None => "lmdb".to_owned(),
-		};
-		let full_path = [root_path.to_owned(), name].join("/");
+		let full_path = Path::new(root_path)
```
wiesche89, line 166: can to_str().unwrap() panic if the path contains non-UTF-8 data?!

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
-		);
-		Ok(())
+	/// Protocol version for the store.
+	pub fn protocol_version(&self) -> ProtocolVersion {
+		self.version
 	}
 
-	/// Gets a value from the db, provided its key.
+	/// Get database from provided key or return default.
+	fn get_db(&self, db_key: Option<u8>) -> &Database<Bytes, Bytes> {
```
wiesche89: function unwraps unknown database keys. If a prefix is missing or misconfigured somewhere, the process can crash.

store/src/lmdb.rs
```diff
@@ -80,434 +82,752 @@ where
...
-		let res = Store {
-			env: Arc::new(env),
-			db: Arc::new(RwLock::new(None)),
-			name: db_name,
+		// Database setup.
+		let r_env_map = env_map.read();
+		let env = r_env_map.get(&full_path).unwrap().env.clone();
+		let mut write = env.write_txn()?;
+		let def_name = db_name.unwrap_or(DEFAULT_ENV_NAME);
+		let def_db = env.create_database(&mut write, Some(def_name))?;
```
wiesche89: Restart/upgrade scenarios should be covered by tests.
Add a test that writes data, drops the store, reopens the store, verifies the data can still be read?

store/src/lib.rs
```diff
@@ -26,62 +26,23 @@ extern crate log;
...
 pub mod prune_list;
 pub mod types;
 
-const SEP: u8 = b':';
-
-use byteorder::{BigEndian, WriteBytesExt};
-
 pub use crate::lmdb::*;
 
-/// Build a db key from a prefix and a byte vector identifier.
```
wiesche89, line 43: It may be worth keeping these helpers as deprecated APIs, even if they are no longer used internally? line 43-71

p2p/src/store.rs
```diff
@@ -118,45 +118,91 @@ pub struct PeerStore {
...
+	}
+}
+
+pub struct PeersIterBatch<'a> {
+	db: grin_store::Batch<'a>,
+}
+
+impl<'a> PeersIterBatch<'a> {
+	/// Iterator over all known peers.
+	pub fn peers_iter(&self) -> Result<impl Iterator<Item = PeerData>, Error> {
```
wiesche89: eer iteration now depends on the new full key snapshot iterator. With a large peer database, this could make normal peer operations significantly more expensive.

But we already had issues around this in the past with very large peer counts (xk peers).

p2p/src/store.rs
```diff
@@ -118,45 +118,91 @@ pub struct PeerStore {
...
+		let protocol_version = self.db.protocol_version();
+		self.db.iter(Some(PEER_PREFIX), move |_, mut v| {
+			ser::deserialize(&mut v, protocol_version, DeserializationMode::default())
+				.map_err(From::from)
+		})
+	}
+
 	/// Find some peers in our local db.
 	pub fn find_peers(
```
wiesche89, line 210: Technically the same issue as around line 198.

chain/src/store.rs
```diff
@@ -275,40 +306,42 @@ impl<'a> Batch<'a> {
...
 	/// original commitment that the key is constructed from. So we need a way of comparing
 	/// a key with another commitment without reconstructing the commitment from the key bytes.
 	pub fn is_match_output_pos_key(&self, key: &[u8], commit: &Commitment) -> bool {
-		let commit_key = to_key(OUTPUT_POS_PREFIX, commit);
-		commit_key == key
+		commit.as_ref() == key
 	}
 
 	/// Iterator over the output_pos index.
 	pub fn output_pos_iter(&self) -> Result<impl Iterator<Item = (Vec<u8>, CommitPos)>, Error> {
```
wiesche89: With the new iterator, this can use a lot of memory for a large UTXO index. Would a streaming iterator be safer here?

chain/src/store.rs
```diff
@@ -423,10 +459,9 @@ impl<'a> Batch<'a> {
 
 	/// Iterator over all full blocks in the db.
 	/// Uses default db serialization strategy via db protocol version.
-	pub fn blocks_iter(&self) -> Result<impl Iterator<Item = Block>, Error> {
-		let key = to_key(BLOCK_PREFIX, "");
+	pub fn blocks_iter(&self) -> Result<impl Iterator<Item = Block> + 'a, Error> {
```
wiesche89: Same scalabilitie issue with the key snapshot?

chain/src/txhashset/txhashset.rs
wiesche89: function init_output_pos_index uses output_pos_iter and is important for recovery/rebuild. If the iterator allocates unbounded memory, this recovery path could become quite memory-heavy on larger databases.

store/tests/lmdb.rs
wiesche89, line 76: I think it would be useful to add dedicated migration tests here

wiesche89, line 106: We could also test the iterator error paths here.

wiesche89, line 143: we can write concurrent-stresstest/trgger resizing 

or

simulate a partial migration, reopen the store, and verify that it does not continue with incomplete data.

chain/tests/store_kernel_pos_index.rs
wiesche89, line 33: The kernel index tests cover the new behavior, but they do not verify migration from old prefixed keys into the new named databases.

store/src/lmdb.rs
ardocrat: we can delete new db in case of fail to sync from 0 at this case

store/src/lmdb.rs
ardocrat: its OK, we check for opened transactions at separate thread to start resize, in case of fail of resize() this method will be called again at next check

store/src/lmdb.rs
ardocrat: we use rwlock here, I will also add Atomicbool instead of usual bool if it will help

store/src/lmdb.rs
ardocrat: problem in migration can happen if some data at old db is corrupted or there is not enough disk space (at 1st case we can delete old db and start from empty db, at 2nd case we can remove records from old db on migration or clean new db and showing an error to user, this can also be checked before migration

store/src/lmdb.rs
ardocrat: question is how to validate and why we need this if there are no errors occurred?

store/src/lmdb.rs
ardocrat: its called before actual migration here, so must be safe, we call write only once after processing all keys.

store/src/lmdb.rs
ardocrat: in the old db PREFIX_KEY_SEPARATOR was unique for keys, so must be safe, we can check only its position here (index 1)

store/src/lmdb.rs
ardocrat: I tried, they are extremely slow, at heed we have no access to raw cursor like it was before (same for iterator if we skip records), loading keys became fastest solution

store/src/lmdb.rs
ardocrat: From what I checked target size is even smaller since we are not storing prefixes, we are writing to new db only after iteration over all existing keys, it can be improved if we'll decide to save data in batches, removing old data, also to not have problem when disk is full

store/src/lmdb.rs
ardocrat: Will keep read txs for iterators at store and remove them on iterator drop

store/src/lmdb.rs
ardocrat: There is no method to close store, we can just destroy struct from memory. Currently all stores are open in lifetime of application for wallet till its open, additional checks for get_db to not crash will be added

store/src/lmdb.rs
ardocrat: it means user specified wrong path at config, we can validate paths on config parse, maybe create separate issue for this?

p2p/src/store.rs
ardocrat: Iterator for DB will be implemented to store maximum 10k keys

p2p/src/store.rs
ardocrat: Iterator for DB will be optimized to store max 10k keys.

chain/src/store.rs
ardocrat: Iterator will be optimized as commented above.

chain/src/store.rs
ardocrat: Iterator will be optimized.

chain/src/txhashset/txhashset.rs
ardocrat: Thank you for finding all weak places, iterator will be optimized.

store/src/lmdb.rs
ardocrat: fixed to atomic

store/src/lmdb.rs
ardocrat: all known prefixes used for data are in constants, if there is unknown prefix means this data can be ignored I guess (it also helped in testing of old db migration)

store/src/lmdb.rs
ardocrat: We use atomic for opened transactions now

store/src/lmdb.rs
ardocrat: Added 10ms delay

store/src/lmdb.rs
ardocrat: Decided to keep max 10k keys at iterator

store/src/lmdb.rs
ardocrat: Question is should we just show an error and continue at iterators or break iteration completely

store/src/lmdb.rs
ardocrat: Same question as above

store/src/lib.rs
ardocrat: Not sure why we need them since we removed such prefixes to avoid many allocations: https://github.com/mimblewimble/grin/pull/3320

store/src/lmdb.rs
wiesche89: I think errors should definitely be reported to the user and not be hidden as a normal end of iteration.
For migration or storage critical paths, I would abort on the first DB read or deserialization error and keep the old DB untouched. Then the user can fix the cause, e.g. disk space or corruption, retry the migration, or explicitly choose to delete the DB and resync from scratch.

store/src/lmdb.rs
wiesche89: One idea would be:
First run a reopen/load smoke test after the migration, possibly by adding or reading a block.
Only if that succeeds, the old data can be deleted.

Otherwise we might delete the old DB and, in the worst case, force users to fully resync.

store/src/lmdb.rs
wiesche89: I was thinking of something like this before each migration step:
 
new multi_lmdb map_size = current multi_lmdb used size + current source DB used size + safety margin

So before the chain DB migration this uses the current target usage plus the old chain DB usage, and before the peer DB migration it recalculates again using the already-filled multi_lmdb plus the old peer DB usage.
That should be safer than sizing the target only from the source environment.

I think this depends on whether LMDB can actually release disk space during migration.

store/src/lmdb.rs
wiesche89: In the normal startup path this is probably safe. 
My concern is only that resize() requires the environment to have no active transactions, so we are relying on that startup ordering implicitly here. 
It would be good to document or assert that assumption.

https://github.com/meilisearch/heed/blob/main/heed/src/envs/env.rs#L647

store/src/lmdb.rs
wiesche89: My remaining concern is that this is still not real streaming, the iterator still counts all keys up front and later reloads them using skip(n), which can become expensive on large DBs. Could we instead keep the read transaction locally inside an iterator tied to the store lifetime, and continue loading from the last seen key?

store/src/lmdb.rs
wiesche89: I think errors should definitely be reported to the user and not be hidden as a normal end of iteration.
For migration or storage critical paths, I would abort on the first DB read or deserialization error and keep the old DB untouched. Then the user can fix the cause, e.g. disk space or corruption, retry the migration, or explicitly choose to delete the DB and resync from scratch.

store/src/lmdb.rs
wiesche89: Answered above. It's just my opinion.

store/src/lmdb.rs
wiesche89: Yes, you're right, it is always open. It is more of a robustness issue (stores_count == 0)
But for the iterator to be safe, it additionally needs to be tied to the store lifetime, and DatabaseIterator should internally hold a reference/guard with the same lifetime.

store/src/lmdb.rs
wiesche89: Yes, we can handle this in a separate ticket.
Do you want to take that one?

store/src/lmdb.rs
ardocrat: problem is we dont know what we are migrating, this thing is universal for all DBs (peers too) since this crate encapsulates only DB methods, so external crates knows nothing about migration, like we dont know about specific databases inside this store crate

store/src/lmdb.rs
ardocrat: this is exact problem, we can not store iterator RoIter itself, it creates from reference of read tx RoTxn. References of read txs should be stored somewhere outside, but can not be shared between threads since we can create a lot iterators from single db.

store/src/lmdb.rs
ardocrat: Validation of path is happening at `one_grin_at_a_time` function on server creation.

store/src/lmdb.rs
ardocrat: This space is for memory, so disk space remained unchanged while we resize, we already add extra space for new db here as for old on resize (65% from total stored, `RESIZE_MIN_TARGET_PERCENT` const is for this)

store/src/lmdb.rs
ardocrat: Now will return an error if db with provided key not found

store/src/lmdb.rs
ardocrat: Added comment here.

store/src/lmdb.rs
ardocrat: Add missed code to iterate over 10k keys and load more keys when needed

store/src/lmdb.rs
ardocrat: Now we handle read error on migration to interrupt process and clear new db in case of error

store/src/lmdb.rs
ardocrat: Handling all errors in migration now

store/src/lmdb.rs
ardocrat: Now we return an error on read next at iterator, so its must be safe to use

store/src/lmdb.rs
ardocrat: Changed to Item = Result<T, Error> in case of read errors from database

store/src/lmdb.rs
ardocrat: Fixed

store/src/lib.rs
ardocrat: Can be used for tests to reproduce data migration

chain/tests/store_kernel_pos_index.rs
ardocrat: added migration tests

store/tests/lmdb.rs
ardocrat: added

store/src/lmdb.rs
ardocrat: data can not be read on drop if commit method of batch not called or returns an error

store/tests/lmdb.rs
ardocrat: to emulate this we need to break process somehow, currently migration is happening synchronously on store object creation, such test can be done only manually by calling ctrl + c in non-tui mode for example

ardocrat: iterator now returns an error on deserialization or database read issue

store/src/lmdb.rs
```diff
@@ -80,434 +83,861 @@ where
...
+			match self.db.get(&self.read, k) {
+				Ok(v) => {
+					if let Some(v) = v {
+						return match (self.deserialize)(k, v) {
+							Ok(v) => {
+								self.skip_total += 1;
+								self.skip_cur += 1;
+								Some(Ok(v))
+							}
+							Err(e) => {
```
wiesche89, line 905: What behavior do we want here? In any case, next() should advance before returning Err. Otherwise callers that continue iterating after Err could loop forever on the same key, right?

store/src/lmdb.rs
```diff
@@ -80,434 +83,861 @@ where
...
+		for kv in db_from.iter(&read_from)? {
+			let (k, v) = kv?;
+			if k.len() > 1 && k[1] == PREFIX_KEY_SEPARATOR {
+				let db_name = k.split_at(1).0;
+				if let Some(db) = self.pre_dbs.get(&db_name[0]) {
+					let key = k.split_at(2).1;
+					db.put(&mut write_to, key, &v)?;
+					count += 1;
+				} else {
+					error!("Migration: unknown DB key: {}", db_name[0]);
```
wiesche89: Should we force and propagate an error here instead of silently accepting this case? If an unknown prefix is skipped, migration can still succeed, the old DB gets removed, and the new DB is incomplete but treated as valid.

store/src/lmdb.rs
```diff
@@ -80,434 +83,861 @@ where
...
+		if resize {
+			// We are sure there are no active txs, cause migration is called on database creation.
+			unsafe {
+				from_env.resize(new_size)?;
+				self.env.resize(new_size)?;
+			}
 		}
+		let db_from = {
+			let mut write = from_env.write_txn()?;
+			let db: Database<Bytes, Bytes> = from_env.create_database(&mut write, from_name)?;
```
wiesche89, line 344: We should fail if the source db is missing, otherwise we may create/open an empty DB and treat the migration as successful.

store/src/lmdb.rs
```diff
@@ -80,434 +83,861 @@ where
...
+							.load(Ordering::Relaxed);
+						if txs_count == 0 {
+							debug!("Start resizing DB {}", env_path);
+							ENV_MAP
+								.get()
+								.unwrap()
+								.write()
+								.get_mut(&env_path)
+								.unwrap()
+								.resizing
```
wiesche89: I think there is still a small race here, right? We check that open_txs_count is 0 first, and only then set resizing to true. In between, a new transaction could start because wait_for_resize() only blocks once resizing is true. Would it be safer to set resizing = true`before waiting for the open transactions to drain, then resize once the count reaches 0?

store/src/lmdb.rs
```diff
@@ -80,434 +83,861 @@ where
...
-				&lmdb::DatabaseOptions::new(lmdb::db::CREATE),
-			)?));
+		// Migrate to default environment if needed.
+		let env_name = env_name.unwrap_or(DEFAULT_ENV_NAME);
+		if env_name != DEFAULT_MULTI_DB_ENV_NAME {
+			let migrate_from = Path::new(root_path).join(env_name);
+			if migrate_from.exists() {
+				match s.migrate_to_default_env(db_name, &migrate_from) {
+					Ok(_) => {
+						let _ = fs::remove_dir_all(&migrate_from);
```
wiesche89: If remove_dir_all fails, the node starts successfully but the old DB remains. Small check?

p2p/src/store.rs
```diff
@@ -118,45 +118,91 @@ pub struct PeerStore {
...
+		} else {
+			peer.last_attempt = Utc::now().timestamp();
+		}
+
+		batch.put_ser(Some(PEER_PREFIX), key.as_bytes(), &peer)?;
 		batch.commit()
 	}
 
+	/// Builds a new iterator batch to be used with this store.
+	pub fn iter_batch(&self) -> Result<PeersIterBatch<'_>, Error> {
```
wiesche89, line 186: Do we need to open a write batch for reading only peer iteration here?

store/src/lmdb.rs
ardocrat: we check db file existence before migration, name is coming from Store creation

store/src/lmdb.rs
ardocrat: we also set resize_checking to true to avoid blocking of existing transactions so its must be safe to resize only when there is really 0 txs count and only 1 check is active. if we'll put resizing before checking txs count it will stuck at the loop of iterator with single opened tx since we are waiting resize to finish. If env.resize method will fail, it will be called again at next check.

store/src/lmdb.rs
ardocrat: the problem can be with bad data written from somewhere outside of grin node, all keys are known at code as constants, any unknown key is getting ignored anyway

store/src/lmdb.rs
ardocrat: Fixed by moving iteration before reading result of reading data

p2p/src/store.rs
ardocrat: There is current limitation of iterator https://github.com/ardocrat/grin/blob/e9e38bd88c39a71a81a68a6959df840c4469925a/store/tests/lmdb.rs#L121

store/src/lmdb.rs
ardocrat: Return an error if removal of old DB file failed

store/src/lmdb.rs
wiesche89: Makes sense. If unknown keys are ignored as out of schema data, I think this should be warn!();

store/src/lmdb.rs
wiesche89: Fair enough. If the DB name is guaranteed by store creation, I’m okay with leaving this as is.

p2p/src/store.rs
ardocrat: Added write transaction support to iterator

store/src/lmdb.rs
```diff
@@ -80,434 +83,890 @@ where
...
-				Some(&res.name),
-				&lmdb::DatabaseOptions::new(lmdb::db::CREATE),
-			)?));
+		// Migrate to default environment if needed.
+		let env_name = env_name.unwrap_or(DEFAULT_ENV_NAME);
+		if env_name != DEFAULT_MULTI_DB_ENV_NAME {
+			let migrate_from = Path::new(root_path).join(env_name);
+			if migrate_from.exists() {
+				match s.migrate_to_default_env(db_name, &migrate_from) {
+					Ok(_) => match fs::remove_dir_all(&migrate_from) {
```
wiesche89: I would be a bit more conservative here and leave the deletion up to the user. Perhaps just rename it and provide an informational message and warning like:

INFO store migration: migration completed successfully!
WARN store migration: old database kept at path xyz,  you may remove it manually after verifying the node runs correctly with the migrated database!

But that's really just my opinion :D

store/src/lmdb.rs
wiesche89: I checked the surrounding code.
resize_checking only prevents another resize check, but new transactions seem to wait only on resizing.
Doesn't that leave a small race window between open_txs_count == 0 and setting resizing = true?
Would it make sense to set resizing = true first, then wait for active transactions to finish before calling resize?

wait is in on_change_tx_count:
image

image

store/src/lmdb.rs
ardocrat: There is problem at mobile, cause users have no access to internal memory to delete manually, also this will increase disk space at all nodes, I guess it can be safe to delete if migration passed without errors, we also clean new db now before migration  (in case of interruption) to make sure all records will be migrated again.

store/src/lmdb.rs
ardocrat: OK, will set resizing flag and will allow to open new txs when there are more than 0 txs to avoid stuck.

store/src/lmdb.rs
```diff
@@ -80,434 +83,909 @@ where
...
-				res.env.clone(),
-				Some(&res.name),
-				&lmdb::DatabaseOptions::new(lmdb::db::CREATE),
-			)?));
+		// Migrate to default environment if needed.
+		let env_name = env_name.unwrap_or(DEFAULT_ENV_NAME);
+		if env_name != DEFAULT_MULTI_DB_ENV_NAME {
+			let migrate_from = Path::new(root_path).join(env_name);
+			if migrate_from.exists() {
+				let _ = s.clear();
```
wiesche89: I still see one migration edge case that could occur. If the new database has already been committed successfully, but deleting the old database directory fails or gets interrupted, the next startup will still see the old path, call s.clear(), and potentially delete the already migrated and valid new database.
A small migration complete key would be enough to prevent the new database from being cleared once the migration has been committed successfully.

ardocrat: Thank you for the patch.

PR #3829, opened 2026-05-02
Include git info into docker build, show git ref at version

ardocrat:
- Include .git directory into Docker build to show git ref into build info: https://github.com/mimblewimble/grin/blob/90dab5fcc68381ce39526c56c7dd25f8bdd33fb7/src/bin/grin.rs#L58
- Remove `actions/checkout` action from CI, since `docker/build-push-action` uses its own git context.
- Show git ref at version and full commit hash at logs instead of last tag to differentiate running docker builds from master branch.

### DIFF
--- a/p2p/src/msg.rs
+++ b/p2p/src/msg.rs
@@ -40,8 +40,19 @@ use std::io::{Read, Write};
 use std::sync::Arc;
 use std::{fmt, thread, time::Duration};
 
+// include build information
+pub mod built_info {
+	include!(concat!(env!("OUT_DIR"), "/built.rs"));
+}
+
 /// Grin's user agent with current version
-pub const USER_AGENT: &str = concat!("MW/Grin ", env!("CARGO_PKG_VERSION"));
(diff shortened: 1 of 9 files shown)


p2p/src/msg.rs
```diff
@@ -40,8 +40,19 @@ use std::io::{Read, Write};
...
+// include build information
+pub mod built_info {
+	include!(concat!(env!("OUT_DIR"), "/built.rs"));
+}
+
 /// Grin's user agent with current version
-pub const USER_AGENT: &str = concat!("MW/Grin ", env!("CARGO_PKG_VERSION"));
+pub fn user_agent() -> String {
+	format!(
+		"MW/Grim {}{}",
```
DavidBurkett: `MW/Grin` not `MW/Grim` 😜

wiesche89 (commented):
Looks good to me. In my opinion this is quite useful, especially with the upcoming new features and staging/testing work, since it allows us to clearly distinguish different node/build versions.

PR #3833, opened 2026-05-12
docs: update contributing process

wiesche89:
Updates the contributing workflow to use staging before master and refreshes community contact links.


Anynomouss (approved):
Adding the 'staging' branch does not truly solve bottlenecks, but it is worth a shot. 
Testing by community members and the merging afterwards will probably remain as bottleneck. Still it is worth a shot.
I agree with the part about making a new branch for any pull request, it is a good habit and I should have done so (if I only knew about it) for my PR regarding config fixes in _grin-wallet_. Without making your own branch you cannot working on multiple issues, so I think this should always be recommended.

Technically, we did something similar in the past using [alpha] versions of the node and wallet for testing. The difference now is that we keep the master branch more clean - only merging PR's that have been tested. One downside might be that this makes adding contributions slightly more complex.
Let's just try and see if this improves things.

PR #3836, opened 2026-05-19
tui: show server initialization status and error

ardocrat:
Show node server initialization progress on launch.

- Show database loading status (TODO: Migration progress)
- Show API server initialization status
- Show P2P server initialization status
- Show an error on server initialization

### DIFF
--- a/src/bin/cmd/server.rs
+++ b/src/bin/cmd/server.rs
@@ -28,6 +28,8 @@ use crate::tui::ui;
 use futures::channel::oneshot;
 use grin_p2p::msg::PeerAddrs;
 use grin_p2p::PeerAddr;
+use grin_servers::common::types::{Error, ServerInitStatus};
+use grin_servers::Server;
 use grin_util::logger::LogEntry;
 use std::sync::mpsc;
 
@@ -46,29 +48,25 @@ fn start_server_tui(
 	logs_rx: Option<mpsc::Receiver<LogEntry>>,
 	api_chan: &'static mut (oneshot::Sender<()>, oneshot::Receiver<()>),
 ) {
-	// Run the UI controller.. here for now for simplicity to access
-	// everything it might need
 	if config.run_tui.unwrap_or(false) {
 		warn!("Starting GRIN in UI mode...");
-		servers::Server::start(
-			config,
-			logs_rx,
-			|serv: servers::Server, logs_rx: Option<mpsc::Receiver<LogEntry>>| {
-				let mut controller = ui::Controller::new(logs_rx.unwrap()).unwrap_or_else(|e| {
-					panic!("Error loading UI controller: {}", e);
-				});
-				controller.run(serv);
-			},
-			None,
-			api_chan,
-		)
-		.unwrap();
+		// Run the UI controller.
+		let (serv_tx, serv_rx) = mpsc::channel::<ServerInitStatus>();
+		let mut controller = ui::Controller::new(logs_rx, serv_rx).unwrap_or_else(|e| {
+			panic!("Error loading UI controller: {}", e);
+		});
+		let serv_tx_clone = serv_tx.clone();
+		thread::spawn(move || {
+			match Server::start(config, None, Some(serv_tx_clone.clone()), api_chan) {
+				Ok(s) => serv_tx_clone.send(ServerInitStatus::FinishedLoading(s)),
+				Err(e) => serv_tx_clone.send(ServerInitStatus::ErrorLoading(e)),
+			}
+		});
+		controller.run();
 	} else {
 		warn!("Starting GRIN w/o UI...");
-		servers::Server::start(
-			config,
-			logs_rx,
-			|serv: servers::Server, _: Option<mpsc::Receiver<LogEntry>>| {
+		match Server::start(config, None, None, api_chan) {
+			Ok(s) => {
 				let running = Arc::new(AtomicBool::new(true));
 				let r = running.clone();
 				ctrlc::set_handler(move || {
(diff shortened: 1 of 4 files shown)


src/bin/tui/ui.rs
```diff
@@ -122,6 +123,11 @@ impl UI {
 		// Configure a callback (shutdown, for the first test)
 		let controller_tx_clone = grin_ui.controller_tx.clone();
 		grin_ui.cursive.add_global_callback('q', move |c| {
+			let v = c.pop_layer();
```
wiesche89: I think this also blocks normal q shutdown. The main UI is a layer too, so pop_layer returns Some even when no dialog is open. Maybe we should check directly whether a dialog is open?

src/bin/tui/ui.rs
```diff
@@ -201,20 +247,33 @@ impl Controller {
 					ControllerMessage::Shutdown => {
 						warn!("Shutdown in progress, please wait");
 						self.ui.stop();
-						server.stop();
-						return;
+						break;
```
wiesche89: It looks like the server is no longer stopped on TUI shutdown. Before this change, server.stop() was called both on shutdown and after the UI loop ended.

src/bin/cmd/server.rs
```diff
@@ -79,12 +77,12 @@ fn start_server_tui(
...
 				warn!("Received SIGINT (Ctrl+C) or SIGTERM (kill).");
-				serv.stop();
-			},
-			None,
-			api_chan,
-		)
-		.unwrap();
+				s.stop();
+			}
+			Err(e) => {
```
wiesche89, line 100: Should startup failure return a non zero exit status here (docker, etc)?

src/bin/cmd/server.rs
```diff
@@ -46,29 +48,25 @@ fn start_server_tui(
...
-			api_chan,
-		)
-		.unwrap();
+		// Run the UI controller.
+		let (serv_tx, serv_rx) = mpsc::channel::<ServerInitStatus>();
+		let mut controller = ui::Controller::new(logs_rx, serv_rx).unwrap_or_else(|e| {
+			panic!("Error loading UI controller: {}", e);
+		});
+		let serv_tx_clone = serv_tx.clone();
+		thread::spawn(move || {
```
wiesche89, line 61: What happens if the user quits the tuiwhile the server is still starting? The startup continues in the background, but there does not seem to be a clean way to stop it if the ui is already gone

src/bin/tui/ui.rs
wiesche89: since this PR touches shutdown handling anyway, maybe we could also avoid the unwrap() on the shutdown send here. If the receiver is already gone, panicking during shutdown is not very helpful.

src/bin/tui/ui.rs
ardocrat: Added a check when dialog is open

src/bin/cmd/server.rs
ardocrat: Added a panic on error

src/bin/tui/ui.rs
ardocrat: Added stop back if server exists

src/bin/cmd/server.rs
ardocrat: Added a check to stop server if tui is not running anymore

src/bin/tui/ui.rs
ardocrat: Do not panic here now

src/bin/tui/ui.rs
```diff
@@ -117,17 +120,21 @@ impl UI {
...
 		grin_ui.cursive.add_global_callback('q', move |c| {
+			if show_dialog_clone.load(Ordering::Relaxed) {
+				return;
+			}
 			let content = StyledString::styled("Shutting down...", Color::Light(BaseColor::Yellow));
 			c.add_layer(CircularFocus::new(Dialog::around(TextView::new(content))).wrap_tab());
-			controller_tx_clone
-				.send(ControllerMessage::Shutdown)
-				.unwrap();
+			let _ = controller_tx_clone.send(ControllerMessage::Shutdown);
```
wiesche89, line 135: Currently, it is not possible to press q and shut down the node while an initialization dialog is being displayed.
Is that acceptable for us?

src/bin/cmd/server.rs
```diff
@@ -46,29 +48,36 @@ fn start_server_tui(
...
-		.unwrap();
+		// Run the UI controller.
+		let (serv_tx, serv_rx) = mpsc::channel::<ServerInitStatus>();
+		let mut controller = ui::Controller::new(logs_rx, serv_rx).unwrap_or_else(|e| {
+			panic!("Error loading UI controller: {}", e);
+		});
+		let tui_running = Arc::new(AtomicBool::new(true));
+		let serv_tx_clone = serv_tx.clone();
+		let tui_running_clone = tui_running.clone();
+		thread::spawn(move || {
```
wiesche89, line 61: Should the startup path detect the cancellation and abort initialization before the server has fully completed its startup process if the TUI is closed during startup?

ardocrat: we can pass stop_state to call it after TUI stop

src/bin/tui/ui.rs
ardocrat: we can just close current dialog and shut it down

src/bin/cmd/server.rs
wiesche89: That makes sense to me.

src/bin/tui/ui.rs
wiesche89: That makes sense to me.

PR #3837, opened 2026-05-20
Change default IP address for P2PConfig to '::'

noobvie:
Feature request: Change default P2P listen host from 0.0.0.0 to :: for dual-stack IPv4+IPv6 support. As tested on several windows/linux machines and not found issue. Worked with ipv4 or ipv6 or dual ipv4+ipv6.


wiesche89:
@noobvie if you want, I can move this to the staging branch, test it there, and merge it.

PR #3785, opened 2024-04-16
chore: fix typos

xiaoxianBoy:
fix some typos

PR #3840, opened 2026-05-26
Clarifies the blinding factor range proof explanation and includes the remaining typo fixes

wiesche89:
This PR replaces PR #3783 by applying the same clarification on top of the current `staging` branch.

It also includes @Anynomouss review suggestion to make the explanation of why the blinding factor range proof is needed clearer.

This PR replaces PR #3785


Anynomouss (approved):
Great to have that finally fixed and clarified.

PR #3800, opened 2024-11-13
[WIP] Snapcraft Definitions

yeastplume:
Goal of this PR is to create snaps of grin as part of the build process, this involves. 
- [ ] Adding ability to create and run snaps within the dev container
- [ ] Developing standalone snap and adding to the build file
- [ ] Adding creation of snap to build process

PR #3844, opened 2026-05-27
Snap addition

wiesche89:
Goal of this PR is to create snaps of grin as part of the build process, this involves.

 Adding ability to create and run snaps within the dev container
 Developing standalone snap and adding to the build file
 Adding creation of snap to build process

This PR completes the following PR #3800

### DIFF
--- a/.github/workflows/snap.yaml
+++ b/.github/workflows/snap.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   build-snap:
     name: Build snap
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
     steps:
       - name: Checkout
         uses: actions/checkout@v6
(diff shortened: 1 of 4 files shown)


.github/workflows/snap.yaml
```diff
@@ -0,0 +1,36 @@
+name: Snap Package
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - master
+      - staging
```
ardocrat, line 8: Seems like `staging` branch should have `grade: devel` quality level

ardocrat: https://github.com/snapcore/action-publish allows to change channel based on ref

wiesche89: > Seems like `staging` branch should have `grade: devel` quality level
Done, staging now builds the snap with grade: devel and master grade: stabel


> https://github.com/snapcore/action-publish allows to change channel based on ref
I will open a separate issue for this, since I currently do not have the Snap Store credentials required for publishing.

ardocrat (approved):

PR #3849, opened 2026-06-07
propagate lmdb iterator page errors

wiesche89:
**Content**
- report lmdb iterator errors instead of silently stopping iteration
- fix iteration across more than 10,000 entries
- isolate the peer handshake test directory
- add a regression test

**Testing**
- cargo test -p grin_store test_iter_pages
- cargo test -p grin_p2p peer_handshake

### DIFF
--- a/store/src/lmdb.rs
+++ b/store/src/lmdb.rs
@@ -635,12 +635,9 @@ impl Store {
 				Ok(read) => {
 					let db_res = self.get_db(db_key);
 					match db_res {
-						Ok(db) => Ok(DatabaseIterator::new(
-							self,
-							Arc::new(db.clone()),
-							read,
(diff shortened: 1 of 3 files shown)


store/src/lmdb.rs
```diff
@@ -894,44 +891,37 @@ where
...
-						error!("db iter: error read value: {}", e);
-						Some(Err(Error::from(e)))
+					Err(e) => {
+						return {
+							error!("db iter: error read value: {}", e);
+							Some(Err(Error::from(e)))
+						}
 					}
 				}
+			} else if self.done {
```
ardocrat: Should we check this before start a loop to avoid iteration at previous condition when no more keys can be loaded at next condition?

PR #3860, opened 2026-06-12
LMDB resize fixes

ardocrat:
- Do not resize source env before migration and on env initialization
- Optimize open txs counter before resize

### DIFF
--- a/store/src/lmdb.rs
+++ b/store/src/lmdb.rs
@@ -99,7 +99,7 @@ static ENV_MAP: OnceLock<RwLock<HashMap<String, EnvState>>> = OnceLock::new();
 struct EnvState {
 	env: Env<WithoutTls>,
 	open_txs_count: AtomicU32,
-	resizing: AtomicBool,
+	resize_pending: AtomicBool,
 	resize_checking: AtomicBool,
 	stores_count: AtomicU32,
 }
@@ -202,7 +202,7 @@ impl Store {
 				EnvState {
 					env,
 					open_txs_count: AtomicU32::new(0),
-					resizing: AtomicBool::new(false),
+					resize_pending: AtomicBool::new(false),
 					resize_checking: AtomicBool::new(false),
 					stores_count: AtomicU32::new(1),
 				},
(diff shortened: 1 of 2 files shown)


store/src/lmdb.rs
```diff
@@ -572,9 +522,8 @@ impl Store {
 		key: &[u8],
 		deser_mode: Option<DeserializationMode>,
 	) -> Result<Option<T>, Error> {
-		self.wait_for_resize();
+		let _ = self.enter_tx();
```
wiesche89: This drops the TxCounter immediately, so this read is not protected from a concurrent lmdb resize.

store/src/lmdb.rs
```diff
@@ -587,15 +536,13 @@ impl Store {
...
 			}
 		};
-		TxCounter::on_change_tx_count(&self.env_path, false);
 		res
 	}
 
 	/// Whether the key exists at the provided database key.
 	pub fn exists(&self, db_key: Option<u8>, key: &[u8]) -> Result<bool, Error> {
-		self.wait_for_resize();
+		let _ = self.enter_tx();
```
wiesche89: This drops the TxCounter immediately, so this read is not protected from a concurrent lmdb resize.

store/src/lmdb.rs
```diff
@@ -627,68 +573,66 @@ impl Store {
...
-		TxCounter::on_change_tx_count(&self.env_path, true);
-		let res = { Batch::new(self) };
-		if res.is_err() {
-			TxCounter::on_change_tx_count(&self.env_path, false);
+	/// Creates tx counter incrementing if resize is not pending.
+	fn enter_tx(&self) -> TxCounter {
+		loop {
+			let mut map = ENV_MAP.get().unwrap().write();
+			let state = map.get_mut(&self.env_path).unwrap();
+			if state.open_txs_count.load(Ordering::Relaxed) > 0
```
wiesche89: While resize is pending, this still allows new transactions as long as another transaction is already open. Under steady read load, resize may keep waiting much longer than intended.

store/src/lmdb.rs
```diff
@@ -189,26 +189,20 @@ impl Store {
 		if !has_env {
 			let env = unsafe {
 				let mut options = EnvOpenOptions::new().read_txn_without_tls();
-				let mut env_options = options.map_size(alloc_chunk_size).max_dbs(24);
+				let mut env_options = options.max_dbs(24);
```
wiesche89, line 199: Does this intentionally delay the initial map sizing until the first write? That first write may now have to perform the resize instead of doing it during env setup.

store/src/lmdb.rs
ardocrat: yes, this is needed to avoid deadlock when there are opened txs at same lifetime block.

store/src/lmdb.rs
ardocrat: it was useless resize after specifying minimal chunk size when DB is already filled, now only empty db will be resized on first write

store/src/lmdb.rs
ardocrat: set a name to returned var

store/src/lmdb.rs
ardocrat: set a name to returned var

store/src/lmdb.rs
ardocrat: Need to brind back separate thread when waiting resizing and closing all txs

ardocrat: Reverted back same logic with separate thread

PR #3861, opened 2026-06-14
Validate bitmap segment indices on decode

wiesche89:
## Description
This fixes a small decode edge case in bitmap segments.
When reading sparse bitmap blocks, we now check that each decoded bit index is within the size of the block before calling BitVec::set(). Invalid data returns CorruptedData instead of panicking.

This does not change valid bitmap segment encoding.

## Tests

Added a test for invalid positive and negative bitmap indices.


ardocrat (approved):

ardocrat:
There are many places at this file where panic can potentially happen (not related to PR title), so we should get rid not only from unwraps, but from assertions and also at external libs.

DavidBurkett (approved):

Issue #3848, opened 2026-05-28
grin seedcheck is not working on Windows (is working on Linux)


Anynomouss:
**Describe the bug**
Running ./grin seedcheck is not returning any result. 

**To Reproduce**
1. rung ./grin seedcheck
2. Expect a json printed with the results per seed

**Relevant Information**
No results are returned. The callchain for seedcheck is:
```
grin.rs (main)
  → check_seeds(is_testnet), location src/bin/tools/seedcheck.rs
    → resolve_dns_to_addrs(dns_records)
      → dns.to_socket_addrs() (OS-dependent DNS resolution)
    → check_seed_health(addr, is_testnet, peers)
      → TcpStream::connect_timeout(&addr.0, Duration::from_secs(5)) (TCP connection)
      → Peer::connect(...) (Handshake)
```
Key Steps in check_seeds:

Determine Seed List:

If is_testnet is true, use TESTNET_DNS_SEEDS and port 13414.
Otherwise, use MAINNET_DNS_SEEDS and port 3414.
Sets the global chain type to Testnet if is_testnet is true.

# Initialize P2P Components:

1) Creates a default P2PConfig.
Initializes a Dummy Adapter and Peers instance for peer management.
2) Iterate Over Seed Domains:
3) For each seed domain, it:
a) Resolves the DNS entry to IP addresses using resolve_dns_to_addrs.
b) If no DNS entries are found, it logs a failure and skips to the next seed.
4) Check Seed Health: For each resolved IP address, it calls check_seed_health to attempt a connection and handshake.
5) Cleanup: Removes the temporary directory .\__grintmp__\ used for peer storage.

Since no Failures are logged, DNS resolving to IP is likely not the issue. Potential issues are handling of file path separator for the temporary file created. .`\__grintmp__\` is hardcoded with forward slashes (/), while Windows uses  back slash for path separator. I will do a quick test if this fixes the issue.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro
 - Version [e.g. 1.0.1]

wiesche89:
Done with PR #3862. Build staging branch

PR #3862, opened 2026-06-15
Improve seedcheck diagnostics and Windows cleanup

wiesche89:
This pr improves the seedcheck command a bit.
It adds support for checking a single DNS seed with --seed, includes connection/handshake errors in the JSON output.

Test commands:
.\target\release\grin.exe seedcheck --testnet --output testnet-seedcheck.json
.\target\release\grin.exe seedcheck --output mainnet-seedcheck.json

.\target\release\grin.exe seedcheck --testnet --seed testnet.grin-coffee.com --output testnet-grin-coffee-seedcheck.json
.\target\release\grin.exe seedcheck --seed mainnet.grin-coffee.com --output mainnet-grin-coffee-seedcheck.json

fix issue:
#3848

### DIFF
--- a/src/bin/tools/seedcheck.rs
+++ b/src/bin/tools/seedcheck.rs
@@ -131,15 +131,12 @@ pub fn check_seeds(is_testnet: bool, seed: Option<&str>) -> Vec<SeedCheckResult>
 		let resolved_dns_entries = resolve_dns_to_addrs(&vec![format!("{}:{}", s, port)]);
 		if resolved_dns_entries.is_empty() {
 			info!("FAIL - No dns entries found for {}", s);
-			eprintln!("  FAIL dns: no records found");
 			result.push(seed_result);
 			continue;
 		}
 		seed_result.dns_resolutions_found = true;
-		eprintln!("  DNS records: {}", resolved_dns_entries.len());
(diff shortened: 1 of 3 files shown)


ardocrat (approved):

src/bin/tools/seedcheck.rs
```diff
@@ -160,8 +182,16 @@ pub fn check_seeds(is_testnet: bool) -> Vec<SeedCheckResult> {
 		result.push(seed_result);
 	}
 
+	drop(peers);
+
 	// Clean up temporary files
-	fs::remove_dir_all(".__grintmp__").expect("Unable to delete temporary files");
+	if let Err(e) = fs::remove_dir_all(tmp_root) {
```
ardocrat: For some reason empty `.__grintmp__` stays after execution in the roof of project when we execute `target/release/grin`.

wiesche89: There was one edge case with parallel seedcheck runs. Each process uses its own seedcheck-pid temp directory, but cleanup removed the shared .__grintmp__ root. This meant one finished run could delete another still-running run’s temp directory. The fix deletes only the current run’s PID directory first, then removes .__grintmp__ only if it is empty.

PR #3864, opened 2026-06-16
p2p: compare peers ports to allow running multiple nodes from single IP

ardocrat:
- Allow to launch several nodes under single non-global IP.
- Do not store private addresses at db
- Set default chain port for peers on parsing error from config
- Ignore port for deny/allow peers when `:0` port specified

Open problem: if you run node behind a proxy, it will use local ip address for peers as native socket API is spreading first available address, we just use port from config at handshake.


p2p/src/types.rs
```diff
@@ -219,37 +226,89 @@ impl std::hash::Hash for PeerAddr {
...
 			self.0 == other.0
 		} else {
 			self.0.ip() == other.0.ip()
 		}
 	}
 }
 
+/// Check if IP address is private.
+/// Implementation taken from `core::net:ip_addr` while `is_global` is unstable.
+fn is_private_ip(ip: &IpAddr) -> bool {
```
wiesche89: Don't gossip or store non routable peer addresses, otherwise private addresses can leak into peer exchange as unreachable noise.

config/src/config.rs
```diff
@@ -363,6 +361,26 @@ impl GlobalConfig {
 			.replace("WARN", "Warning")
 			.replace("ERROR", "Error")
 	}
+
+	/// Generate random p2p port for provided chain type.
+	/// Return default after 5 unsuccessful attempts.
+	fn p2p_port_for_chain(chain_type: &global::ChainTypes) -> u16 {
+		let default_host = "::";
+		for _ in 0..5 {
+			let port = thread_rng().gen_range(1024, 49151);
```
wiesche89: I think it would be cleaner to keep the configured port as the first choice and only fall back when the actual bind fails with address in use. The listener should be kept and the real bound port should be advertised. That preserves the standard mainnet/testnet ports and avoids the check then bind race.

ardocrat: I will revert this commit, cause Docker can have problems when we are not using `network_mode: host`.

p2p/src/store.rs
```diff
@@ -233,6 +233,13 @@ impl<'a> PeersIterBatch<'a> {
 		Ok(peers)
 	}
 
+	/// Delete a peer by provided address.
+	pub fn delete_peer(&mut self, peer_addr: PeerAddr) -> Result<(), Error> {
```
wiesche89: delete_peer never commits its batch

p2p/src/peers.rs
```diff
@@ -180,6 +185,12 @@ impl Peers {
 		// check if peer exist
 		self.get_peer(peer_addr)?;
 		if self.is_banned(peer_addr) {
+			// delete banned private peer
+			if is_private_ip(&peer_addr.0.ip()) {
+				if let Ok(mut batch) = self.store.iter_batch() {
+					batch.delete_peer(peer_addr)?;
```
wiesche89: The delete here doesn't take effect. see comment below

p2p/src/peers.rs
```diff
@@ -180,6 +185,12 @@ impl Peers {
...
+				}
+			}
 			self.update_state(peer_addr, State::Healthy)
```
wiesche89: After deleting we set the peer to Healthy, but Healthy is exactly what gets gossiped

ardocrat: fixed

p2p/src/peers.rs
ardocrat: fixed

p2p/src/store.rs
ardocrat: fixed

p2p/src/peers.rs
```diff
@@ -86,9 +86,14 @@ impl Peers {
 				peers.insert(peer_data.addr, peer);
 			}
 		}
-		debug!("Saving newly connected peer {}.", peer_data.addr);
-		if let Err(e) = self.save_peer(&peer_data) {
-			error!("Could not save connected peer address: {:?}", e);
+		// Do not save private peer.
+		if !is_private_ip(&peer_data.addr.0.ip()) {
```
wiesche89, line 90: I think this changes an assumption used by ban_peer(). Private peers are no longer saved here, but ban_peer() still starts by updating the peer in the store. If a connected private peer later sends bad data, the ban may fail before the peer is actually disconnected. Could we make ban_peer() handle missing store entries for private peers?

p2p/src/peers.rs
wiesche89, line 123: Connected private peers are no longer stored, but private peers that fail the handshake still get stored through add_banned(). Is that intentional, or should the private-address filtering live in one central save path?

p2p/src/types.rs
```diff
@@ -219,37 +226,89 @@ impl std::hash::Hash for PeerAddr {
 }
 
 impl PartialEq for PeerAddr {
-	/// If loopback address then we care about ip and port.
+	/// If private address then we care about ip and port.
 	/// If regular address then we only care about the ip and ignore the port.
 	fn eq(&self, other: &PeerAddr) -> bool {
-		if self.0.ip().is_loopback() {
+		if is_private_ip(&self.0.ip()) {
```
wiesche89: This makes private peer identity depend on the port as well as the IP. That seems useful for multiple local/private nodes, but it also changes a few existing assumptions: inbound peers can influence the advertised port, duplicate checks may behave differently because the final peer address is resolved after the handshake, and peers_allow/peers_deny now require an exact port match for private addresses. Could we add tests for these cases, or document the intended behavior clearly?

p2p/src/peers.rs
ardocrat: only banned private peers can be stored, they are getting deleted on unban to do not send this info to other peers

p2p/src/peers.rs
ardocrat: I will check for missing records on ban

p2p/src/types.rs
ardocrat: Added tests

p2p/src/handshake.rs
```diff
@@ -215,7 +215,7 @@ impl Handshake {
 		// so check if we are configured to explicitly allow or deny it.
 		// If denied then we want to close the connection
 		// (without providing our peer with any details why).
-		if Peer::is_denied(&self.config, peer_info.addr) {
+		if Peer::is_denied(&self.config, hand.sender_addr) {
```
wiesche89: Should this stay on the resolved address here? hand.sender_addr comes from the peer, so for inbound connections it feels safer to keep using peer_info.addr, real TCP IP, but still the advertised port.

ardocrat: then for private peers this check will not work, cause for them we are using port for comparison, here it will come as random, will revert this change.

p2p/src/types.rs
wiesche89: Do we still need a fixed limit for private inbound peers?


Private allow/deny entries now also depend on the port. Could we add a short note for operators?

ardocrat: > Private allow/deny entries now also depend on the port. Could we add a short note for operators?

Will add comment here, I guess operators used it mostly for public IPs

p2p/src/types.rs
```diff
@@ -219,37 +226,89 @@ impl std::hash::Hash for PeerAddr {
...
+				&& ip.octets()[3] != 9 && ip.octets()[3] != 10
+			// this address is part of the Shared Address Space defined in
+			// [IETF RFC 6598] (`100.64.0.0/10`).
+			|| ip.octets()[0] == 100 && (ip.octets()[1] & 0b1100_0000 == 0b0100_0000)
+		)
+		}
+		IpAddr::V6(ip) => {
+			ip.is_loopback() || ip.is_unspecified()
+			// IPv4-mapped Address (`::ffff:0:0/96`)
+			|| matches!(ip.segments(), [0, 0, 0, 0, 0, 0xffff, _, _])
```
wiesche89, line 309: This looks risky for default "::" listeners. Inbound IPv4 peers may arrive as IPv4-mapped IPv6 ::ffff:a.b.c.d, and this marks them as private.

That would make public IPv4 inbound peers port-sensitive, skip persistence/gossip, and weaken same-IP dedup because the ephemeral source port becomes part of the identity.

Could we normalize IPv4-mapped IPv6 to IPv4 before is_private_ip?

p2p/src/types.rs
wiesche89: One concrete case here. Ban checks can key the same inbound peer differently.
Pre handshake checks use the socket source port, while the final peer addr uses the advertised port. For private or mapped IPv4 peers this means a banned peer can reconnect with a different source/advertised port and miss the stored ban.

Could we add a post handshake ban check on the resolved/canonical peer addr?

wiesche89: For private peers, allow/deny/ban now depends on the advertised port. So 192.168.x.y:3414 will not match if the same peer advertises another port.

If this is intentional, I think it needs a direct test and operator note.

p2p/src/peers.rs
```diff
@@ -328,7 +352,11 @@ impl Peers {
 
 	/// Saves updated information about mulitple peers in batch
 	pub fn save_peers(&self, p: Vec<PeerData>) -> Result<(), Error> {
-		self.store.save_peers(p).map_err(From::from)
+		let public_peers = p
+			.iter()
+			.filter(|p| !is_private_ip(&p.addr.0.ip()))
```
wiesche89, line 362: This changes the store key for existing non loopback private peer rows from IP only to ip:port.
Do we need a small migration/prune for old private rows? Otherwise they may no longer be found by get_peer, but can still remain visible in full scans.

config/src/comments.rs
```diff
@@ -295,9 +295,9 @@ fn comments() -> HashMap<String, String> {
 #seeds = [\"192.168.0.1:3414\",\"192.168.0.2:3414\"]
 
 #hardcoded peer lists for allow/deny
-#will *only* connect to peers in allow list
+#will *only* connect to peers in allow list, ports will be ignored only for public IPs
```
wiesche89: Since this PR changes port semantics, should invalid allow/deny entries fail loudly? A port-less literal like "1.2.3.4" appears to be silently dropped during parsing instead of denying that IP.

ardocrat: We need to check if port is 0 at these lists then to block/allow by IP.

p2p/src/peers.rs
ardocrat: They will be deleted after 7 days on cleanup at `peers_monitor` as non-active.

p2p/src/types.rs
ardocrat: Need to call `to_ipv4_mapped` https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#method.to_ipv4_mapped.

ardocrat: Fixed

config/src/comments.rs
ardocrat: Currently peer with empty port at config is getting ignored at all (not parsed).

ardocrat: Will add default current chain port if peer form `seeds`, `peers_preferred`, `peers_deny` or `peers_allow` list specified without port

p2p/src/peer.rs
```diff
@@ -152,8 +152,12 @@ impl Peer {
 	}
 
 	pub fn is_denied(config: &P2PConfig, peer_addr: PeerAddr) -> bool {
+		// Check if IP and non-empty port of peers are equal.
+		let peer_filter = |p: &PeerAddr| {
+			p.0.ip() == peer_addr.0.ip() && (p.0.port() == peer_addr.0.port() || p.0.port() == 0)
```
wiesche89: Just to clarify: should public IPs ignore ports by default, or only when config uses :0? The config comment sounds like public IP ports are always ignored, but this only ignores them for :0.

p2p/src/types.rs
```diff
@@ -185,12 +192,33 @@ impl<'de> Visitor<'de> for PeerAddrs {
...
+					eprintln!(
+						"Address {} parse error, trying to resolve DNS: {}",
+						entry, e
+					);
+					let socket_addrs: Result<std::vec::IntoIter<SocketAddr>, M::Error> = {
+						match entry.to_socket_addrs() {
+							Ok(r) => Ok(r),
+							Err(_) => (
+								entry,
+								if global::is_testnet() {
```
wiesche89: since this runs during config loading, can we rely on global::is_testnet() here yet? Maybe the default port should come from the chain/config being loaded.

p2p/src/types.rs
```diff
@@ -185,12 +192,33 @@ impl<'de> Visitor<'de> for PeerAddrs {
...
+								.map_err(|e| {
+									let err_msg =
+										format!("Unable to resolve DNS for {}: {}", entry, e);
+									eprintln!("{}", err_msg);
+									serde::de::Error::custom(err_msg)
+								}),
+						}
+					};
 					if let Ok(socket_addrs) = socket_addrs {
+						println!("resolved DNS for {:?}", socket_addrs);
```
wiesche89: Errors on stderr make sense before logging is initialized. But should successful DNS resolves print to stdout too? That feels more like debug output.

p2p/src/peer.rs
ardocrat: Fixed, comparing non-private peers only by IP.

p2p/src/types.rs
ardocrat: Fixed by initializing global chain type before config parse.

p2p/src/types.rs
ardocrat: I guess it can spam output with a lot peers

src/bin/grin.rs
```diff
@@ -97,6 +97,9 @@ fn real_main() -> i32 {
 		global::ChainTypes::Mainnet
 	};
 
+	// Initialize our global chain_type.
+	global::init_global_chain_type(chain_type);
```
wiesche89: This fixes the config parsing issue, but it also changes the source of truth for chain_type. global is now initialized from the CLI flags before the config is loaded, while server.rs still picks the genesis from config.chain_type. So a testnet config loaded from the working directory or via --config_file without --testnet can run with testnet genesis but mainnet globals/consensus params/seeds. Could we add a hard mismatch check after config load, or avoid making the PeerAddrs deserializer depend on global::is_testnet()?

p2p/src/types.rs
```diff
@@ -222,37 +250,100 @@ impl std::hash::Hash for PeerAddr {
...
 		PeerAddr(SocketAddr::new(addr, port))
 	}
 
-	/// If the ip is loopback then our key is "ip:port" (mainly for local usernet testing).
-	/// Otherwise we only care about the ip (we disallow multiple peers on the same ip address).
+	/// If the ip is private then our key is "ip:port".
+	/// Otherwise, we only care about the ip (we disallow multiple peers on the same ip address).
 	pub fn as_key(&self) -> String {
-		if self.0.ip().is_loopback() {
+		if is_private_ip(&self.0.ip()) {
```
wiesche89, line 387: Small consistency concern for local/private peers. Now that connected peer identity is ip:port, ban entries can end up keyed by different ports depending on the path. Pre handshake checks see the TCP source port, while manual/internal bans usually use the advertised listen port. That may make local bans less predictable and can leave stale rows from failed handshakes. Should ban identity for private IPs be handled separately from connected peer identity?

p2p/src/types.rs
wiesche89: Two edge cases here. Usernet falls into the mainnet-port branch even though USERNET_PEER_PORT exists, and portless private entries get a concrete default port. With the new port aware private matching, peers_deny = ["192.168.1.5"] becomes :3414 and no longer denies the same LAN node on another port. Would it be safer to map portless private entries to :0 as the wildcard, and use a full ChainTypes match for defaults?

p2p/src/peer.rs
```diff
@@ -152,8 +152,17 @@ impl Peer {
 	}
 
 	pub fn is_denied(config: &P2PConfig, peer_addr: PeerAddr) -> bool {
+		// Check if IP and non-empty port of peers are equal.
+		let peer_filter = |p: &PeerAddr| {
+			let same_ip = p.0.ip() == peer_addr.0.ip();
+			if is_private_ip(&p.0.ip()) {
+				same_ip && (p.0.port() == peer_addr.0.port() || p.0.port() == 0)
```
wiesche89: The :0 wildcard only exists in this local filter. Other deny filtering, especially in seed.rs via PeerAddr::eq /difference, does not know about it, so a :0 deny entry can be rejected at handshake but still selected for dialing. Could we centralize this matching in PeerAddrs/PeerAddr and use the same semantics everywhere?

p2p/tests/peer_handshake.rs
```diff
@@ -83,32 +88,98 @@ fn peer_handshake() {
...
+	let test_dir = "target/peer_handshake";
+	clean_output_dir(test_dir);
 
-	let server_peer = server.peers.get_connected_peer(my_addr).unwrap();
-	assert_eq!(server_peer.info.total_difficulty(), Difficulty::min_dma());
-	assert!(server.peers.iter().connected().count() > 0);
+	// Start peers and connect to check handshake, checking ping/pong exchange.
+	{
+		let (_, server) = p2p_server(test_dir, vec![], vec![], None);
+		let (peer_addr, _) = p2p_server(test_dir, vec![], vec![], Some(5000));
```
wiesche89: These fixed ports make the test fragile. Port 5000 is already occupied on my macOS machine, so this test would fail for reasons unrelated to the PR. Since open_port() already exists, could we allocate the ports up front and use those values in the allow/deny lists?

src/bin/grin.rs
ardocrat: Will add check after config load, when --config_file specified it must be a priority for sure, but its hard to setup peers without ports without global setup

wiesche89: fixed, please review

p2p/src/types.rs
wiesche89: fixed, please review

p2p/src/types.rs
wiesche89: fixed, please review.

p2p/src/peer.rs
wiesche89: fixed, please review

p2p/tests/peer_handshake.rs
wiesche89: fixed, please review

wiesche89 (approved):

src/bin/grin.rs
```diff
@@ -89,26 +89,44 @@ fn real_main() -> i32 {
...
 		if let ("config", Some(_)) = server_args.subcommand() {
-			cmd::config_command_server(&chain_type, SERVER_CONFIG_FILE_NAME);
+			global::init_global_chain_type(cli_chain_type);
+			cmd::config_command_server(&cli_chain_type, SERVER_CONFIG_FILE_NAME);
 			return 0;
 		}
 	}
 
+	let config_file_path = match args.subcommand() {
+		("server", Some(server_args)) => server_args.value_of("config_file"),
```
ardocrat, line 112: Here config_path can be reused from the above (added commit).

src/bin/grin.rs
```diff
@@ -89,29 +89,50 @@ fn real_main() -> i32 {
...
+			panic!(
+				"Chain type mismatch: CLI selected {:?}, but configuration selected {:?}",
+				cli_chain_type, chain_type
+			);
+		}
+	}
+
+	// Initialize our global chain_type.
+	global::init_global_chain_type(chain_type);
+	global::set_local_chain_type(chain_type);
```
ardocrat: Seems like its not needed as we set it while getting `chain_type`:

Image

p2p/src/store.rs
```diff
@@ -158,10 +158,18 @@ impl PeerStore {
...
+			return peer;
+		}
+		let ban_key = peer_addr.as_ban_key();
+		if ban_key == key {
+			return peer;
+		}
+		self.get_banned_peer(peer_addr)
 	}
 
 	pub fn get_banned_peer(&self, peer_addr: PeerAddr) -> Result<PeerData, Error> {
```
ardocrat: This function just returns a peer with IP as key, will be renamed

p2p/src/types.rs
```diff
@@ -209,55 +206,195 @@ impl<'de> Deserialize<'de> for PeerAddrs {
...
+			same_ip && (self.0.port() == peer_addr.0.port() || self.0.port() == 0)
 		} else {
-			format!("{}", self.0.ip())
+			same_ip
 		}
 	}
+
+	/// Ban entries are IP-wide so pre-handshake checks do not depend on source ports.
+	pub fn as_ban_key(&self) -> String {
```
ardocrat: This function just returns a peer, ignoring port, will be renamed

p2p/src/store.rs
```diff
@@ -133,32 +133,72 @@ impl PeerStore {
...
 			|| format!("Peer at address: {}", peer_addr),
+		);
+		if peer.is_ok() {
+			return peer;
+		}
+		let ban_key = peer_addr.as_ban_key();
+		if ban_key == key {
+			return peer;
+		}
+		self.get_banned_peer(peer_addr)
```
ardocrat: This call repeats same logic as we try to get peer by key 2 times if selection only by ip fails.

p2p/src/store.rs
```diff
@@ -133,32 +133,72 @@ impl PeerStore {
...
+	}
+
+	pub fn get_banned_peer(&self, peer_addr: PeerAddr) -> Result<PeerData, Error> {
```
ardocrat: This call just select peer by IP, must be renamed.

p2p/src/store.rs
```diff
@@ -133,32 +133,72 @@ impl PeerStore {
 		debug!("save_peer: {:?} marked {:?}", p.addr, p.flags);
 
 		let mut batch = self.db.batch()?;
-		let key = p.addr.as_key();
+		let key = if p.flags == State::Banned {
+			p.addr.as_ban_key()
```
ardocrat: Banned peer can have port here, I will do single selection point at get_peer

p2p/src/store.rs
```diff
@@ -178,7 +218,38 @@ impl PeerStore {
 			peer.last_attempt = Utc::now().timestamp();
 		}
 
-		batch.put_ser(Some(PEER_PREFIX), key.as_bytes(), &peer)?;
+		let new_key = if new_state == State::Banned {
+			peer_addr.as_ban_key()
```
ardocrat: Banned peer can have port here, I will do single selection point at get_peer

p2p/src/store.rs
```diff
@@ -251,7 +322,11 @@ impl<'a> PeersIterBatch<'a> {
 		// Delete peers in single batch
 		if !to_remove.is_empty() {
 			for peer in to_remove {
-				let key = peer.addr.as_key();
+				let key = if peer.flags == State::Banned {
+					peer.addr.as_ban_key()
```
ardocrat: Banned peer can have port here, I will do single selection point at get_peer

p2p/src/store.rs
```diff
@@ -133,32 +133,72 @@ impl PeerStore {
...
 	}
 
-	pub fn save_peers(&self, p: Vec<PeerData>) -> Result<(), Error> {
+	pub fn save_peers(&self, p: Vec<&PeerData>) -> Result<(), Error> {
 		let mut batch = self.db.batch()?;
 		for pd in p {
 			debug!("save_peers: {:?} marked {:?}", pd.addr, pd.flags);
-			let key = pd.addr.as_key();
+			let key = if pd.flags == State::Banned {
+				pd.addr.as_ban_key()
```
ardocrat: Banned peer can have port here, I will do single selection point at get_peer

p2p/src/store.rs
```diff
@@ -133,32 +133,72 @@ impl PeerStore {
...
 		let key = peer_addr.as_key();
-		self.db.exists(Some(PEER_PREFIX), key.as_bytes())
+		if self.db.exists(Some(PEER_PREFIX), key.as_bytes())? {
+			return Ok(true);
+		}
+		let ban_key = peer_addr.as_ban_key();
+		if ban_key == key {
+			return Ok(false);
+		}
+		self.db.exists(Some(PEER_PREFIX), ban_key.as_bytes())
```
ardocrat: I will make this function to return key, so we can update/delete with right key

p2p/src/store.rs
```diff
@@ -133,32 +133,72 @@ impl PeerStore {
...
+			p.addr.as_ban_key()
+		} else {
+			p.addr.as_key()
+		};
 		batch.put_ser(Some(PEER_PREFIX), key.as_bytes(), p)?;
 		batch.commit()
 	}
 
-	pub fn save_peers(&self, p: Vec<PeerData>) -> Result<(), Error> {
+	pub fn save_peers(&self, p: Vec<&PeerData>) -> Result<(), Error> {
```
ardocrat, line 148: We call this function when we receive a list of peers, state can not be banned here and seems like we are losing Banned state here, will add a check for existed peer here

p2p/src/store.rs
```diff
@@ -133,52 +133,102 @@ impl PeerStore {
...
 		let mut batch = self.db.batch()?;
 		for pd in p {
-			debug!("save_peers: {:?} marked {:?}", pd.addr, pd.flags);
-			let key = pd.addr.as_key();
-			batch.put_ser(Some(PEER_PREFIX), key.as_bytes(), &pd)?;
+			if let Ok((peer, key)) = self.get_peer(pd.addr) {
```
wiesche89, line 151: I think we may lose a small recovery path here. If a peer is already stored as Defunct and later shows up again through a peer list, peer_addrs_received() tries to move it back to Unknown. With this skip it stays Defunct. Could we preserve Banned entries specifically, but still allow Defunct Unknown?

p2p/src/store.rs
```diff
@@ -133,52 +133,102 @@ impl PeerStore {
...
 		peer.flags = new_state;
 		if new_state == State::Banned {
 			peer.last_banned = Utc::now().timestamp();
 		} else {
 			peer.last_attempt = Utc::now().timestamp();
 		}
 
-		batch.put_ser(Some(PEER_PREFIX), key.as_bytes(), &peer)?;
+		let mut batch = self.db.batch()?;
+		let new_key = peer_addr.as_key();
```
wiesche89, line 214: Small migration concern, we write this row under the new key, but the PeerData still keeps the old peer.addr. If a private peer changed ports, the DB key and the stored address can drift apart. Would it be simpler to set peer.addr = peer_addr before saving?

p2p/src/peers.rs
```diff
@@ -86,16 +86,29 @@ impl Peers {
...
+		} else {
+			debug!("Do not save connected private peer {}.", peer_data.addr);
 		}
 		Ok(())
 	}
 
 	/// Add a peer as banned to block future connections, usually due to failed
 	/// handshake
 	pub fn add_banned(&self, addr: PeerAddr, ban_reason: ReasonForBan) -> Result<(), Error> {
+		// Avoid banning every local/LAN node behind the same private IP.
```
wiesche89, line 104: One note on the new private peer behavior, a ban for a LAN peer now only takes effect after the handshake, not at connect time. That makes sense for supporting multiple local nodes, but a banned local peer can still force repeated handshakes. Could we check the advertised address once after the handshake, or at least document this tradeoff?


I do not think the ban is ineffective here. The tradeoff is that for private peers we cannot match the persisted ip:advertised_port ban before the handshake, because we only know the TCP source port at that point, so the ban takes effect after the advertised address is known.

p2p/src/store.rs
ardocrat: Unknown comes from peers, so filter to update only Defunct added, also added deletion of old record with ip-only key if exists.

p2p/src/store.rs
ardocrat: added

p2p/src/peers.rs
ardocrat: we can check banned peer just after handshake as we do with denied, not only on message consume

ardocrat: we can check banned peer just after handshake as we do with denied, not only on message consume

ardocrat: check_undesirable takes an address from the stream, not after handshake

p2p/src/msg.rs
```diff
@@ -573,15 +573,17 @@ impl PeerAddrs {
 		self.peers.as_slice()
 	}
 
+	/// Match using peer-filter rules, not exact set membership.
 	pub fn contains(&self, addr: &PeerAddr) -> bool {
```
wiesche89: contains() is no longer exact set membership, it now applies peer filter matching, including port wildcard / port ignoring. Maybe matches_addr() or similar would make this harder to misread at future call sites.

p2p/src/store.rs
```diff
@@ -133,52 +133,106 @@ impl PeerStore {
...
-	pub fn save_peers(&self, p: Vec<PeerData>) -> Result<(), Error> {
+	pub fn save_peers(&self, p: Vec<&PeerData>) -> Result<(), Error> {
 		let mut batch = self.db.batch()?;
 		for pd in p {
-			debug!("save_peers: {:?} marked {:?}", pd.addr, pd.flags);
-			let key = pd.addr.as_key();
-			batch.put_ser(Some(PEER_PREFIX), key.as_bytes(), &pd)?;
+			if let Ok((peer, key)) = self.get_peer(pd.addr) {
+				if peer.flags == State::Defunct {
+					let new_key = peer.addr.as_key();
```
wiesche89: Could we derive the new key from pd.addr here? The current reachable path works, but migration or recovery is easier to reason about if the old key comes from get_peer() and the new key consistently belongs to the PeerData being stored.

p2p/src/msg.rs
ardocrat: renamed

p2p/src/store.rs
ardocrat: fixed

PR #3865, opened 2026-06-16
Optimize LMDB migration size check

ardocrat:
Optimize needed size calculation to avoid possible f32 precision loss.

### DIFF
--- a/store/src/lmdb.rs
+++ b/store/src/lmdb.rs
@@ -337,7 +337,10 @@ impl Store {
 		let to_map_size = self.env.info().map_size;
 
 		// Leave headroom so the migrated env is not immediately above the resize threshold.
-		let required = ((to_used + from_used) as f32 / RESIZE_MIN_TARGET_PERCENT) as usize;
+		let required = ((to_used.saturating_add(from_used) as u128 * 100
+			+ (RESIZE_MIN_TARGET_PERCENT * 100.0 - 1.0) as u128)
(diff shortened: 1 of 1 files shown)


store/src/lmdb.rs
```diff
@@ -337,7 +337,10 @@ impl Store {
 		let to_map_size = self.env.info().map_size;
 
 		// Leave headroom so the migrated env is not immediately above the resize threshold.
-		let required = ((to_used + from_used) as f32 / RESIZE_MIN_TARGET_PERCENT) as usize;
+		let required = ((to_used.saturating_add(from_used) as u128 * 100
+			+ (RESIZE_MIN_TARGET_PERCENT * 100.0 - 1.0) as u128)
```
DavidBurkett: This still uses floating point arithmetic. The correct fix using `RESIZE_MIN_TARGET_PERCENT` would be:
```
let used = to_used.saturating_add(from_used) as u128;
let required = ((used * 100 + RESIZE_MIN_TARGET_PERCENT - 1) / RESIZE_MIN_TARGET_PERCENT).min(usize::MAX as u128) as usize;
```

ardocrat: `RESIZE_MIN_TARGET_PERCENT`  is still `f32`, we can not add we can not add `f32` to `u128`. It means we should change RESIZE_MIN_TARGET_PERCENT to 65

DavidBurkett (approved):

PR #3868, opened 2026-06-16
Fix TUI server startup shutdown handling

DavidBurkett:
- stop the server when the TUI exits through non-`q` paths
- join the async startup thread and drain any queued initialized server before exiting
- replace active startup dialogs on error/status changes instead of stacking them
- update stale `Server::start` documentation

### DIFF
--- a/src/bin/cmd/server.rs
+++ b/src/bin/cmd/server.rs
@@ -50,7 +50,7 @@ pub fn start_server(
 		let serv_tx_clone = serv_tx.clone();
 		let stop_state = Arc::new(StopState::new());
 		let stop_state_clone = stop_state.clone();
-		thread::spawn(move || {
+		let server_thread = thread::spawn(move || {
 			match Server::start(
 				config,
 				Some(stop_state_clone.clone()),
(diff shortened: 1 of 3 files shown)


src/bin/cmd/server.rs
```diff
@@ -71,6 +71,10 @@ pub fn start_server(
 		});
 		let exit_code = controller.run();
 		stop_state.stop();
+		if let Err(e) = server_thread.join() {
```
wiesche89, line 74: This can make shutdown wait (join() can block shutdown) until server startup finishes, so a slow or stuck startup could still delay exiting.

src/bin/tui/ui.rs
wiesche89: These can panic if the UI receiver is already gone

wiesche89: These can panic if the UI receiver is already gone

src/bin/cmd/server.rs
DavidBurkett: Your concern is valid. But my code is just intended to go back to the behavior we had before https://github.com/mimblewimble/grin/commit/c1455296fb3001afbd3d3a35164aaf1336b478c8 while still keeping Ardocrat's startup progress TUI

src/bin/tui/ui.rs
DavidBurkett: The receiver is `self.ui.ui_rx`, which is owned by the same `UI` struct as `self.ui.ui_tx`. So it shouldn't be a problem. I don't mind replacing them with a defensive `let _ = self.ui.ui_tx.send(...)` though.

DavidBurkett: https://github.com/mimblewimble/grin/pull/3868#discussion_r3424011573

wiesche89 (approved):

PR #3835, opened 2026-05-16
Improve PIHD sync and PIBD segment handling

wiesche89:
# Description
This PR adds PIHD (Parallel Initial Header Download) header segment sync support and hardens PIBD segment handling during initial sync.
It also incorporates the relevant PIBD fixes from @ardocrat #3823, including improved segment request selection, peer height slack handling, timeout handling, and safer retry behavior.

# Main changes:
* Add PIHD header segment sync flow
* Improve PIBD segment selection and validation
* Avoid requesting output/rangeproof/kernel segments before the required bitmap state is available
* Track rejected PIBD segments and handle timed-out peers more defensively
* Improve PIHD/PIBD sync logging for start, abort/fallback, and completion

# Compatibility
Legacy header sync and existing fallback behavior remain available for peers that do not support PIHD/PIBD.

# Testing
Tested locally on testnet with fresh sync runs.
Tested in testnet farm with grafana.
reachable gateway node grinffindor.org with port 23414, please sync :)

Also verified:
* PIHD starts and completes
* PIBD starts and progresses through segment download and validation
* Invalid/unusable PIBD data is rejected earlier
* Fallback paths remain available

# Related Issues

*Fixes #3830
*Partially addresses #3082
*Partially addresses #3289
*Related to #3083
*Related to #3767
*Related to #3292

PR #3875, opened 2026-06-19
Harden PIBD segment retry and recovery

wiesche89:
**Summary**
1/4: PIBD PR #3835 

This PR makes the existing PIBD sync path more resilient.
It improves how we track, retry, and recover PIBD segment requests, especially when peers time out, send bad data, or sync is interrupted and later resumed.

**Changes**
- Track pending PIBD segments per peer
- Retry timed-out or dropped segments more reliably
- Avoid retrying recently rejected bad segments from the same peer
- Increase the segment timeout to 30s
- Improve recovery after interrupted PIBD/header sync
- Discard in-memory prune list changes when rolling back PMMR state


chain/src/types.rs
```diff
@@ -334,6 +384,53 @@ impl SyncState {
...
+		peer_addr: SocketAddr,
+	) -> bool {
+		self.requested_pibd_segments
+			.read()
+			.iter()
+			.any(|i| &i.identifier == id && i.last_peer == Some(peer_addr))
+	}
+
+	/// Mark a requested PIBD segment as rejected for this peer.
+	pub fn reject_pibd_segment_from(&self, id: &SegmentTypeIdentifier, peer_addr: SocketAddr) {
```
ardocrat, line 581: I dont see where we are using this function, maybe missed migration from https://github.com/mimblewimble/grin/pull/3835

servers/src/grin/sync/state_sync.rs
```diff
@@ -377,109 +458,111 @@ impl StateSync {
...
-				.and_then(|(_, addr)| *addr);
+			// Choose a random "most work" peer, excluding peer from stale/retry segment
+			// and preferring outbound if at all possible.
 			let peer = available_pibd_peers()
 				.outbound()
-				.with_filter(|p| !peers.is_blocked(p.info.addr))
-				.exclude(excluded_peer)
+				.with_filter(|p| {
+					!peers.is_blocked(p.info.addr)
+						&& !sync_state.rejected_pibd_segment_from(
```
ardocrat: we check rejecting, but `reject_pibd_segment_from` is never called

servers/src/grin/sync/state_sync.rs
```diff
@@ -77,15 +82,11 @@ impl StateSync {
 	pub fn check_run(
 		&mut self,
 		header_head: &chain::Tip,
-		head: &chain::Tip,
-		tail: &chain::Tip,
+		_head: &chain::Tip,
+		_tail: &chain::Tip,
```
ardocrat: do we really need this variables if they are not used?

ardocrat (commented):
Now it takes ~48mins with 8 peers to download whole state (next time is for validation). Fixed some logs and warnings on build.

chain/src/types.rs
```diff
@@ -334,6 +427,77 @@ impl SyncState {
...
+	) -> bool {
+		let cutoff_time = Utc::now() - Duration::seconds(reject_seconds);
+		let rejected = self.rejected_pibd_segments.read();
+		rejected
+			.iter()
+			.any(|i| &i.identifier == id && i.peer_addr == peer_addr && i.reject_time > cutoff_time)
+	}
+
+	/// Check whether this peer has too many recent PIBD rejections.
+	pub fn rejected_pibd_peer(&self, peer_addr: SocketAddr, reject_seconds: i64) -> bool {
```
ardocrat: Argument `reject_seconds` is constant here, not sure if needs to be passed, also this function is always used with `rejected_pibd_segment_from`, can be united into single function for simplification?

chain/src/types.rs
```diff
@@ -334,6 +427,77 @@ impl SyncState {
...
+				reject_count: 1,
+			});
+		}
+		if rejected_peers.len() > MAX_REJECTED_PIBD_PEERS {
+			rejected_peers.remove(0);
+		}
+	}
+
+	/// Check whether this peer recently provided invalid data for this PIBD segment.
+	pub fn rejected_pibd_segment_from(
```
ardocrat: This function is always used with `rejected_pibd_peer`, logic can be united into single function for simplification?

servers/src/common/adapters.rs
```diff
@@ -705,13 +797,16 @@ where
...
+		let (pibd_segment_tx, pibd_segment_rx) = mpsc::sync_channel(PIBD_SEGMENT_QUEUE_CAP);
+		spawn_pibd_segment_worker(sync_state.clone(), Arc::downgrade(&chain), pibd_segment_rx);
 		NetToChainAdapter {
 			sync_state,
 			chain: Arc::downgrade(&chain),
 			tx_pool,
 			peers: OneTime::new(),
 			config,
 			hooks,
+			pibd_segment_tx,
```
ardocrat: Do you think we should make this as an option and drop on pibd completion, so worker thread can stop properly? as receiver works till last sender lives

chain/src/types.rs
wiesche89: Fixed by combining both checks into rejected_pibd_segment_from_peer() and using REJECTED_SEGMENT_RETRY_SECS internally.

chain/src/types.rs
wiesche89: Fixed, combined into rejected_pibd_segment_from_peer()

servers/src/common/adapters.rs
wiesche89: Yes, you are right. The worker currently lives for the NetToChainAdapter lifetime, so this log only appears when the channel is closed, not when pibd completes. I think it is okay to keep it as is for now. The worker is just waiting and not doing any work after PIBD is done, and it will shut down when the adapter channel is dropped. Stopping it exactly at pibd completion would add extra lifecycle handling, especially if PIBD needs to start again. Or what do you think?
I can adjust the log message here to:
"PIBD receive worker shutting down after channel close"
That would make it clearer that the worker stops when the channel is closed, not when PIBD completes.

ardocrat: I made this worker more universal for future usage.

ardocrat (approved):

Issue #3867, opened 2026-06-16
Stratum server leaks connections


ardocrat:
There are several reports how Stratum server leaks connections:

https://github.com/GetGrin/grim/issues/19

And `Too many open files` on Linux with time.

Rework `stratum.rs` to handle connections properly

PR #3876, opened 2026-06-19
Refactor stratum server leak handling

syntaxjak:
Addressing issue: https://github.com/mimblewimble/grin/issues/3867

- Reworked the Stratum connection handling so reads and writes can run independently
- Added bounded response queues, connection limits, and idle timeouts
- Added an RAII cleanup guard so workers are removed correctly on disconnects and errors
- Stratum connections and listener tasks now stop cleanly when the node shuts down
- Added tests for connection cleanup, limits, backpressure, idle handling, and shutdown
- Removed the unused async-stream dependency

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -17,7 +17,7 @@
 use futures::{SinkExt, StreamExt, TryStreamExt};
 use tokio::net::{TcpListener, TcpStream};
 use tokio::runtime::Runtime;
-use tokio::sync::{mpsc, OwnedSemaphorePermit, Semaphore};
+use tokio::sync::{mpsc, Semaphore};
 use tokio::task::JoinSet;
 use tokio::time::{timeout, Instant};
 use tokio_util::codec::{Framed, LinesCodec};
@@ -639,20 +639,7 @@ impl Drop for WorkerCleanup {
 	}
 }
 
-async fn handle_connection(
-	socket: TcpStream,
-	handler: Arc<Handler>,
-	_permit: OwnedSemaphorePermit,
-	idle_timeout: Duration,
-) {
-	handle_connection_with_idle_timeout(socket, handler, idle_timeout).await;
-}
-
-async fn handle_connection_with_idle_timeout(
-	socket: TcpStream,
-	handler: Arc<Handler>,
-	idle_timeout: Duration,
-) {
+async fn handle_connection(socket: TcpStream, handler: Arc<Handler>, idle_timeout: Duration) {
 	let peer_addr = socket.peer_addr().ok();
 	let (tx, mut rx) = mpsc::channel(WORKER_QUEUE_SIZE);
 	let (shutdown_tx, mut shutdown_rx) = mpsc::channel::<()>(1);
@@ -774,10 +761,11 @@ async fn accept_connections_loop(
 						};
 						let handler = handler.clone();
 						connections.spawn(async move {
+							let _permit = permit;
 							if let Err(e) = socket.set_nodelay(true) {
 								debug!("Stratum: set_nodelay failed for {}: {}", peer_addr, e);
 							}
-							handle_connection(socket, handler, permit, idle_timeout).await;
+							handle_connection(socket, handler, idle_timeout).await;
 						});
 					}
 					Err(e) => {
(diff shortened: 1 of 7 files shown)


iho:
Heads-up: #3889 also targets #3867 with the same structural approach (`handle_connection` + Drop guard + simpler accept loop).

That PR additionally adds idle timeout, max workers, and bounded write queues for the half-open / reconnect-storm failure mode. Worth coordinating so the two don’t land overlapping refactors without combining the resource bounds.

.gitignore
```diff
@@ -13,5 +13,6 @@ wallet_data
 wallet/db
 .idea/
 .vscode/
+.venv/
```
wiesche89: This looks unrelated to the Stratum change. Could we leave it out of this PR?

servers/src/mining/stratumserver.rs
```diff
@@ -599,72 +598,92 @@ impl Handler {
...
+						error!("error reading line: {}", e);
+						break;
+					}
+					None => break,
+				}
+			}
+			outgoing = rx.next() => {
+				match outgoing {
+					Some(line) => {
+						if let Err(e) = framed.send(line).await {
```
wiesche89: While framed.send() is blocked, the socket is no longer read. In my test, this kept a half closed socket alive, while the previous split reader and writer closed it immediately. Could we keep reads and writes independent?

servers/src/mining/stratumserver.rs
```diff
@@ -599,72 +598,92 @@ impl Handler {
...
+
+impl Drop for WorkerCleanup {
+	fn drop(&mut self) {
+		self.workers.remove_worker(self.worker_id);
+		info!("Worker {} disconnected", self.worker_id);
+	}
+}
+
+async fn handle_connection(socket: TcpStream, handler: Arc<Handler>) {
+	let (tx, mut rx) = mpsc::unbounded();
```
wiesche89: Since #3889 is closed, I think the limits should be included here. Idle connections otherwise stay open indefinitely and the unbounded queue can keep growing. Could we add an idle timeout, a connection limit, and a bounded queue?

servers/src/mining/stratumserver.rs
```diff
@@ -599,72 +598,92 @@ impl Handler {
...
-					Ok((socket, _)) => yield socket,
-					Err(e) => {
-						error!("accept error = {:?}", e);
-						continue;
-					}
+		loop {
+			match listener.accept().await {
+				Ok((socket, _)) => {
+					let handler = handler.clone();
+					tokio::spawn(async move {
```
wiesche89: Could we make the listener and connection tasks cancellable and keep their handles? That would also make clean start and stop support through the Owner API possible later.

servers/src/mining/stratumserver.rs
```diff
@@ -599,72 +598,92 @@ impl Handler {
...
 // Worker Factory Thread Function
+
+struct WorkerCleanup {
+	worker_id: usize,
+	workers: Arc<WorkersList>,
+}
+
+impl Drop for WorkerCleanup {
+	fn drop(&mut self) {
+		self.workers.remove_worker(self.worker_id);
```
wiesche89, line 644: The worker is removed from the map, but its worker_stats entry remains forever. Could we remove or reuse these entries during cleanup?

servers/src/mining/stratumserver.rs
```diff
@@ -599,72 +598,92 @@ impl Handler {
...
+		info!("Worker {} disconnected", self.worker_id);
+	}
+}
+
+async fn handle_connection(socket: TcpStream, handler: Arc<Handler>) {
```
wiesche89: Could we add lifecycle tests for this? The existing Stratum tests only cover JSON serialization and never reach this code.

servers/src/mining/stratumserver.rs
```diff
@@ -599,72 +598,92 @@ impl Handler {
...
+			match listener.accept().await {
+				Ok((socket, _)) => {
+					let handler = handler.clone();
+					tokio::spawn(async move {
+						handle_connection(socket, handler).await;
+					});
+				}
+				Err(e) => {
+					error!("accept error = {:?}", e);
+					continue;
```
wiesche89: Could we add a short backoff for accept errors? With EMFILE, this loop would otherwise spin and flood the log.

servers/src/mining/stratumserver.rs
```diff
@@ -599,72 +598,92 @@ impl Handler {
...
 		});
-		let server = async_stream::stream! {
-			loop {
-				match listener.accept().await {
-					Ok((socket, _)) => yield socket,
-					Err(e) => {
-						error!("accept error = {:?}", e);
-						continue;
-					}
+		loop {
```
wiesche89: This was the last use of async stream. Could we also remove the dependency from servers/Cargo.toml?

wiesche89 (commented):
I tested this with a miner and additional connection tests. The cleanup works for normal disconnects, but it does not fully solve the reported connection leak yet. I’ve left a few comments on the remaining cases.

Image
https://github.com/GetGrin/grim/issues/19

PR #3877, opened 2026-06-23
Improve macOS TUI support with crossterm backend

wiesche89:
This PR switches the Grin TUI from the pancurses backend to the crossterm backend.

The goal is to remove the native ncurses dependency and make the TUI easier to build and run on macOS. It also adjusts the shutdown order so the server is stopped before the TUI exits, avoiding leftover terminal output after pressing q.

Tested locally on macOS, the TUI starts and exits cleanly.

Linux and Windows still need testing.


ardocrat (approved):
Works well on Linux, Windows and MacOS

PR #3879, opened 2026-06-26
add PIHD header segment protocol

wiesche89:
2/4: PIBD PR #3835

This adds the P2P message types needed for PIHD header segment exchange.

The PR wires header segment requests and responses through the protocol and adapter layers, adds the related peer capability, and covers the new serialization and capability behavior in tests.

This is not consensus breaking. Peers without the new capability continue using the existing sync paths.

### DIFF
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -16,7 +16,9 @@
 //! events to consumers of those events.
 
 use crate::util::RwLock;
+use std::collections::HashMap;
 use std::fs::File;
+use std::net::SocketAddr;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::path::PathBuf;
 use std::sync::{mpsc, Arc, Weak};
@@ -41,7 +43,7 @@ use crate::core::pow::Difficulty;
 use crate::core::ser::ProtocolVersion;
 use crate::core::{core, global};
 use crate::p2p;
-use crate::p2p::types::PeerInfo;
+use crate::p2p::types::{HeaderSegmentAcceptance, PeerInfo};
 use crate::pool::{self, BlockChain, PoolAdapter};
 use crate::util::secp::pedersen::RangeProof;
 use crate::util::OneTime;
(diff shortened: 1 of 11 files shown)


servers/src/common/adapters.rs
```diff
@@ -393,6 +398,60 @@ where
...
+		if !peer_info
+			.capabilities
+			.contains(p2p::Capabilities::PIHD_HIST)
+			|| id.height != p2p::PIHD_HEADER_SEGMENT_HEIGHT
+		{
+			return Ok(None);
+		}
+		if !self.header_segment_request_allowed(peer_info.addr.0) {
+			warn!(
+				"throttling PIHD header segment request {:?} from {}",
```
ardocrat, line 415: Does it mean we will ignore request from single peer if there are more than 2 segments per sec at 60 sec window to avoid requests spamming?

wiesche89: Yes, exactly. It is just a  anti spam limit. Max 120 header segment requests per peer per 60s, roughly 2 per sec.

servers/src/common/adapters.rs
```diff
@@ -393,6 +398,60 @@ where
...
 
+	fn locate_header_segment(
+		&self,
+		id: SegmentIdentifier,
+		peer_info: &PeerInfo,
+	) -> Result<Option<Vec<core::BlockHeader>>, chain::Error> {
+		if !peer_info
+			.capabilities
+			.contains(p2p::Capabilities::PIHD_HIST)
+			|| id.height != p2p::PIHD_HEADER_SEGMENT_HEIGHT
```
ardocrat, line 409: seems like we send height here as constant always, why should it be variable then?

wiesche89: It should not be variable for pihd. We reuse SegmentIdentifier, so the field is present, but this path only accepts the fixed PIHD_HEADER_SEGMENT_HEIGHT.

So yes, the variable looking part is a bit misleading. I can add a short code comment here?, to make it clear that other heights are intentionally rejected.

ardocrat: Reuse seems fine then, then for PIHD it can be simply ignored as peer can provide any value here and we can not validate if its right height actually?

wiesche89: I think we should keep the check.
The height here is not a block height, but the fixed PIHD segment size. With PIHD_HEADER_SEGMENT_HEIGHT = 9, each segment maps to 512 headers. If we ignore the value from the peer, the same idx could map to a different header range.

So this is mainly a cheap sanity check to make sure both sides use the agreed PIHD segment size.

ardocrat: OK, then we'll need to change protocol version (new capability) if this value will be changed.

servers/src/common/adapters.rs
```diff
@@ -610,6 +669,21 @@ where
 	) -> Result<bool, chain::Error> {
 		self.queue_pibd_segment(PIBDSegment::Kernel(block_hash, segment), peer_info)
 	}
+
+	fn receive_header_segment(
```
ardocrat, line 673: Seems like missing implementation:
https://github.com/mimblewimble/grin/pull/3835/changes#diff-172148fdc6d0a8a940347d1429aa3519410b304d29d8c81a8c9544b5bd2dcf31R827

ardocrat:
If you have a RFC about PIHD, please create a PR to https://github.com/mimblewimble/grin-rfcs

servers/src/common/adapters.rs
ardocrat: * Will be added at next branch as discussed

PR #3881, opened 2026-06-28
Create API secrets at configured paths

wiesche89:
Fixes #3821.

API secret files are now created and checked using the paths from grin-server.toml, instead of defaulting to ~/.grin/"chain" before the config is loaded.
Also applies to explicit --config_file usage.

Test:
* cargo test -p grin_config

Tested on Mac:
./grin server config
rm -f .api_secret .foreign_api_secret
./grin server run
ls -la .api_secret .foreign_api_secret

Config-path:
rm -f .api_secret .foreign_api_secret
./grin server --config_file "$PWD/grin-server.toml" run
ls -la .api_secret .foreign_api_secret

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -94,28 +94,39 @@ pub fn check_api_secret(api_secret_path: &PathBuf) -> Result<(), ConfigError> {
 	Ok(())
 }
 
-/// Check that the api secret files exist and are valid
-fn check_api_secret_files(
-	chain_type: &global::ChainTypes,
-	secret_file_name: &str,
-) -> Result<(), ConfigError> {
-	let grin_path = get_grin_path(chain_type)?;
-	let mut api_secret_path = grin_path;
-	api_secret_path.push(secret_file_name);
+/// Check that the api secret file exists and is valid
+fn check_api_secret_file(api_secret_path: &PathBuf) -> Result<(), ConfigError> {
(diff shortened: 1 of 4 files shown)


config/src/config.rs
```diff
@@ -94,28 +94,39 @@ pub fn check_api_secret(api_secret_path: &PathBuf) -> Result<(), ConfigError> {
...
-		check_api_secret(&api_secret_path)
+		check_api_secret(api_secret_path)
 	}
 }
 
+/// Check that the configured api secret files exist and are valid
+fn check_api_secret_files(config: &GlobalConfig) -> Result<(), ConfigError> {
+	let server_config = &config.members.as_ref().unwrap().server;
+	if let Some(ref api_secret_path) = server_config.api_secret_path {
+		check_api_secret_file(&PathBuf::from(api_secret_path))?;
```
ardocrat: If we specify path in config as `api_secret_path = ".api_secret"` file is getting created at directory from where we are launching binary, should we create/use it from `grin_path/chain_type` instead?

wiesche89: Yes, agreed. Relative secret paths should not depend on the process cwd.
I'll resolve relative api_secret_path and foreign_api_secret_path against the chain-specific grin path and keep absolute paths unchanged.

ardocrat (approved):

PR #3882, opened 2026-06-29
Implement PIHD header sync

wiesche89:
3/4: PR #3835 
Adds PIHD-based header sync using deterministic header segments from PIHD capable peers.

*It requests segments in parallel
*validates and processes them in order
*falls back to legacy header sync on stalls/timeouts 
*exposes the active header sync mode in the TUI and status API.


servers/src/common/adapters.rs
```diff
@@ -769,6 +842,18 @@ where
 			.expect("Failed to upgrade weak ref to our peers.")
 	}
 
+	fn block_bad_header_segment_peer(
```
ardocrat: Not implemented?
https://github.com/mimblewimble/grin/pull/3835/changes#diff-172148fdc6d0a8a940347d1429aa3519410b304d29d8c81a8c9544b5bd2dcf31R945

servers/src/grin/sync/header_sync.rs
```diff
@@ -185,22 +389,143 @@ impl HeaderSync {
...
+				.iter()
+				.find(|peer| {
+					peer.info.height() >= start_height
+						&& self
+							.pending_pihd
+							.iter()
+							.filter(|req| req.peer_addr == peer.info.addr)
+							.count() < pihd_params::MAX_IN_FLIGHT_SEGMENTS_PER_PEER
+				})
+				.or_else(|| {
```
ardocrat, line 539: This code is just doing the same as above (lines 464-471).

servers/src/grin/sync/header_sync.rs
```diff
@@ -63,36 +93,210 @@ impl HeaderSync {
...
 
+	fn cleanup_pending_requests(&mut self, header_head: chain::Tip) {
+		let now = Utc::now();
+		let peers = self.peers.clone();
+		if header_head.height > self.prev_header_sync.1 {
+			self.pihd_failure_count = 0;
+			self.pihd_stalling_ts = None;
+		}
+
+		let mut failed = 0;
```
ardocrat: Seems like we should we make it equal to `self.pihd_failure_count` to count failures form previous loop?

servers/src/common/adapters.rs
wiesche89: This is already implemented via the return value.
This is implemented one layer up in p2p::Peers::receive_header_segment.
block_bad_header_segment_peer() returns HeaderSegmentAcceptance::Ban, and Peers::receive_header_segment() maps that to ban_peer(peer_info.addr, ReasonForBan::BadBlockHeader).
So the peer is banned when this path is hit.

servers/src/grin/sync/header_sync.rs
wiesche89: Good point, thanks. I cleaned this up and removed the duplicated logic.

servers/src/grin/sync/header_sync.rs
wiesche89: Good catch, thanks. I adjusted this so failures from previous cleanup loops are considered, without counting the same failures more than once.

servers/src/common/adapters.rs
ardocrat: Before we just blocked peer like at PIHD inside:
image

it means we can just replace call of this function with `HeaderSegmentAcceptance::Ban` for now

servers/src/grin/sync/header_sync.rs
```diff
@@ -63,36 +93,210 @@ impl HeaderSync {
...
+				.identifier
+				.idx
+				.saturating_mul(req.identifier.segment_capacity())
+				.saturating_add(req.identifier.segment_capacity());
+			let completed_height = completed_height.min(req.target_height);
+			let connected = peers.get_connected_peer(PeerAddr(req.peer_addr)).is_some();
+			let complete = header_head.height >= completed_height;
+			let timeout = now
+				> req.request_time + Duration::seconds(pihd_params::HEADER_REQUEST_TIMEOUT_SECS);
+			!complete && connected && !timeout
```
ardocrat: This logic is the same as above (lines 201-211) except we are adding `failed_peers` above, can we extract this as separate function to avoid code duplicate?

servers/src/common/adapters.rs
wiesche89: okay i understand..
You are right, the name is confusing. The helper is only used to log the reason and return the Ban decision, the actual ban happens in the P2P layer. I’ll rename it to make that clearer.

wiesche89: Fixed

servers/src/grin/sync/header_sync.rs
wiesche89: Fixed, thx

servers/src/grin/sync/header_sync.rs
```diff
@@ -185,22 +386,138 @@ impl HeaderSync {
...
+				None => return,
+			};
+			if peer.send_header_segment_request(identifier).is_ok() {
+				let target_height = peer.info.height();
+				self.sync_state.add_pihd_header_segment(
+					identifier,
+					peer.info.addr.0,
+					target_height,
+				);
+				self.pending_pihd.push(PIHDHeaderSegmentContainer::new(
```
ardocrat: Seems like we can use data from sync_state to not duplicate same data at different parts

servers/src/common/adapters.rs
```diff
@@ -863,6 +934,168 @@ where
...
+			}
+		}
+	}
+
+	fn prune_pihd_header_cache(&self, sync_head: chain::Tip) {
+		let next_idx = sync_head.height / p2p::pihd_header_segment_capacity();
+		let max_idx = next_idx.saturating_add(PIHD_HEADER_CACHE_LOOKAHEAD_SEGMENTS);
+		let clear_cache = {
+			let mut anchor = self.pihd_header_cache_anchor.write();
+			let clear_cache = if let Some(prev) = anchor.as_ref() {
```
ardocrat: This can be simplified:

```
let clear_cache = match anchor.as_ref() {
				None => false,
				Some(prev) => sync_head.height < prev.height
					|| (sync_head.height == prev.height && sync_head.last_block_h != prev.hash)
			};

Comment (wiesche89) [reply servers/src/common/adapters.rs]: fixed

File: servers/src/grin/sync/header_sync.rs [hunk 7]
```diff
@@ -185,22 +392,141 @@ impl HeaderSync {
...
+				.find(|peer| can_request(peer, pihd_params::MAX_IN_FLIGHT_SEGMENTS_PER_PEER))
+				.or_else(|| {
+					peers
+						.iter()
+						.find(|peer| can_request(peer, pihd_params::MAX_IN_FLIGHT_SEGMENTS))
+				}) {
+				Some(peer) => peer.clone(),
+				None => return,
+			};
+			if peer.send_header_segment_request(identifier).is_ok() {
```
ardocrat, line 563: Need to add logging

servers/src/grin/sync/header_sync.rs
```diff
@@ -185,22 +392,145 @@ impl HeaderSync {
...
+		{
+			let identifier = SegmentIdentifier {
+				height: PIHD_HEADER_SEGMENT_HEIGHT,
+				idx: segment_idx,
+			};
+			let start_height = match p2p::pihd_header_segment_start_height(identifier) {
+				Some(height) => height,
+				None => return,
+			};
+			if self.sync_state.contains_pihd_header_segment(identifier) {
```
ardocrat, line 512: segment with provided id can be already requested, but not exists after retaining, so at logs can be seen how segments with same id are getting requested repeatedly and as result responses are getting ignored with `ignoring unsolicited PIHD header segment SegmentIdentifier` message.

p2p/src/types.rs
```diff
@@ -76,6 +76,36 @@ const PEER_MIN_PREFERRED_OUTBOUND_COUNT: u32 = 8;
...
+		height: PIHD_HEADER_SEGMENT_HEIGHT,
+		idx: 0,
+	};
+	let capacity = id.segment_capacity();
+	debug_assert_eq!(capacity, MAX_BLOCK_HEADERS as u64);
+	capacity
+}
+
+/// Segment index containing the next header after the given sync height.
+pub fn next_pihd_header_segment_idx(height: u64) -> u64 {
```
ardocrat, line 91: seems like this function is not giving us exact next segment id, so repeated requests are possible, can we store last requested segment id?

servers/src/grin/sync/header_sync.rs
```diff
@@ -164,6 +162,8 @@ impl HeaderSync {
 			});
 			self.pihd_header_sync(sync_head, pihd_peers);
 			self.syncing_peer = None;
+
+			self.cleanup_pending_requests(sync_head);
```
wiesche89: This cleanup needs to run before scheduling, and also outside PIHD only flow. Otherwise legacy pending requests never timeout, and PIHD stale requests still affect this tick’s scheduling.

wiesche89 (commented):
need revert

servers/src/grin/sync/header_sync.rs
wiesche89: fixed

p2p/src/types.rs
wiesche89: I think the function is fine, the issue was removing next_idx from pending too early. Keeping it pending avoids storing last_requested_segment_id. Thx

servers/src/grin/sync/header_sync.rs
wiesche89: Addressed by keeping the segment pending/responded until it is processed, so the same id is not scheduled again too early.

PR #3883, opened 2026-06-29
Update API dependencies

ardocrat:
- Update to `hyper` 1.x
- Update `rustls`, `hyper-rustls` dependencies
- Code formatting, typos fix
- Get rid of direct `ring` dependency, use default comparison for slices https://doc.rust-lang.org/1.52.1/src/core/slice/cmp.rs.html#69-75


api/src/rest.rs
```diff
@@ -212,81 +231,100 @@ impl ApiServer {
...
+						Ok(l) => loop {
+							tokio::select! {
+								Ok((s, _)) = l.accept() => {
+									let router = router.clone();
+									let tls_acceptor = tls_acceptor.clone();
+									let tls_stream = match tls_acceptor.accept(s).await {
+										Ok(tls_stream) => tls_stream,
+										Err(err) => {
+											error!("failed to perform TLS handshake: {err:#}");
+											return;
```
wiesche89: This return exits the whole TLS api server loop, not just the failed connection. A single bad TLS handshake would stop the https api until restart. We should log the error and continue accepting connections instead.

api/src/rest.rs
```diff
@@ -167,30 +170,46 @@ impl ApiServer {
...
-					server.await
+					let graceful = hyper_util::server::graceful::GracefulShutdown::new();
+					// When this signal completes, start shutdown.
+					let mut signal = std::pin::pin!(shutdown_signal(rx));
+
+					// Start server loop.
+					match TcpListener::bind(addr).await {
+						Ok(l) => loop {
+							tokio::select! {
+								Ok((s, _)) = l.accept() => {
```
wiesche89: We should handle accept() errors explicitly here. With this refutable pattern, an Err can disable the accept branch and leave the server waiting only for shutdown. Better to match on l.accept() and log and continue on errors.

api/src/rest.rs
```diff
@@ -212,81 +231,100 @@ impl ApiServer {
...
-									continue;
+					let graceful = hyper_util::server::graceful::GracefulShutdown::new();
+					// When this signal completes, start shutdown.
+					let mut signal = std::pin::pin!(shutdown_signal(rx));
+
+					// Start server loop.
+					match TcpListener::bind(addr).await {
+						Ok(l) => loop {
+							tokio::select! {
+								Ok((s, _)) = l.accept() => {
```
wiesche89: Same concern here as in the non TLS path, accept() errors should be handled explicitly, otherwise a transient accept error can stop the listener from accepting future connections.

api/src/auth.rs
```diff
@@ -66,11 +68,7 @@ impl Handler for BasicAuthMiddleware {
 			}
 		}
 		if req.headers().contains_key(AUTHORIZATION)
-			&& verify_slices_are_equal(
-				req.headers()[AUTHORIZATION].as_bytes(),
-				&self.api_basic_auth.as_bytes(),
-			)
-			.is_ok()
+			&& req.headers()[AUTHORIZATION].as_bytes() == self.api_basic_auth.as_bytes()
```
wiesche89: This replaces the previous constant time secret comparison with normal slice equality. Since this is an auth path, I think we should keep constant time comparison, using a small alternative like subtle or constant_time_eq instead of ring.

api/src/auth.rs
```diff
@@ -104,23 +102,19 @@ impl BasicAuthURIMiddleware {
...
 			return next_handler.call(req, handlers);
 		}
 		if req.uri().path() == self.target_uri {
 			if req.headers().contains_key(AUTHORIZATION)
-				&& verify_slices_are_equal(
-					req.headers()[AUTHORIZATION].as_bytes(),
-					&self.api_basic_auth.as_bytes(),
-				)
-				.is_ok()
+				&& req.headers()[AUTHORIZATION].as_bytes() == self.api_basic_auth.as_bytes()
```
wiesche89: Same here: this should stay constant time rather than using normal slice equality.

api/src/client.rs
```diff
@@ -215,9 +220,13 @@ where
...
 }
 
-async fn send_request_async(req: Request<Body>, timeout: TimeOut) -> Result<String, Error> {
+async fn send_request_async(
+	req: Request<BoxBody<Bytes, Infallible>>,
+	timeout: TimeOut,
+) -> Result<String, Error> {
 	let https = hyper_rustls::HttpsConnectorBuilder::new()
 		.with_native_roots()
+		.unwrap()
```
wiesche89: Can we avoid unwrapping with_native_roots() here? If loading native roots fails, this should return a normal client error instead of panicking.

servers/src/common/hooks.rs
```diff
@@ -222,13 +227,14 @@ impl WebHook {
 
 		let https = hyper_rustls::HttpsConnectorBuilder::new()
 			.with_native_roots()
+			.unwrap()
```
wiesche89: Same concern here, failing to load native roots should not panic during webhook setup.

api/src/router.rs
```diff
@@ -12,64 +12,69 @@
...
-use std::task::{Context, Poll};
 
 lazy_static! {
 	static ref WILDCARD_HASH: u64 = calculate_hash(&"*");
 	static ref WILDCARD_STOP_HASH: u64 = calculate_hash(&"**");
 }
 
-pub type ResponseFuture =
-	Pin<Box<dyn Future<Output = Result<Response<Body>, hyper::Error>> + Send>>;
+pub type ResponseFuture = Pin<
```
wiesche89: Since this migration introduces BoxBody<Bytes, Infallible> in many signatures, could we add shared aliases like ApiResponse and use them across the API code? It would make the hyper body type easier to change later and reduce a lot of repeated type noise.

api/src/rest.rs
```diff
@@ -167,30 +170,46 @@ impl ApiServer {
...
-						.serve(make_service_fn(move |_| {
-							let router = router.clone();
-							async move { Ok::<_, Infallible>(router) }
-						}))
-						.with_graceful_shutdown(async {
-							rx.await.ok();
-						});
-
-					server.await
+					let graceful = hyper_util::server::graceful::GracefulShutdown::new();
```
wiesche89: The http and https paths now have very similar accept loop logic. Since the error handling needs to stay in sync between both, it might be worth extracting the shared loop shape and keeping only the TLS handshake as the path specific part.

api/src/router.rs
```diff
@@ -333,45 +339,47 @@ mod tests {
...
-					.unwrap();
-				resp.status().as_u16()
-			};
-			block_on(task)
+			let res = client::get::<u16>(format!("http://{}{}", server_addr, url).as_str(), None);
+			thread::sleep(time::Duration::from_millis(500));
+			res.unwrap()
 		};
 
+		println!("1");
```
wiesche89: Looks like leftover debug output. Can we remove these printl calls from the test?

api/src/router.rs
```diff
@@ -333,45 +339,47 @@ mod tests {
...
 		assert_eq!(call_handler("/v1/users"), 101);
-		assert_eq!(call_handler("/v1/users/xxx"), 103);
+
+		println!("2");
```
wiesche89: Looks like leftover debug output. Can we remove these printl calls from the test?

api/src/web.rs
```diff
@@ -126,8 +145,8 @@ impl From<Option<&str>> for QueryParams {
 	}
 }
 
-impl From<Request<Body>> for QueryParams {
-	fn from(req: Request<Body>) -> Self {
+impl From<Request<BoxBody<Bytes, Infallible>>> for QueryParams {
```
wiesche89: This impl looks unused after the hyper migration.

api/src/router.rs
ardocrat: Added `ApiBody` type

api/src/auth.rs
ardocrat: Using subtle now

api/src/rest.rs
ardocrat: Added continue, so exit will happen only on shutdown signal

api/src/rest.rs
ardocrat: Added logging on accept error

api/src/rest.rs
ardocrat: Added logging on accept error

api/src/rest.rs
ardocrat: Moved server creation to single function

api/src/rest.rs
```diff
@@ -212,81 +194,124 @@ impl ApiServer {
...
+	let server = async move {
+		let graceful = hyper_util::server::graceful::GracefulShutdown::new();
+		// When this signal completes, start shutdown.
+		let mut signal = std::pin::pin!(shutdown_signal(rx));
+
+		// Start server loop.
+		match TcpListener::bind(addr).await {
+			Ok(l) => loop {
+				tokio::select! {
+					Ok(s) = async {
```
wiesche89: The Ok(s) = async { ... } pattern is still dangerous here. If accept() returns Err once, this branch does not match and select! waits only on the shutdown signal, so the API server stops accepting new connections until restart. Logging the error inside the async block is not enough, we should match on l.accept() in the branch body and continue the loop on Err, like the previous implementation did.

api/src/rest.rs
```diff
@@ -212,81 +194,124 @@ impl ApiServer {
...
+							let conn = http1::Builder::new().serve_connection(io, router.clone());
+							let fut = graceful.watch(conn);
+							tokio::spawn(async move {
+								if let Err(e) = fut.await {
+									error!("API HTTP server error: {:?}", e);
+								}
+							});
+						};
+					}
+					_ = &mut signal => {
```
wiesche89: We register connections with graceful.watch(conn), but after this break we never call graceful.shutdown().await. rt.block_on(server) returns and the runtime is dropped, so in-flight requests can be aborted rather than drained. Could we call graceful.shutdown().await after leaving the accept loop, optionally with a timeout?

api/src/router.rs
```diff
@@ -333,45 +340,45 @@ mod tests {
...
-					.next()
-					.unwrap()
-					.get(Request::new(Body::default()))
-					.await
-					.unwrap();
-				resp.status().as_u16()
-			};
-			block_on(task)
+			let res = client::get::<u16>(format!("http://{}{}", server_addr, url).as_str(), None);
+			thread::sleep(time::Duration::from_millis(500));
```
wiesche89: This sleep happens after the response has already been received, so it only slows the test down. Could we remove the sleep and stop the server at the end, or keep this as an in process router test and cover the server path in tests/rest.rs?

servers/src/common/hooks.rs
```diff
@@ -44,7 +47,12 @@ pub fn init_net_hooks(config: &ServerConfig) -> Vec<Box<dyn NetEvents + Send + S
 		|| config.webhook_config.tx_received_url.is_some()
 		|| config.webhook_config.header_received_url.is_some()
 	{
-		list.push(Box::new(WebHook::from_config(&config.webhook_config)));
+		match WebHook::from_config(&config.webhook_config) {
```
wiesche89: This is nicer than panicking, but it also changes the behavior quite a bit. If webhook setup fails, the node now keeps running without the configured hooks. Is that intentional? For setups that depend on block_accepted_url or tx/header notifications, it may be better to fail startup loudly than to run with webhooks disabled.

api/src/rest.rs
```diff
@@ -21,18 +21,20 @@
...
+use rustls::pki_types::pem::{PemObject, SectionKind};
+use rustls::pki_types::{CertificateDer, PrivateKeyDer};
 use rustls_pemfile as pemfile;
-use std::convert::Infallible;
 use std::fs::File;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use std::{io, thread};
-use tokio::net::TcpListener;
+use tokio::net::{TcpListener, TcpStream};
```
wiesche89: TcpStream and TlsStream are unused now and add two new grin_api warnings.

servers/src/common/hooks.rs
ardocrat: Will return an error on server creation then

api/src/router.rs
ardocrat: Removed delay, added server stop

api/src/rest.rs
ardocrat: Fixed

servers/Cargo.toml
```diff
@@ -9,8 +9,8 @@ keywords = [ "crypto", "grin", "mimblewimble" ]
 edition = "2021"
 
 [dependencies]
-hyper = { version = "0.14", features = ["full"] }
-hyper-rustls = "0.23"
+hyper = { version = "1.10.1", features = ["full"] }
+hyper-rustls = "0.27.9"
 fs2 = "0.4"
 futures = "0.3"
 http = "0.2"
```
wiesche89: http = "0.2" does not seem to be used anymore after the upgrade.

api/src/handlers/pool_api.rs
```diff
@@ -23,11 +23,13 @@ use crate::types::*;
...
 use crate::util::RwLock;
 use crate::web::*;
-use hyper::{Body, Request, StatusCode};
+use hyper::body::Incoming;
+use hyper::{Request, StatusCode};
 use std::sync::Weak;
 
 /// Get basic information about the transaction pool.
 /// GET /v1/pool
+#[allow(dead_code)]
```
wiesche89: These v1 handlers do not seem to be registered anywhere. Instead of adding #[allow(dead_code)], should we remove them or add a short note why they are intentionally kept?

api/src/handlers/peers_api.rs
```diff
@@ -18,16 +18,18 @@ use crate::p2p::{self, PeerData};
 use crate::rest::*;
 use crate::router::{Handler, ResponseFuture};
 use crate::web::*;
-use hyper::{Body, Request, StatusCode};
+use hyper::body::Incoming;
+use hyper::{Request, StatusCode};
 use std::net::SocketAddr;
 use std::sync::Weak;
 
+#[allow(dead_code)]
```
wiesche89: These v1 handlers do not seem to be registered anywhere. Instead of adding #[allow(dead_code)], should we remove them or add a short note why they are intentionally kept?

api/src/handlers/server_api.rs
```diff
@@ -19,21 +19,23 @@ use crate::rest::*;
...
+use hyper::Request;
 use serde_json::json;
 use std::convert::TryInto;
 use std::sync::Weak;
 
-// RESTful index of available api endpoints
-// GET /v1/
+/// RESTful index of available api endpoints
+/// GET /v1/
+#[allow(dead_code)]
```
wiesche89: These v1 handlers do not seem to be registered anywhere. Instead of adding #[allow(dead_code)], should we remove them or add a short note why they are intentionally kept?

servers/Cargo.toml
ardocrat: removed also `rustls`

api/src/handlers/peers_api.rs
ardocrat: seems like before it was used to get all peers from db over API

api/src/handlers/pool_api.rs
ardocrat: this just returns pool size, no extra information

api/src/handlers/server_api.rs
ardocrat: before it was list of all available API endpoints

api/src/rest.rs
ardocrat: Also added a timeout to wait

api/src/rest.rs
ardocrat: Looking at this example https://hyper.rs/guides/1/server/graceful-shutdown/ seems like tokio::select! just waits when first branch completes, with error or not https://docs.rs/tokio/latest/tokio/macro.select.html

api/src/rest.rs
```diff
@@ -212,81 +194,137 @@ impl ApiServer {
...
+		let graceful = hyper_util::server::graceful::GracefulShutdown::new();
+		// When this signal completes, start shutdown.
+		let mut signal = std::pin::pin!(shutdown_signal(rx));
+
+		// Start server loop.
+		match TcpListener::bind(addr).await {
+			Ok(l) => {
+				loop {
+					tokio::select! {
+						Ok(s) = async {
```
wiesche89, line 248: If accept() returns Err, we log it but the result does not match the Ok(s) pattern, so tokio disables this branch for the current select! call and then waits only for shutdown. Could we use res = l.accept() and match Ok/Err inside the branch body, logging and continue on Err?

api/src/rest.rs
```diff
@@ -212,81 +194,137 @@ impl ApiServer {
...
+							match l.accept().await {
+								Ok((s, _)) => Ok(s),
+								Err(e) => {
+									error!("Failed to accept connection: {e:#}");
+									Err(e)
+								}
+							}
+						} => {
+							if let Some(tls) = tls.clone() {
+								let tls_stream = match tls.accept(s).await {
```
wiesche89: This tls handshake is awaited inside the select branch body, so while it is pending we cannot accept new connections or process the shutdown signal. A client that connects and never sends a Client Hello can stall the TLS API. Could we move the handshake into the spawned task and wrap it in a timeout?

api/Cargo.toml
```diff
@@ -10,30 +10,29 @@ edition = "2021"
...
-tokio-rustls = "0.23"
-http = "0.2"
-hyper-rustls = "0.23"
-hyper-timeout = "0.4"
+tokio-rustls = "0.26.4"
+hyper-rustls = "0.27.9"
+hyper-timeout = "0.5.2"
 futures = "0.3"
-rustls = "0.20"
+rustls = "0.23.41"
```
wiesche89: rustls 0.23 pulls in aws-lc-rs by default, which adds native build requirements, while ring still appears to stay in the dependency tree anyway. Should we explicitly choose the rustls crypto provider, e.g. default-features = false with the intended provider feature, so future feature unification cannot accidentally enable both providers?

api/src/router.rs
```diff
@@ -333,45 +339,45 @@ mod tests {
...
-			.add_route("/v1/zzz/*", Arc::new(HandlerImpl(103)))
+			.add_route("/v1/zzz/*", Arc::new(HandlerImpl(104)))
 			.unwrap();
 		routes
-			.add_route("/v1/zzz/*/zzz", Arc::new(HandlerImpl(106)))
+			.add_route("/v1/zzz/*/zzz", Arc::new(HandlerImpl(105)))
 			.unwrap();
 
+		let mut server = ApiServer::new();
+		let server_addr = "127.0.0.1:14424";
```
wiesche89: Can we avoid the hardcoded 14424 here? This can get flaky when the port is taken or tests overlap. We already use an open_port() helper in p2p/tests/peer_handshake.rs, so the same pattern should work here.

api/src/client.rs
```diff
@@ -246,15 +251,19 @@ async fn send_request_async(req: Request<Body>, timeout: TimeOut) -> Result<Stri
...
-	let raw = body::to_bytes(resp.into_body())
+	let body_resp = resp
+		.into_body()
+		.collect()
 		.await
 		.map_err(|e| Error::RequestError(format!("Cannot read response body: {}", e)))?;
 
-	Ok(String::from_utf8_lossy(&raw).to_string())
+	let body = body_resp.to_bytes();
+	let body_text = String::from_utf8(body.to_vec())
```
wiesche89: This changes response decoding from lossy UTF-8 to strict UTF-8. That is probably fine for JSON APIs, but it is a small behavior change. Is that intentional, or should we keep the old from_utf8_lossy behavior?

api/src/handlers/pool_api.rs
```diff
@@ -104,13 +79,15 @@ where
...
 /// Dummy wrapper for the hex-encoded serialized transaction.
+#[allow(dead_code)]
 #[derive(Serialize, Deserialize)]
 struct TxWrapper {
 	tx_hex: String,
 }
 
 /// Push new transaction to our local transaction pool.
 /// POST /v1/pool/push_tx
+#[allow(dead_code)]
```
wiesche89: Some unused v1 handlers were removed, but PoolPushHandler and its helper code now remain behind #[allow(dead_code)]. Should we remove this too, or add a short note explaining why this one is intentionally kept?

ardocrat: removed, `push_transaction` exists at new handler

api/src/router.rs
ardocrat: fixed

api/Cargo.toml
ardocrat: added `default-features = false`

api/src/client.rs
ardocrat: replaced `from_utf8` with `from_utf8_lossy`

api/src/rest.rs
ardocrat: impossible to use graceful inside another task, timeout can be good idea

api/src/rest.rs
ardocrat: we can return Option then to continue on None

api/src/router.rs
```diff
@@ -12,64 +12,71 @@
...
+use http_body_util::combinators::BoxBody;
+use http_body_util::{BodyExt, Empty};
+use hyper::body::Incoming;
 use hyper::service::Service;
-use hyper::{Body, Method, Request, Response, StatusCode};
+use hyper::{Method, Request, Response, StatusCode};
 use std::collections::hash_map::DefaultHasher;
+use std::convert::Infallible;
 use std::hash::{Hash, Hasher};
+use std::net::TcpListener;
```
wiesche89: This import is only used inside the test module now, so normal grin_api builds emit an unused import warning. Could we move it into mod tests?

wiesche89: fixed

wiesche89 (approved):

PR #3885, opened 2026-07-06
Fix tests using rustls

ardocrat:
Initialize default rustls provider so external process tests will not fail (grin-wallet).

### DIFF
--- a/api/tests/rest.rs
+++ b/api/tests/rest.rs
@@ -73,6 +73,9 @@ fn open_port(host: &str) -> u16 {
 
 #[test]
 fn test_start_api() {
+	rustls::crypto::ring::default_provider()
+		.install_default()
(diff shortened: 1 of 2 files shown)


api/tests/rest.rs
```diff
@@ -73,6 +73,9 @@ fn open_port(host: &str) -> u16 {
 
 #[test]
 fn test_start_api() {
+	rustls::crypto::ring::default_provider()
+		.install_default()
+		.expect("Failed to install rustls crypto provider");
```
wiesche89: install_default() can only succeed once per process. Since both tests in this binary call it, the second one will panic when running with --include-ignored. Could we use let _ = ...install_default() here? “Already installed” is fine in this test setup.

ardocrat: fixed

Issue #3210, opened 2020-01-28
Implement missing unit tests for stratumserver


WhoNeedszZz:
Currently `stratumserver` has no unit tests. This will improve the quality and allow future commits to be TDD safe. All units will be tested to ensure the module as a unit itself is tested.

iho:
https://github.com/mimblewimble/grin/pull/3886 I am working on it

zaverpon:
@WhoNeedszZz Hi, is this issue still available? I'm interested in working on it.

PR #3886, opened 2026-07-09
test: add unit tests for stratumserver module

iho:
## Summary

- Expand `stratumserver` unit tests from 8 (RPC serde only) to 30, covering the main testable units without consensus or full PoW mining.
- Adds coverage for:
  - `RpcError` constructors and JSON conversion
  - `parse_params` success/failure paths (`LoginParams`, `SubmitParams`)
  - `WorkersList` add/login/remove/re-login, stats updates, broadcast vs unicast, network stats, unknown-worker error paths
  - `Handler` RPC routing: `keepalive`, unknown method, login, `status`, `getjobtemplate` (synced + syncing), stale/invalid/missing-params/invalid-edge_bits `submit`, `build_block_template`, `last_seen`

Addresses #3210.

**Behavior fix (found via these tests, not just a test change):** `RpcError::internal_error()` used a positive JSON-RPC error code (`32603`). Per the JSON-RPC 2.0 spec (and `api/src/json_rpc.rs`, which already uses the correct value), this is now `-32603`. This changes the wire response a stratum client receives on an internal error from `{"code": 32603, ...}` to `{"code": -32603, ...}`.

RPC routing tests share a single lazily-initialized test chain (via `OnceLock`) instead of opening one LMDB env per test, to avoid intermittent `Invalid argument` failures under default test-runner parallelism.

## Test plan

- [x] `cargo test -p grin_servers --lib mining::stratumserver::tests` (30 passed, default parallelism; also verified with `--test-threads=16`)
- [x] CI on this PR

Not covered in this PR (possible follow-ups): valid share/block submit with real PoW, TCP accept-loop integration tests. See also the review comment about a pre-existing panic on out-of-range `worker_id` in `handle_rpc_requests`.

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -928,6 +928,8 @@ mod tests {
 	// ----------------------------------------
 	// Helpers
 
+	const TEST_MINIMUM_SHARE_DIFFICULTY: u64 = 1;
+
 	fn dummy_tx() -> Tx {
 		let (tx, _rx) = mpsc::unbounded();
 		tx
@@ -960,7 +962,7 @@ mod tests {
 	}
 
 	/// Build a Handler backed by the shared test chain for RPC routing tests.
-	fn setup_handler(minimum_share_difficulty: u64) -> Handler {
+	fn setup_handler() -> Handler {
 		global::set_local_chain_type(ChainTypes::AutomatedTesting);
 		let chain = shared_test_chain();
 		let stratum_stats = Arc::new(RwLock::new(StratumStats::default()));
@@ -971,7 +973,7 @@ mod tests {
 			String::from("test"),
 			stratum_stats,
 			sync_state,
-			minimum_share_difficulty,
+			TEST_MINIMUM_SHARE_DIFFICULTY,
 			chain,
 		)
 	}
@@ -1314,7 +1316,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_keepalive() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		let resp = parse_rpc_response(
@@ -1327,7 +1329,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_method_not_found() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		let resp = parse_rpc_response(
@@ -1341,7 +1343,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_login_ok() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		let params = serde_json::json!({
@@ -1363,7 +1365,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_login_invalid_params() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		let resp =
@@ -1375,7 +1377,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_getjobtemplate_while_syncing() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		// Force syncing state
 		handler.sync_state.update(SyncStatus::HeaderSync {
 			sync_head: handler.chain.head().unwrap(),
@@ -1396,7 +1398,10 @@ mod tests {
 
 	#[test]
 	fn test_handle_getjobtemplate_ok() {
-		let handler = setup_handler(7);
+		let handler = setup_handler();
+		// Non-default difficulty so the template path is not only asserting the const.
+		let job_difficulty = 7;
+		handler.current_state.write().minimum_share_difficulty = job_difficulty;
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		let resp = parse_rpc_response(
@@ -1406,7 +1411,7 @@ mod tests {
 		let result = resp.result.unwrap();
 		assert_eq!(result["height"], 0);
 		assert_eq!(result["job_id"], 0);
-		assert_eq!(result["difficulty"], 7);
+		assert_eq!(result["difficulty"], job_difficulty);
 		// pre_pow is hex-encoded header bytes.
 		let pre_pow = result["pre_pow"].as_str().unwrap();
 		assert!(!pre_pow.is_empty());
@@ -1415,7 +1420,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_status() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 		handler.workers.update_stats(worker_id, |ws| {
 			ws.num_accepted = 10;
@@ -1439,7 +1444,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_submit_too_late() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		// Wrong height vs current block version (height 0) => stale share
@@ -1463,7 +1468,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_submit_invalid_job_id() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		// job_id out of range of current_block_versions
@@ -1484,7 +1489,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_submit_missing_params() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		let resp = parse_rpc_response(
@@ -1496,7 +1501,7 @@ mod tests {
 
 	#[test]
 	fn test_handle_submit_invalid_edge_bits() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 
 		// edge_bits below the AutomatedTesting minimum (10) and not the
@@ -1522,7 +1527,7 @@ mod tests {
 
 	#[test]
 	fn test_last_seen_updates() {
-		let handler = setup_handler(1);
+		let handler = setup_handler();
 		let worker_id = handler.workers.add_worker(dummy_tx());
 		// Force a known baseline instead of racing a real clock read against
 		// the update below.
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -918,7 +918,7 @@ where
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use crate::chain::types::{NoopAdapter, SyncStatus};
+	use crate::chain::types::{HeaderSyncMode, NoopAdapter, SyncStatus};
 	use crate::core::genesis;
 	use crate::core::global::{self, ChainTypes};
 	use crate::core::pow::Difficulty;
@@ -1405,6 +1405,7 @@ mod tests {
 		// Force syncing state
 		handler.sync_state.update(SyncStatus::HeaderSync {
 			sync_head: handler.chain.head().unwrap(),
+			sync_mode: HeaderSyncMode::Legacy,
 			highest_height: 100,
 			highest_diff: Difficulty::from_num(1000),
 		});
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -1137,32 +1137,6 @@ mod tests {
 
 	#[test]
 	fn test_rpc_error_constructors() {
-		let cases = vec![
-			(RpcError::internal_error(), -32603, "Internal error"),
-			(
-				RpcError::node_is_syncing(),
-				-32000,
-				"Node is syncing - Please wait",
-			),
-			(RpcError::method_not_found(), -32601, "Method not found"),
-			(RpcError::too_late(), -32503, "Solution submitted too late"),
-			(
-				RpcError::cannot_validate(),
-				-32502,
-				"Failed to validate solution",
-			),
-			(
-				RpcError::too_low_difficulty(),
-				-32501,
-				"Share rejected due to low difficulty",
-			),
-			(RpcError::invalid_request(), -32600, "Invalid Request"),
-		];
-		for (err, code, message) in cases {
-			assert_eq!(err.code, code);
-			assert_eq!(err.message, message);
-		}
-
 		// Regression: internal_error() must serialize with the negative
 		// JSON-RPC 2.0 code, matching api/src/json_rpc.rs.
 		let value: Value = RpcError::internal_error().into();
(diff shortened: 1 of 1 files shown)


servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1119,453 @@ mod tests {
...
 		assert_eq!(expected_deserialized, actual_deserialized);
 	}
+
+	// ----------------------------------------
+	// RpcError
+
+	#[test]
+	fn test_rpc_error_constructors() {
+		let cases = vec![
+			(RpcError::internal_error(), 32603, "Internal error"),
```
wiesche89: Should this be -32603? JSON-RPC internal error is negative, and api/src/json_rpc.rs uses -32603 too.

servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1119,453 @@ mod tests {
...
+		clean_output_dir(dir);
+	}
+
+	#[test]
+	fn test_handle_login_invalid_params() {
+		let dir = ".grin_stratum_test_login_bad";
+		let handler = setup_handler(dir, 1);
+		let worker_id = handler.workers.add_worker(dummy_tx());
+
+		let resp = parse_rpc_response(
```
wiesche89: Looks like cargo fmt wants to rewrap this line.

servers/src/mining/stratumserver.rs
```diff
@@ -918,6 +918,67 @@ where
...
+		let _ = fs::remove_dir_all(dir_name);
+	}
+
+	fn dummy_tx() -> Tx {
+		let (tx, _rx) = mpsc::unbounded();
+		tx
+	}
+
+	/// Build a Handler backed by a temporary chain for RPC routing tests.
+	fn setup_handler(dir_name: &str, minimum_share_difficulty: u64) -> Handler {
```
wiesche89: could use tempdir() or a path under target/ so interrupted tests do not leave .grin_stratum_test_ dirs in the repo root.

iho: Yes, sure. Thanks for suggestion!

servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1124,452 @@ mod tests {
...
+		let dir = test_chain_dir("grin_stratum_test_last_seen");
+		let handler = setup_handler(&dir, 1);
+		let worker_id = handler.workers.add_worker(dummy_tx());
+		let before = handler.workers.get_stats(worker_id).unwrap().last_seen;
+
+		// Any RPC call updates last_seen via handle_rpc_requests
+		thread::sleep(Duration::from_millis(5));
+		let _ = handler.handle_rpc_requests(rpc_request("keepalive", None), worker_id);
+		let after = handler.workers.get_stats(worker_id).unwrap().last_seen;
+		assert!(after >= before);
```
wiesche89: after >= before would still pass if last_seen were not updated at all. Could we set it to UNIX_EPOCH before the request and assert that it changed? That would also avoid the sleep.

servers/src/mining/stratumserver.rs
```diff
@@ -918,6 +918,72 @@ where
...
+		let (tx, _rx) = mpsc::unbounded();
+		tx
+	}
+
+	/// Build a Handler backed by a temporary chain for RPC routing tests.
+	fn setup_handler(dir_name: &str, minimum_share_difficulty: u64) -> Handler {
+		global::set_local_chain_type(ChainTypes::AutomatedTesting);
+		clean_output_dir(dir_name);
+		let chain = Arc::new(
+			chain::Chain::init(
```
wiesche89: I can reproduce intermittent LMDB Invalid argument failures when these tests run with higher parallelism. Could the handlers share a test chain, or avoid creating a full chain for RPC paths that do not use it? CI runs the suite with default parallelism.


StoreErr(LmdbErr("Invalid argument (os error 22)"), ...)

servers/src/mining/stratumserver.rs
```diff
@@ -84,7 +84,7 @@ struct RpcError {
 impl RpcError {
 	pub fn internal_error() -> Self {
 		RpcError {
-			code: 32603,
+			code: -32603,
```
wiesche89, line 87: This fix is correct and consistent with JSON-RPC and api/src/json_rpc.rs. Could you mention it in the PR description as well? It is a wire visible behavior change that is easy to miss in a test-focused PR.

servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1124,452 @@ mod tests {
...
+		assert_eq!(result["difficulty"], 5);
+		assert_eq!(result["accepted"], 10);
+		assert_eq!(result["rejected"], 2);
+		assert_eq!(result["stale"], 1);
+
+		clean_output_dir(&dir);
+	}
+
+	#[test]
+	fn test_handle_submit_too_late() {
```
wiesche89, line 1446: Would it be worth adding the two cheap submit error cases here as well: missing params returning -32600, and invalid edge_bits returning -32502 with num_rejected incremented? Neither requires mining a valid proof.

servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1124,452 @@ mod tests {
...
+		assert_eq!(result["height"], 0);
+		assert_eq!(result["job_id"], 0);
+		assert_eq!(result["difficulty"], 7);
+		assert!(result["pre_pow"].as_str().unwrap().len() > 0);
+
+		clean_output_dir(&dir);
+	}
+
+	#[test]
+	fn test_handle_status() {
```
wiesche89, line 1422: Could we also cover status with an unknown worker ID? That would exercise the get_stats failure through the RPC layer and verify the -32603 response.

servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1124,452 @@ mod tests {
...
+		assert_eq!(workers.count(), 0);
+
+		let id0 = workers.add_worker(dummy_tx());
+		let id1 = workers.add_worker(dummy_tx());
+		assert_eq!(id0, 0);
+		assert_eq!(id1, 1);
+		assert_eq!(workers.count(), 2);
+		assert_eq!(stats.read().num_workers, 2);
+
+		workers
```
wiesche89, line 1205: could this test log the worker in a second time and verify that the login and agent are replaced? That would document the current re authentication behavior.

iho:
Pushed a follow-up commit addressing the remaining open comments:

- Shared a single lazily-initialized test chain (`OnceLock`) across the RPC routing tests instead of opening one LMDB env per test, to fix the intermittent `Invalid argument` failures under default parallelism. None of these tests reach `chain::Chain::process_block`, so the chain is read-only from their perspective and safe to share.
- `test_last_seen_updates` now seeds `last_seen` to `UNIX_EPOCH` and asserts strict `>` instead of `>=` after a sleep, so a no-op update would actually fail the test.
- Added the two cheap `submit` error cases: missing params (`-32600`) and an `edge_bits` value that's neither primary nor secondary PoW (`-32502`), neither requiring a mined proof.
- Added a re-login test documenting that logging in again replaces the previous `login`/`agent` rather than being rejected.
- Also updated the PR description to call out the `-32603` fix as a wire-visible behavior change.

**One thing I could not implement as suggested — found a real bug while trying:** the request to add a `status` test with an unknown worker ID, asserting the `-32603` response, doesn't work as written. `handle_rpc_requests` unconditionally calls `self.workers.last_seen(worker_id)` *before* dispatching to `handle_status`. `last_seen` → `update_stats` does `stratum_stats.worker_stats[worker_id]`, an unchecked `Vec` index, whereas `get_stats` (which `handle_status` uses) does a checked `.get(worker_id)`. So calling `handle_rpc_requests` with an out-of-range `worker_id` panics at the `last_seen` step with `index out of bounds`, rather than returning the graceful `-32603` `RpcError` that `handle_status` alone would produce.

I confirmed this with a throwaway test (`handler.handle_rpc_requests(rpc_request("status", None), 999)` on a handler with no workers added) — it panicked with `index out of bounds: the len is 0 but the index is 999` at the `update_stats` call site, rather than returning an error response.

Since fixing that unchecked indexing is a production-code behavior change beyond this test-only PR's scope, I left it alone here and instead added `test_workers_list_get_stats_missing_worker`, which exercises `WorkersList::get_stats`'s `-32603` path directly (mirroring the existing `test_workers_list_login_missing_worker` pattern) without going through the panicking `handle_rpc_requests` path.

Happy to open a separate PR/issue for the panic if that's useful — a connected worker's `worker_id` is normally always in range, so this likely isn't reachable in practice today, but it's a footgun for any future caller of `handle_rpc_requests` (or `update_stats`/`last_seen`) with an id that isn't guaranteed to be currently registered.

servers/src/mining/stratumserver.rs
```diff
@@ -918,6 +918,91 @@ where
...
+	use std::sync::OnceLock;
+
+	// ----------------------------------------
+	// Helpers
+
+	fn clean_output_dir(dir_name: &str) {
+		let _ = fs::remove_dir_all(dir_name);
+	}
+
+	/// Path under `target/tmp/` so interrupted tests do not leave dirs in the repo root.
```
wiesche89: These helpers are only used by shared_test_chain. Could we inline them there to keep the test setup smaller?

servers/src/mining/stratumserver.rs
```diff
@@ -918,6 +918,91 @@ where
...
+	fn test_chain_dir(name: &str) -> String {
+		format!("target/tmp/{}", name)
+	}
+
+	fn dummy_tx() -> Tx {
+		let (tx, _rx) = mpsc::unbounded();
+		tx
+	}
+
+	/// A single chain instance shared by every RPC routing test below.
```
wiesche89: This comment is quite long. Could we shorten it and just note that write-path tests need their own chain?

servers/src/mining/stratumserver.rs
```diff
@@ -918,6 +918,91 @@ where
...
+			params,
+		}
+	}
+
+	fn parse_rpc_response(json: &str) -> RpcResponse {
+		serde_json::from_str(json).unwrap()
+	}
+
+	// ----------------------------------------
+	// RpcRequest / RpcResponse serde (existing)
```
wiesche89: “Existing” is review context rather than useful source documentation. Could we drop it?

servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1143,492 @@ mod tests {
...
+	#[test]
+	fn test_workers_list_relogin_replaces_login_and_agent() {
+		let stats = Arc::new(RwLock::new(StratumStats::default()));
+		let workers = WorkersList::new(stats);
+		let id0 = workers.add_worker(dummy_tx());
+
+		workers
+			.login(id0, "alice".to_string(), "agent-a".to_string())
+			.unwrap();
+		// Logging in again (e.g. after a reconnect) replaces the previous
```
wiesche89: This is a second login for the same worker, not a reconnect, which would create a new worker ID. Could we drop the reconnect example?

wiesche89 (commented):
The behavioral coverage is useful and the earlier review points are addressed. The test module still feels broader and more heavily documented than the surrounding code, though. Could we trim tests that only mirror constructor field assignments, inline single-use helpers, and keep comments for non obvious behavior and setup constraints?

wiesche89 (commented):
1) Thanks for trimming this down, the test module looks much cleaner now. I think a few tests could still be consolidated: test_rpc_error_into_value, the successful parse_params cases, and test_workers_list_update_stats mostly cover Serde or direct assignments, while test_build_block_template overlaps with test_handle_getjobtemplate_ok.
Maybe keep one focused regression test that checks internal_error() is serialized as -32603.

2) The relative test path ends up under servers/target/tmp/, rather than the workspace target/. That’s fine, but calling it the “crate-local target directory” would make the comment a bit clearer (shared_test_chain)

iho:
Pushed a follow-up commit addressing the latest review:

- Folded `test_rpc_error_into_value` into `test_rpc_error_constructors`, keeping one focused regression check that `internal_error()` serializes as `-32603` (the actual wire-visible bug this PR fixes) rather than testing `method_not_found()`'s `Into<Value>` path, which wasn't adding anything beyond the constructor table.
- Merged the two successful `parse_params` tests (`LoginParams`, `SubmitParams`) into a single `test_parse_params_ok`.
- Dropped `test_workers_list_update_stats` — `update_stats` is already exercised through production code paths in `test_handle_status` and the `submit` error-path tests.
- Dropped `test_build_block_template`, which overlapped `test_handle_getjobtemplate_ok`; folded its hex-digit `pre_pow` check into the latter so that coverage isn't lost.
- Reworded the `shared_test_chain` directory comment to call out that `target/tmp/...` is the crate-local target directory (resolves under `servers/target/`), not the workspace one.

`cargo test -p grin_servers --lib mining::stratumserver::tests` passes (30 tests now, down from 34, with no coverage loss per the above), and `cargo fmt`/`cargo clippy` are clean on the file.

servers/src/mining/stratumserver.rs
```diff
@@ -1058,4 +1131,433 @@ mod tests {
 
 		assert_eq!(expected_deserialized, actual_deserialized);
 	}
+
+	// ----------------------------------------
+	// RpcError
+
+	#[test]
+	fn test_rpc_error_constructors() {
+		let cases = vec![
```
wiesche89: Thanks for consolidating these tests. One small part of the previous review remains here. The cases table still mirrors the literals from all seven constructors. Could we remove it and keep only the wire regression check below? That protects the actual -32603 change without maintaining the codes and messages twice.

wiesche89 (commented):
After the consolidation there are now 30 Stratum tests, but the summary and test plan still say 34. Could you update both numbers to 30? The current command passes 30/30 tests for me.


I added one more review comment. Otherwise, everything looks good.

iho:
Pushed a follow-up commit and updated the description:

- Dropped the \`RpcError\` constructor cases table in \`test_rpc_error_constructors\`, keeping only the \`internal_error()\` → \`-32603\` wire regression check. The other constructors' codes are already exercised through the RPC-routing tests (\`method_not_found\` via \`test_handle_method_not_found\`, \`invalid_request\` via \`test_handle_login_invalid_params\`/\`test_handle_submit_missing_params\`, \`too_late\` via \`test_handle_submit_too_late\`, \`cannot_validate\` via \`test_handle_submit_invalid_edge_bits\`, \`node_is_syncing\` via \`test_handle_getjobtemplate_while_syncing\`), so nothing is being tracked in two places anymore.
- Updated the PR description/test plan from 34 to 30 tests, matching the post-consolidation count.

\`cargo test -p grin_servers --lib mining::stratumserver::tests\` passes 30/30.

wiesche89 (approved):

servers/src/mining/stratumserver.rs
```diff
@@ -918,6 +918,79 @@ where
...
+						None,
+					)
+					.unwrap(),
+				)
+			})
+			.clone()
+	}
+
+	/// Build a Handler backed by the shared test chain for RPC routing tests.
+	fn setup_handler(minimum_share_difficulty: u64) -> Handler {
```
ardocrat: `minimum_share_difficulty` is always passing as 1, can be const

servers/src/mining/stratumserver.rs
ardocrat: Seems like we need to move these variables in constants, grin-wallet also uses them, at 2 crates

servers/src/mining/stratumserver.rs
iho: Done — `setup_handler` now uses a `TEST_MINIMUM_SHARE_DIFFICULTY` const (1) with no parameter. The getjobtemplate test still overrides difficulty to a non-default value so that path is not only asserting the const.

ardocrat (approved):

Issue #2447, opened 2019-01-22
Update grin/pool.md documentation


lehnberg:
[pool.md](https://github.com/mimblewimble/grin/blob/master/doc/internal/pool.md) documentation claims we are using a DAG to represent mempool transactions, which is no longer correct:

> Hashmap: We don’t use DAG for mempool anymore

PR #3887, opened 2026-07-09
docs: rewrite pool.md for current txpool/stempool design

iho:
## Summary

- Rewrite `doc/internal/pool.md` so it matches the current `pool` crate instead of the old DAG + orphan-graph design.
- Document the **one big transaction** validation model (`aggregate(pool ∪ new_tx)` vs chain state), nested as:

  `[ new_tx + [ stempool + [ txpool + [ chain_state ] ] ] ]`

- Use consistent **txpool** / **stempool** terminology, plus reorg cache, mining bucket selection, reconciliation, config, and a short code map.

Addresses #2447 (maintainer notes on hashmap / one-big-tx representation and stempool terminology).

## Test plan

- [x] Docs-only change; content cross-checked against `pool/src/transaction_pool.rs`, `pool/src/pool.rs`, and `pool/src/types.rs`
- [ ] Maintainer skim for accuracy (especially Dandelion stem fluff edge cases)

### DIFF
--- a/doc/internal/pool.md
+++ b/doc/internal/pool.md
@@ -1,58 +1,167 @@
 # Transaction Pool
 
-This document describes some of the basic functionality and requirements of grin's transaction pool.
+This document describes the design and behavior of Grin's transaction pool as implemented today in the `pool` crate (`pool/src/`).
 
-## Overview of Required Capabilities
+For Dandelion stem/fluff propagation details, see [dandelion.md](../dandelion/dandelion.md).
 
-The primary purpose of the memory pool is to maintain a list of mineable transactions to be supplied to the miner service while building new blocks. The design will center around ensuring correct behavior here, especially around tricky conditions like head switching.
+## Purpose
 
-For standard (non-mining) nodes, the primary purpose of the memory pool is to serve as a moderator for transaction broadcasts by requiring connectivity to the blockchain. Secondary uses include monitoring incoming transactions, for example for giving early notice of an unconfirmed transaction to the user's wallet.
+The transaction pool keeps a set of unconfirmed transactions that:
 
-Given the focus of grin (and mimblewimble) on reduced resource consumption, the memory pool should be an optional but recommended component for non-mining nodes.
+1. Can be selected by the mining service when building a new block.
+2. Can be relayed to peers (after any stem embargo ends).
+3. Moderate broadcast behavior so only transactions valid against current chain state (plus the pool itself) are accepted.
 
-## Design Overview
+The pool is required for mining and for normal transaction relay. It is implemented as two related layers plus a short re-org cache:
 
-The primary structure of the transaction pool is a pair of Directed Acyclic Graphs. Since each transaction is rooted directly by its inputs in a non-cyclic way, this structure naturally encompasses the directionality of the chains of unconfirmed transactions. Defining this structure has a few other nice properties: descendent invalidation (when a conflicting transaction is accepted for a given input) is nearly free, and the mineability of a given transaction is clearly depicted in its location in the hierarchy.
+| Layer | Name in code | Visibility | Role |
+|-------|--------------|------------|------|
+| **txpool** | `TransactionPool.txpool` | Public | Fluffed transactions ready for relay and mining |
+| **stempool** | `TransactionPool.stempool` | Private | Stem-phase Dandelion transactions under embargo |
+| **reorg cache** | `TransactionPool.reorg_cache` | Internal | Recent fluffed txs kept briefly to repopulate after a reorg |
 
-Another, non-obvious reason for the choice of a DAG is that the acyclic nature of transactions is a necessary property but must be explicitly verified in a way that is not true of other UTXO-based cryptocurrencies. Consider the following loop of single-input single-output transactions in BTC:
+Both `txpool` and `stempool` are instances of the same `Pool` type (`pool/src/pool.rs`), parameterized by a `BlockChain` adapter for UTXO and lock-height checks.
 
-A->B->C->A
+## Design: one big transaction
 
-Because each input in Bitcoin specifically references the hash and output index of the output in a preceding transaction, for a loop to exist, a transaction must reference (and know the hash of) a transaction that does not yet exist (C, in the trivial example.) Furthermore, the hash and output index pair (called an "outpoint" in Bitcoin) is covered by the transaction hash of A, such that any change to either causes the hash of A to change. Therefore, attempting to build such a loop by amending A with the proper outpoint in C after C has been built causes A's hash to change, invalidating B, and so forth.
+Grin does **not** store the mempool as a DAG of transactions.
 
-In grin, an input references an output by the output's own hash. Thus, the backreference does not include the situation the output was generated in, which allows (from a purely mechanical point of view) the creation of a loop without the ability to generate a specific hash from a tightly constrained preimage.
+Earlier design notes (and older versions of this document) described a pair of directed acyclic graphs for connected and orphan transactions. That model was abandoned.
 
-The pair of graphs represents the connected graph and the orphans graph. (While it is possible to represent both groups of transactions in a single graph, it makes determination of orphan status of a given transaction non-trivial, requiring either the maintenance of a flag or traversal upwards of potentially many inputs.)
+The current model treats each pool as a flat list of `PoolEntry` values (transaction + source + timestamp), stored in insertion order:
 
-A transaction reference in the pool has parents, one for each input. The parents fall into one of four states:
+```text
+Pool.entries: Vec<PoolEntry>
+```
 
-* Unknown
-* Blockchain transaction
-* Pool transaction
-* Orphan transaction
+Validation does not walk a graph. Instead, **all transactions already in a pool are aggregated into a single transaction** (Mimblewimble cut-through / aggregation). Accepting a new transaction means:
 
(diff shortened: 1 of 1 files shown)


doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
 
-### Adding a Transaction
+This is why `Pool::all_transactions_aggregate` and `transaction::aggregate` are central to the implementation: the pool is always considered as “one big tx” (or empty) on top of the UTXO set.
 
-The most basic task of the transaction pool is to add an incoming transaction to the graph.
+### Why aggregation instead of a DAG
 
-The first step is the validation of the transaction itself. This involves the enforcement of all consensus rules surrounding the construction of the transaction itself, and the verification of all relevant signatures and proofs.
+- Mimblewimble already supports transaction aggregation and cut-through at the protocol level.
+- A single aggregate validation reuses the same rules as block validation (kernel sums, UTXO membership, coinbase maturity, NRD rules, etc.).
```
wiesche89, line 61: coinbase maturity isn't checked inside the aggregate validation, it runs separately beforehand in the add path using the result of locate_spends, as step 7 below correctly describes. I'd drop coinbase maturity from this list so it doesn't read as part of validate_raw_tx.

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
+Failed stem adapter acceptance falls back to fluff (add to txpool).
+
+## Mining selection
+
+`prepare_mineable_transactions` (txpool only):
+
+1. Sort and group pool txs with **bucket** logic (`bucket_transactions`):
+   - Prefer keeping dependency order and maximizing cut-through within a bucket.
+   - Prefer higher aggregate fee rate; avoid merging if it would lower fee rate.
+   - Txs with multiple in-pool parents are skipped for this block (picked up later).
```
wiesche89, line 113: The bucket logic rejects a tx on the second input spending a pool output, even when both outputs come from the same parent. Could we describe this as more than one input spending unconfirmed pool outputs instead?

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
+2. Re-validate the ordered list incrementally as aggregates against chain state under the miner’s `mineable_max_weight`.
+3. Return the list of individually mineable transactions for block building.
+
+Eviction under capacity pressure uses the same bucket ordering and removes a last (low fee-rate, non-dependent) transaction.
```
wiesche89, line 117: Non dependent is a little ambiguous because the evicted tx may itself depend on a parent. Could we say a low-fee-rate transaction that no other pool tx depends on instead?

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
-Given the focus of grin (and mimblewimble) on reduced resource consumption, the memory pool should be an optional but recommended component for non-mining nodes.
+1. Can be selected by the mining service when building a new block.
+2. Can be relayed to peers (after any stem embargo ends).
+3. Moderate broadcast behavior so only transactions valid against current chain state (plus the pool itself) are accepted.
 
-## Design Overview
+The pool is required for mining and for normal transaction relay. It is implemented as two related layers plus a short re-org cache:
 
-The primary structure of the transaction pool is a pair of Directed Acyclic Graphs. Since each transaction is rooted directly by its inputs in a non-cyclic way, this structure naturally encompasses the directionality of the chains of unconfirmed transactions. Defining this structure has a few other nice properties: descendent invalidation (when a conflicting transaction is accepted for a given input) is nearly free, and the mineability of a given transaction is clearly depicted in its location in the hierarchy.
+| Layer | Name in code | Visibility | Role |
```
wiesche89, line 17: All three fields are pub in Rust, so Visibility may be easy to misread as code visibility. Could we call this column Network exposure or something similar?

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
+- On successful fluff acceptance, the entry is also pushed into the reorg cache and the pool adapter’s `tx_accepted` hook runs (P2P broadcast, etc.).
 
-Under adversarial situations, the primary concerns to the transaction pool are denial-of-service attacks. The greatest concern should be maintaining the ability of the node to provide services to miners, by supplying ready made transactions to the mining service for inclusion in blocks. Resource consumption should be constrained as well. As we've seen on other chains, miners often have little incentive to include transactions if doing so impacts their ability to collect their primary reward.
+### stempool
+
+- Private holding area for Dandelion **stem** transactions.
+- Stem txs are validated with the current **txpool aggregate** as `extra_tx`, so they cannot double-spend against fluffed txs.
+- Contents are not used for compact-block reconstruction or mining (`retrieve_transactions` and `prepare_mineable_transactions` use **txpool only**).
+- The Dandelion monitor (`servers/src/grin/dandelion_monitor.rs`) periodically fluffs stem entries when embargo or epoch timers expire by moving them into the txpool path.
```
wiesche89, line 79: Could we mention the aggregation timer here as well? During a fluff epoch, the monitor also fluffs when an entry exceeds aggregation_secs, while embargo expiry is handled separately per entry.

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
-### Adding a Transaction
+This is why `Pool::all_transactions_aggregate` and `transaction::aggregate` are central to the implementation: the pool is always considered as “one big tx” (or empty) on top of the UTXO set.
 
-The most basic task of the transaction pool is to add an incoming transaction to the graph.
+### Why aggregation instead of a DAG
 
-The first step is the validation of the transaction itself. This involves the enforcement of all consensus rules surrounding the construction of the transaction itself, and the verification of all relevant signatures and proofs.
+- Mimblewimble already supports transaction aggregation and cut-through at the protocol level.
+- A single aggregate validation reuses the same rules as block validation (kernel sums, UTXO membership, coinbase maturity, NRD rules, etc.).
+- Dependency ordering is not required for storage; it is reconstructed only when selecting txs for a block (see [Mining selection](#mining-selection)).
```
wiesche89, line 62: Only seems slightly too strong here because the same bucket dependency ordering is also reconstructed for eviction. Could we mention mining and eviction, or simply drop only?

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
-The primary purpose of the memory pool is to maintain a list of mineable transactions to be supplied to the miner service while building new blocks. The design will center around ensuring correct behavior here, especially around tricky conditions like head switching.
+## Purpose
 
-For standard (non-mining) nodes, the primary purpose of the memory pool is to serve as a moderator for transaction broadcasts by requiring connectivity to the blockchain. Secondary uses include monitoring incoming transactions, for example for giving early notice of an unconfirmed transaction to the user's wallet.
+The transaction pool keeps a set of unconfirmed transactions that:
 
-Given the focus of grin (and mimblewimble) on reduced resource consumption, the memory pool should be an optional but recommended component for non-mining nodes.
+1. Can be selected by the mining service when building a new block.
+2. Can be relayed to peers (after any stem embargo ends).
+3. Moderate broadcast behavior so only transactions valid against current chain state (plus the pool itself) are accepted.
```
wiesche89, line 13: this breaks the transactions that can... structure of the list. Maybe Act as a broadcast moderator, so only transactions valid against the current chain and pool state are accepted?

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
+| Component | Location |
+|-----------|----------|
+| `TransactionPool` (txpool + stempool + reorg cache) | `pool/src/transaction_pool.rs` |
+| `Pool` (entries, aggregate validate, buckets, mining prep) | `pool/src/pool.rs` |
+| Config, `PoolEntry`, errors, adapters | `pool/src/types.rs` |
+| Dandelion stem timers / fluff | `servers/src/grin/dandelion_monitor.rs` |
+| Block building from pool | `servers/src/mining/mine_block.rs` |
+| Reconcile on new block / reorg | `servers/src/common/adapters.rs` |
+
+## Historical note
```
wiesche89, line 165: This mostly repeats the DAG explanation near the start of the document. Would it be simpler to keep the historical context there and drop this final section?

doc/internal/pool.md
```diff
@@ -1,58 +1,167 @@
...
+
+### reorg cache
+
+- Stores recent successfully fluffed `PoolEntry` values (same size bound as `max_pool_size`).
+- On a reorg that increases total work, `reconcile_reorg_cache` attempts to re-add those entries to the txpool against the new tip.
+- Entries older than the configured retention window (`reorg_cache_period`, default 30 minutes) are truncated.
+
+## Adding a transaction
+
+High-level flow in `TransactionPool::add_to_pool`:
```
wiesche89, line 91: This is accurate today, but the exact ten step ordering is fairly implementation specific and may become stale after a refactor. Could we keep this section focused on the important validation boundaries rather than mirroring the function step by step?

wiesche89 (commented):
Overall, this is a substantial improvement over the old DAG based document and matches the current pool design well. Most of my comments are small precision or wording points. My only broader concern is keeping the implementation level walkthroughs maintainable as the code changes. With the validation boundaries and bucket behavior clarified, this looks good to me.

doc/internal/pool.md
wiesche89: On second thought, a better fit for the existing list structure might be: Have been validated against the current chain and pool state before acceptance and relay.

Issue #3647, opened 2021-07-01
Clean old txhashset_snapshot_*.zip periodically


noobvie:
**Is your feature request related to a problem? Please describe.**
Sometimes the grinnode generated many zip files in ~/.grin/main/chain_data which cost lot of space (around 700MB each) and it consumes space of all nodes in the network. For example:
│ -rw-rw-r-- 1 root root 741396605 Jun 30 20:21 txhashset_snapshot_00032512d663.zip                                                                                                                  │
│ -rw-rw-r-- 1 root root 742231585 Jul  1 07:38 txhashset_snapshot_00033291faad.zip                                                                                                                  │
│ -rw-rw-r-- 1 root root 741166818 Jun 30 16:39 txhashset_snapshot_00052138ecfe.zip                                                                                                                  │

**Describe the solution you'd like**
I suggest grinnode should have feature to remove those files periodically beacause most nodes are running on low end boxes with low disk/cpu/ram.

Thanks.

phyro:
Thank you for reporting this. There is an RFC that solves this called PIBD (Parallel Initial Block Download) which should resolve the issue of having a large zip file for seeding new nodes. The main guy that was working on it is unfortunately not contributing right now so it is stale at the moment. Hopefully someone picks up from where they left and finishes the implementation.

phyro:
Sorry, I seem to have written this a bit wrong. The zip files won't go away for some time as there will be nodes on the version that does not support PIBD. So they will be around for some time until we know PIBD works really well and people have upgraded.

mojitoo:
i also ran out of space and node stopped working because of this

wiesche89:
When creating a new txhashset, the previous snapshot file is lazily cleaned up and removed automatically:
[chain/src/txhashset/txhashset.rs#L2053](https://github.com/mimblewimble/grin/blob/db2923a147e1e68aa858d91b9f121ca5a3090cec/chain/src/txhashset/txhashset.rs#L2053)
So this issue appears to already be handled and can likely be closed.

wiesche89:
A periodic cleanup could therefore be implemented.

- Old txhashset_snapshot_*.zip files are cleaned up when a node needs to generate a new txhashset zip for a peer.
- Only files with the txhashset_snapshot_ prefix older than 24 hours are removed.

Cleanup does not happen if:
- no peer requests a new txhashset zip for a long time
- the same already existing zip file is repeatedly requested, because the function returns early in that case

[...]

PR #3888, opened 2026-07-09
fix: clean old txhashset snapshot zips on every zip_read

iho:
## Summary

- Fix disk growth from leftover `txhashset_snapshot_*.zip` files (#3647).
- On every `zip_read` (reuse **and** create), remove other snapshot zips and keep only the one for the requested header.
- Previous behavior cleaned age-based (24h) files only when creating a **new** zip, so:
  - reusing the same zip never cleaned anything
  - multiple multi-GB zips could accumulate within 24 hours
- Adds `cleanup_old_txhashset_zips` + unit test.

## Test plan

- [x] `cargo test -p grin_chain --test test_txhashset -- --test-threads=1` (2 passed)
- [ ] CI on this PR

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -37,7 +37,7 @@ use crate::util::secp::pedersen::{Commitment, RangeProof};
 use crate::util::{file, secp_static, zip, StopState};
 use crate::SyncState;
 use croaring::Bitmap;
-use grin_store::pmmr::{clean_files_by_prefix, PMMRBackend};
+use grin_store::pmmr::PMMRBackend;
 use std::cmp::Ordering;
 use std::fs::{self, File};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


chain/src/txhashset/txhashset.rs
```diff
@@ -2056,6 +2056,61 @@ impl<'a> Extension<'a> {
...
+		}
+		if keep_name.is_some() && path.file_name() == keep_name {
+			continue;
+		}
+		let name = match entry.file_name().into_string() {
+			Ok(n) => n,
+			Err(_) => continue,
+		};
+		// Match `txhashset_snapshot_<hash>.zip` only.
+		if !name.starts_with(&prefix) || !name.ends_with(".zip") {
```
wiesche89, line 2095: The comment says <hash>.zip, but this matches any text between the prefix and .zip. 

Generated snapshot names use a 12-character hex hash. Could we restrict cleanup to that exact format?

chain/tests/test_txhashset.rs
```diff
@@ -115,3 +115,37 @@ fn test_unexpected_zip() {
...
+	let unrelated = db_root.join("other_file.bin");
+	let not_zip = db_root.join("txhashset_snapshot_cccc.txt");
+
+	File::create(&keep).unwrap();
+	File::create(&old_a).unwrap();
+	File::create(&old_b).unwrap();
+	File::create(&unrelated).unwrap();
+	File::create(&not_zip).unwrap();
+
+	let removed = txhashset::cleanup_old_txhashset_zips(db_root, &keep);
```
wiesche89, line 137: Could this also exercise zip_read() with the current archive already present and one stale archive beside it? The original regression was specifically that the reuse path skipped cleanup, while this test only covers the helper directly.

chain/src/txhashset/txhashset.rs
```diff
@@ -2056,6 +2056,61 @@ impl<'a> Extension<'a> {
...
+			continue;
+		}
+		match fs::remove_file(&path) {
+			Ok(()) => {
+				debug!("cleanup_old_txhashset_zips: removed {:?}", path);
```
wiesche89, line 2100: zip_read() already logs the total number removed. Could this per file message be trace! or be dropped to avoid duplicate debug output during a rollover?

chain/src/txhashset/txhashset.rs
```diff
@@ -2056,6 +2056,61 @@ impl<'a> Extension<'a> {
 	}
 }
 
+/// Remove other `txhashset_snapshot_*.zip` files in `data_dir`, keeping only
```
wiesche89, line 2059: Could we shorten this? The key detail here is that cleanup runs before both reuse and creation

PR #3889, opened 2026-07-09
fix: stop stratum TCP connection leak under reconnect load

iho:
## Summary

Fixes [#3867](https://github.com/mimblewimble/grin/issues/3867) / reports of stratum leaving thousands of TCP sockets open until the process exits (`Too many open files`, Windows loopback connection storms).

### Root causes addressed

1. **Sessions never timed out** — half-open / abandoned peers could sit forever with no TCP keepalive and no app idle limit.
2. **Unbounded write queue** — `futures` unbounded mpsc + job broadcast kept feeding dead peers without backpressure.
3. **No concurrent worker cap** — reconnect storms could open unlimited FDs.
4. **Cleanup was easy to skip** — `select` + best-effort `remove_worker` (and `expect` panics) were fragile; worker_stats also grew forever.

### Changes

| Change | Why |
|--------|-----|
| `handle_connection` + `Drop` guard | Always `remove_worker` when the task ends |
| 5-minute idle timeout | Close silent / half-open sessions |
| Max 256 workers | Hard FD / resource bound |
| Tokio bounded channel (64) | Backpressure; drop full/slow peers on broadcast |
| 30s write timeout | Don't pin tasks on stalled sockets |
| `LinesCodec` max 64 KiB | Avoid huge-line memory abuse |
| Worker stats slot reuse | Reconnect storms don't grow the stats vec forever |
| Idempotent `remove_worker` | Safe double-cleanup |

### Tests

- `test_worker_slot_reuse_after_disconnect`
- `test_try_send_to_missing_full_and_ok`
- `test_remove_worker_is_idempotent`
- Existing RPC serde tests

```sh
cargo test -p grin_servers --lib mining::stratumserver -- --test-threads=1
```

## Test plan

- [x] Unit tests above (11 passed)
- [ ] Run a node with stratum enabled and a reconnecting miner; confirm worker count and FDs stay bounded
- [ ] Confirm idle miners are disconnected after ~5 minutes without traffic
- [ ] CI

### DIFF
--- a/scripts/stratum_reconnect_soak.sh
+++ b/scripts/stratum_reconnect_soak.sh
@@ -1,0 +1,306 @@
+#!/usr/bin/env bash
+# Full-node soak: run grin with stratum enabled + reconnecting "miner" clients.
+# Verifies worker connections and process FDs stay bounded (issue #3867 / PR #3889).
+#
+# Usage (from repo root):
+#   ./scripts/stratum_reconnect_soak.sh
+#
+# Env overrides:
+#   GRIN_BIN       path to grin binary (default: target/debug/grin)
+#   CYCLES         connect/login/drop cycles (default: 120)
+#   CONCURRENT     concurrent held connections (default: 32)
+#   FD_GROWTH_MAX  max allowed FD growth over soak (default: 100)
+
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "$0")/.." && pwd)"
+cd "$ROOT"
+
+GRIN_BIN="${GRIN_BIN:-$ROOT/target/debug/grin}"
+CYCLES="${CYCLES:-120}"
+CONCURRENT="${CONCURRENT:-32}"
+FD_GROWTH_MAX="${FD_GROWTH_MAX:-100}"
+STRATUM_HOST="127.0.0.1"
+# Ephemeral high port avoids collisions with leftover usernet nodes.
+STRATUM_PORT="${STRATUM_PORT:-$(( 24000 + (RANDOM % 1000) ))}"
+API_PORT="${API_PORT:-$(( 25000 + (RANDOM % 1000) ))}"
+P2P_PORT="${P2P_PORT:-$(( 26000 + (RANDOM % 1000) ))}"
+WORKDIR="$(mktemp -d "${TMPDIR:-/tmp}/grin-stratum-soak.XXXXXX")"
+LOG="$WORKDIR/grin.log"
+PID=""
+
+cleanup() {
+  if [[ -n "${PID}" ]] && kill -0 "$PID" 2>/dev/null; then
+    echo "==> Stopping grin (pid $PID)"
+    kill -TERM "$PID" 2>/dev/null || true
+    for _ in $(seq 1 30); do
+      kill -0 "$PID" 2>/dev/null || break
+      sleep 0.2
+    done
+    kill -KILL "$PID" 2>/dev/null || true
+    wait "$PID" 2>/dev/null || true
+  fi
+}
+trap cleanup EXIT
+
+count_fds() {
+  local pid="$1"
+  if command -v lsof >/dev/null 2>&1; then
+    # Count open files for the process (includes sockets).
+    lsof -nP -p "$pid" 2>/dev/null | tail -n +2 | wc -l | tr -d ' '
+  elif [[ -d "/proc/$pid/fd" ]]; then
+    ls -1 "/proc/$pid/fd" 2>/dev/null | wc -l | tr -d ' '
+  else
+    echo "0"
+  fi
+}
+
+count_tcp_established_stratum() {
+  local pid="$1"
+  if command -v lsof >/dev/null 2>&1; then
+    # ESTABLISHED only (exclude LISTEN). Match local or remote stratum port.
+    lsof -nP -p "$pid" -iTCP 2>/dev/null \
+      | awk -v p=":${STRATUM_PORT}" '
+          NR>1 && $0 ~ /ESTABLISHED/ && index($0, p) { c++ }
+          END { print c+0 }
+        '
+  else
+    echo "0"
+  fi
+}
+
+wait_port() {
+  local host="$1" port="$2" tries="${3:-60}"
+  for i in $(seq 1 "$tries"); do
+    if (echo >/dev/tcp/"$host"/"$port") >/dev/null 2>&1; then
+      return 0
+    fi
+    # bash /dev/tcp may be unavailable; fall back to python
+    if python3 - "$host" "$port" <<'PY' 2>/dev/null; then
+import socket, sys
+s = socket.socket()
+s.settimeout(0.3)
+try:
+    s.connect((sys.argv[1], int(sys.argv[2])))
+    sys.exit(0)
+except Exception:
+    sys.exit(1)
+finally:
+    s.close()
+PY
+      return 0
+    fi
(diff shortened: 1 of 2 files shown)


iho:
### Note: overlap with #3876

Both this PR and #3876 address #3867 with the same structural fix:

- drop `async_stream` accept path
- per-connection `handle_connection` + `tokio::select!`
- RAII cleanup so workers are always removed on disconnect

**#3876** is a smaller, cleaner refactor of that path (still uses unbounded queues; no idle timeout / worker cap).

**This PR** builds on the same idea and also bounds resources that keep half-open / abandoned sessions alive under reconnect load:

- 5m idle timeout
- max 256 workers
- bounded write queue + drop slow peers
- write timeout, max RPC line length
- worker_stats slot reuse + idempotent `remove_worker`

Happy to adjust the numeric limits if maintainers prefer different defaults. Also removed a stray `// temp` line that had snuck into the branch tip.

iho:
### Live reconnect / FD tests (now automated)

I initially only ran unit tests; the “run a node + reconnecting miner” items were manual and not executed yet. That was a gap.

They are now covered by automated tests in-tree:

```sh
cargo test -p grin_servers --lib mining::stratumserver::tests::test_live -- --test-threads=1
```

- `test_live_reconnect_storm_workers_and_fds_bounded` — 150 connect/login/drop cycles + 40 concurrent holds; asserts workers drain to 0, FD count does not track reconnects, worker_stats slots are reused  
- `test_live_max_workers_cap` — holds `MAX_STRATUM_WORKERS + 16` connections; asserts count ≤ 256 then drains to 0  

Both **passed** locally (`2 passed` in ~1s after compile).

Still not a full mainnet node + grin-miner run, but it exercises the real accept/`handle_connection` path under reconnect load.

iho:
### Full-node soak now run

```sh
./scripts/stratum_reconnect_soak.sh
```

**Result: PASS** (local)

| Metric | Value |
|--------|--------|
| Mode | `grin --usernet --no-tui` with stratum enabled, burn_reward, no seeds |
| Cycles | 120 connect/login/drop |
| Concurrent | 32 held then released |
| FD before → after | **40 → 40** (delta 0) |
| ESTABLISHED stratum TCP after close | **0** |
| panics / EMFILE | none |

This exercises the real binary accept path (not only the unit-test harness). Script lives at `scripts/stratum_reconnect_soak.sh`.

iho:
### Idle miner disconnect confirmed

Added `test_live_idle_miner_disconnected`:

- Uses the **same** `handle_connection` idle path as production
- Production constant remains **`WORKER_IDLE_TIMEOUT = 5 minutes`**
- Test injects an **800ms** idle timeout so CI does not wait 5 wall-clock minutes
- Asserts: worker registers after login → still present mid-idle → **count 0 after timeout** → peer sees EOF/error

```sh
cargo test -p grin_servers --lib mining::stratumserver::tests::test_live_idle_miner_disconnected -- --nocapture
# ok
```

Full suite: **14 passed**.

servers/src/mining/stratumserver.rs
```diff
@@ -605,70 +622,164 @@ fn accept_connections(listen_addr: SocketAddr, handler: Arc<Handler>) {
...
-				match listener.accept().await {
-					Ok((socket, _)) => yield socket,
-					Err(e) => {
-						error!("accept error = {:?}", e);
+
+		loop {
+			match listener.accept().await {
+				Ok((socket, peer_addr)) => {
+					// Hard cap concurrent workers to avoid FD exhaustion.
+					if handler.workers.count() >= MAX_STRATUM_WORKERS {
```
wiesche89, line 630: This check runs before the spawned task registers the worker, so a burst of accepts can exceed the limit. Could the cap be enforced atomically inside add_worker?

servers/src/mining/stratumserver.rs
```diff
@@ -777,19 +912,31 @@ impl WorkersList {
...
+			let workers = self.workers_list.read();
+			for (id, worker) in workers.iter() {
+				if worker.tx.try_send(msg.clone()).is_err() {
+					slow.push(*id);
+				}
+			}
+		}
+		for id in slow {
+			warn!("Stratum: dropping slow/disconnected worker {}", id);
+			self.remove_worker(id);
```
wiesche89, line 939: Since worker IDs are reused, the old connection guard could remove a new worker that inherited the same ID. Could removal verify a generation token, or be left to the connection task?

servers/src/mining/stratumserver.rs
```diff
@@ -711,28 +822,52 @@ impl WorkersList {
...
 	pub fn remove_worker(&self, worker_id: usize) {
-		self.update_stats(worker_id, |ws| ws.is_connected = false);
-		let mut stratum_stats = self.stratum_stats.write();
 		let mut workers_list = self.workers_list.write();
-		workers_list
-			.remove(&worker_id)
-			.expect("Stratum: no such addr in map");
+		if workers_list.remove(&worker_id).is_none() {
+			// Already removed (e.g. concurrent cleanup); still refresh counts.
+			let mut stratum_stats = self.stratum_stats.write();
```
wiesche89, line 858: add_worker takes stratum_stats before workers_list, but this branch takes them in the opposite order. Could we release workers_list before locking stratum_stats to avoid a possible deadlock?

servers/src/mining/stratumserver.rs
```diff
@@ -605,70 +622,164 @@ fn accept_connections(listen_addr: SocketAddr, handler: Arc<Handler>) {
...
-					handler.workers.remove_worker(worker_id);
-					info!("Worker {} disconnected", worker_id);
-				};
-				tokio::spawn(task);
+			msg = rx.recv() => {
+				match msg {
+					Some(line) => {
+						idle_deadline = Instant::now() + idle_timeout;
+						// Bound write time so a stalled peer cannot pin the task forever.
+						match timeout(Duration::from_secs(30), writer.send(line)).await {
```
wiesche89, line 746: could we make these 30 seconds a named constant as well, so all connection limits live in one place?

servers/src/mining/stratumserver.rs
```diff
@@ -605,70 +622,164 @@ fn accept_connections(listen_addr: SocketAddr, handler: Arc<Handler>) {
...
-					Result::<_, ()>::Ok(())
-				};
+	let framed = Framed::new(socket, LinesCodec::new_with_max_length(MAX_RPC_LINE_BYTES));
+	let (mut writer, mut reader) = framed.split();
+
+	let mut idle_deadline = Instant::now() + idle_timeout;
+
+	loop {
+		tokio::select! {
+			biased;
```
wiesche89, line 704: Is the read first bias intentional? A continuously readable client can starve the writer until its response queue fills and the connection is dropped.

scripts/stratum_reconnect_soak.sh
```diff
@@ -0,0 +1,306 @@
+#!/usr/bin/env bash
```
wiesche89, line 1: Please remove this script from the PR. The relevant reconnect, FD cleanup, worker cap, and idle behavior is already covered by the automated Rust tests.

wiesche89 (commented):
Thanks for addressing the leak. Since #3876 already provides the same structural fix in a smaller, more focused change, I think it would be the better base. We could add the relevant Rust regression tests there and handle the resource limits separately, as the current combined approach introduces new concurrency issues. The one off soak script should also stay out of the repository.

Issue #3550, opened 2021-01-21
Node API (3413) becomes unresponsive while compressing txhashset_zip


bladedoyle:
**Describe the bug**
Node stops responding to foreign api requests while doing a grin_util::zip - compress on hain_data/txhashset_zip

**To Reproduce**
Run a node with default settings
In a loop query the foreign api (get chain height for example) in a loop, every 5 seconds, reporting connection errors.
Match up the client connection timeouts with the node logs for compressing txhashset_zip

Its unacceptable for the node to become unresponsive.

PR #3890, opened 2026-07-09
fix: keep Node API responsive during txhashset zip

iho:
## Summary

Fixes [#3550](https://github.com/mimblewimble/grin/issues/3550): Node API becomes unresponsive while packaging a `txhashset` zip for a syncing peer.

### Cause

`Chain::txhashset_read` held **write** locks on `header_pmmr` and `txhashset` for the entire rewind + snapshot + multi-GB zip path. API handlers that need any chain access blocked for the whole packaging duration.

### Fix

1. **Write locks** only for rewind + snapshot (short, consistency-critical).
2. **Read locks** for `zip_read` (copy + package). Concurrent readers (Node API, headers, etc.) can proceed. Compaction/writers still wait until the zip finishes so backend files are not rewritten mid-package.

## Test plan

- [x] `cargo test -p grin_chain --test test_txhashset -- --test-threads=1`
- [ ] Manual: request txhashset from a peer while polling foreign API (chain height) every second — requests should continue to succeed during zip packaging

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -858,25 +858,36 @@ impl Chain {
 	/// Provides a reading view into the current txhashset state as well as
 	/// the required indexes for a consumer to rewind to a consistent state
 	/// at the provided block hash.
+	///
+	/// Write locks are only held while rewinding/snapshotting. Packaging the
+	/// multi-GB zip is done under *read* locks so the Node API and other
+	/// readers remain responsive while a peer is served a txhashset archive
+	/// (see #3550). Compaction and other writers wait until the zip finishes.
 	pub fn txhashset_read(&self, h: Hash) -> Result<(u64, u64, File), Error> {
-		// now we want to rewind the txhashset extension and
-		// sync a "rewound" copy of the leaf_set files to disk
-		// so we can send these across as part of the zip file.
-		// The fast sync client does *not* have the necessary data
-		// to rewind after receiving the txhashset zip.
+		// Rewind the txhashset extension and sync a "rewound" copy of the
+		// leaf_set files to disk so we can send these across as part of the zip.
+		// The fast sync client does *not* have the necessary data to rewind
+		// after receiving the txhashset zip.
 		let header = self.get_block_header(&h)?;
 
-		let mut header_pmmr = self.header_pmmr.write();
-		let mut txhashset = self.txhashset.write();
+		{
+			let mut header_pmmr = self.header_pmmr.write();
+			let mut txhashset = self.txhashset.write();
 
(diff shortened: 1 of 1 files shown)


chain/src/chain.rs
```diff
@@ -858,25 +858,36 @@ impl Chain {
...
-			// prepare the zip
-			txhashset::zip_read(self.db_root.clone(), &header)
-				.map(|file| (header.output_mmr_size, header.kernel_mmr_size, file))
-		})
+		// Shared locks: allow concurrent chain reads (API, header queries) while
+		// still preventing compaction from rewriting backend files mid-zip.
+		let _header_pmmr = self.header_pmmr.read();
+		let _txhashset = self.txhashset.read();
+
+		let file = txhashset::zip_read(self.db_root.clone(), &header)?;
```
wiesche89, line 889: zip_read was previously serialized by the write locks. It can now run concurrently for the same temp directory and zip path, so requests may clobber each other or reuse a partial archive. Could we serialize archive creation separately or use unique temp files with an atomic rename?

chain/src/chain.rs
```diff
@@ -858,25 +858,36 @@ impl Chain {
 	/// Provides a reading view into the current txhashset state as well as
 	/// the required indexes for a consumer to rewind to a consistent state
 	/// at the provided block hash.
+	///
+	/// Write locks are only held while rewinding/snapshotting. Packaging the
+	/// multi-GB zip is done under *read* locks so the Node API and other
```
wiesche89, line 863: The height request mentioned in #3550 goes through Chain::head() and does not take either of these locks. Since the multi-GB copy, zip write, and sync_all() remain unchanged, could we reproduce the original timeout before claiming this fixes it?

chain/src/chain.rs
```diff
@@ -858,25 +858,36 @@ impl Chain {
...
+				ext.extension.snapshot(batch)?;
+				Ok(())
+			})?;
+		}
 
-			// prepare the zip
-			txhashset::zip_read(self.db_root.clone(), &header)
-				.map(|file| (header.output_mmr_size, header.kernel_mmr_size, file))
-		})
+		// Shared locks: allow concurrent chain reads (API, header queries) while
```
wiesche89, line 884: This only allows concurrent readers until a writer is waiting. With parking_lot::RwLock, later readers then queue behind that writer for the rest of the zip. Could we test the API with a writer waiting during packaging?

wiesche89 (commented):
This also needs coordination with #3888, which changes cleanup in the same zip_read path. Concurrent creation and cleanup should share clear serialization and ownership.

Issue #3588, opened 2021-03-07
p2p: Replace Load&Store with CAS to avoid possible atomicity violation


BurtonQin:
**Describe the bug**
https://github.com/mimblewimble/grin/blob/35740204a1b212131afea502b49af4632aec8913/p2p/src/protocol.rs#L260-L265
Perhaps we can use `compare_exchange()` to replace the separate `load()` and `store()` to avoid atomicity violation ( `load()` and `store()` interleaved by another `store()` from another thread).

**To Reproduce**
Found it through static analysis.

**Relevant Information**
None

**Screenshots**
None

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version [18.04]

**Additional context**
None

antiochp:
Hi @BurtonQin thanks for the issue!
Let us look into this but I think your suggestion makes a lot of sense.

PR #3891, opened 2026-07-09
fix: claim state_sync_requested with compare_exchange

iho:
## Summary

Fixes [#3588](https://github.com/mimblewimble/grin/issues/3588).

When handling `TxHashSetArchive`, the code used a non-atomic check-then-clear:

```rust
if !state_sync_requested.load(...) { return Err(...) }
state_sync_requested.store(false, ...)
```

Two concurrent archive messages could both observe `true` and proceed.

### Change

Use `compare_exchange(true → false)` so only one handler claims the outstanding state-sync request.

## Test plan

- [x] `cargo test -p grin_p2p --lib -- --test-threads=1`

### DIFF
--- a/p2p/src/protocol.rs
+++ b/p2p/src/protocol.rs
@@ -289,12 +289,17 @@ impl MessageHandler for Protocol {
 					);
 					return Err(Error::BadMessage);
 				}
-				if !self.state_sync_requested.load(Ordering::Relaxed) {
+				// Atomically claim the outstanding state-sync request so two
+				// concurrent archive messages cannot both observe "requested"
+				// and proceed (load+store race; see #3588).
+				if self
+					.state_sync_requested
+					.compare_exchange(true, false, Ordering::SeqCst, Ordering::Relaxed)
+					.is_err()
+				{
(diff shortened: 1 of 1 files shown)


p2p/src/protocol.rs
```diff
@@ -289,12 +289,17 @@ impl MessageHandler for Protocol {
 					);
 					return Err(Error::BadMessage);
 				}
-				if !self.state_sync_requested.load(Ordering::Relaxed) {
+				// Atomically claim the outstanding state-sync request so two
+				// concurrent archive messages cannot both observe "requested"
+				// and proceed (load+store race; see #3588).
+				if self
+					.state_sync_requested
+					.compare_exchange(true, false, Ordering::SeqCst, Ordering::Relaxed)
```
wiesche89, line 297: Issue #3588 raised a static analysis concern about another thread updating the flag between the load and store. The race described in this PR does not appear reachable, since archive responses from a peer are processed one at a time. Could we clarify which concrete interleaving this fixes? If it is a delayed response overlapping a newer request, this boolean still cannot tell them apart.

PR #3893, opened 2026-07-09
fix: fall back to defaults for missing config keys

iho:
## Summary

Fixes [#3002](https://github.com/mimblewimble/grin/issues/3002).

Operators who trim `grin-server.toml` to only non-default keys hit parse failures for missing fields such as `log_to_file`, even though the code has sensible defaults.

### Change

- `LoggingConfig`: `#[serde(default)]` + field defaults matching `Default` (so omitting `log_to_file` → `true`, etc.)
- `ConfigMembers`: optional `logging` / `config_file_version` with serde default
- `ServerConfig` / `StratumServerConfig`: missing keys use the same values as `Default`

Minimal configs can now omit keys safely; present keys still override.

## Test plan

- [x] `cargo test -p grin_config -- --test-threads=1` (9 passed), including:
  - `test_logging_config_missing_keys_use_defaults`
  - `test_logging_section_optional`
  - `test_stratum_partial_config_defaults`

### DIFF
--- a/config/src/types.rs
+++ b/config/src/types.rs
@@ -87,12 +87,15 @@ pub struct GlobalConfig {
 /// internal state that we don't necessarily
 /// want serialised or deserialised
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[serde(default)]
 pub struct ConfigMembers {
 	/// Config file version (None == version 1)
+	#[serde(default)]
 	pub config_file_version: Option<u32>,
 	/// Server config
 	#[serde(default)]
 	pub server: ServerConfig,
-	/// Logging config
+	/// Logging config (optional section; missing keys use LoggingConfig defaults)
+	#[serde(default)]
 	pub logging: Option<LoggingConfig>,
 }
--- a/config/src/types.rs
+++ b/config/src/types.rs
@@ -87,7 +87,7 @@ pub struct GlobalConfig {
 /// internal state that we don't necessarily
 /// want serialised or deserialised
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
-#[serde(default)]
+#[serde(default, deny_unknown_fields)]
 pub struct ConfigMembers {
 	/// Config file version (None == version 1)
 	#[serde(default)]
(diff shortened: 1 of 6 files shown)


config/src/types.rs
```diff
@@ -87,12 +87,15 @@ pub struct GlobalConfig {
...
 pub struct ConfigMembers {
 	/// Config file version (None == version 1)
+	#[serde(default)]
 	pub config_file_version: Option<u32>,
 	/// Server config
 	#[serde(default)]
 	pub server: ServerConfig,
-	/// Logging config
+	/// Logging config (optional section; missing keys use LoggingConfig defaults)
+	#[serde(default)]
```
wiesche89: A missing logging section now becomes None, but node startup still calls .logging.clone().unwrap(). The minimal config therefore parses and then panics. Should this default to Some(LoggingConfig::default()) instead?

servers/src/common/types.rs
```diff
@@ -138,23 +138,32 @@ impl Default for ChainValidationMode {
 
 /// Full server configuration, aggregating configurations required for the
 /// different components.
+///
+/// Missing optional keys fall back to [`Default`] (see #3002).
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+#[serde(default)]
```
wiesche89: These are always the mainnet defaults. A trimmed testnet or usernet config therefore misses the network-specific API, P2P, and stratum values from GlobalConfig::for_chain(). Could we apply defaults for the selected chain instead?

wiesche89: With these defaults, a misspelled key can be silently ignored and replaced by its default, which is risky for settings such as archive_mode. Could we reject or at least report unknown fields?

config/src/types.rs
iho: Done in 98b73d44. Omitted `[logging]` now deserializes to `Some(LoggingConfig::default())` (via `default_logging` / post-parse guard), so `logging.clone().unwrap()` on startup is safe.

servers/src/common/types.rs
iho: Done in 98b73d44. Config load now deep-merges the file onto `GlobalConfig::for_chain(chain_type)` (chain type taken from the file, defaulting to Mainnet). A trimmed testnet/usernet config therefore gets the network-specific API, P2P, and stratum defaults while still overriding only the keys present in the file.

iho: Done in 98b73d44. Added `#[serde(deny_unknown_fields)]` on the config structs (`ConfigMembers`, `ServerConfig`, stratum, webhooks, logging, p2p, pool, dandelion). Typos such as `archive_mod` now fail to parse instead of being ignored. Covered by `test_unknown_config_key_rejected`.

Issue #3546, opened 2021-01-14
Node starts API and stratum servers too soon


bladedoyle:
**Describe the bug**
The grin node starts API and stratum servers before its ready to handle requests.
The result is confusing because the connections are accepted but result in an error, so the node seems broken at that point. 
 For example, the wallet reports this:

```
$ ./grin-wallet info
Password: 
20201229 10:39:47.871 WARN grin_wallet_libwallet::internal::updater - Not updating outputs as the height of the node's chain is less than the last reported wallet update height.
20201229 10:39:47.871 WARN grin_wallet_libwallet::internal::updater - Please wait for sync on node to complete or fork to resolve and try again.
20201229 10:39:47.943 WARN grin_wallet_libwallet::api_impl::owner_updater - Scanning - 0% complete
20201229 10:39:48.214 ERROR grin_wallet_impls::node_clients::http - Response { result: Some(Object({"Err": String("NotFound")})), error: None, id: Number(1), jsonrpc: Some("2.0") }
20201229 10:39:48.214 ERROR grin_wallet_impls::node_clients::http - Unable to parse response for get_pmmr_indices: JSON decode error: invalid type: null, expected struct OutputListing
Wallet command failed: LibWallet Error: Client Callback Error: Unable to parse response for get_pmmr_indices: JSON decode error: invalid type: null, expected struct OutputListing
```

**To Reproduce**
1. Start grin node
2. Immediately after the node reports "grin_api::handlers - HTTP Node listener started." run "grin-wallet info"
3. see errors 

**Relevant Information**
The api (and other servers) are started before the node is ready to handle requests
We should either not start the servers until the node is ready, or return an error message that it is "initializing" or something informative.

PR #3894, opened 2026-07-09
fix: clearer readiness while node is still syncing

iho:
## Summary

Fixes [#3546](https://github.com/mimblewimble/grin/issues/3546).

API and stratum used to accept connections immediately while the chain was still initialising/syncing. Wallets then saw opaque `NotFound` / null decode errors (e.g. on `get_pmmr_indices`); miners connected only to get “node is syncing” job failures.

### Changes

| Area | Behaviour |
|------|-----------|
| **Stratum** | Do not bind/listen until `!sync_state.is_syncing()`; set `is_running` only then |
| **Foreign API** | `get_pmmr_indices` / `get_outputs` / `get_unspent_outputs` return `Error::Unavailable` (“node is still syncing…”) while syncing |
| **REST** | Map `Unavailable` → **HTTP 503** |
| **Status / tip / version** | Unchanged so monitoring still works during sync |
| **PMMR map errors** | Surface the real chain error instead of empty NotFound |

## Test plan

- [x] `cargo test -p grin_api --lib -- --test-threads=1` (includes `ensure_not_syncing_*`)
- [x] `cargo test -p grin_servers --lib mining::stratumserver -- --test-threads=1`
- [ ] Start node mid-sync: wallet `get_pmmr_indices` should report syncing / 503, not null NotFound
- [ ] Stratum port should not accept until sync completes

### DIFF
--- a/api/src/foreign.rs
+++ b/api/src/foreign.rs
@@ -21,6 +21,7 @@ use crate::handlers::blocks_api::{BlockHandler, HeaderHandler};
 use crate::handlers::chain_api::{ChainHandler, KernelHandler, OutputHandler};
 use crate::handlers::pool_api::PoolHandler;
 use crate::handlers::transactions_api::TxHashSetHandler;
+use crate::handlers::utils::ensure_not_syncing;
 use crate::handlers::version_api::VersionHandler;
 use crate::pool::{self, BlockChain, PoolAdapter, PoolEntry};
 use crate::types::{
@@ -236,6 +237,7 @@ where
 		include_proof: Option<bool>,
 		_include_merkle_proof: Option<bool>,
 	) -> Result<Vec<OutputPrintable>, Error> {
+		ensure_not_syncing(&self.sync_state)?;
 		let output_handler = OutputHandler {
 			chain: self.chain.clone(),
 		};
@@ -267,6 +269,7 @@ where
 		max: u64,
 		include_proof: Option<bool>,
 	) -> Result<OutputListing, Error> {
+		ensure_not_syncing(&self.sync_state)?;
 		let output_handler = OutputHandler {
 			chain: self.chain.clone(),
 		};
@@ -288,6 +291,8 @@ where
 		start_block_height: u64,
 		end_block_height: Option<u64>,
 	) -> Result<OutputListing, Error> {
+		// Wallet scan/info hits this early; return a clear signal while syncing (#3546).
+		ensure_not_syncing(&self.sync_state)?;
 		let txhashset_handler = TxHashSetHandler {
 			chain: self.chain.clone(),
 		};
--- a/api/src/handlers/transactions_api.rs
+++ b/api/src/handlers/transactions_api.rs
@@ -106,7 +106,13 @@ impl TxHashSetHandler {
 		let chain = w(&self.chain)?;
 		let range = chain
 			.block_height_range_to_pmmr_indices(start_block_height, end_block_height)
-			.map_err(|_| Error::NotFound)?;
+			.map_err(|e| {
+				// Prefer a descriptive error over empty NotFound (wallet misreports null).
+				Error::Argument(format!(
+					"cannot map block heights {}..{:?} to PMMR indices: {}",
+					start_block_height, end_block_height, e
+				))
+			})?;
 		let out = OutputListing {
 			last_retrieved_index: range.0,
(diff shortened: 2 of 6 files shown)


api/src/foreign.rs
```diff
@@ -288,6 +291,8 @@ where
 		start_block_height: u64,
 		end_block_height: Option<u64>,
 	) -> Result<OutputListing, Error> {
+		// Wallet scan/info hits this early; return a clear signal while syncing (#3546).
+		ensure_not_syncing(&self.sync_state)?;
```
wiesche89, line 295: The current grin-wallet client only deserializes result["Ok"] and ignores result["Err"]. This therefore still becomes null and produces the same OutputListing decode error. Could we test this against the actual wallet response handling?

api/src/web.rs
```diff
@@ -52,6 +52,7 @@ where
 			Error::Argument(msg) => response(StatusCode::BAD_REQUEST, msg),
 			Error::RequestError(msg) => response(StatusCode::BAD_REQUEST, msg),
 			Error::NotFound => response(StatusCode::NOT_FOUND, "".into()),
+			Error::Unavailable(msg) => response(StatusCode::SERVICE_UNAVAILABLE, msg),
```
wiesche89, line 55: None of the changed endpoints reaches this REST path. /v2/foreign still returns HTTP 200. Is the 503 mapping meant to be used elsewhere?

api/src/handlers/transactions_api.rs
```diff
@@ -106,7 +106,13 @@ impl TxHashSetHandler {
 		let chain = w(&self.chain)?;
 		let range = chain
 			.block_height_range_to_pmmr_indices(start_block_height, end_block_height)
-			.map_err(|_| Error::NotFound)?;
+			.map_err(|e| {
+				// Prefer a descriptive error over empty NotFound (wallet misreports null).
+				Error::Argument(format!(
```
wiesche89, line 111: Internal store or PMMR errors can also reach this path. Could we avoid reporting all chain errors as bad arguments?

servers/src/mining/stratumserver.rs
```diff
@@ -878,11 +878,23 @@ impl StratumServer {
...
+		while self.sync_state.is_syncing() {
+			thread::sleep(Duration::from_millis(50));
+		}
+
 		let _listener_th = thread::spawn(move || {
 			accept_connections(listen_addr, h);
 		});
 
-		// We have started
+		// We have started (only after sync so is_running matches actual readiness).
```
wiesche89, line 897: The listener binds in the spawned thread, so this flag can still become true before bind succeeds. Could startup report success before we mark Stratum as running?

servers/src/mining/stratumserver.rs
```diff
@@ -878,11 +878,23 @@ impl StratumServer {
 		let handler = Arc::new(Handler::from_stratum(&self));
 		let h = handler.clone();
 
+		// Wait until the node has finished syncing before accepting miner connections
```
wiesche89, line 881: Could we shorten this to describe the current behavior and leave the issue history in the PR?

api/src/foreign.rs
```diff
@@ -288,6 +291,8 @@ where
 		start_block_height: u64,
 		end_block_height: Option<u64>,
 	) -> Result<OutputListing, Error> {
+		// Wallet scan/info hits this early; return a clear signal while syncing (#3546).
```
wiesche89, line 294: Could we shorten this to describe the current behavior and leave the issue history in the PR?

api/src/foreign.rs
```diff
@@ -236,6 +237,7 @@ where
 		include_proof: Option<bool>,
 		_include_merkle_proof: Option<bool>,
 	) -> Result<Vec<OutputPrintable>, Error> {
+		ensure_not_syncing(&self.sync_state)?;
```
wiesche89, line 240: The current grin-wallet client only reads result["Ok"], so these Err responses still become null and produce decode errors. Could we test all three methods through the actual wallet RPC handling?

wiesche89 (commented):
The readiness direction makes sense, but the wallet still sees null decode errors because its JSON-RPC client only reads result["Ok"]. The new 503 mapping is not reached by these methods. I think the actual wallet/RPC response should be tested before this is considered fixed.

Issue #3842, opened 2026-05-26
Wrong root on launch if node will be stopped on compaction


ardocrat:
Node launch can take a lot of time with low RAM amount, so if it will be stopped manually or killed we get an error after second launch: `Error returned, discarding txhashset extension: Invalid Root` and starting to rewind a header till the end. It happens while compaction is going (this process is very slow and can take several minutes on launch), also as we can see at logs this process is not stopping after server shutdown:

<img width="1902" height="1294" alt="Image" src="https://github.com/user-attachments/assets/6325e062-0dbc-419e-be4f-90a52c826f92" />

PR #3895, opened 2026-07-09
fix: abort chain compaction cleanly on node shutdown

iho:
## Summary

Fixes [#3842](https://github.com/mimblewimble/grin/issues/3842).

Long compaction runs (especially on low-RAM machines / startup) continued after the node was asked to stop. Killing the process while live PMMR files were being replaced left the chain with **Invalid Root** on the next launch.

### Changes

| Layer | Behaviour |
|-------|-----------|
| `NetToChainAdapter` | Holds `StopState`; does not spawn compactors when stopped; passes stop into compact |
| `Syncer` | Uses `compact_with_stop` on transition to NoSync |
| `Chain::compact_with_stop` | Aborts early when stop is set |
| `PMMRBackend::check_compact_until` | After writing `.tmp` files, if aborting: **discard tmp**, do **not** replace live files |
| `AppendOnlyFile::discard_tmp` | Removes leftover `.tmp` companions |

`Chain::compact()` remains for API/tests (no stop handle).

## Test plan

- [x] `cargo test -p grin_store --test pmmr pmmr_compact_abort_before_replace`
- [x] `cargo test -p grin_store --test pmmr -- --test-threads=1`
- [x] `cargo test -p grin_chain --test mine_simple_chain compact -- --test-threads=1`
- [x] `cargo test -p grin_servers --lib -- --test-threads=1`
- [ ] Manual: trigger compact, stop node mid-run, confirm no Invalid Root on restart

### DIFF
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1329,6 +1329,23 @@ impl Chain {
 	/// * removes historical blocks and associated data from the db (unless archive mode)
 	///
 	pub fn compact(&self) -> Result<(), Error> {
+		self.compact_with_stop(None)
+	}
+
+	/// Compact the chain, optionally aborting if `stop_state` is stopped.
+	/// Used on shutdown so long compaction does not keep rewriting files after
+	/// the node was asked to stop (#3842).
+	pub fn compact_with_stop(
+		&self,
+		stop_state: Option<Arc<crate::util::StopState>>,
+	) -> Result<(), Error> {
+		let should_abort = || stop_state.as_ref().map(|s| s.is_stopped()).unwrap_or(false);
+
+		if should_abort() {
+			info!("compact: not starting, node is stopping");
+			return Ok(());
+		}
+
 		// A node may be restarted multiple times in a short period of time.
 		// We compact at most once per 60 blocks in this situation by comparing
 		// current "head" and "tail" height to our cut-through horizon and
@@ -1364,7 +1381,13 @@ impl Chain {
 			let horizon_hash = header_pmmr.get_header_hash_by_height(horizon_height)?;
 			let horizon_header = batch.get_block_header(&horizon_hash)?;
 
-			txhashset.compact(&horizon_header, &batch)?;
+			txhashset.compact_until(&horizon_header, &batch, should_abort)?;
+		}
+
+		if should_abort() {
+			info!("compact: aborted after txhashset compact (node stopping)");
+			// Do not commit further DB changes if we aborted mid-flight.
+			return Ok(());
 		}
 
 		// If we are not in archival mode remove historical blocks from the db.
--- a/chain/src/chain.rs
+++ b/chain/src/chain.rs
@@ -1334,7 +1334,7 @@ impl Chain {
 
 	/// Compact the chain, optionally aborting if `stop_state` is stopped.
 	/// Used on shutdown so long compaction does not keep rewriting files after
-	/// the node was asked to stop (#3842).
+	/// the node was asked to stop.
 	pub fn compact_with_stop(
 		&self,
 		stop_state: Option<Arc<crate::util::StopState>>,
@@ -1381,7 +1381,7 @@ impl Chain {
 			let horizon_hash = header_pmmr.get_header_hash_by_height(horizon_height)?;
 			let horizon_header = batch.get_block_header(&horizon_hash)?;
 
-			txhashset.compact_until(&horizon_header, &batch, should_abort)?;
+			txhashset.compact_with_abort(&horizon_header, &batch, should_abort)?;
 		}
 
 		if should_abort() {
--- a/servers/src/common/adapters.rs
+++ b/servers/src/common/adapters.rs
@@ -15,14 +15,14 @@
 //! Adapters connecting new block, new transaction, and accepted transaction
 //! events to consumers of those events.
 
-use crate::util::{RwLock, StopState};
+use crate::util::{Mutex, RwLock, StopState};
 use std::collections::HashMap;
 use std::fs::File;
 use std::net::SocketAddr;
 use std::panic::{catch_unwind, AssertUnwindSafe};
 use std::path::PathBuf;
 use std::sync::{mpsc, Arc, Weak};
-use std::thread;
+use std::thread::{self, JoinHandle};
 use std::time::Instant;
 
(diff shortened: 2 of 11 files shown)


store/src/pmmr.rs
```diff
@@ -447,6 +467,15 @@ impl<T: PMMRable> PMMRBackend<T> {
 			self.data_file.write_tmp_pruned(&pos_to_rm)?;
 		}
 
+		// Critical section: do not replace live files if we are shutting down.
```
wiesche89: A stop can arrive immediately after this check, while the files are replaced separately. Is shutdown guaranteed to wait until the whole replace section finishes?

servers/src/common/adapters.rs
```diff
@@ -966,15 +970,25 @@ where
...
+			let stop_state = self.stop_state.clone();
 			let _ = thread::Builder::new()
 				.name("compactor".to_string())
 				.spawn(move || {
-					if let Err(e) = chain.compact() {
+					if stop_state.is_stopped() {
+						debug!("compactor: not starting, node is stopping");
+						return;
+					}
+					if let Err(e) = chain.compact_with_stop(Some(stop_state)) {
```
wiesche89, line 1042: This compactor still runs on a detached thread, so shutdown cannot wait if it has already entered the replace phase. Could the thread be tracked and joined?

store/tests/pmmr.rs
```diff
@@ -214,6 +214,36 @@ fn pmmr_compact_leaf_sibling() {
...
+		let mmr_size = load(0, &elems[..], &mut backend);
+		backend.sync().unwrap();
+		let root_before = {
+			let pmmr = PMMR::at(&mut backend, mmr_size);
+			pmmr.root().unwrap()
+		};
+
+		// Always abort: should not replace live files.
+		let done = backend
+			.check_compact_until(2, &Bitmap::new(), || true)
```
wiesche89: || true trips the first check before any tmp file is written, so the discard path never runs. Could it return false first, then true, and verify that no tmp files remain?

chain/src/chain.rs
```diff
@@ -1329,6 +1329,23 @@ impl Chain {
 	/// * removes historical blocks and associated data from the db (unless archive mode)
 	///
 	pub fn compact(&self) -> Result<(), Error> {
+		self.compact_with_stop(None)
```
wiesche89, line 1332: The owner API still reaches this uncancellable path. Could an API-triggered compaction hit the same shutdown problem?

store/src/types.rs
```diff
@@ -521,6 +527,16 @@ where
 		Ok(())
 	}
 
+	/// Drop any `.tmp` companion file without replacing the live file.
+	pub fn discard_tmp(&self) {
```
wiesche89: Could this return io::Result<()>? Otherwise the caller reports a clean abort even when tmp cleanup fails.

store/src/pmmr.rs
```diff
@@ -414,8 +414,28 @@ impl<T: PMMRable> PMMRBackend<T> {
 	/// will have a suitable output_pos. This is used to enforce a horizon
 	/// after which the local node should have all the data to allow rewinding.
 	pub fn check_compact(&mut self, cutoff_pos: u64, rewind_rm_pos: &Bitmap) -> io::Result<bool> {
+		self.check_compact_until(cutoff_pos, rewind_rm_pos, || false)
+	}
+
+	/// Compact backend files, optionally aborting before live files are replaced
+	/// when `should_abort` returns true (e.g. node shutdown). Aborted compaction
+	/// discards `.tmp` files and leaves the live PMMR files untouched (#3842).
+	pub fn check_compact_until<F>(
```
wiesche89: Small naming thought, until sounds like a position limit. Would check_compact_with_abort describe this callback more clearly?

store/tests/pmmr.rs
```diff
@@ -214,6 +214,36 @@ fn pmmr_compact_leaf_sibling() {
 	teardown(data_dir);
 }
 
+/// Aborting compaction before replace must leave live files intact (#3842).
```
wiesche89: Could we keep this focused on the invariant and leave the issue reference in the PR?

wiesche89 (commented):
Good direction, but shutdown still needs to wait for the replace phase, and the test should exercise the actual tmp discard path.

store/src/pmmr.rs
iho: Done in b12bfca3. Once the pre-replace abort check decides to proceed, replace always runs to completion (no further abort polls). Shutdown now joins the compactor thread (`CompactorTracker` / `Server::stop`) so the process does not exit mid-replace.

servers/src/common/adapters.rs
iho: Done in b12bfca3. Added `CompactorTracker` that stores the `JoinHandle`. The adapter reaps finished handles and only starts one compact at a time; `Server::stop` joins the tracker after signalling stop so any in-flight compact can finish or abort cleanly.

store/tests/pmmr.rs
iho: Done in b12bfca3. The abort callback returns false on the first call (allow tmp write) and true on the second (abort before replace). The test asserts the second check ran and that no `.tmp` files remain under the data dir.

chain/src/chain.rs
iho: Done in b12bfca3. `stop_state` is plumbed into `Owner` / `ChainCompactHandler`; API compact uses `compact_with_stop` so a shutdown during an owner-triggered compact can abort before replace the same way as the background path.

store/src/types.rs
iho: Done in b12bfca3. `discard_tmp` now returns `io::Result<()>`, and the abort path propagates cleanup failures with `?` instead of treating them as a clean abort.

store/src/pmmr.rs
iho: Done in b12bfca3. Renamed to `check_compact_with_abort` (and `compact_with_abort` on the txhashset) so it is clear the callback is an abort flag, not a position limit.

store/tests/pmmr.rs
iho: Done in b12bfca3. Test comment now states the invariant only (abort after tmp write must discard tmp and leave live roots intact); the issue reference stays in the PR description.

PR #3896, opened 2026-07-09
fix: use FutureTimeLimit instead of CorruptedData for clock skew

iho:
## Summary

Fixes [#3360](https://github.com/mimblewimble/grin/issues/3360).

`UntrustedBlockHeader` rejected headers past `future_time_limit` with `ser::Error::CorruptedData`. That label is misleading: the bytes are well-formed; local clocks may simply disagree (miners near the FTL boundary). Maintainers already noted we need more informative deserialization errors for this case.

### Change

- Add `ser::Error::FutureTimeLimit`
- Use it (with a `warn!`) when a header timestamp is beyond FTL
- Keep real corruption / invalid PoW paths on `CorruptedData`
- Unit test: future header deserializes to `FutureTimeLimit`, not `CorruptedData`

P2P still treats all serialization errors as non-bannable (`Error::Serialization`); this mainly improves logging/clarity and future handling if finer peer policy is added.

## Test plan

- [x] `cargo test -p grin_core --test block deserialize_untrusted_header_future_time`
- [x] `cargo test -p grin_core --test block -- --test-threads=1` (28 passed)

### DIFF
--- a/core/src/core/block.rs
+++ b/core/src/core/block.rs
@@ -448,14 +448,16 @@ impl Readable for UntrustedBlockHeader {
 		let header = read_block_header(reader)?;
 		let ftl = global::get_future_time_limit();
 		if header.timestamp > Utc::now() + Duration::seconds(ftl as i64) {
-			// refuse blocks whose timestamp is too far in the future
-			// this future_time_limit (FTL) is specified in grin-server.toml
-			// TODO add warning in p2p code if local time is too different from peers
-			error!(
-				"block header {} validation error: block time is more than {} seconds in the future",
+			// Refuse blocks whose timestamp is too far in the future.
+			// FTL is configured in grin-server.toml. This is *not* corrupted data —
+			// local clocks differ; use a dedicated error so peers are not treated
+			// as sending garbage (#3360).
+			// TODO: warn in p2p if local time is systematically different from peers
+			warn!(
+				"block header {} rejected: block time is more than {} seconds in the future (clock skew?)",
 				header.hash(), ftl
 			);
-			return Err(ser::Error::CorruptedData);
+			return Err(ser::Error::FutureTimeLimit);
 		}
 
 		// Check the block version before proceeding any further.
--- a/core/src/ser.rs
+++ b/core/src/ser.rs
@@ -72,6 +72,10 @@ pub enum Error {
 	InvalidBlockVersion,
 	/// Unsupported protocol version
 	UnsupportedProtocolVersion,
+	/// Block/header timestamp is beyond the future time limit.
+	/// Not corrupted data — peer clocks may simply disagree; treat as non-fatal
+	/// for peer reputation (see #3360).
+	FutureTimeLimit,
 }
 
 impl From<io::Error> for Error {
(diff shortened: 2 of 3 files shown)


core/src/core/block.rs
```diff
@@ -448,14 +448,16 @@ impl Readable for UntrustedBlockHeader {
...
+			// FTL is configured in grin-server.toml. This is *not* corrupted data —
+			// local clocks differ; use a dedicated error so peers are not treated
+			// as sending garbage (#3360).
+			// TODO: warn in p2p if local time is systematically different from peers
+			warn!(
+				"block header {} rejected: block time is more than {} seconds in the future (clock skew?)",
 				header.hash(), ftl
 			);
-			return Err(ser::Error::CorruptedData);
+			return Err(ser::Error::FutureTimeLimit);
```
wiesche89, line 460: This still becomes a generic P2P serialization error and closes the connection. Is this intentionally diagnostics-only, or should future timestamps be handled without dropping the peer?

core/src/ser.rs
```diff
@@ -72,6 +72,10 @@ pub enum Error {
 	InvalidBlockVersion,
 	/// Unsupported protocol version
 	UnsupportedProtocolVersion,
+	/// Block/header timestamp is beyond the future time limit.
+	/// Not corrupted data — peer clocks may simply disagree; treat as non-fatal
```
wiesche89, line 76: P2P currently treats all serialization errors the same, so this variant does not change peer reputation yet. Could we keep the comment focused on the timestamp condition?

core/tests/block.rs
```diff
@@ -466,6 +466,30 @@ fn deserialize_untrusted_header_weight() {
...
+#[test]
+fn deserialize_untrusted_header_future_time() {
+	test_setup();
+	let keychain = ExtKeychain::from_random_seed(false).unwrap();
+	let builder = ProofBuilder::new(&keychain);
+	let prev = BlockHeader::default();
+	let key_id = ExtKeychain::derive_key_id(1, 1, 0, 0, 0);
+	let mut b = new_block(&[], &keychain, &builder, &prev, &key_id);
+
+	// Far beyond future_time_limit (default 300s for AutomatedTesting? check global)
```
wiesche89, line 479: The default is 300 seconds, so the question can go. Since timestamp is checked before PoW, could this use a plain header and avoid mining a proof?

core/src/core/block.rs
```diff
@@ -448,14 +448,16 @@ impl Readable for UntrustedBlockHeader {
 		let header = read_block_header(reader)?;
 		let ftl = global::get_future_time_limit();
 		if header.timestamp > Utc::now() + Duration::seconds(ftl as i64) {
-			// refuse blocks whose timestamp is too far in the future
-			// this future_time_limit (FTL) is specified in grin-server.toml
-			// TODO add warning in p2p code if local time is too different from peers
-			error!(
-				"block header {} validation error: block time is more than {} seconds in the future",
+			// Refuse blocks whose timestamp is too far in the future.
```
wiesche89, line 451: Could we shorten this to describe the rejection reason and leave the issue history in the PR?

wiesche89 (commented):
The new error is clearer, but it remains diagnostics-only. P2P still closes the connection for every serialization error. The PR description should make that scope clear, and the test can be simplified because timestamp is checked before PoW.

PR #3897, opened 2026-07-09
feat: expose stratum mining stats via Owner API

### LINKED ISSUE #3201: Add mining endpoint to Node API
Feature request: add a "Mining Endpoint" to the [Node API](https://github.com/mimblewimble/grin/blob/master/doc/api/node_api.md), that could be used to retrieve the stats currently available on the Mining tab of the TUI.
Would also be handy to get this into `grin client`.

Currently, when [running grin in headless mode](https://github.com/mimblewimble/docs/wiki/Build-and-run-automation#set-up-grin-as-a-service), there is no way to retrieve the mining statistics such as the connected workers.

iho:
## Summary

Adds an Owner API endpoint for stratum mining statistics so operators running headless (no TUI) can inspect connected workers and share stats — the same data shown on the TUI mining tab.

Closes #3201

### Changes
- New `get_mining_status` method on Owner JSON-RPC (`/v2/owner`)
- New types: `MiningStatus` / `WorkerInfo` in the API crate
- Server wires live `StratumStats` into the Owner API via a provider callback
- CLI: `grin client miningstatus`
- Brief docs in `doc/api/api.md`

### Example

```json
{
  "jsonrpc": "2.0",
  "method": "get_mining_status",
  "params": [],
  "id": 1
}
```

Returns whether stratum is enabled/running, worker count, block height, network difficulty/hashrate, blocks found, and per-worker share stats for currently connected workers. When stratum is disabled, a provider still returns `is_enabled: false`. If no mining stats provider is wired into the Owner API, `get_mining_status` returns an internal error instead of masking the misconfiguration as disabled.

## Test plan
- [x] Unit tests: `MiningStatus` serde roundtrip
- [x] Unit tests: `Owner::get_mining_status` without provider (error) and with provider via generated OwnerRpc JSON-RPC handler (wire response)
- [x] Unit tests: `StratumStats` → `MiningStatus` conversion (filters disconnected workers)
- [x] `cargo check -p grin --bin grin`
- [ ] Manually: enable stratum, connect a miner, run `grin client miningstatus`

### DIFF
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -62,6 +62,39 @@ pub fn node_apis<B, P>(
 	tls_config: Option<TLSConfig>,
 	api_chan: (mpsc::Sender<()>, mpsc::Receiver<()>),
 	stop_state: Arc<StopState>,
+) -> Result<(), Error>
+where
+	B: BlockChain + 'static,
+	P: PoolAdapter + 'static,
+{
+	node_apis_with_mining_stats(
+		addr,
+		chain,
+		tx_pool,
+		peers,
+		sync_state,
+		api_secret,
+		foreign_api_secret,
+		tls_config,
+		api_chan,
+		stop_state,
+		None,
+	)
+}
+
+/// Same as [`node_apis`] but additionally wires a provider of live stratum
+/// mining stats into the Owner API `get_mining_status` method.
+pub fn node_apis_with_mining_stats<B, P>(
+	addr: &str,
+	chain: Arc<Chain>,
+	tx_pool: Arc<RwLock<pool::TransactionPool<B, P>>>,
+	peers: Arc<p2p::Peers>,
+	sync_state: Arc<SyncState>,
+	api_secret: Option<String>,
+	foreign_api_secret: Option<String>,
+	tls_config: Option<TLSConfig>,
+	api_chan: (mpsc::Sender<()>, mpsc::Receiver<()>),
+	stop_state: Arc<StopState>,
 	mining_stats: Option<MiningStatsProvider>,
 ) -> Result<(), Error>
 where
--- a/api/src/handlers.rs
+++ b/api/src/handlers.rs
@@ -27,7 +27,7 @@ use crate::auth::{
 use crate::chain::{Chain, SyncState};
 use crate::foreign::Foreign;
 use crate::foreign_rpc::ForeignRpc;
-use crate::owner::Owner;
+use crate::owner::{MiningStatsProvider, Owner};
 use crate::owner_rpc::OwnerRpc;
 use crate::pool;
 use crate::pool::{BlockChain, PoolAdapter};
@@ -62,6 +62,7 @@ pub fn node_apis<B, P>(
 	tls_config: Option<TLSConfig>,
 	api_chan: (mpsc::Sender<()>, mpsc::Receiver<()>),
 	stop_state: Arc<StopState>,
+	mining_stats: Option<MiningStatsProvider>,
 ) -> Result<(), Error>
 where
 	B: BlockChain + 'static,
@@ -85,6 +86,7 @@ where
 		Arc::downgrade(&chain),
 		Arc::downgrade(&peers),
 		Arc::downgrade(&sync_state),
+		mining_stats,
 	);
 	router.add_route("/v2/owner", Arc::new(api_handler))?;
 
@@ -142,25 +144,33 @@ pub struct OwnerAPIHandlerV2 {
 	pub chain: Weak<Chain>,
 	pub peers: Weak<p2p::Peers>,
 	pub sync_state: Weak<SyncState>,
+	pub mining_stats: Option<MiningStatsProvider>,
 }
 
 impl OwnerAPIHandlerV2 {
 	/// Create a new owner API handler for GET methods
-	pub fn new(chain: Weak<Chain>, peers: Weak<p2p::Peers>, sync_state: Weak<SyncState>) -> Self {
+	pub fn new(
+		chain: Weak<Chain>,
+		peers: Weak<p2p::Peers>,
+		sync_state: Weak<SyncState>,
+		mining_stats: Option<MiningStatsProvider>,
+	) -> Self {
 		OwnerAPIHandlerV2 {
 			chain,
 			peers,
 			sync_state,
+			mining_stats,
 		}
 	}
 }
 
 impl crate::router::Handler for OwnerAPIHandlerV2 {
 	fn post(&self, req: Request<Incoming>) -> ResponseFuture {
-		let api = Owner::new(
+		let api = Owner::with_mining_stats(
 			self.chain.clone(),
 			self.peers.clone(),
 			self.sync_state.clone(),
+			self.mining_stats.clone(),
 		);
 
 		Box::pin(async move {
--- a/api/src/owner.rs
+++ b/api/src/owner.rs
@@ -22,9 +22,12 @@ use crate::handlers::server_api::StatusHandler;
 use crate::p2p::types::PeerInfoDisplay;
 use crate::p2p::{self, PeerData};
 use crate::rest::*;
-use crate::types::Status;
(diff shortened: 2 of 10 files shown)


servers/src/common/stats.rs
```diff
@@ -283,3 +284,91 @@ impl Default for StratumStats {
...
+			is_enabled: stats.is_enabled,
+			is_running: stats.is_running,
+			num_workers: stats.num_workers,
+			block_height: stats.block_height,
+			network_difficulty: stats.network_difficulty,
+			edge_bits: stats.edge_bits,
+			blocks_found: stats.blocks_found,
+			network_hashrate: stats.network_hashrate,
+			minimum_share_difficulty: stats.minimum_share_difficulty,
+			worker_stats: stats.worker_stats.iter().map(WorkerInfo::from).collect(),
```
wiesche89: worker_stats keeps every disconnected worker, so this response grows for the lifetime of the node while the CLI only shows connected workers. Should the API filter to active workers too?

api/src/owner.rs
```diff
@@ -78,6 +99,25 @@ impl Owner {
...
+	/// # Returns
+	/// * Result Containing:
+	/// * A [`MiningStatus`](types/struct.MiningStatus.html)
+	/// * or [`Error`](struct.Error.html) if an error is encountered.
+	///
+
+	pub fn get_mining_status(&self) -> Result<MiningStatus, Error> {
+		match &self.mining_stats {
+			Some(provider) => Ok(provider()),
+			None => Ok(MiningStatus::default()),
```
wiesche89: A disabled server already has a provider returning is_enabled, false. Could a missing provider return an internal error instead of silently looking disabled?

api/src/handlers.rs
```diff
@@ -62,6 +62,7 @@ pub fn node_apis<B, P>(
 	tls_config: Option<TLSConfig>,
 	api_chan: &'static mut (oneshot::Sender<()>, oneshot::Receiver<()>),
 	stop_state: Arc<StopState>,
+	mining_stats: Option<MiningStatsProvider>,
```
wiesche89, line 98: node_apis is publicly re-exported, so this required argument breaks external callers. Could we preserve the existing wrapper and add a provider-aware variant?

api/src/types.rs
```diff
@@ -104,6 +104,72 @@ impl Status {
...
+}
+
+/// Mining / stratum status available via the Owner API (and `grin client miningstatus`).
+/// Mirrors the stats shown on the TUI mining tab so headless operators can query them.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct MiningStatus {
+	/// Whether the stratum server is enabled in config
+	pub is_enabled: bool,
+	/// Whether the stratum server is currently running
+	pub is_running: bool,
```
wiesche89, line 136: The source flag is set on startup but never cleared on listener failure or shutdown. Is “currently running” too strong for the API contract?

api/src/owner.rs
```diff
@@ -78,6 +99,25 @@ impl Owner {
...
+	/// When stratum is disabled or no stats provider is wired up, returns a default
+	/// (disabled) [`MiningStatus`](types/struct.MiningStatus.html).
+	///
+	/// # Returns
+	/// * Result Containing:
+	/// * A [`MiningStatus`](types/struct.MiningStatus.html)
+	/// * or [`Error`](struct.Error.html) if an error is encountered.
+	///
+
+	pub fn get_mining_status(&self) -> Result<MiningStatus, Error> {
```
wiesche89, line 110: Could we trim the empty doc line and shorten the returns section to match the neighboring methods?

wiesche89 (commented):
This is a nice and useful feature, and the endpoint is wired to the live StratumStats correctly. Before merging, I’d clarify whether the API should return only active workers, avoid masking a missing provider as disabled, and preserve the public node_apis signature.

servers/src/common/stats.rs
```diff
@@ -292,3 +293,100 @@ impl Default for StratumStats {
...
+			is_enabled: stats.is_enabled,
+			is_running: stats.is_running,
+			num_workers: stats.num_workers,
+			block_height: stats.block_height,
+			network_difficulty: stats.network_difficulty,
+			edge_bits: stats.edge_bits,
+			blocks_found: stats.blocks_found,
+			network_hashrate: stats.network_hashrate,
+			minimum_share_difficulty: stats.minimum_share_difficulty,
+			// worker_stats retains every worker that ever connected; only expose
```
wiesche89, line 329: The response is bounded now, but each request still scans the full worker history under the stats lock. Could we keep a bounded active-worker view and derive num_workers from it too?

api/src/types.rs
wiesche89: A port conflict still leaves this as true after the listener exits. Could we set it after a successful bind, clear it on exit, and test the bind-failure case?

api/src/types.rs
```diff
@@ -104,6 +104,71 @@ impl Status {
...
+	pub blocks_found: u16,
+	/// Estimated network hashrate for the current edge_bits
+	pub network_hashrate: f64,
+	/// Minimum share difficulty requested from miners
+	pub minimum_share_difficulty: u64,
+	/// Per-worker status for currently connected workers
+	pub worker_stats: Vec<WorkerInfo>,
+}
+
+impl Default for MiningStatus {
```
wiesche89: This has no caller besides its own test and duplicates StratumStats::default(). Could we remove the impl and test to avoid default drift?

api/src/owner.rs
```diff
@@ -201,3 +239,45 @@ impl Owner {
...
+	use std::sync::Arc;
+
+	#[test]
+	fn get_mining_status_without_provider_returns_error() {
+		let owner = Owner::new(Weak::new(), Weak::new(), Weak::new());
+		assert!(owner.get_mining_status().is_err());
+	}
+
+	#[test]
+	fn get_mining_status_with_provider() {
```
wiesche89: Could we test this once through the generated JSON RPC handler? The current tests only call Owner directly, so the wire response is not covered.

api/src/owner.rs
```diff
@@ -78,6 +99,23 @@ impl Owner {
...
+	/// # Returns
+	/// * Result Containing:
+	/// * A [`MiningStatus`](types/struct.MiningStatus.html)
+	/// * or [`Error`](struct.Error.html) if no mining stats provider is configured.
+	///
+
+	pub fn get_mining_status(&self) -> Result<MiningStatus, Error> {
+		match &self.mining_stats {
+			Some(provider) => Ok(provider()),
+			None => Err(Error::Internal(
```
wiesche89, line 113: This is fixed now. Could you also update the PR description? It still says a missing provider returns a disabled default.

wiesche89:
I’ve written a prototype Grin miner in Rust https://github.com/wiesche89/rust-miner . It can be used to test the new Owner API method end to end.
With a synced testnet node running, Stratum enabled on port 13416, and the Owner API listening on 13413, start a lightweight worker with:

```
./target/release/grin-cuckatoo-miner mine \
  --backend cpu \
  --edge-bits 25 \
  --node-port 13416 \
  --login api-test-1
```

I tested four connected miners this way. The API reported num_workers: 4 with four worker entries, then returned num_workers: 0 and an empty worker list after disconnecting them. C25 is only intended as a lightweight connection test, valid testnet mining requires C32.

api/src/types.rs
iho: Removed the `Default` impl and its unit test so we do not duplicate / drift from `StratumStats::default`.

api/src/owner.rs
iho: Done — `get_mining_status` is now covered through the generated `OwnerRpc` JSON-RPC handler (success wire payload + missing-provider error path).

api/src/owner.rs
iho: PR description updated: missing provider returns an internal error; disabled stratum still returns `is_enabled: false` via the wired provider.

PR #3898, opened 2026-07-09
fix: update peer difficulty on header/block receipt

### LINKED ISSUE #3167: Peer total_difficulty should be updated on Header/Block/CompactBlock (not just Ping/Pong)
## Problem

We track `height` and `total_difficulty` for all our connected peers on the `PeerLiveInfo` struct.

These are updated when we handle `Ping` and `Pong` messages. These are the *only* places we update `height` and `total_difficulty` for a connected peer.

Surprisingly we do not update these when we process and accept a new block header... and we must wait (potentially up to 10s) for the next `Ping` or `Pong` to arrive to update the peer in our list of connected peers.

I am not entirely sure what the implications of this unnecessary delay means but its probably not great. 

## Proposal

1) We should update `height` and `total_difficulty` as soon as we accept a valid block header (or compact block or full block) from a peer (if this increases total_difficulty).

2) We should then update `height` and `total_difficulty` for any peer that subsequently sends us that header, compact_block or full block, even if we treat this as "already known" on our side.

Why update peer on block header and not just a full block? As soon as a peer sends us a new block header it is advertising that it indeed has this full block. In the same way as a Ping/Pong advertises the existence of a block with certain total_difficulty and height.

## Related 

The `Hand` and `Shake` msgs include `total_difficulty` but do not include `height`. 
We do not use this `total_difficulty` when initializing the `PeerLiveInfo` for the connected peer.
We should probably initialize `PeerLiveInfo` with the `total_difficulty` from the initial `Hand` or `Shake` and either add `height` to these msgs for completeness or maybe just fire off a `Ping` as soon as the handshake negotiation has completed successfully.

iho:
## Summary

`PeerLiveInfo` height and `total_difficulty` were only updated from **Ping/Pong**, so after a peer sent a new tip header/block we could still treat them as stale for up to ~10 seconds (until the next ping cycle). That delays peer selection, sync decisions, and TUI/API peer stats.

This updates peer live info as soon as we accept a header, compact block, or full block from that peer — and also when they re-send an already-known tip (they are still advertising that work).

Closes #3167

### Changes
- Add `PeerInfo::update_if_better` — only increases advertised work (higher difficulty, or same difficulty with higher height)
- Call it from `NetToChainAdapter` on:
  - already-known full/compact blocks and headers
  - successfully processed headers / blocks
  - orphans and other non-ban refusals when we already have the header
  - header-sync batches after successful processing
- Unit test covering increase / ignore / ping overwrite behavior

Ping/Pong still uses absolute `update()` as before.

## Test plan
- [x] `cargo test -p grin_p2p --lib update_if_better`
- [x] `cargo check -p grin_servers`
- [ ] Manual: watch peer height/difficulty in TUI or `get_connected_peers` immediately after a peer announces a new tip (should update without waiting for ping)

### DIFF
--- a/p2p/src/types.rs
+++ b/p2p/src/types.rs
@@ -671,6 +671,25 @@ impl PeerInfo {
 		live_info.total_difficulty = total_difficulty;
 		live_info.last_seen = Utc::now()
 	}
+
+	/// Update height/difficulty only when the peer advertises *more* work
+	/// (higher total_difficulty, or same difficulty at a greater height).
+	/// Used when a peer sends us a header/block/compact block so we do not wait
+	/// for the next ping/pong (up to ~10s) to reflect their chain tip.
+	pub fn update_if_better(&self, height: u64, total_difficulty: Difficulty) {
+		let mut live_info = self.live_info.write();
+		let better = total_difficulty > live_info.total_difficulty
+			|| (total_difficulty == live_info.total_difficulty && height > live_info.height);
+		if !better {
+			return;
+		}
+		if total_difficulty != live_info.total_difficulty {
+			live_info.stuck_detector = Utc::now();
+		}
+		live_info.height = height;
+		live_info.total_difficulty = total_difficulty;
+		live_info.last_seen = Utc::now()
+	}
 }
 
 /// Flatten out a PeerInfo and nested PeerLiveInfo (taking a read lock on it)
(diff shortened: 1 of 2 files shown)


wiesche89 (commented):
please fix merge conflicts

iho (commented):
Rebased onto current `staging` and resolved the conflict in `servers/src/common/adapters.rs`.

Staging had extracted header-batch handling into `process_header_batch`; the peer tip update now runs after that helper returns success (including the PIHD header-cache path).

`cargo test -p grin_p2p --lib update_if_better` and `cargo check -p grin_servers` both pass.

PR #3899, opened 2026-07-09
perf: cache packed proof nonces for cheaper header/block hash

### LINKED ISSUE #3372: block.hash() unnecessarily expensive to call
We call `write()` on `Writeable` when serializing things to bytes.
Not just for serializing full data but also when we serialize during a call to `hash()`.

Block hash -> BlockHeader hash -> ProofOfWork hash -> Proof hash

Every call to `block.hash()` or `header.hash()` delegates to the following code to produce the data to be hashed -

https://github.com/mimblewimble/grin/blob/098d25e5696d1d97e4fbfabb80f27063caccd0f3/core/src/pow/types.rs#L455-L472

We build the bitvec "on the fly" every time we call `hash()`.
Ideally `hash()` on any type is as cheap as possible.

I _suspect_, but have no benchmarks or data to confirm, that this is relatively expensive over time. We make a *lot* of calls to `block.hash()` and `header.hash()`.

Possible improvement - 
1. Consider initializing the bitvec when we initially `read` the Proof instance. 
1. Use this "cached" value when we write it in Hash serialization mode.
1. Ignore the "cached" value when writing full data.
1. Make sure `nonces` or `edge_bits` cannot be written to or modified in unexpected ways, leaving the cache stale.

iho:
## Summary

`block.hash()` / `header.hash()` ultimately hash the PoW `Proof` by bit-packing all nonces on every call. That packing is relatively expensive and runs very often during sync and validation.

This caches the packed form when a `Proof` is **deserialized** (the on-wire bytes we already read), so later hash/write uses that buffer instead of re-running `pack_bits`.

Closes #3372

### Approach
- Add private `packed_nonces: Option<Vec<u8>>` on `Proof` (ignored by `PartialEq` / serde)
- `Readable::read` stores the packed bytes alongside extracted nonces
- `Writeable::write` prefers the cache; otherwise packs on demand (mining / constructed proofs)
- `Proof::clear_packed_cache()` for safe mutation of `edge_bits` / `nonces`
- Stratum share path clears cache after applying a solution
- Unit tests: round-trip cache, hash equivalence, mutation + clear

Locally constructed proofs (mining templates, tests, genesis) keep `None` and behave as before.

## Test plan
- [x] `cargo test -p grin_core --lib pow::types`
- [x] `cargo test -p grin_core --lib genesis`
- [x] `cargo check -p grin_servers`


core/src/pow/types.rs
```diff
@@ -330,12 +330,23 @@ impl ProofOfWork {
...
+	///
+	/// Populated when the proof is deserialized from the on-wire / on-disk bit
+	/// packing, so subsequent `hash()` / `write()` calls do not re-run the
+	/// relatively expensive bit-packing step. Not populated for locally
+	/// constructed proofs (mining / tests) unless filled explicitly.
+	///
+	/// **Important:** if you mutate `edge_bits` or `nonces` after construction,
+	/// call [`Proof::clear_packed_cache`] so a stale cache cannot poison hashes.
+	#[serde(skip)]
+	packed_nonces: Option<Vec<u8>>,
```
wiesche89, line 345: The cache depends on edge_bits and nonces, but both are still publicly mutable. A missed clear_packed_cache() makes verification and hashing use different proof data. Could we make mutation invalidate the cache automatically in this PR?

core/src/pow/types.rs
```diff
@@ -525,7 +584,12 @@ impl Writeable for Proof {
 		if writer.serialization_mode() != ser::SerializationMode::Hash {
 			writer.write_u8(self.edge_bits)?;
 		}
-		writer.write_fixed_bytes(&self.pack_nonces())
+		// Prefer the deserialized packed form (hash-hot path); otherwise pack on demand.
+		if let Some(ref packed) = self.packed_nonces {
```
wiesche89, line 607: Could we use the cache only for hash serialization and always repack full wire/disk writes from the fields? That matches #3372 and prevents stale cached bytes from being emitted.

core/src/pow/types.rs
```diff
@@ -555,11 +620,43 @@ mod tests {
...
 
+	#[test]
+	fn packed_cache_matches_recomputed_and_hash() {
+		global::set_local_chain_type(global::ChainTypes::Mainnet);
+		let mut proof = Proof::new(gen_proof(29));
+		proof.edge_bits = 29;
+		let hash_before = proof.hash();
+		let packed = proof.compute_packed_nonces();
+
+		// Simulate a deserialized proof: same nonces, cache filled.
```
wiesche89: Could this use a real serialize/deserialize round trip instead of setting the private cache directly? That would cover the actual read => cache => hash/write path.

core/src/pow/types.rs
```diff
@@ -402,8 +436,27 @@ impl Proof {
...
+
+	/// Whether this proof currently holds a packed-nonces cache.
+	#[cfg(test)]
+	pub fn has_packed_cache(&self) -> bool {
+		self.packed_nonces.is_some()
+	}
+
+	/// Pack the nonces of the proof to their exact bit size as described above.
+	/// Uses the deserialized cache when available.
 	pub fn pack_nonces(&self) -> Vec<u8> {
```
wiesche89, line 471: Could you add before/after numbers for repeated header hashing? The cache keeps and clones an extra allocation per proof, so it would be good to confirm the net win.

wiesche89 (commented):
This addresses the main idea of #3372, but not all of its proposed safeguards yet. The cache is populated on read and reused for hashing, but it is also used for full serialization, and the public fields can still be changed without invalidating it. I think those two points should be resolved in this PR before closing the issue. Some before/after numbers would also help confirm the expected benefit.

core/src/pow/types.rs
iho: Done in f5a4df03. `edge_bits` and `nonces` are private; mutations go through `set_edge_bits` / `set_nonces` / `nonces_mut`, which always clear `packed_nonces`. Call sites (stratum, tests, cuckoo solvers, etc.) were updated accordingly, so a missed `clear_packed_cache` is no longer possible.

core/src/pow/types.rs
iho: Done in f5a4df03. Full wire/disk writes always repack from the fields. The cache is used only in `SerializationMode::Hash`, matching #3372 so stale cached bytes cannot be emitted on the network or to disk.

core/src/pow/types.rs
iho: Done in f5a4df03. Tests go through a real `write` → `read` round trip (`round_trip`) to populate the cache, then check hash equivalence and that setters drop the cache. Also added `full_write_does_not_emit_stale_cache` to lock in hash-only use of the cache.

core/src/pow/types.rs
iho: Done in f5a4df03. Added ignored bench `packed_cache_hash_bench` (50k `proof.hash()` iters).

**Release numbers** (local, Apple Silicon):
- uncached ~533 ns/op
- cached ~351 ns/op
- **~1.52×** faster when the deserialized cache is used

```sh
cargo test -p grin_core --release --lib packed_cache_hash_bench -- --nocapture --ignored
```

Net win is modest per call (hashing itself dominates) but compounds over the very high volume of header/block hashes during sync. Extra cost is one `Vec<u8>` per deserialized proof, only retained while the proof is in memory.

PR #3900, opened 2026-07-09
fix: rebroadcast txs recovered from reorg cache

### LINKED ISSUE #3489: Increase "reorg cache" timer limit?

iho:
## Summary

On reorg, the reorg cache correctly re-adds previously seen txs to the **local** mempool, but it never rebroadcast them. Peers that did not keep those txs (or never saw them) would not learn about them until someone else re-announced.

As noted on #3489 by @antiochp: every tx successfully re-accepted from the reorg cache should also be treated as new and rebroadcast.

Closes #3489

### Changes
- `reconcile_reorg_cache`: call `adapter.tx_accepted` after a successful `add_to_txpool`
- Skip rebroadcast when re-add fails (already in pool / invalid on new tip)
- Integration test with a recording pool adapter

Retention period remains configurable via `reorg_cache_period` (default 30 minutes). This PR focuses on the missing rebroadcast path rather than changing the default window.

## Test plan
- [x] `cargo test -p grin_pool --test reorg_cache_rebroadcast`

### DIFF
--- a/pool/src/transaction_pool.rs
+++ b/pool/src/transaction_pool.rs
@@ -283,6 +283,11 @@ where
 		debug!("truncate_reorg_cache: size: {}", cache.len());
 	}
 
+	/// Re-apply cached transactions to the mempool after a reorg.
+	///
+	/// Txs that are successfully re-accepted are also rebroadcast via the pool
+	/// adapter so peers learn about them again. Previously we only re-added to
+	/// the local pool and never announced them on the network (see #3489).
 	pub fn reconcile_reorg_cache(&mut self, header: &BlockHeader) -> Result<(), PoolError> {
 		let entries = self.reorg_cache.read().iter().cloned().collect::<Vec<_>>();
 		debug!(
@@ -290,12 +295,19 @@ where
 			entries.len(),
 			header.hash(),
 		);
+		let mut rebroadcast = 0usize;
 		for entry in entries {
-			let _ = self.add_to_txpool(&entry, header);
+			// Only rebroadcast txs that were actually re-accepted (already-present
+			// or invalid-on-new-tip txs are skipped without network spam).
+			if self.add_to_txpool(&entry, header).is_ok() {
+				self.adapter.tx_accepted(&entry);
+				rebroadcast += 1;
+			}
 		}
 		debug!(
-			"reconcile_reorg_cache: block: {:?} ... done.",
-			header.hash()
+			"reconcile_reorg_cache: block: {:?} ... done (rebroadcast {})",
+			header.hash(),
+			rebroadcast
 		);
 		Ok(())
 	}
--- a/pool/tests/common.rs
+++ b/pool/tests/common.rs
@@ -17,7 +17,7 @@
 use self::chain::types::{NoopAdapter, Options};
 use self::chain::Chain;
 use self::core::consensus;
-use self::core::core::hash::Hash;
+use self::core::core::hash::{Hash, Hashed};
 use self::core::core::{
 	Block, BlockHeader, BlockSums, Inputs, KernelFeatures, OutputIdentifier, Transaction, TxKernel,
 };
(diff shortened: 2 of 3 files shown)


pool/src/transaction_pool.rs
```diff
@@ -283,19 +283,31 @@ where
 		debug!("truncate_reorg_cache: size: {}", cache.len());
 	}
 
+	/// Re-apply cached transactions to the mempool after a reorg.
+	///
+	/// Txs that are successfully re-accepted are also rebroadcast via the pool
```
wiesche89, line 288: The later discussion in #3489 explicitly questions unconditional rebroadcasting and suggests opt-in/default-off. This PR also leaves the issue’s retention-period request unchanged. Could we settle that policy here and use Refs #3489 rather than closing it unless both parts are handled?

pool/src/transaction_pool.rs
```diff
@@ -283,19 +283,31 @@ where
...
 			entries.len(),
 			header.hash(),
 		);
+		let mut rebroadcast = 0usize;
 		for entry in entries {
-			let _ = self.add_to_txpool(&entry, header);
+			// Only rebroadcast txs that were actually re-accepted (already-present
+			// or invalid-on-new-tip txs are skipped without network spam).
+			if self.add_to_txpool(&entry, header).is_ok() {
+				self.adapter.tx_accepted(&entry);
```
wiesche89, line 303: tx_accepted() is fine for a single tx, but here it runs in a loop under the pool write lock. Each peer queue has at most 100 free slots and silently drops overflow while reporting success, so a reorg recovering more than SEND_CHANNEL_CAP txs can lose later announcements. Could we schedule these through a paced bulk-relay path outside the pool lock?

pool/tests/reorg_cache_rebroadcast.rs
```diff
@@ -0,0 +1,97 @@
...
+		.unwrap();
+	pool.add_to_pool(test_source(), tx_b.clone(), false, &header)
+		.unwrap();
+	// Initial accept calls: one per tx.
+	assert_eq!(adapter.accepted_count(), 2);
+	assert_eq!(pool.total_size(), 2);
+	assert_eq!(pool.reorg_cache.read().len(), 2);
+
+	// Simulate reorg: empty the pool while leaving reorg_cache intact.
+	pool.txpool.entries.clear();
```
wiesche89, line 72: The recording adapter only counts callbacks. Could we add a real fork/reorg test for the production trigger, plus a bounded-queue adapter using SEND_CHANNEL_CAP to cover overflow? Otherwise the test misses the actual delivery failure mode.

pool/tests/reorg_cache_rebroadcast.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright 2021 The Grin Developers
```
wiesche89, line 1: 2026?

pool/tests/common.rs
```diff
@@ -170,6 +170,57 @@ where
...
+impl PoolAdapter for RecordingPoolAdapter {
+	fn tx_accepted(&self, entry: &PoolEntry) {
+		self.accepted.lock().unwrap().push(entry.tx.hash());
+	}
+	fn stem_tx_accepted(&self, _entry: &PoolEntry) -> Result<(), PoolError> {
+		Ok(())
+	}
+}
+
+pub fn init_transaction_pool_with_adapter<B, P>(
```
wiesche89, line 203: this duplicates the full PoolConfig literal from init_transaction_pool for one caller. Could the existing helper delegate here, or both use a shared test config, so they cannot drift?

PR #3901, opened 2026-07-09
fix: prune dead stratum workers instead of unbounded growth

### LINKED ISSUE #2413: Please provide an option to prune dead workers from grin-server
I would love a button to kick out all instances of dead workers, who connected and disconnect. With all my rigs, and fiddling around the list becomes very long, very quickly. It would be nice to be able to clean it up without having to shut it down and restart it.

My other wish, would be to be able to reset the stats (accepted, rejected. stale, found etc...) for one or more workers. Perhaps hitting 'R' follwed by a number (worker ID)  or "R" followed by "A" to reset em all.

Just my 2 cents, i am sure you have bigger fish to fry right now.

Thanks for all the good work.

iho:
## Summary

Stratum `worker_stats` was append-only: every reconnect left a permanent disconnected entry. With rigs cycling connections the TUI mining table (and mining status API) grew without bound until node restart.

Closes #2413

### Changes
- **Slot reuse**: new workers fill the first disconnected slot (stats reset for that slot)
- **Idle clear**: when the last live worker disconnects, drop all historical worker stats
- **Auto-prune**: if more than 32 dead slots remain, compact and remap live worker IDs
- **`prune_dead_workers()`**: explicit compact/remap helper (returns removed count)
- Unit tests for reuse, idle clear, and remapping

Live worker IDs only change during prune/remap (under write locks). Day-to-day operation uses slot reuse without remapping.

## Test plan
- [x] `cargo test -p grin_servers --lib stratumserver::tests`

### DIFF
--- a/servers/src/mining/stratumserver.rs
+++ b/servers/src/mining/stratumserver.rs
@@ -711,28 +711,104 @@ impl WorkersList {
 
 	pub fn add_worker(&self, tx: Tx) -> usize {
 		let mut stratum_stats = self.stratum_stats.write();
-		let worker_id = stratum_stats.worker_stats.len();
-		let worker = Worker::new(worker_id, tx);
 		let mut workers_list = self.workers_list.write();
+
+		// Reuse a disconnected worker slot when available so the stats list
+		// does not grow without bound under reconnect churn (#2413).
+		let worker_id = stratum_stats
+			.worker_stats
+			.iter()
+			.position(|ws| !ws.is_connected)
+			.unwrap_or_else(|| stratum_stats.worker_stats.len());
+
+		let worker = Worker::new(worker_id, tx);
 		workers_list.insert(worker_id, worker);
 
 		let mut worker_stats = WorkerStats::default();
 		worker_stats.is_connected = true;
 		worker_stats.id = worker_id.to_string();
 		worker_stats.pow_difficulty = stratum_stats.minimum_share_difficulty;
-		stratum_stats.worker_stats.push(worker_stats);
+		// Reset counters for the (re)used slot.
+		if worker_id < stratum_stats.worker_stats.len() {
+			stratum_stats.worker_stats[worker_id] = worker_stats;
+		} else {
+			stratum_stats.worker_stats.push(worker_stats);
+		}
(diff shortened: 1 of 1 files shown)


servers/src/mining/stratumserver.rs
```diff
@@ -711,28 +711,104 @@ impl WorkersList {
...
+			let n = self.prune_dead_workers();
+			if n > 0 {
+				debug!("Stratum: pruned {} disconnected worker stats", n);
+			}
+		}
+	}
+
+	/// Remove disconnected worker stats and reassign live worker IDs densely.
+	///
+	/// Safe to call while workers are connected: live workers are remapped
```
wiesche89, line 775: I don't think remapping live workers is safe here. Each connection task captures its worker_id at accept time and keeps using it for RPCs, replies, stats, and disconnect. After a remap that ID can point to another worker or outside the stats vector. Could we keep connection IDs stable and add a test that continues using the original ID after cleanup?

servers/src/mining/stratumserver.rs
```diff
@@ -711,28 +711,104 @@ impl WorkersList {
...
 		let mut stratum_stats = self.stratum_stats.write();
-		let worker_id = stratum_stats.worker_stats.len();
-		let worker = Worker::new(worker_id, tx);
 		let mut workers_list = self.workers_list.write();
+
+		// Reuse a disconnected worker slot when available so the stats list
+		// does not grow without bound under reconnect churn (#2413).
+		let worker_id = stratum_stats
+			.worker_stats
+			.iter()
```
wiesche89, line 720: remove_worker() marks the stats disconnected before removing the worker from the map. A concurrent accept can reuse and replace that slot in between, after which the old disconnect removes the new connection. Could we make removal and the disconnected-state update atomic under the same locks?

wiesche89 (commented):
This addresses the growth problem behind #2413 automatically, rather than adding the requested manual prune action, and the separate stats reset request remains out of scope. Could the PR description clarify that distinction and use Refs #2413 unless the automatic replacement is considered sufficient to close the issue?

Issue #3827, opened 2026-04-23
capacity overflow error


aglkm:
Compiled master branch and got the following error during node syncing:

```
20260423 16:38:33.770 ERROR grin_util::logger - 
thread 'peer_read' panicked at 'capacity overflow': /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/alloc/src/raw_vec/mod.rs:28   0: grin_util::logger::send_panic_to_log::{{closure}}
   1: std::panicking::panic_with_hook
   2: std::panicking::panic_handler::{closure#0}
   3: std::sys::backtrace::__rust_end_short_backtrace::<std::panicking::panic_handler::{closure#0}, !>
   4: __rustc::rust_begin_unwind
   5: core::panicking::panic_fmt
   6: alloc::raw_vec::capacity_overflow
   7: alloc::raw_vec::handle_error
   8: <grin_core::core::pmmr::segment::Segment<T> as grin_core::ser::Readable>::read
   9: <grin_p2p::msg::SegmentResponse<T> as grin_core::ser::Readable>::read
  10: grin_p2p::codec::Codec::read
  11: std::sys::backtrace::__rust_begin_short_backtrace
  12: core::ops::function::FnOnce::call_once{{vtable.shim}}
  13: <std::sys::thread::unix::Thread>::new::thread_start
  14: start_thread
             at ./nptl/pthread_create.c:448:8
  15: __GI___clone3
             at ./misc/../sysdeps/unix/sysv/linux/x86_64/clone3.S:78

```

Syncing stopped at: Sync step 2/7: Downloading Tx state (PIBD) - 21101792 / 24156692 entries - 87%

ardocrat:
Related PR: https://github.com/mimblewimble/grin/pull/3807

PR #3902, opened 2026-07-09
fix: bound MerkleProof path length on deserialize

iho:
## Summary

`MerkleProof::read` used the untrusted `path_len` directly in `Vec::with_capacity`. A crafted proof with a huge path length could panic (`capacity overflow`) or attempt a massive allocation — the same class of issue fixed for PMMR segments in #3850 / reported in #3827.

### Changes
- Cap Merkle proof sibling path length at **64** (O(log n) in practice; far above real proofs)
- Return `TooLargeReadErr` for oversized `path_len` before allocating
- Cap path preallocation to 16 entries
- Align `StreamingReader::read_fixed_bytes` with BinReader’s existing **100k** single-read limit
- Tests for oversized and near-`u64::MAX` path lengths

## Test plan
- [x] `cargo test -p grin_core --test merkle_proof`
- [x] `cargo test -p grin_core --test segment`

### DIFF
--- a/core/src/core/merkle_proof.rs
+++ b/core/src/core/merkle_proof.rs
@@ -18,6 +18,7 @@ use crate::core::hash::Hash;
 use crate::core::pmmr;
 use crate::ser;
 use crate::ser::{PMMRIndexHashable, Readable, Reader, Writeable, Writer};
+use std::cmp::min;
 use util::ToHex;
 
 /// Merkle proof errors.
@@ -26,6 +27,15 @@ pub enum MerkleProofError {
 	/// Merkle proof root hash does not match when attempting to verify.
 	RootMismatch,
 }
+
+/// Maximum sibling path length accepted when deserializing a Merkle proof.
+///
+/// Path length is O(log n) for an MMR of size n (plus a few peak hashes).
+/// 64 is far larger than any legitimate proof and prevents a malicious
+/// `path_len` from causing a capacity overflow / huge allocation.
+pub const MAX_MERKLE_PROOF_PATH: u64 = 64;
+
+const MERKLE_PROOF_PATH_PREALLOC: u64 = 16;
 
 /// A Merkle proof that proves a particular element exists in the MMR.
 #[derive(Serialize, Deserialize, Debug, Eq, PartialEq, Clone, PartialOrd, Ord)]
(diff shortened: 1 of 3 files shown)


core/src/core/merkle_proof.rs
```diff
@@ -50,7 +60,11 @@ impl Readable for MerkleProof {
 	fn read<R: Reader>(reader: &mut R) -> Result<MerkleProof, ser::Error> {
 		let mmr_size = reader.read_u64()?;
 		let path_len = reader.read_u64()?;
-		let mut path = Vec::with_capacity(path_len as usize);
+		if path_len > MAX_MERKLE_PROOF_PATH {
+			return Err(ser::Error::TooLargeReadErr);
+		}
+		// Cap preallocation even when path_len is within the allowed range.
+		let mut path = Vec::with_capacity(min(path_len, MERKLE_PROOF_PATH_PREALLOC) as usize);
```
wiesche89, line 67: With path_len capped at 64, the full allocation is at most 2 KiB. Could we allocate path_len directly and drop the second limit and the min import?

core/src/ser.rs
```diff
@@ -595,6 +595,10 @@ impl<'a> Reader for StreamingReader<'a> {
 
 	/// Read a fixed number of bytes.
 	fn read_fixed_bytes(&mut self, len: usize) -> Result<Vec<u8>, Error> {
+		// Match BinReader / BufReader: refuse huge single reads that would OOM or panic.
+		if len > 100_000 {
```
wiesche89, line 599: This is now the third copy of the 100k reader limit. Could we use one shared constant and add a direct StreamingReader boundary test here?

core/tests/merkle_proof.rs
```diff
@@ -14,19 +14,54 @@
...
+
 #[test]
 fn empty_merkle_proof() {
 	let proof = MerkleProof::empty();
 	assert_eq!(proof.path, vec![]);
 	assert_eq!(proof.mmr_size, 0);
 }
 
+#[test]
+fn merkle_proof_read_rejects_huge_path_len() {
```
wiesche89, line 35: Could we also cover the accepted side of the boundary: 64 hashes accepted and 65 rejected? The current tests only exercise rejection.

core/tests/merkle_proof.rs
```diff
@@ -14,19 +14,54 @@
...
+		&mut &bytes[..],
+		ProtocolVersion(1),
+		DeserializationMode::default(),
+	);
+	assert_eq!(res.err(), Some(ser::Error::TooLargeReadErr));
+}
+
+#[test]
+fn merkle_proof_read_rejects_capacity_overflow_path_len() {
+	// Historical bug: path_len near u64::MAX caused Vec::with_capacity to panic.
```
wiesche89, line 52: Could we trim these tests a little? The names already describe the cases, and the historical context is covered by the PR description. A compact bounds test covering 64, 65, and u64::MAX would avoid the duplicated setup.

PR #3904, opened 2026-07-09
docs: accurate input/output linking in grin4bitcoiners.md

### LINKED ISSUE #2427: Update grin4bitcoiners.md
https://github.com/mimblewimble/grin/blob/master/doc/grin4bitcoiners.md needs some rewording to more accurately reflect the contents of https://github.com/mimblewimble/docs/wiki/Grin-Privacy-Primer. 

Specifically related to input and output linking.

iho:
## Summary

`doc/grin4bitcoiners.md` previously suggested that cut-through / block aggregation means **all** input↔output association is lost. That overstates Mimblewimble privacy relative to the [Grin Privacy Primer](https://github.com/mimblewimble/docs/wiki/Grin-Privacy-Primer), which treats **input/output linking** as residual leakage (taint analysis, relay observation) even while amounts and addresses are strong wins.

Closes #2427

### Changes
- Clarify CT (no amounts), no addresses, cut-through/aggregation **without** claiming perfect unlinkability
- New subsection on what cut-through does **not** erase (relay graph, taint, kernels)
- FAQ on listening to relay rewritten to match the primer
- Link to the Privacy Primer for deeper reading
- Light touch-ups (scripting, emission, quantum) for accuracy and tone

Translations (`grin4bitcoiners_KR.md`, `ZH-CN`) left as-is for native-speaker follow-up.

## Test plan
- [x] Doc-only change; review rendered markdown

### DIFF
--- a/doc/grin4bitcoiners.md
+++ b/doc/grin4bitcoiners.md
@@ -1,60 +1,73 @@
 # Grin/Mimblewimble for Bitcoiners
 
-*Read this in other languages:[Korean](translations/grin4bitcoiners_KR.md)
+*Read this in other languages: [Korean](translations/grin4bitcoiners_KR.md), [Chinese](translations/grin4bitcoiners_ZH-CN.md)*
+
+This page is a short comparison for readers familiar with Bitcoin. For a deeper treatment of what Grin does and does not hide, see the
+[Grin Privacy Primer](https://github.com/mimblewimble/docs/wiki/Grin-Privacy-Primer).
 
 ## Privacy and Fungibility
 
-There are 3 main properties of Grin transactions that make them private:
+Grin transactions are private in ways that Bitcoin’s transparent UTXO model is not. Three properties matter most:
 
-1. There are no addresses.
-1. There are no amounts.
-1. 2 transactions, one spending the other, can be merged in a block to form only one, removing all intermediary information.
+1. **No addresses.** Outputs are one-shot curve points. There is no reusable “payment address” that links payments over time.
+2. **No amounts.** Every transaction uses confidential transactions (CT). Amounts sit inside Pedersen commitments with range proofs, so outsiders cannot read transfer values.
+3. **Cut-through and aggregation.** When an output is spent, Mimblewimble can remove the spent input/output pair from the long-term chain state. Inside a block, many transactions are merged into one aggregate set of inputs, outputs, and kernels, so a confirmed block does not look like a list of labeled payments.
 
-The 2 first properties mean that all transactions are indistinguishable from one another. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
+Because of (1) and (2), unspent outputs and kernels all look like random-looking data unless you participated in building them. Nodes can still verify that no money was created out of thin air by checking that commitments balance (homomorphic structure) and that range proofs and signatures are valid.
 
-Moreover, there are no more transactions in a block. A Grin block looks just like one giant transaction and all original association between inputs and outputs is lost.
+### What cut-through does *not* erase
+
+It is easy to overstate the third point. **Cut-through improves scalability and reduces long-term history, but it does not make input↔output linking impossible.**
+
+- While a transaction is **relayed** (before or while it is mined), observers can still see which commitments are spent and which new ones appear. That is a real information channel.
+- **Taint / hop analysis** can follow known “marked” outputs across spends if an adversary can introduce or learn those outputs (for example by paying you, or by watching the network closely). Aggregation and Dandelion make this harder, but they are not perfect anonymity.
+- After cut-through, the **kernel** of a transaction remains as a permanent ~100-byte footprint. The chain still records that “some” balanced state change occurred, even when the intermediate UTXOs are gone.
+
+So: Grin hides amounts and addresses strongly; it **weakens but does not eliminate** graph-style analysis of which outputs feed which later spends. That matches the Privacy Primer’s view that addresses and amounts are checked off as strong wins, while **input/output linking** remains an area of residual leakage and ongoing research.
 
 ## Scalability
 
-As explained in the previous section, thanks to the Mimblewimble transaction and block format we can merge transactions when an output is directly spent by the input of another. It's as if when Alice gives money to Bob, and then Bob gives it all to Carol, Bob was never involved and his transaction is actually never even seen on the blockchain.
+Because spent outputs can be removed from the active UTXO set, a Grin node’s long-term storage grows primarily with **unspent outputs** (users holding coins), not with every historical payment. Full verification still needs kernels and headers; kernels are the main “per-transaction” residue that remains after cut-through.
 
-Pushing that further, between blocks, most outputs end up being spent sooner or later by another input. So *all spent outputs can be safely removed*. And the whole blockchain can be stored, downloaded and fully verified in just a few gigabytes or less (assuming a number of transactions similar to bitcoin).
-
-This means that the Grin blockchain scales with the number of users (unspent outputs), not the number of transactions. At the moment, there is one caveat to that: a small piece of data (called a *kernel*, about 100 bytes) needs to stay around for each transaction. But we're working on optimizing that as well.
+In practice the chain stays far smaller than a transparent UTXO chain with the same activity, but it is not free—range proofs and kernels have a cost, and sync still requires downloading state (for example via PIBD or a txhashset snapshot).
 
 ## Scripting
 
-Maybe you've heard that Mimblewimble doesn't support scripts. And in some way, that's true. But thanks to cryptographic trickery, many contracts that in Bitcoin would require a script can be achieved with Grin using properties of Elliptic Curve Cryptography. So far, we know how to do:
+Mimblewimble does not carry Bitcoin-style scripts on the chain. Many contracts that Bitcoin implements with Script can still be built using elliptic-curve techniques and interactive protocols, for example:
 
-* Multi-signature transactions.
-* Atomic swaps.
-* Time-locked transactions and outputs.
-* Lightning Network
+* Multi-signature transactions
+* Atomic swaps
+* Time-locked transactions and outputs
+* Payment-channel style constructions (e.g. Lightning-like designs)
+
+Those constructions live more in wallet protocols than in on-chain script programs.
 
 ## Emission Rate
 
-Bitcoin's 10 minute block time has its initial 50 btc reward cut in half every 4 years until there are 21 million bitcoin in circulation. Grin's emission rate is linear, meaning it never drops. The block reward is currently set at 60 grin with a block goal of 60 seconds. This still works because 1) dilution trends toward zero and 2) a non-negligible amount of coins gets lost or destroyed every year.
+Bitcoin’s block subsidy halves over time toward a fixed supply. Grin’s base emission is **linear** (a constant block reward), so supply grows without a hard cap. Dilution trends toward zero as the monetary base grows, and coins are also lost over time. The block target is on the order of one minute (see current consensus parameters for the exact reward).
 
 ## FAQ
 
 ### Wait, what!? No address?
 
-Nope, no address. All outputs in Grin are unique and have no common data with any previous output. Instead of relying on a known address to send money, transactions have to be built interactively, with two (or more) wallets exchanging data with one another. This interaction **does not require both parties to be online at the same time**. Practically speaking, there are many ways for two programs to interact privately and securely. This interaction could even take place over email or Signal (or carrier pigeons).
+Correct—no reusable addresses. Outputs are unique and do not share an address field with earlier outputs. Wallets build transactions **interactively** (or via a slate/exchange of messages). Both parties do not need to be online at the same instant; the handshake can happen over any private channel (including offline media).
 
 ### If transaction information gets removed, can I just cheat and create money?
 
-No, and this is where Mimblewimble and Grin shine. Confidential transactions are a form of [homomorphic encryption](https://en.wikipedia.org/wiki/Homomorphic_encryption). Without revealing any amount, Grin can verify that the sum of all transaction inputs equal the sum of transaction outputs, plus the fee. Going even further, comparing the sum of all money created by mining with the total sum of money that's being held, Grin nodes can check the correctness of the total money supply.
+No. Confidential transactions are designed so nodes can check that the sum of inputs equals the sum of outputs plus the fee **without** learning the amounts. Across the whole chain, nodes also check consistency of total supply against coinbase issuance. Cut-through removes spent UTXOs; it does not remove the need for those balance checks.
 
-### If I listen to transaction relay, can't I just figure out who they belong to before being cut-through?
+### If I listen to transaction relay, can’t I just figure out who they belong to before cut-through?
 
-You can figure out which outputs are being spent by which transaction, but the trail of data stops here. All inputs and outputs look like random pieces of data, so you can't tell if the money was transferred, still belongs to the same person, which output is the actual transfer and which is the change, etc. Grin transactions are built with *no identifiable piece of information*.
+You can observe **which** outputs are spent and **which** new outputs appear in a given transaction or stem. You generally **cannot** read amounts or reuse addresses to label “who paid whom” the way you can on Bitcoin.
 
-In addition, Grin leverages [Dandelion relay](dandelion/dandelion.md), which provides additional anonymity as to which IP or client the transaction originated from, and allows for transactions to be aggregated.
+What you *can* still do, with extra information, is **link** spends: for example by paying someone a known output and watching how related commitments move, or by combining relay timing with other metadata. Dandelion stem relay reduces the reliability of “this IP originated this tx,” and stem aggregation can blur some couplings, but they do not make the network a black box.
 
-### What about the quantum computaggedon?
+For a careful split of “what Grin hides” vs “what still leaks,” read the [Privacy Primer](https://github.com/mimblewimble/docs/wiki/Grin-Privacy-Primer).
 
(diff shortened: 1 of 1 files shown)


doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,73 @@
...
 
-Moreover, there are no more transactions in a block. A Grin block looks just like one giant transaction and all original association between inputs and outputs is lost.
+### What cut-through does *not* erase
 
-## Scalability
+It is easy to overstate the third point. **Cut-through improves scalability and reduces long-term history, but it does not make input↔output linking impossible.**
 
-As explained in the previous section, thanks to the Mimblewimble transaction and block format we can merge transactions when an output is directly spent by the input of another. It's as if when Alice gives money to Bob, and then Bob gives it all to Carol, Bob was never involved and his transaction is actually never even seen on the blockchain.
+- While a transaction is **relayed** (before or while it is mined), observers can still see which commitments are spent and which new ones appear. That is a real information channel.
+- **Taint / hop analysis** can follow known “marked” outputs across spends if an adversary can introduce or learn those outputs (for example by paying you, or by watching the network closely). Aggregation and Dandelion make this harder, but they are not perfect anonymity.
```
ardocrat: Worth to mention payjoins, they are in contracts branch

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,73 @@
...
 
 ## FAQ
 
 ### Wait, what!? No address?
 
-Nope, no address. All outputs in Grin are unique and have no common data with any previous output. Instead of relying on a known address to send money, transactions have to be built interactively, with two (or more) wallets exchanging data with one another. This interaction **does not require both parties to be online at the same time**. Practically speaking, there are many ways for two programs to interact privately and securely. This interaction could even take place over email or Signal (or carrier pigeons).
+Correct—no reusable addresses. Outputs are unique and do not share an address field with earlier outputs. Wallets build transactions **interactively** (or via a slate/exchange of messages). Both parties do not need to be online at the same instant; the handshake can happen over any private channel (including offline media).
```
ardocrat: What do you mean "address field" here? Public key of account can be used for interaction as "address"

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,73 @@
...
+No. Confidential transactions are designed so nodes can check that the sum of inputs equals the sum of outputs plus the fee **without** learning the amounts. Across the whole chain, nodes also check consistency of total supply against coinbase issuance. Cut-through removes spent UTXOs; it does not remove the need for those balance checks.
+
+### If I listen to transaction relay, can’t I just figure out who they belong to before cut-through?
 
-### If I listen to transaction relay, can't I just figure out who they belong to before being cut-through?
+You can observe **which** outputs are spent and **which** new outputs appear in a given transaction or stem. You generally **cannot** read amounts or reuse addresses to label “who paid whom” the way you can on Bitcoin.
```
ardocrat: still not sure whats addresses you mean here

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,73 @@
...
+What you *can* still do, with extra information, is **link** spends: for example by paying someone a known output and watching how related commitments move, or by combining relay timing with other metadata. Dandelion stem relay reduces the reliability of “this IP originated this tx,” and stem aggregation can blur some couplings, but they do not make the network a black box.
 
-In addition, Grin leverages [Dandelion relay](dandelion/dandelion.md), which provides additional anonymity as to which IP or client the transaction originated from, and allows for transactions to be aggregated.
+For a careful split of “what Grin hides” vs “what still leaks,” read the [Privacy Primer](https://github.com/mimblewimble/docs/wiki/Grin-Privacy-Primer).
 
-### What about the quantum computaggedon?
+### What about quantum computers?
```
ardocrat, line 71: Worth to mention that while the obfuscation can be removed by a quantum attacker, the attacker has no way of knowing whether the transactions that they are attempting to view are of significant enough value to warrant performing a quantum attack.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,73 @@
...
+- **Taint / hop analysis** can follow known “marked” outputs across spends if an adversary can introduce or learn those outputs (for example by paying you, or by watching the network closely). Aggregation and Dandelion make this harder, but they are not perfect anonymity.
+- After cut-through, the **kernel** of a transaction remains as a permanent ~100-byte footprint. The chain still records that “some” balanced state change occurred, even when the intermediate UTXOs are gone.
 
-Pushing that further, between blocks, most outputs end up being spent sooner or later by another input. So *all spent outputs can be safely removed*. And the whole blockchain can be stored, downloaded and fully verified in just a few gigabytes or less (assuming a number of transactions similar to bitcoin).
+So: Grin hides amounts and addresses strongly; it **weakens but does not eliminate** graph-style analysis of which outputs feed which later spends. That matches the Privacy Primer’s view that addresses and amounts are checked off as strong wins, while **input/output linking** remains an area of residual leakage and ongoing research.
```
ardocrat: What do you mean "addresses" here?

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,73 @@
...
 ## Privacy and Fungibility
 
-There are 3 main properties of Grin transactions that make them private:
+Grin transactions are private in ways that Bitcoin’s transparent UTXO model is not. Three properties matter most:
 
-1. There are no addresses.
-1. There are no amounts.
-1. 2 transactions, one spending the other, can be merged in a block to form only one, removing all intermediary information.
+1. **No addresses.** Outputs are one-shot curve points. There is no reusable “payment address” that links payments over time.
+2. **No amounts.** Every transaction uses confidential transactions (CT). Amounts sit inside Pedersen commitments with range proofs, so outsiders cannot read transfer values.
```
ardocrat: Links to explain Pedersen commitments and Range proofs can be added

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,73 @@
...
+
+## Scalability
 
-This means that the Grin blockchain scales with the number of users (unspent outputs), not the number of transactions. At the moment, there is one caveat to that: a small piece of data (called a *kernel*, about 100 bytes) needs to stay around for each transaction. But we're working on optimizing that as well.
+Because spent outputs can be removed from the active UTXO set, a Grin node’s long-term storage grows primarily with **unspent outputs** (users holding coins), not with every historical payment. Full verification still needs kernels and headers; kernels are the main “per-transaction” residue that remains after cut-through.
+
+In practice the chain stays far smaller than a transparent UTXO chain with the same activity, but it is not free—range proofs and kernels have a cost, and sync still requires downloading state (for example via PIBD or a txhashset snapshot).
```
ardocrat, line 32: PIBD is term that requires a link https://github.com/mimblewimble/grin-rfcs/blob/master/text/0022-pibd-deployment.md, txhashset is a term too

doc/grin4bitcoiners.md
ardocrat: also need to explain stem (mention Dandelion once + link will be enough)

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
 
 ## Privacy and Fungibility
 
-There are 3 main properties of Grin transactions that make them private:
+Grin transactions are private in ways that Bitcoin’s transparent UTXO model is not. Three properties matter most:
 
-1. There are no addresses.
-1. There are no amounts.
-1. 2 transactions, one spending the other, can be merged in a block to form only one, removing all intermediary information.
+1. **No on-chain addresses.** Outputs are one-shot curve points. There is no reusable on-chain “payment address” (no Bitcoin-style address or scriptPubKey reuse) that links payments over time. Wallets may still use an *off-chain* contact handle for interaction—for example a slatepack address, Tor onion, or account public key—but that identifier never appears on the chain as a field tying outputs together.
```
Anynomouss, line 12: I would remove "Outputs are one-shot curve points.", probably to complicated for an average reader.
**Simplify:**
There are no on-chain "payment address" and as such no address reuse that links payments over time.
**Rephrase:** 
Wallets may still use an *off-chain* contact handle for interaction between users, such as a slatepack address, but that identifier never appears on the chain.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
+2. **No amounts.** Every transaction uses confidential transactions (CT). Amounts sit inside [Pedersen commitments](intro.md#balance) with [range proofs](intro.md#range-proofs) (Grin uses [Bulletproofs](https://eprint.iacr.org/2017/1066.pdf)), so outsiders cannot read transfer values.
```
Anynomouss, line 13: **No amounts.** Mimblewimble builds on the concept of Confidential Transactions. Amounts are hidden inside [Pedersen commitments](intro.md#balance), so outsiders cannot read transfer values.

I think the part about rangeproofs should not be put here since they have nothing to do with hiding the amounts, they are only used to proof no negative values are used. Best to avoid too much details that complicates reading.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
+3. **Cut-through and aggregation.** When an output is spent, Mimblewimble can remove the spent input/output pair from the long-term chain state. Inside a block, many transactions are merged into one aggregate set of inputs, outputs, and kernels, so a confirmed block does not look like a list of labeled payments.
```
Anynomouss, line 14: In your writing style, try to combine the minimalism of the original with being complete like in your suggested changes (this feedback can be applied to all changes you propose).

I would rewrite slightly:
**Cut-through and aggregation.** When an output is spent, Mimblewimble can freely forget (remove) the spent output from the blockchain. Transactions inside a block are merged into one aggregate set of inputs, outputs, and kernels.  Therefore, block information does does not look like a list of labeled payments but like one large transaction.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
 
-The 2 first properties mean that all transactions are indistinguishable from one another. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points).
+Because of (1) and (2), unspent outputs and kernels all look like random-looking data unless you participated in building them. Nodes can still verify that no money was created out of thin air by checking that commitments balance (homomorphic structure) and that range proofs and signatures are valid.
```
Anynomouss, line 16: I merged the original and your version based on which parts I found best:

Because of properties (1) and (2), transactions are indistinguishable from one another. Unless you directly participated in the transaction, all inputs and outputs look like random pieces of data (in lingo, they're all random curve points). Nodes can still verify that no money was created out of thin air by checking that commitments balance (homomorphic structure) and that range proofs and signatures are valid.
Moreover, there are no more transactions in a block. A Grin block looks just like one giant transaction and all original association between inputs and outputs is lost.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
+- After cut-through, the **kernel** of a transaction remains as a permanent ~100-byte footprint. The chain still records that “some” balanced state change occurred, even when the intermediate UTXOs are gone.
+
+So: Grin hides amounts and *on-chain* addresses strongly; it **weakens but does not eliminate** graph-style analysis of which outputs feed which later spends. That matches the Privacy Primer’s view that addresses and amounts are checked off as strong wins, while **input/output linking** remains an area of residual leakage and ongoing research.
 
-As explained in the previous section, thanks to the Mimblewimble transaction and block format we can merge transactions when an output is directly spent by the input of another. It's as if when Alice gives money to Bob, and then Bob gives it all to Carol, Bob was never involved and his transaction is actually never even seen on the blockchain.
```
Anynomouss, line 19: Why remove this? I think it is best to simply keep these parts.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
+## Scalability
 
-Pushing that further, between blocks, most outputs end up being spent sooner or later by another input. So *all spent outputs can be safely removed*. And the whole blockchain can be stored, downloaded and fully verified in just a few gigabytes or less (assuming a number of transactions similar to bitcoin).
```
Anynomouss, line 21: Again, why remove this part, it is good and true, right?

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
+Because spent outputs can be removed from the active UTXO set, a Grin node’s long-term storage grows primarily with **unspent outputs** (users holding coins), not with every historical payment. Full verification still needs kernels and headers; kernels are the main “per-transaction” residue that remains after cut-through.
 
-This means that the Grin blockchain scales with the number of users (unspent outputs), not the number of transactions. At the moment, there is one caveat to that: a small piece of data (called a *kernel*, about 100 bytes) needs to stay around for each transaction. But we're working on optimizing that as well.
+In practice the chain stays far smaller than a transparent UTXO chain with the same activity, but it is not free—range proofs and kernels have a cost, and sync still requires downloading state (for example via PIBD or a txhashset snapshot).
```
Anynomouss, line 32: Again, I would keep the original, mimblewimble has mostly benfits at hardly any costs. I mean do you have to say "range proofs and kernels have a cost, and sync still requires downloading state (for example via PIBD or a txhashset snapshot)."
Obviously any blockchain involves downloading data, that is not a cost just the obvious, especially if the whole blockchain is only a few gigabytes. I understand you do not want to write overtly optimistic, but neither should downsides be exaggerated.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
-* Multi-signature transactions.
-* Atomic swaps.
-* Time-locked transactions and outputs.
-* Lightning Network
+Those constructions live more in wallet protocols than in on-chain script programs.
 
 ## Emission Rate
 
-Bitcoin's 10 minute block time has its initial 50 btc reward cut in half every 4 years until there are 21 million bitcoin in circulation. Grin's emission rate is linear, meaning it never drops. The block reward is currently set at 60 grin with a block goal of 60 seconds. This still works because 1) dilution trends toward zero and 2) a non-negligible amount of coins gets lost or destroyed every year.
+Bitcoin’s block subsidy halves over time toward a fixed supply. Grin’s base emission is **linear** (a constant block reward), so supply grows without a hard cap. Dilution trends toward zero as the monetary base grows, and coins are also lost over time. The block target is on the order of one minute (see current consensus parameters for the exact reward).
```
Anynomouss, line 48: I mixed and matched your version and the original:

Bitcoin's 10-minute block time starts with an initial 50 BTC reward that cuts in half roughly every four years until 21 million bitcoin reaches circulation. Grin's emission rate is linear, meaning the block reward stays fixed and supply grows without a hard cap. The reward is currently set at 60 GRIN with a block goal of 60 seconds. This still works because: 1) dilution trends toward zero as the monetary base expands, and 2) a non-negligible amount of coins gets lost or destroyed every year.

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
 
 ### If transaction information gets removed, can I just cheat and create money?
 
-No, and this is where Mimblewimble and Grin shine. Confidential transactions are a form of [homomorphic encryption](https://en.wikipedia.org/wiki/Homomorphic_encryption). Without revealing any amount, Grin can verify that the sum of all transaction inputs equal the sum of transaction outputs, plus the fee. Going even further, comparing the sum of all money created by mining with the total sum of money that's being held, Grin nodes can check the correctness of the total money supply.
+No. Confidential transactions are designed so nodes can check that the sum of inputs equals the sum of outputs plus the fee **without** learning the amounts. Across the whole chain, nodes also check consistency of total supply against coinbase issuance. Cut-through removes spent UTXOs; it does not remove the need for those balance checks.
+
```
Anynomouss, line 61: Again, hard to see which is better, the original, or the new version of this section. I think I might prefer the original 🤔

doc/grin4bitcoiners.md
```diff
@@ -1,60 +1,76 @@
...
+For a careful split of “what Grin hides” vs “what still leaks,” read the [Privacy Primer](https://github.com/mimblewimble/docs/wiki/Grin-Privacy-Primer).
 
-### What about the quantum computaggedon?
+### What about quantum computers?
 
-In every Grin output, we also include a bit of hashed data, which is quantum safe. If quantum computing was to become a reality, we can safely introduce additional verification that would protect existing coins from being hacked.
+Commitments and signatures used today are not quantum-safe in the usual sense. Outputs also carry hash material that can support future migration plans if quantum threats become practical. Even if a quantum attacker could strip the cryptographic hiding of historical commitments, amounts are not known in advance—so there is no reliable way to pick “high-value” transactions to attack first. Treat long-range quantum risk like other cryptocurrencies: monitor standards and upgrades; do not assume current crypto is forever.
```
Anynomouss, line 72: Mmmm. I get your point. My problem is that many of these "subtle changes" changes the tone or sentiment of the document as well as change emphasis. For example your version contains a _scent of fear_, .... oooooo... be afraid "Treat long-range quantum risk like other cryptocurrencies: monitor standards and upgrades; do not assume current crypto is forever." 
I would suggest removing that part in any case. Probably that part is AI generated. AI's are trained to overstate the risk of quantageddon and essentially create FUD. I know, mine did after which I completely flipped its reasoning by pointing out all the fallacies and assumptions it made.

Anynomouss (changes requested):
I found this pull hard to review. At some points, there are improvements, at others I feel perhaps too much details is use, while in some places I simply prefer the original text.

In quite a few cases I found myself mixing parts of the original and the new to get the best of both worlds.

I think that 1)  feedback needs to be incorporated2) I would propose others might review this PR as well since the original documents was also quite good and minimal. In some case I do not like the change of emphasis/tone, see comments.

PR #3905, opened 2026-07-10
Migrate TUI from cursive to ratatui

### LINKED ISSUE #3841: Migrate tui from cursive to ratatui
Current TUI lib depends on external `ncurses` lib that should be installed. It also has some problems:
- Cross-compilation fails (even with zig)
- MacOS crashing on mouse clicks
- Windows is crashing on 1st run

Ratatui has big support at Rust community, has more features and has no external dependencies: https://github.com/ratatui/ratatui

iho:
## Summary

- Replaces the `cursive`/`cursive_table_view` TUI (which pulls in `pancurses`/`ncurses`) with `ratatui` + `crossterm`, which have no external C dependency
- Fixes the cross-compilation failures, macOS mouse-click crashes, and Windows first-run crashes described in the issue
- All 5 screens (Basic Status, Peers and Sync, Mining, Logs, Version Info) and their keybindings (`j`/`k`/arrows, `Tab`, `Enter`, `Esc`, `q`, plus `w`/`d` to toggle the Mining sub-view) are preserved
- New: mouse support (click a menu item to switch tabs, scroll wheel moves table selection — safe now that ncurses is gone), `PageUp`/`PageDown`/`Home`/`End` navigation in tables, and throttled redraws (repaint only on input or new data, 250ms cap) to reduce idle CPU
- Interactive column sorting (`cursive_table_view`) is dropped, since `ratatui::widgets::Table` has no equivalent — tables render in the order `ServerStats` provides. Can be revisited as a follow-up if wanted.
- `Controller::new`/`Controller::run()`'s public API (used by `src/bin/cmd/server.rs`) is unchanged
- Windows note: key events are filtered to `KeyEventKind::Press`, avoiding the classic doubled-keystroke issue with crossterm on Windows

## Test plan

- [x] `cargo build --bin grin` — 0 errors
- [x] `cargo test --bin grin tui::status` — both `update_sync_status` unit tests pass unmodified
- [x] `cargo clippy --bin grin` — zero findings in `src/bin/tui`
- [x] Confirmed via `otool -L` the built binary has no `ncurses`/`pancurses` linkage
- [x] Manually drove the TUI against a live testnet node in a tmux pty: all 5 tabs render with real peer/mining/sync data; keyboard navigation, mining workers/difficulty toggle, table selection clamping, `Home`/`End`/`PageUp`/`PageDown`, mouse tab-clicks and wheel scrolling (via SGR escape injection) all verified; error-dialog path renders correctly; `q` exits cleanly with exit code 0 and the terminal fully restored (raw mode off, alternate screen left, mouse capture disabled)

Fixes #3841


src/bin/tui/app.rs
```diff
@@ -0,0 +1,152 @@
+// Copyright 2024 The Grin Developers
```
ardocrat: 2026?

iho: Yes, thanks

wiesche89 (commented):
please rebase (merge conflicts)

src/bin/tui/ui.rs
```diff
@@ -15,168 +15,318 @@
...
+				Ok(Event::Resize(_, _)) => self.needs_redraw = true,
+				_ => {}
 			}
 		}
 
-		// Step the UI
-		self.cursive.step();
+		if self.needs_redraw || self.last_draw.elapsed() >= MAX_REDRAW_INTERVAL {
+			let app = &mut self.app;
+			let _ = self.terminal.draw(|f| draw(f, app));
```
wiesche89: Could we propagate terminal input and render errors instead of dropping them? A failed poll, read, or draw currently leaves the controller running without a usable UI, and a failed draw is still marked as completed.

src/bin/tui/ui.rs
```diff
@@ -15,168 +15,318 @@
...
-		grin_ui.cursive.add_fullscreen_layer(main_layer);
-
-		// Configure a callback (shutdown, for the first test)
-		let controller_tx_clone = grin_ui.controller_tx.clone();
-		let show_dialog_clone = grin_ui.show_dialog.clone();
-		grin_ui.cursive.add_global_callback('q', move |c| {
-			if show_dialog_clone.load(Ordering::Relaxed) {
-				c.pop_layer();
+	fn handle_key(&mut self, code: KeyCode) {
+		if let Some(dialog) = &self.app.dialog {
```
wiesche89, line 122: Mouse input is blocked for every dialog, but keyboard input is only blocked for errors. During startup, Tab/j/Enter can still change the view behind the info dialog. Could we make these dialogs modal while still allowing q to shut down?

src/bin/tui/logs.rs
```diff
@@ -12,93 +12,104 @@
...
-	buffer: VecDeque<LogEntry>,
+/// Word-wraps `text` to `width` columns, hard-breaking words that don't fit
+/// on their own. Empty input produces a single empty line.
+fn wrap_text(text: &str, width: usize) -> Vec<String> {
+	let width = width.max(1);
+	let mut out = Vec::new();
+	for raw_line in text.split('\n') {
+		let mut current = String::new();
+		let mut current_len = 0usize;
+		for word in raw_line.split(' ') {
```
wiesche89: Could we use Paragraph::wrap() and line_count() here instead of maintaining a separate wrapping implementation? That would preserve terminal display width and whitespace while still allowing the log view to stay bottom aligned.

src/bin/tui/ui.rs
```diff
@@ -15,168 +15,318 @@
...
+				MiningSubview::Difficulty => &mut self.app.mining_diff_table,
+			}),
+			_ => None,
+		};
+		if let Some(state) = state {
+			if len == 0 {
+				state.select(None);
+				return;
+			}
+			let current = state.selected().unwrap_or(0) as i64;
```
wiesche89: TableState starts at None, so the first Down/j/scroll jumps straight to row 1. Could we anchor the first move to row 0 and cover it with a small test?

src/bin/tui/constants.rs
```diff
@@ -12,55 +12,29 @@
...
 // Logo (not final, to be used somewhere eventually
-pub const _WELCOME_LOGO: &str = "                 GGGGG                      GGGGGGG         
-               GGGGGGG                      GGGGGGGGG      
-             GGGGGGGGG         GGGG         GGGGGGGGGG     
-           GGGGGGGGGGG       GGGGGGGG       GGGGGGGGGGG    
-          GGGGGGGGGGGG       GGGGGGGG       GGGGGGGGGGGG   
-         GGGGGGGGGGGGG       GGGGGGGG       GGGGGGGGGGGGG  
-        GGGGGGGGGGGGGG       GGGGGGGG       GGGGGGGGGGGGGG 
-        GGGGGGGGGGGGGG       GGGGGGGGGGGGGGGGGGGGGGGGGGGGG 
+pub const _WELCOME_LOGO: &str = "                 GGGGG                      GGGGGGG
```
wiesche89: After the migration this module only contains the unused _WELCOME_LOGO. Could we remove the file and mod constant,  as part of the cleanup?

src/bin/tui/ui.rs
```diff
@@ -15,168 +15,318 @@
...
+		}
 		true
 	}
 
-	/// Stop the UI
+	/// Stop the UI and restore the terminal
 	pub fn stop(&mut self) {
-		self.cursive.quit();
+		self.app.should_quit = true;
+		let _ = disable_raw_mode();
```
wiesche89: The terminal restore sequence is now repeated in the panic hook, stop(), and Drop. Could we centralize it and make the cleanup so these paths cannot drift?

src/bin/tui/ui.rs
```diff
@@ -15,168 +15,318 @@
...
-					UIMessage::UpdateStatus(update) => match *selection {
-						VIEW_BASIC_STATUS => {
-							status::TUIStatusView::update(&mut self.cursive, &update)
-						}
-						VIEW_MINING => mining::TUIMiningView::update(&mut self.cursive, &update),
-						VIEW_PEER_SYNC => peers::TUIPeerView::update(&mut self.cursive, &update),
-						_ => {}
-					},
+			match message {
+				UIMessage::UpdateStatus(update) => self.app.stats = Some(update),
```
wiesche89: Now that the UI owns App and status updates happen on the same controller thread, could we update app.stats directly and remove the single-message UI channel? It looks like leftover Cursive plumbing.

src/bin/tui/status.rs
```diff
@@ -15,339 +15,240 @@
...
+						"{} ({})",
+						tx_stats.stem_pool_size, tx_stats.stem_pool_kernels
+					),
+				));
+			} else {
+				lines.push(line("Transaction Pool Size:", "0 (0)"));
+				lines.push(line("Stem Pool Size:", "0 (0)"));
+			}
+			lines.push(Line::from(SEPARATOR));
+			// These three lines are reserved for mining config/status/network info,
```
wiesche89: These fields were never populated in the old view. Could we drop the reserved blank rows instead of carrying the unused placeholders into the new renderer?

src/bin/tui/ui.rs
```diff
@@ -15,168 +15,318 @@
...
+use ratatui::{Frame, Terminal};
+use std::io::{self, Stdout};
+use std::sync::mpsc;
+use std::time::{Duration, Instant};
+
+type Backend = CrosstermBackend<Stdout>;
+
+/// Redraw at least this often even without input or new data, so
+/// time-based fields (peer "last seen" ages, etc.) stay fresh.
+const MAX_REDRAW_INTERVAL: Duration = Duration::from_millis(250);
```
wiesche89: The time based fields only change once per second, while 250 ms still rebuilds the full UI four times per second when idle. Could this use a one-second refresh, or is there a measurement supporting the shorter interval?

src/bin/tui/logs.rs
```diff
@@ -12,93 +12,104 @@
...
-				level: Level::Info,
-			},
-		);
+/// Draw the logs view, bottom-anchoring the newest log lines.
+pub fn draw(f: &mut Frame, area: Rect, app: &App) {
+	let width = area.width as usize;
+	let height = area.height as usize;
 
-		LogBufferView { buffer }
+	// Walk entries newest-first, wrapping each until we have enough rows to
```
wiesche89: There’s quite a bit of commentary walking through this rendering step by step. Could we simplify it by using Ratatui’s wrapping and line counting support, then keep only the non obvious bottom alignment note?

wiesche89 (commented):
Ratatui looks like the right direction and works well on macOS. I couldn’t test Linux or Windows. I left a few comments that I’d prefer to address in this PR so the migration lands cleanly.

src/bin/tui/ui.rs
iho: Done in 6f78dcb5. `UI::step()` now returns `io::Result<bool>` and propagates failures from `event::poll`, `event::read`, and `terminal.draw` instead of swallowing them. On error the controller logs the failure, sets exit code 1, and leaves the loop so we no longer keep running with a dead UI or mark a failed draw as completed.

src/bin/tui/ui.rs
iho: Done in 6f78dcb5. All dialogs are modal now: keyboard input is blocked while either an info or error dialog is up. Error dialogs still accept Enter / q / Esc to exit; info dialogs (startup status) only allow `q` so you can shut down without Tab/j/Enter changing the view behind them. Mouse input was already blocked for every dialog.

src/bin/tui/logs.rs
iho: Done in 6f78dcb5. The custom `wrap_text` helper is gone. Logs now use `Paragraph::wrap(Wrap { trim: false })` plus `line_count()` (via the `unstable-rendered-line-info` feature) so wrapping matches terminal display width. Bottom alignment is kept by scrolling when content overflows, or rendering into a bottom sub-area when it is shorter than the pane.

src/bin/tui/ui.rs
iho: Done in 6f78dcb5. Selection logic is in `next_table_selection`: when `TableState` is still `None`, the first move anchors to row 0 (or the last row for End). Unit tests cover the unselected → 0 case, empty tables, and clamp/move behavior.

src/bin/tui/constants.rs
iho: Done in 6f78dcb5. Removed `src/bin/tui/constants.rs` and its `mod constants` entry; `_WELCOME_LOGO` was unused after the migration.

src/bin/tui/ui.rs
iho: Done in 6f78dcb5. Terminal teardown is centralized in `restore_terminal()`, shared by the panic hook, `stop()`, and `Drop`. `UI` also tracks a `restored` flag so the sequence only runs once and cannot drift between those paths.

src/bin/tui/ui.rs
iho: Done in 6f78dcb5. Removed the `UIMessage` channel and `types.rs`. Stats are written straight to `app.stats` on the controller thread (same thread that owns `App`), and the redraw flag is set there.

src/bin/tui/status.rs
iho: Done in 6f78dcb5. Dropped the three reserved blank rows (and the trailing separator that only framed them) from the status view so we no longer carry unused placeholders forward.

src/bin/tui/ui.rs
iho: Done in 6f78dcb5. Switched `MAX_REDRAW_INTERVAL` from 250 ms to 1 s so idle redraws match the cadence of the time-based fields (uptime, peer “last seen”, etc.) and the 1 s stats update. No measurement justified the shorter interval.

src/bin/tui/logs.rs
iho: Done in 6f78dcb5 (together with the wrap/`line_count` change above). Step-by-step wrapping commentary is gone; the log draw path uses ratatui wrapping and only documents the non-obvious bottom-alignment behavior.

Issue #3872, opened 2026-06-18
Hardware resource much increased after upgrading to v5.5


noobvie:
**Describe the bug**
I did few tests on some lowend vps (2 CPUs, 2GB RAM, 30 GB SSD), i notice the hardware resources CPU/IO increased a lot.

**To Reproduce**
Steps to reproduce the behavior, I did two kinds of tests
1. Copy from migrated chain_data of other 5.5, it's working at step Loading database... really long, could up to nearly 1 hour to be in running status
 <img width="1161" height="547" alt="Image" src="https://github.com/user-attachments/assets/e3551d77-9b98-4f0f-b697-df2c8c8c4d3e" />
sar stats

```
12:40:01        CPU     %user     %nice   %system   %iowait    %steal     %idle
12:50:00        all     13.82      0.00     31.85      9.60      0.51     44.22
13:00:01        all     12.90      0.00     13.78      0.06      0.42     72.85
13:10:01        all     22.14      0.00     21.91      3.94      0.70     51.31
13:20:01        all     11.97      0.00     32.68     21.82      0.63     32.89
Average:        all     20.70      0.03     46.88     17.61      0.63     14.15
```

iostat
Linux 6.8.0-124-generic        06/18/26        _x86_64_        (2 CPU)

```
avg-cpu:  %user   %nice %system %iowait  %steal   %idle
          20.59    0.03   46.55   17.55    0.63   14.65

Device             tps    kB_read/s    kB_wrtn/s    kB_dscd/s    kB_read    kB_wrtn    kB_dscd
loop0             0.00         0.00         0.00         0.00         14          0          0
sr0               0.00         0.01         0.00         0.00        334          0          0
vda            1785.30    130181.04      2398.40       527.72 5391771784   99335547   21856688
```

2. I synced chain_data by default, it's stuck at PIBD step again, not really sure if it's due to high intensive hardware consumption, but the CPU/IO all is used.

**Relevant Information**
Some logs: A lot of PeerNotBanned
```
20260618 12:43:00.203 ERROR grin_servers::grin::seed - failed to unban peer 66.154.127.84:3414: PeerNotBanned
20260618 12:43:20.375 ERROR grin_servers::grin::seed - failed to unban peer 66.154.127.84:3414: PeerNotBanned
20260618 12:43:40.602 ERROR grin_servers::grin::seed - failed to unban peer 66.154.127.84:3414: PeerNotBanned
[...]

PR #3906, opened 2026-07-10
Speed up chain compaction, harden PIBD validation, crash-safe compact

iho:
Three related fixes for low-end hardware reliability and post-PIBD stability, plus a stress harness for kill-during-compact.

## Compaction speed (#3872)

**Avoid quadratic scan when rewriting PMMR files.** `write_tmp_pruned` checked `prune_pos.contains` for every element against the whole remaining prune list. Both are sorted, so this was O(elements × prune_count). Now only the head of the remaining slice is compared: O(elements + prune_count).

**Bound the header walk in `init_output_pos_index`.** When any output was missing from the index, the rebuild walked every height from genesis to head and kept walking after all missing outputs were placed. It now binary-searches for the first relevant header and breaks once the last missing entry is indexed.

## PIBD segment hash validation (PR #3820 fallout)

Segment root validation only checked hashes used to rebuild the segment root, while apply writes every carried hash into the local PMMR. A peer could pass a valid root proof with poisoned intermediate hashes; failure only appeared at full state validation as a bad root (PIBD restart loop).

After the merkle proof succeeds, every carried hash must be bound to that root (consumed during reconstruction, or a child of a bound parent that hashes correctly). Dense segments from uncompacted peers still pass when consistent; unbound/wrong hashes fail early as `UnverifiableHash`.

## Crash-safe compaction

A hard kill between replacing hash/data and flushing the prune list left compacted files with a stale prune list → wrong shifts → bad root on restart. Compaction now stages all new files, fsyncs, writes a journal marker, then renames. On open, a present journal rolls renames forward; temps without a journal are discarded as abandoned prep.

## QEMU / weak-VPS stress harness

`etc/qemu-compact-stress/` + `store/examples/compact_crash_stress.rs`:

- Build a large pruned PMMR and time compaction (slow path under load)
- `SIGKILL` the compact process at `hash_tmp` / `data_tmp` / `journal` and reopen
- Modes: native, `qemu-user` (TCG), Docker `linux/amd64` with 0.5 CPU / 512MB (QEMU on arm hosts)
- Env-gated pause hooks in `check_compact` (no-op unless set)

```bash
./etc/qemu-compact-stress/run.sh              # native kill at journal
./etc/qemu-compact-stress/run.sh --all-phases
./etc/qemu-compact-stress/run.sh --weak       # weak VPS via Docker+QEMU
```

Manual workflow: `.github/workflows/compact-stress.yaml` (`workflow_dispatch`).

## Testing

- `cargo test -p grin_store -p grin_core -p grin_chain`
- `segment_carried_hash_verification`
- `compact_clears_journal`, `compact_journal_recovery_roll_forward`, `compact_orphan_temps_discarded`
- `spend_in_fork_and_compact`
- `./etc/qemu-compact-stress/run.sh --all-phases` (native smoke)

### DIFF
--- a/chain/src/txhashset/txhashset.rs
+++ b/chain/src/txhashset/txhashset.rs
@@ -714,9 +714,29 @@ impl TxHashSet {
 		let total_outputs = outputs_pos.len();
 		let max_height = batch.head()?.height;
 
+		// outputs_pos is sorted by pos, so binary search for the first header
+		// that can contain the first missing output. This avoids walking the
+		// entire chain when only a few recent index entries are missing.
+		let first_pos = outputs_pos[0].1;
+		let mut start_height = 1;
+		let mut end_height = max_height;
+		while start_height < end_height {
+			let search_height = (start_height + end_height) / 2;
+			let hash = header_pmmr.get_header_hash_by_height(search_height)?;
+			let h = batch.get_block_header(&hash)?;
+			if h.output_mmr_size < first_pos {
+				start_height = search_height + 1;
+			} else {
+				end_height = search_height;
+			}
+		}
+
 		let mut i = 0;
-		for search_height in 0..max_height {
-			let hash = header_pmmr.get_header_hash_by_height(search_height + 1)?;
+		for search_height in start_height..=max_height {
+			if i >= total_outputs {
+				break;
+			}
+			let hash = header_pmmr.get_header_hash_by_height(search_height)?;
 			let h = batch.get_block_header(&hash)?;
 			while i < total_outputs {
 				let (commit, pos1) = outputs_pos[i];
--- a/core/src/core/pmmr/segment.rs
+++ b/core/src/core/pmmr/segment.rs
@@ -19,6 +19,7 @@ use crate::core::pmmr::{self, Backend, ReadablePMMR, ReadonlyPMMR};
 use crate::ser::{Error, PMMRIndexHashable, PMMRable, Readable, Reader, Writeable, Writer};
 use croaring::Bitmap;
 use std::cmp::min;
+use std::collections::HashMap;
 use std::fmt::Debug;
 
 const MAX_SEGMENT_READ_ITEMS: u64 = 1_000_000;
@@ -102,6 +103,9 @@ pub enum SegmentError {
 	/// Mismatch between expected and actual root hash
 	#[error("Root hash mismatch")]
 	Mismatch,
+	/// A hash included in the segment could not be verified against the segment root
+	#[error("Unverifiable hash at pos {0}")]
+	UnverifiableHash(u64),
 }
 
 /// Tuple that defines a segment of a given PMMR
@@ -415,6 +419,19 @@ where
 		mmr_size: u64,
 		bitmap: Option<&Bitmap>,
 	) -> Result<Option<Hash>, SegmentError> {
+		self.root_with_known(mmr_size, bitmap).map(|(root, _)| root)
+	}
+
+	/// As `root`, but also returns the hash of every position that was bound
+	/// during the reconstruction: leaf hashes, computed parent hashes and the
+	/// segment hashes consumed at pruning boundaries. Used to verify that the
+	/// remaining hashes carried by the segment are consistent with the root.
+	fn root_with_known(
+		&self,
+		mmr_size: u64,
+		bitmap: Option<&Bitmap>,
+	) -> Result<(Option<Hash>, HashMap<u64, Hash>), SegmentError> {
+		let mut known = HashMap::new();
 		let (segment_first_pos, segment_last_pos) = self.segment_pos_range(mmr_size);
 		let mut hashes = Vec::<Option<Hash>>::with_capacity(2 * (self.identifier.height as usize));
 		let mut leaves0 = self.leaf_pos.iter().zip(&self.leaf_data);
@@ -462,10 +479,12 @@ where
 						(Some(l), Some(r)) => Some((l, r).hash_with_index(pos0)),
 						(None, Some(r)) => {
 							let l = self.get_hash(left_child_pos - 1)?;
+							known.insert(left_child_pos - 1, l);
 							Some((l, r).hash_with_index(pos0))
 						}
 						(Some(l), None) => {
 							let r = self.get_hash(right_child_pos - 1)?;
+							known.insert(right_child_pos - 1, r);
 							Some((l, r).hash_with_index(pos0))
 						}
 					}
@@ -481,12 +500,15 @@ where
 					)
 				}
 			};
+			if let Some(h) = hash {
+				known.insert(pos0, h);
+			}
 			hashes.push(hash);
 		}
 
 		if self.full_segment(mmr_size) {
 			// Full segment: last position of segment is subtree root
-			Ok(hashes.pop().unwrap())
+			Ok((hashes.pop().unwrap(), known))
 		} else {
 			// Not full (only final segment): peaks in segment, bag them together
 			let peaks = pmmr::peaks(mmr_size)
@@ -500,7 +522,9 @@ where
 					.ok_or_else(|| SegmentError::MissingHash(1 + pos0))?;
 				if lhash.is_none() && bitmap.is_some() {
 					// If this entire peak is pruned, load it from the segment hashes
-					lhash = Some(self.get_hash(pos0)?);
+					let h = self.get_hash(pos0)?;
+					known.insert(pos0, h);
+					lhash = Some(h);
 				}
 				let lhash = lhash.ok_or_else(|| SegmentError::MissingHash(1 + pos0))?;
 
@@ -509,7 +533,7 @@ where
 					Some(rhash) => Some((lhash, rhash).hash_with_index(mmr_size)),
 				};
 			}
-			Ok(Some(hash.unwrap()))
+			Ok((Some(hash.unwrap()), known))
 		}
 	}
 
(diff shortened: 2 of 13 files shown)


core/src/core/pmmr/segment.rs
```diff
@@ -548,7 +584,61 @@ where
...
+
+	/// Verify that every hash carried by this segment is bound to the segment
+	/// root that the merkle proof validates. Root reconstruction only consumes
+	/// hashes at pruning boundaries: any additional hashes would otherwise be
+	/// applied to the local PMMR unchecked, allowing a peer to hand us a
+	/// segment that passes root validation but corrupts the reconstructed MMR,
+	/// which only surfaces (without attribution) at full state validation.
+	/// A hash is accepted if it was consumed during reconstruction, or if it
+	/// hashes into an already accepted parent together with its sibling.
+	fn verify_carried_hashes(&self, mut known: HashMap<u64, Hash>) -> Result<(), SegmentError> {
```
ardocrat, line 600: Pruned hashes are still accepted when we are syncing against a node with https://github.com/mimblewimble/grin/pull/3820, so in the end we get `Invalid Root` before validation

store/tests/pmmr.rs
```diff
@@ -813,6 +813,170 @@ fn compact_twice() {
...
+		assert!(!dir.join("pmmr_prun.bin.compact").exists());
+		assert!(!dir.join("pmmr_leaf.bin.compact").exists());
+	}
+	teardown(data_dir);
+}
+
+/// If a journal is present with staged files, open rolls renames forward and
+/// yields a consistent PMMR (simulates hard kill after journal, before apply).
+#[test]
+fn compact_journal_recovery_roll_forward() {
```
ardocrat, line 847: For some reason test is crashing on Windows

store/tests/pmmr.rs
```diff
@@ -813,6 +813,170 @@ fn compact_twice() {
...
+		assert_eq!(root, pmmr.root().unwrap());
+		assert_eq!(pmmr.get_data(4).unwrap(), TestElem(4));
+	}
+	teardown(data_dir);
+}
+
+/// Staging files without a journal are abandoned prep — discarded on open,
+/// leaving the pre-compact live files intact.
+#[test]
+fn compact_orphan_temps_discarded() {
```
ardocrat, line 915: Test is crashing on Windows

store/src/pmmr.rs
```diff
@@ -446,27 +601,39 @@ impl<T: PMMRable> PMMRBackend<T> {
...
+		let prune_path = self.data_dir.join(PMMR_PRUN_FILE);
+		let prune_staging = compact_staging_path(&prune_path);
+		let new_prune = PruneList::new(Some(prune_path), new_prune_bitmap);
+		new_prune.write_to(&prune_staging)?;
+
+		let leaf_staging = compact_staging_path(&self.data_dir.join(PMMR_LEAF_FILE));
+		self.leaf_set.write_to(&leaf_staging)?;
+
+		// --- Phase 2: journal commit point (temps are durable) ---
+		write_compact_journal(&self.data_dir)?;
```
wiesche89, line 618: The journal is the commit point, but the staging directory entries are not synced before it, and the replacements are not synced before the marker is removed. Could we make these directory syncs mandatory and ordered? Otherwise a power loss can leave mixed files with no journal.

store/src/pmmr.rs
```diff
@@ -446,27 +601,39 @@ impl<T: PMMRable> PMMRBackend<T> {
...
+		compact_test_hook("journal");
+
+		// --- Phase 3: replace live files, then clear journal ---
+		debug!("compact: applying journaled file replacements...");
+		// Release mmaps/handles so renames succeed on all platforms.
+		self.hash_file.release();
```
wiesche89, line 624: If apply or reinit fails after these releases, the backend stays without file handles and callers only log the error or restart the sync loop. Could we complete recovery and reopen it before returning, or propagate this as a fatal node restart condition?

store/src/pmmr.rs
```diff
@@ -42,6 +83,105 @@ pub const PMMR_FILES: [&str; 4] = [
...
+		fs::remove_file(to)?;
+	}
+	fs::rename(from, to)?;
+	Ok(())
+}
+
+/// Apply any durable compact temps into place and clear the journal.
+/// Safe to call when some temps are already gone (partial prior apply).
+fn apply_compact_temps(data_dir: &Path) -> io::Result<()> {
+	replace_if_present(
```
wiesche89, line 110: The kill phases stop before any replacement or after all of them, but never in between. Could we add kill points after each hash/data/prune/leaf replacement? That is the main partial apply recovery guarantee here.

core/src/core/pmmr/segment.rs
```diff
@@ -548,7 +584,61 @@ where
...
+			let height = pmmr::bintree_postorder_height(pos0);
+			if height == 0 {
+				continue;
+			}
+			let left_pos0 = pos0 - (1 << height);
+			let right_pos0 = pos0 - 1;
+			let left = known
+				.get(&left_pos0)
+				.copied()
+				.or_else(|| self.get_hash(left_pos0).ok());
```
wiesche89, line 624: get_hash() scans hash_pos linearly and is called for many bound nodes here, making dense peer segments O(n^2). Since this is peer controlled PIBD input, could we index the carried hashes once or use binary search?

store/src/pmmr.rs
```diff
@@ -27,11 +29,50 @@ use croaring::Bitmap;
...
+/// qemu stress harness). Production runs leave these env vars unset so this
+/// is a pure no-op.
+///
+/// - `GRIN_COMPACT_READY_FILE`: path written with the current phase name when
+///   that phase is reached (`hash_tmp`, `data_tmp`, `journal`, `applied`).
+/// - `GRIN_COMPACT_PAUSE_PHASE`: if set to a phase name, sleep while at that
+///   phase so an external watchdog can SIGKILL the process.
+/// - `GRIN_COMPACT_CRASH_PAUSE_MS`: sleep duration in ms (default 5000 when
+///   pause phase is set).
+fn compact_test_hook(phase: &str) {
```
wiesche89, line 42: Could we gate these test hooks behind a Cargo feature? That keeps env reads and sleeps out of the production compaction path and prevents a stray GRIN_COMPACT_PAUSE_PHASE from pausing a real node.

store/examples/compact_crash_stress.rs
```diff
@@ -0,0 +1,417 @@
+// Copyright 2021 The Grin Developers
```
wiesche89, line 1: 2026

store/examples/compact_crash_stress.rs
```diff
@@ -0,0 +1,417 @@
...
+	if !saw {
+		let _ = child.kill();
+		let _ = child.wait();
+		panic!(
+			"timed out waiting for compact phase '{}' (ready file: {:?})",
+			phase, ready_file
+		);
+	}
+
+	// Hard kill: no graceful shutdown (simulates OOM killer / kill -9 / power loss).
```
wiesche89, line 355: SIGKILL covers process death, but not power loss: the page cache and directory metadata remain. Could we narrow this claim and keep power loss durability separate?

store/src/leaf_set.rs
```diff
@@ -181,6 +181,23 @@ impl LeafSet {
 		Ok(())
 	}
 
+	/// Write the current bitmap to an arbitrary path (used for journaled compact).
+	/// Does not update the in-memory backup; call `mark_flushed` after the write
+	/// is committed via rename.
+	pub fn write_to<P: AsRef<Path>>(&self, path: P) -> io::Result<()> {
```
wiesche89, line 187: This duplicates PruneList::write_to() and both methods are only used by compaction. Could we use one shared crate private bitmap writer instead of expanding both public APIs?

store/src/pmmr.rs
```diff
@@ -27,11 +29,50 @@ use croaring::Bitmap;
...
 const PMMR_HASH_FILE: &str = "pmmr_hash.bin";
 const PMMR_DATA_FILE: &str = "pmmr_data.bin";
 const PMMR_LEAF_FILE: &str = "pmmr_leaf.bin";
 const PMMR_PRUN_FILE: &str = "pmmr_prun.bin";
 const PMMR_SIZE_FILE: &str = "pmmr_size.bin";
+/// Marker written after all compact temps are durable and before any live
+/// file is replaced. Presence means an interrupted compact must roll forward.
+const PMMR_COMPACT_JOURNAL: &str = "pmmr_compact.journal";
+/// Suffix appended to leaf/prune paths for pre-commit compact temps
+/// (`pmmr_prun.bin.compact`). Distinct from `.tmp` used by single-file flushes.
```
wiesche89, line 74: Hash/data compaction still uses .tmp, while prune/leaf uses .compact, so this comment is a little misleading. Could we either use one convention or explain why both suffixes belong to the same journal transaction?

store/examples/compact_crash_stress.rs
```diff
@@ -0,0 +1,417 @@
...
+	drop(mmr);
+	backend.sync().unwrap();
+	eprintln!(
+		"prepare: push done in {:.1}s, pruning...",
+		t0.elapsed().as_secs_f64()
+	);
+
+	// Keep every Nth leaf so (prune_pct)% are removed. N = 100 / (100 - pct).
+	// e.g. prune_pct=50 → keep every 2nd; prune_pct=90 → keep every 10th.
+	let keep_every = (100u32 / (100 - prune_pct)).max(1);
```
wiesche89, line 190: This only approximates the requested percentage. For example, 60 selects keep_every = 2 and actually prunes about 50%, while the output still reports 60%. Could we calculate or report the actual percentage?

chain/src/txhashset/txhashset.rs
```diff
@@ -714,9 +714,29 @@ impl TxHashSet {
 		let total_outputs = outputs_pos.len();
 		let max_height = batch.head()?.height;
 
+		// outputs_pos is sorted by pos, so binary search for the first header
```
wiesche89, line 717: Could we add an equivalence test for missing entries near genesis and at the head, including consecutive headers with the same output_mmr_size? That would cover the boundaries of the new binary search.

wiesche89 (commented):
This contains three independent core changes: compaction performance, PIBD validation, and crash recovery with its test harness. Could we split them into three PRs? Each touches a different critical path and would be much easier to review, test, and revert separately.

Could you add before/after timings for both write_tmp_pruned and init_output_pos_index to the PR description? The stress harness is useful, but the expected improvement for #3872 is not recorded yet.

PR #3907, opened 2026-07-10
feat: optional TLS for P2P connections (#1420)

### LINKED ISSUE #1420: Add TLS to peer-to-peer connections
The goal is to improve privacy by making packet inspections a lot harder. In countries where cryptocurrencies or anonymous cryptocurrencies are illegal, this could be very important.

Ulttimately we could use Kovri or something similar, but this gets us 80% with what should just be very little code.

iho:
## Summary

Implements optional TLS for peer-to-peer connections ([#1420](https://github.com/mimblewimble/grin/issues/1420)).

- Encrypts P2P traffic with **rustls** (same stack as the HTTP API) for privacy against passive packet inspection
- **Opt-in** via `p2p_config.tls_enabled` (default `false`) so the existing plaintext network stays compatible
- Self-signed certificates are auto-generated under `{data_dir}/p2p_tls/` when paths are not configured
- Clients intentionally **do not verify** peer certificates — goal is privacy encryption, not PKI authentication
- Optional PEM paths: `tls_certificate_file` / `tls_certificate_key`
- Config comments updated in `grin-server.toml` template

### Peer discovery: `Capabilities::TLS` + `tls_required`

- New capability bit `Capabilities::TLS` (`0b1_0000_0000`), advertised when `tls_enabled = true`
- **Not** part of default capabilities (TLS remains opt-in for network compatibility)
- `p2p_config.tls_required` (default `false`): when set with `tls_enabled`, only attempt outbound connections to peers known to advertise TLS, and request `PEER_LIST | TLS` in peer-list queries
- Seeds, preferred, and allow-listed addresses are still tried for bootstrap (capabilities unknown until first successful handshake)
- Unknown peers (no stored capabilities) are skipped when `tls_required` is on

### Design notes

- New `Stream` abstraction wraps plain `TcpStream` or a mutex-shared rustls session (needed because reader/writer threads cannot `try_clone` a TLS session the way they do TCP)
- Both peers must enable TLS to connect; mixed TLS/plaintext will fail the handshake
- Not consensus-breaking; pure transport layer change

## How to enable

```toml
[server.p2p_config]
tls_enabled = true
# Optional: only connect to peers known to support TLS
# tls_required = true
# optional PEM paths:
# tls_certificate_file = "/path/to/p2p.crt"
# tls_certificate_key = "/path/to/p2p.key"
```

## Testing

- [x] `cargo test -p grin_p2p` (includes `peer_handshake_tls`, capability / `tls_required` helpers)
- [x] `cargo check -p grin_config -p grin_servers`
- [ ] Manual: two nodes with `tls_enabled = true` connecting on testnet/local

## Consensus / compatibility

- **Not consensus breaking**
- Default remains plaintext; operators must opt in
- Nodes with TLS enabled will not successfully complete Grin handshake with plaintext-only peers
- With `tls_required = true`, automatic outbound selection prefers the TLS peer mesh once capabilities are known

Closes #1420

### DIFF
--- a/servers/src/grin/seed.rs
+++ b/servers/src/grin/seed.rs
@@ -217,25 +217,30 @@ fn monitor_peers(
 	let default_peers = PeerAddrs::default();
 	let enough_outbound = peers.enough_outbound_peers();
 
+	// When tls_required, request PEER_LIST|TLS so returned addresses are TLS-capable when known.
+	let peer_list_caps = config.peer_list_request_capabilities();
+
 	if !enough_outbound {
 		// loop over connected peers that can provide peer lists
-		// ask them for their list of peers
+		// ask them for their list of peers (optionally TLS-capable only)
 		for p in peers
 			.iter()
 			.with_capabilities(p2p::Capabilities::PEER_LIST)
 			.connected()
 		{
 			trace!(
-				"monitor_peers: {}:{} ask {} for more peers",
+				"monitor_peers: {}:{} ask {} for more peers ({:?})",
 				config.host,
 				config.port,
 				p.info.addr,
+				peer_list_caps,
 			);
-			let _ = p.send_peer_request(p2p::Capabilities::PEER_LIST);
+			let _ = p.send_peer_request(peer_list_caps);
 		}
 	}
 
 	// Attempt to connect to any preferred peers even if all outbound slots are full.
+	// Preferred peers are bootstrap; do not apply the TLS capability filter.
 	let peers_preferred = config.peers_preferred.as_ref().unwrap_or(&default_peers);
 	let peers_deny = config.peers_deny.as_ref().unwrap_or(&default_peers);
 	let connected_peers: Vec<_> = peers.iter().connected().into_iter().collect();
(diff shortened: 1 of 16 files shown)


servers/src/grin/seed.rs
```diff
@@ -272,10 +277,15 @@ fn monitor_peers(
 		new_peers.push(hp.addr);
 	}
 	// always check min 32 (max 96, if there are no healthy) random unknown peers received from peer list request.
-	let req_unk_count = cmp::max(
-		max_peer_attempts / 2 - new_peers.len() + max_peer_attempts / 4,
-		max_peer_attempts / 4,
-	);
+	// When TLS is required, skip unknown peers — we only know TLS support after a successful handshake.
```
ardocrat: `Unknown` state is valid for all new peers that are coming from peer list request

wiesche89 (commented):
please fix merge conflicts

servers/src/grin/seed.rs
iho: Agreed — fixed in the rebase. Unknown peers from peer-list responses are dialed again so handshake can discover TLS (and other) capabilities. `tls_required` only filters peers that already have stored capabilities (healthy/defunct) via `accepts_outbound_peer_capabilities`.

iho (commented):
Rebased onto current `staging` and resolved conflicts in `servers/src/grin/seed.rs` and `servers/src/grin/server.rs`.

Also addressed @ardocrat's note: `State::Unknown` peers from peer-list requests are dialed again under `tls_required` (capabilities are only known after handshake). Preferred-peer handling from staging is preserved.

`cargo test -p grin_p2p` and `cargo check -p grin_config -p grin_servers` pass.

PR #3911, opened 2026-07-13
rest: use tokio channel for shutdown signal

ardocrat:
- use `tokio` channel for shutdown signal, so it can be called remotely from another thread

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ name = "grin"
 path = "src/bin/grin.rs"
 
 [dependencies]
-blake2-rfc = "0.2"
 chrono = "0.4.11"
 thiserror = "1"
 clap = { version = "2.33", features = ["yaml"] }
@@ -29,16 +28,15 @@ cursive_table_view = "0.15.0"
 humansize = "1.1.0"
 serde = "1"
 serde_derive = "1"
-futures = "0.3.19"
 serde_json = "1"
 log = "0.4"
 term = "0.6"
+tokio = { version = "1", features = ["full"] }
 
 grin_api = { path = "./api", version = "5.5.1-alpha.0" }
 grin_config = { path = "./config", version = "5.5.1-alpha.0" }
 grin_chain = { path = "./chain", version = "5.5.1-alpha.0" }
 grin_core = { path = "./core", version = "5.5.1-alpha.0" }
-grin_keychain = { path = "./keychain", version = "5.5.1-alpha.0" }
 grin_p2p = { path = "./p2p", version = "5.5.1-alpha.0" }
 grin_servers = { path = "./servers", version = "5.5.1-alpha.0" }
 grin_util = { path = "./util", version = "5.5.1-alpha.0" }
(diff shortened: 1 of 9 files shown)


api/src/rest.rs
```diff
@@ -141,77 +141,31 @@ impl ApiServer {
...
 
 	/// Stops the API server.
 	pub fn stop(&mut self) -> bool {
-		if self.shutdown_sender.is_some() {
-			let tx = self.shutdown_sender.as_mut().unwrap();
-			let m = oneshot::channel::<()>();
-			let tx = std::mem::replace(tx, m.0);
-			match tx.send(()) {
+		if let Some(tx) = self.shutdown_sender.take() {
+			let rt = Runtime::new().unwrap();
```
wiesche89: Could we avoid creating a new runtime here? stop() currently panics when called from an existing Tokio runtime. A synchronous send would be safer and simpler.

Cargo.toml
```diff
@@ -29,16 +28,15 @@ cursive_table_view = "0.15.0"
 humansize = "1.1.0"
 serde = "1"
 serde_derive = "1"
-futures = "0.3.19"
 serde_json = "1"
 log = "0.4"
 term = "0.6"
+tokio = { version = "1", features = ["full"] }
```
wiesche89: The binary only uses Tokio’s channel, so would the sync feature be enough here instead of full?

api/src/rest.rs
```diff
@@ -141,77 +141,31 @@ impl ApiServer {
...
-		&mut self,
-		addr: SocketAddr,
-		router: Router,
-		conf: TLSConfig,
-		api_chan: &'static mut (oneshot::Sender<()>, oneshot::Receiver<()>),
+		api_chan: (mpsc::Sender<()>, mpsc::Receiver<()>),
 	) -> Result<thread::JoinHandle<()>, Error> {
 		if self.shutdown_sender.is_some() {
 			return Err(Error::Internal(
 				"Can't start HTTPS API server, it's running already".to_string(),
```
wiesche89: This message now also applies to the HTTP server. Could we say “API server” instead of “HTTPS API server”?

wiesche89 (commented):
Since #3912 is now merged, this PR conflicts with staging. Could you rebase it and keep the provider setup in the common start() path? I’ll recheck the final diff afterwards.

Cargo.toml
ardocrat: Changed, dependencies were not changed here

api/src/rest.rs
ardocrat: changed

api/src/rest.rs
ardocrat: made stop async

api/src/handlers.rs
```diff
@@ -121,7 +121,7 @@ where
 			loop {
 				thread::sleep(std::time::Duration::from_millis(100));
 				if stop_state.is_stopped() {
-					apis.stop();
+					let _ = apis.stop();
```
wiesche89, line 124: stop() is async now, but this future is dropped immediately, so none of its code runs. The server only stops because apis is dropped and closes the channel. Could we keep stop() synchronous and use try_send() here? That would also let an exit or restart reason be delivered later.

ardocrat: removed async from stop, try_send is a solution

PR #3912, opened 2026-07-14
rest: install rustls provider at single place

ardocrat:
To avoid installing at multiple places and manually when using this crate.

### DIFF
--- a/api/src/rest.rs
+++ b/api/src/rest.rs
@@ -143,6 +143,7 @@ impl ApiServer {
 		conf: Option<TLSConfig>,
 		api_chan: &'static mut (oneshot::Sender<()>, oneshot::Receiver<()>),
 	) -> Result<thread::JoinHandle<()>, Error> {
+		let _ = rustls::crypto::ring::default_provider().install_default();
 		match conf {
 			Some(conf) => self.start_tls(addr, router, conf, api_chan),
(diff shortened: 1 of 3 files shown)


api/src/rest.rs
```diff
@@ -197,6 +197,9 @@ impl ApiServer {
 		let tx = std::mem::replace(tx, m.0);
 		self.shutdown_sender = Some(tx);
 
+		rustls::crypto::ring::default_provider()
+			.install_default()
```
wiesche89: Could we treat an already installed provider as okay and finish the TLS setup before marking the server as running? Otherwise a failed start can leave the server stuck in the running state.

api/src/rest.rs
```diff
@@ -197,6 +197,9 @@ impl ApiServer {
 		let tx = std::mem::replace(tx, m.0);
 		self.shutdown_sender = Some(tx);
 
+		rustls::crypto::ring::default_provider()
```
wiesche89: Should this also cover the API client? Right now the client can still fail unless a TLS server has already been started or the provider was set up manually.

api/src/rest.rs
ardocrat: OK, will not return an error here.

api/src/rest.rs
ardocrat: do not return error here now

PR #3914, opened 2026-07-15
Do not disconnect from preferred peers when there is enough outbound

ardocrat:
- check if peer is preferred after connection to not disconnect if its preferred
- add logging of reason to stop peer
- do not store stopped peer at memory when there is enough outbound

### DIFF
--- a/p2p/src/peers.rs
+++ b/p2p/src/peers.rs
@@ -60,39 +60,50 @@ impl Peers {
 		}
 	}
 
-	/// Adds the peer to our internal peer mapping. Note that the peer is still
-	/// returned so the server can run it.
+	/// Adds the peer to our internal peer mapping and records the connection.
 	pub fn add_connected(&self, peer: Arc<Peer>) -> Result<(), Error> {
-		let peer_data: PeerData;
+		let peer_data = Self::connected_peer_data(&peer.info);
 		{
 			// Scope for peers vector lock - don't hold the peers lock while adding to db.
 			let mut peers = self.peers.try_write_for(LOCK_TIMEOUT).ok_or_else(|| {
 				error!("add_connected: failed to get peers lock");
 				Error::Timeout
 			})?;
-			peer_data = PeerData {
-				addr: peer.info.addr,
-				capabilities: peer.info.capabilities,
-				user_agent: peer.info.user_agent.clone(),
-				flags: State::Healthy,
-				last_banned: 0,
-				ban_reason: ReasonForBan::None,
-				last_connected: Utc::now().timestamp(),
-				last_attempt: Utc::now().timestamp(),
-			};
 			debug!("Adding newly connected peer {}.", peer_data.addr);
 			peers.insert(peer_data.addr, peer);
 		}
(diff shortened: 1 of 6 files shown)


p2p/src/serv.rs
```diff
@@ -208,8 +209,15 @@ impl Server {
 					&self.handshake,
 					self.peers.clone(),
 				)?;
-				if self.peers.enough_outbound_peers() {
-					peer.stop();
+				if self.peers.enough_outbound_peers()
```
wiesche89, line 213: Peers::add_connected() still skips outbound peers when enough_outbound_peers() is true, so the preferred peer is not retained in the connected-peer map. Also, monitor_peers() only explicitly queues preferred peers while !enough_outbound, which can delay reconnects at full capacity. Could we fix the admission path here and handle reconnects in a follow-up?

p2p/src/peers.rs
```diff
@@ -489,7 +489,7 @@ impl Peers {
 		match peers.remove(&peer_addr) {
 			Some(peer) => {
 				warn!("disconnecting peer {} ({})", peer_addr, reason);
-				peer.stop();
+				peer.stop("disconnect_peer");
```
wiesche89: Can we pass the existing reason to stop() here instead of replacing it with "disconnect_peer"?

p2p/src/peers.rs
```diff
@@ -295,7 +295,7 @@ impl Peers {
 				};
 				// Mark peer as defunct after ping failure.
 				let _ = self.update_state(p.info.addr, State::Defunct);
-				p.stop();
+				p.stop("Error pinging");
```
wiesche89: could we keep the new reason strings consistently lowercase and descriptive, e.g. ping error? They currently mix styles such as Error pinging, error broadcast, and clean peers.

p2p/src/serv.rs
ardocrat: yes, also this method is called for inbound, so enough_outbound must be provided as optional check to skip

p2p/src/peers.rs
ardocrat: changed

p2p/src/peers.rs
ardocrat: changed

p2p/src/serv.rs
ardocrat: fixed, to do not store stopped

servers/src/grin/seed.rs
```diff
@@ -233,23 +232,22 @@ fn monitor_peers(
...
 		}
 	}
 
+	// Attempt to connect to any preferred peers even if all outbound slots are full.
+	let peers_preferred = config.peers_preferred.as_ref().unwrap_or(&default_peers);
 	let peers_deny = config.peers_deny.as_ref().unwrap_or(&default_peers);
+	for preferred in peers_preferred.peers.iter() {
+		let connected = peers
+			.iter()
+			.connected()
```
ardocrat: we can use iter only once here out of loop

PR #3915, opened 2026-07-18
Make the sparse PIBD segment optimization (#3820) safe to apply

iho:
# Make the sparse PIBD segment optimization (#3820) safe to apply

This PR carries #3820's segment size optimization together with the fixes it needs to
be deployable. It is intended either to supersede #3820 or to be merged into it -
happy to go whichever way the maintainers and @mkorovkin2 prefer.

## Why #3820 needs this

#3820 optimizes `Segment::from_pmmr` to only ship hashes at pruning boundaries derived
from the unspent bitmap. Testing it against a live node reproduced @ardocrat's report
on the PR thread: PIBD sync from a #3820 peer ends in `error validating PIBD state:
Invalid Root` (peer present from the start) or a `PIBD Reported Failure - Restarting
Sync` loop (peer appears mid-sync), with no peer attribution.

The root cause is that #3820 decides **hash** inclusion from the bitmap (spent status
at the archive header) while **leaf** inclusion is decided from on-disk presence -
and on any node between compaction passes those two signals disagree. For a subtree
that is fully spent per the bitmap but only partially compacted on disk (some leaves
spent recently, still in the files; others compacted away), the producer ships the
spent-but-present leaves as data **and** collapses the whole region into a single
higher boundary hash. The same positions are described twice, in incompatible ways.

Crucially, **no receiver-side validation can catch this**:

- Root reconstruction (`Segment::root`) ignores leaf data at bitmap-spent positions
  and consumes the boundary hash, which is genuinely bound to the proof-validated
  root - the segment is cryptographically consistent. This also holds under #3906's
  carried-hash hardening (`verify_carried_hashes`): every carried hash *is* bound to
  the root. There is nothing wrong with the segment as a commitment; it is wrong as
  an *application plan*.
- Apply (`TxHashSet::apply_output_segment`) then pushes the spent leaves at the
  frontier, and on reaching the boundary hash (guarded only by `pos0 >= size`) calls
  `push_pruned_subtree` for a subtree that **overlaps the leaves it just appended**.
  `prune_list.append` registers positions as pruned that are physically present in
  the hash file, desyncing every subsequent shift computation. Apply reports success;
  the corruption only surfaces at full state validation as Invalid Root.

Reproduced deterministically in `store/tests/segment.rs::
redundant_leaf_and_boundary_hash_segment_corrupts_apply`: the redundant segment shape
passes `validate()` and produces a divergent root through the exact
`sort_pmmr_hashes_and_leaves`/apply insertion rules.

This is also why the fix belongs in the **producer**, not behind a `PIBD_HIST_2`
capability bit (discussed on the #3820 thread): with the producer emitting only
segments that satisfy the apply invariant, existing deployed receivers handle sparse
segments correctly as-is - the wire format is already self-describing. A capability
bit would only be needed to protect old nodes from the *broken* producer.

## What this PR changes

**1. `Segment::from_pmmr` derives its hash set from the receiver's own reconstruction
walk instead of a bitmap heuristic** (`core/src/core/pmmr/segment.rs`)

Two passes:
- Pass 1 selects exactly the leaves receiver-side `root()` consumes: present on disk
  and required by the bitmap, its sibling, or the final uneven leaf.
- Pass 2 walks the segment bottom-up in postorder - the same traversal `root()`
  performs - tracking which positions are derivable from the selected leaves. At each
  parent with exactly one derivable child it emits the other child's hash (mirroring
  the `(None, Some)`/`(Some, None)` lookups in `root()`). Subtree roots still
  underivable at the end of the walk (fully pruned peaks of the final partial
  segment, or the root of a fully pruned segment) are emitted directly, matching
  where `root()` and `first_unpruned_parent()` probe.

Producer and receiver can no longer disagree about which positions are covered by
which hash. As a bonus the redundant spent-but-uncompacted leaves fold into the
boundary hash instead of being shipped twice, so segments get *smaller* than #3820's
in exactly the mixed regions where #3820 broke.

**2. Receiver-side guard: reject overlapping pruned-subtree collapses instead of
corrupting state** (`chain/src/txhashset/txhashset.rs`)

`apply_output_segment`/`apply_rangeproof_segment` now check that a carried hash's
subtree lies entirely beyond the local frontier (`bintree_leftmost(pos0) >= size`)
before `push_pruned_subtree`, returning `Error::InvalidSegment` otherwise.
Defense-in-depth: any future producer bug of this class fails fast at apply time with
attribution, instead of surfacing hours later as an unattributable Invalid Root and a
restart loop.

**3. Desegmenter fixes for small chains** (`chain/src/txhashset/desegmenter.rs`)

Found while trying to run PIBD end-to-end on test chains; both are unreachable on
mainnet but panic or stall any chain with <= 1024 outputs, including the canned
chains for `test_pibd_copy`:
- `calc_bitmap_mmr_sizes`: `unwrap_or` evaluates its fallback eagerly, panicking on
  the empty peak set of a single-chunk bitmap MMR. Now evaluated lazily, degrading
  to 0.
- `next_desired_segments`: the bitmap branch's strict `>` size comparison never
  requests the first segment of a single-leaf bitmap MMR, silently stalling sync
  (positions `0..size-1` are present locally, so a segment whose last position
  *equals* the local size still contains new data).

## Testing

- `cargo test -p grin_core -p grin_store -p grin_chain`: 221 passed.
- New: `sparse_segment_mixed_compaction_round_trip` - a segment from a
  partially-compacted region (spent+compacted sibling pair, spent-but-present pair
  under the same bitmap boundary) validates and round-trips through the desegmenter
  apply rules to the identical root, and asserts the mixed region collapses to the
  single boundary hash with no leaf data beneath it.
- New: `redundant_leaf_and_boundary_hash_segment_corrupts_apply` - regression test
  pinning down why the bitmap-only heuristic was unsafe: the redundant segment shape
  validates but must not silently reproduce the source root through apply.
- All pre-existing segment tests pass unchanged, including #3820's updated
  `ser_round_trip` (the optimized sizes are preserved) and the
  `prunable_mmr`/`pruned_segment` pruning-pattern batteries, which caught two edge
  cases in earlier iterations of the selection walk (fully pruned segment roots and
  pruned peaks of the final partial segment) - both covered by the final version.

Note: `test_pibd_copy_sample` (ignored, canned-chain end-to-end) currently fails on
master before reaching any segment logic - first run panics in
`calc_bitmap_mmr_sizes` (fixed here), reruns then fail opening the fixture because
the first open migrates/mutates the checked-in chain data in place. With this PR the
desegmenter panic is fixed; refreshing the canned fixtures is left as a separate
task.

### DIFF
--- a/chain/src/txhashset/desegmenter.rs
+++ b/chain/src/txhashset/desegmenter.rs
@@ -486,9 +486,13 @@ impl Desegmenter {
 				self.bitmap_mmr_size,
 				self.default_bitmap_segment_height,
 			);
-			// Advance iterator to next expected segment
+			// Advance iterator to next expected segment. `local_pmmr_size` is
+			// a count (positions 0..size-1 are present locally), so a segment
+			// whose last position equals it still contains new data and must
+			// be requested: with a strict `>` a single-leaf bitmap MMR (last
+			// position 0, local size 0) is never requested and sync stalls.
 			while let Some(id) = identifier_iter.next() {
-				if id.segment_pos_range(self.bitmap_mmr_size).1 > local_pmmr_size {
+				if id.segment_pos_range(self.bitmap_mmr_size).1 >= local_pmmr_size {
 					if !self.has_bitmap_segment_with_id(id) {
 						return_vec.push(SegmentTypeIdentifier::new(SegmentType::Bitmap, id));
 						if return_vec.len() >= max_elements {
@@ -672,18 +676,23 @@ impl Desegmenter {
 			"pibd_desegmenter - expected number of leaves in bitmap MMR: {}",
 			self.bitmap_mmr_leaf_count
 		);
-		// Total size of Bitmap PMMR
+		// Total size of Bitmap PMMR. `unwrap_or` evaluates its argument
+		// eagerly, so the fallback used to run - and panic on an empty peak
+		// set - whenever the bitmap MMR had a single leaf (output MMR with
+		// <= 1024 leaves). Evaluate it lazily and degrade to 0 instead of
+		// panicking when even the reduced peak set is empty.
 		self.bitmap_mmr_size =
 			1 + pmmr::peaks(pmmr::insertion_to_pmmr_index(self.bitmap_mmr_leaf_count))
 				.last()
-				.unwrap_or(
-					&(pmmr::peaks(pmmr::insertion_to_pmmr_index(
-						self.bitmap_mmr_leaf_count - 1,
+				.copied()
+				.unwrap_or_else(|| {
+					pmmr::peaks(pmmr::insertion_to_pmmr_index(
+						self.bitmap_mmr_leaf_count.saturating_sub(1),
 					))
 					.last()
-					.unwrap()),
-				)
-				.clone();
+					.copied()
+					.unwrap_or(0)
+				});
 
 		trace!(
 			"pibd_desegmenter - expected size of bitmap MMR: {}",
--- a/chain/src/txhashset/segmenter.rs
+++ b/chain/src/txhashset/segmenter.rs
@@ -57,7 +57,7 @@ impl Segmenter {
 		let now = Instant::now();
 		let txhashset = self.txhashset.read();
 		let kernel_pmmr = txhashset.kernel_pmmr_at(&self.header);
-		let segment = Segment::from_pmmr(id, &kernel_pmmr, false)?;
+		let segment = Segment::from_pmmr(id, &kernel_pmmr, None)?;
 		debug!(
 			"kernel_segment: id: ({}, {}), leaves: {}, hashes: {}, proof hashes: {}, took {}ms",
 			segment.id().height,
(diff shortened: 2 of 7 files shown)


ardocrat (commented):
This PR syncing against `staging` branch without issue, but https://github.com/mimblewimble/grin/pull/3915 sync against https://github.com/mimblewimble/grin/pull/3820 gives an error:

```
20260719 23:34:58.334 DEBUG grin_chain::txhashset::txhashset - Error returned, discarding txhashset extension: Invalid segment: rangeproof segment hash at pos 10282164 would overwrite 3 already-applied position(s) starting at 10282134 (sparse/dense segment mismatch)
20260719 23:34:58.609 ERROR grin_servers::grin::sync::state_sync - error applying segment: Invalid segment: rangeproof segment hash at pos 10282164 would overwrite 3 already-applied position(s) starting at 10282134 (sparse/dense segment mismatch)
20260719 23:34:58.622 ERROR grin_servers::grin::sync::state_sync - PIBD Reported Failure - Restarting Sync
```

PR #3917, opened 2026-07-24
REST address bind fix

ardocrat:
- check if address is free before server start

### DIFF
--- a/api/src/rest.rs
+++ b/api/src/rest.rs
@@ -148,7 +148,6 @@ impl ApiServer {
 				"Can't start API server, it's running already".to_string(),
 			));
 		}
-		self.shutdown_sender = Some(api_chan.0);
 
 		let _ = rustls::crypto::ring::default_provider().install_default();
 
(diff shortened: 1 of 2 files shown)


api/src/rest.rs
```diff
@@ -143,6 +143,14 @@ impl ApiServer {
 		conf: Option<TLSConfig>,
 		api_chan: (mpsc::Sender<()>, mpsc::Receiver<()>),
 	) -> Result<thread::JoinHandle<()>, Error> {
+		// Check if provided address is free.
+		{
+			match std::net::TcpListener::bind(addr) {
```
wiesche89: The listener is dropped here and start_server binds again later. I reproduced two concurrent starts both returning Ok, although only one can bind. Could we bind once, pass that listener to Tokio, and cover this with a test?

api/src/rest.rs
```diff
@@ -143,6 +143,14 @@ impl ApiServer {
 		conf: Option<TLSConfig>,
 		api_chan: (mpsc::Sender<()>, mpsc::Receiver<()>),
 	) -> Result<thread::JoinHandle<()>, Error> {
+		// Check if provided address is free.
```
wiesche89: This check runs before the existing “already running” guard, so a second start now only returns “address in use”. Could we keep the guard first?

api/src/rest.rs
ardocrat: converted

api/src/rest.rs
ardocrat: check moved after existing

PR #3920, opened 2026-07-29
Better init of global variables

ardocrat:
Change from OneTime's `init` to `init_if_unset` to do not check if it was already set to do not panic and avoid data races in multi-threading (at tests).

### DIFF
--- a/core/src/global.rs
+++ b/core/src/global.rs
@@ -178,13 +178,13 @@ thread_local! {
 
 /// One time initialization of the global chain_type.
 /// Will panic if we attempt to re-initialize this (via OneTime).
-pub fn init_global_chain_type(new_type: ChainTypes) {
-	GLOBAL_CHAIN_TYPE.init(new_type)
+pub fn init_global_chain_type(new_type: ChainTypes) -> bool {
+	GLOBAL_CHAIN_TYPE.init_if_unset(new_type)
 }
 
 /// Set the global chain_type using an override
 pub fn set_global_chain_type(new_type: ChainTypes) {
-	GLOBAL_CHAIN_TYPE.set(new_type, true);
+	GLOBAL_CHAIN_TYPE.set(new_type);
 }
 
 /// Set the chain type on a per-thread basis via thread_local storage.
@@ -194,16 +194,15 @@ pub fn set_local_chain_type(new_type: ChainTypes) {
 
 /// Get the chain type via thread_local, fallback to global chain_type.
 pub fn get_chain_type() -> ChainTypes {
-	CHAIN_TYPE.with(|chain_type| match chain_type.get() {
-		None => {
+	CHAIN_TYPE.with(|chain_type| {
+		chain_type.get().unwrap_or_else(|| {
 			if !GLOBAL_CHAIN_TYPE.is_init() {
 				panic!("GLOBAL_CHAIN_TYPE and CHAIN_TYPE unset. Consider set_local_chain_type() in tests.");
 			}
 			let chain_type = GLOBAL_CHAIN_TYPE.borrow();
 			set_local_chain_type(chain_type);
 			chain_type
-		}
-		Some(chain_type) => chain_type,
+		})
 	})
 }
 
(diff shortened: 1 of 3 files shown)


util/src/lib.rs
```diff
@@ -84,17 +84,19 @@ where
 	}
 
 	/// Initializes the OneTime, should only be called once after construction.
-	/// Will panic (via assert) if called more than once.
-	pub fn init(&self, value: T) {
-		self.set(value, false);
+	/// Returns `false` if already was initialized.
+	pub fn init_if_unset(&self, value: T) -> bool {
```
wiesche89, line 89: init_if_unset() atomically fixes the reproduced wallet race. Please keep init() and the existing set() API and add the new method alongside them: duplicate adapter initialization is still a programming error.

servers/src/common/adapters.rs
```diff
@@ -841,8 +830,8 @@ where
 
 	/// Initialize a NetToChainAdaptor with reference to a Peers object.
 	/// Should only be called once.
-	pub fn init(&self, peers: Arc<p2p::Peers>) {
-		self.peers.init(Arc::downgrade(&peers));
+	pub fn init(&self, peers: Arc<p2p::Peers>) -> bool {
```
wiesche89: These adapters are wired once per server and are unrelated to the wallet test race. Please keep using strict init() here so accidental duplicate initialization remains immediately visible.

servers/src/common/adapters.rs
```diff
@@ -1401,14 +1390,14 @@ where
...
-			hooks: hooks,
+			hooks,
 		}
 	}
 
 	/// Initialize a ChainToPoolAndNetAdapter instance with handle to a Peers
 	/// object. Should only be called once.
-	pub fn init(&self, peers: Arc<p2p::Peers>) {
-		self.peers.init(Arc::downgrade(&peers));
+	pub fn init(&self, peers: Arc<p2p::Peers>) -> bool {
```
wiesche89: These adapters are wired once per server and are unrelated to the wallet test race. Please keep using strict init() here so accidental duplicate initialization remains immediately visible.

servers/src/common/adapters.rs
```diff
@@ -1499,8 +1488,8 @@ impl PoolToNetAdapter {
 	}
 
 	/// Setup the p2p server on the adapter
-	pub fn init(&self, peers: Arc<p2p::Peers>) {
-		self.peers.init(Arc::downgrade(&peers));
+	pub fn init(&self, peers: Arc<p2p::Peers>) -> bool {
```
wiesche89: These adapters are wired once per server and are unrelated to the wallet test race. Please keep using strict init() here so accidental duplicate initialization remains immediately visible.

servers/src/common/adapters.rs
```diff
@@ -1528,8 +1517,8 @@ impl PoolToChainAdapter {
 	}
 
 	/// Set the pool adapter's chain. Should only be called once.
-	pub fn set_chain(&self, chain_ref: Arc<chain::Chain>) {
-		self.chain.init(Arc::downgrade(&chain_ref));
+	pub fn set_chain(&self, chain_ref: Arc<chain::Chain>) -> bool {
```
wiesche89: These adapters are wired once per server and are unrelated to the wallet test race. Please keep using strict init() here so accidental duplicate initialization remains immediately visible.

util/src/lib.rs
```diff
@@ -84,17 +84,19 @@ where
...
+		let mut inner = self.inner.write();
+		if inner.is_none() {
+			*inner = Some(value);
+			return true;
+		}
+		false
```
wiesche89, line 95: Please add a concurrency test with parallel init_if_unset() calls, exactly one must return true, and that caller’s value must remain stored.

core/src/global.rs
```diff
@@ -218,86 +217,84 @@ pub fn get_genesis_block() -> Block {
 
 /// One time initialization of the global future time limit
 /// Will panic if we attempt to re-initialize this (via OneTime).
-pub fn init_global_future_time_limit(new_ftl: u64) {
-	GLOBAL_FUTURE_TIME_LIMIT.init(new_ftl)
+pub fn init_global_future_time_limit(new_ftl: u64) -> bool {
```
wiesche89, line 220: These comments still promise a panic. Please document that the first value is preserved and later calls return false.

core/src/global.rs
```diff
@@ -218,86 +217,84 @@ pub fn get_genesis_block() -> Block {
...
 pub fn set_global_future_time_limit(new_ftl: u64) {
-	GLOBAL_FUTURE_TIME_LIMIT.set(new_ftl, true)
+	GLOBAL_FUTURE_TIME_LIMIT.set(new_ftl)
 }
 
 /// One time initialization of the global accept fee base
 /// Will panic if we attempt to re-initialize this (via OneTime).
-pub fn init_global_accept_fee_base(new_base: u64) {
-	GLOBAL_ACCEPT_FEE_BASE.init(new_base)
+pub fn init_global_accept_fee_base(new_base: u64) -> bool {
```
wiesche89, line 231: These comments still promise a panic. Please document that the first value is preserved and later calls return false.

core/src/global.rs
```diff
@@ -218,86 +217,84 @@ pub fn get_genesis_block() -> Block {
...
-		Some(ftl) => ftl,
+		})
 	})
 }
 
 /// One time initialization of the global NRD feature flag.
 /// Will panic if we attempt to re-initialize this (via OneTime).
-pub fn init_global_nrd_enabled(enabled: bool) {
-	GLOBAL_NRD_FEATURE_ENABLED.init(enabled)
+pub fn init_global_nrd_enabled(enabled: bool) -> bool {
```
wiesche89, line 286: These comments still promise a panic. Please document that the first value is preserved and later calls return false.

ardocrat: fixed

core/src/global.rs
ardocrat: fixed

core/src/global.rs
ardocrat: fixed

servers/src/common/adapters.rs
ardocrat: added

servers/src/common/adapters.rs
ardocrat: added

servers/src/common/adapters.rs
ardocrat: added

servers/src/common/adapters.rs
ardocrat: added

util/src/lib.rs
ardocrat: added

util/src/lib.rs
ardocrat: added

util/src/lib.rs
```diff
@@ -173,3 +174,22 @@ impl StopState {
...
+fn one_time() {
+	use std::thread;
+
+	let one_time: OneTime<u8> = OneTime::new();
+	one_time.init(1);
+	assert!(!one_time.init_if_unset(2));
+
+	let one_time_parallel: OneTime<u8> = OneTime::new();
+	let one_time_parallel_clone1 = one_time_parallel.clone();
+	thread::spawn(move || {
```
wiesche89: Please join both threads so their panics fail the test. Use different values and verify that exactly one returns true and its value is stored.

ardocrat: fixed

PR #3921, opened 2026-07-30
Deserialize without limits

ardocrat:
Add ability to deserialize without 100k bytes limit to select big tx context for wallet from db


ardocrat:
Closed in favor of https://github.com/mimblewimble/grin/pull/3922

PR #3932, opened 2026-08-31
Fix PIBD resume stall with newer archive target

### LINKED ISSUE #3931: staging: PIBD can stall when resuming with a newer archive target
This currently only affects staging.
When resuming PIBD with a newer archive target, the node may stop requesting output, rangeproof and kernel segments if the local and target PMMR sizes fall within the same segment. Headers continue to sync, but body sync gets stuck.
Checking the actual PMMR sizes instead of only the segment count fixes it. Resetting the node data is a workaround for affected nodes.

wiesche89:
PIBD could stop requesting output, rangeproof and kernel segments when
resuming against a newer archive target if both pmmr sizes were within the
same segment.

This could be fixed with three small changes, but the surrounding logic was
duplicated for all three pmmr. I moved it into a shared helper to keep the
behavior consistent and make it easier to test.

Tests cover the reported resume case, segment boundaries, completed targets
and Genesis (https://forum.grin.mw/t/grim-cross-platform-gui-for-grin/11128/115?u=wiesche).

Fixes #3931

### DIFF
--- a/chain/src/txhashset/desegmenter.rs
+++ b/chain/src/txhashset/desegmenter.rs
@@ -860,7 +860,7 @@ impl Desegmenter {
 	}
 
 	fn next_required_segment_index(
-		segment_type: &str,
+		segment_type: SegmentType,
 		local_mmr_size: u64,
 		target_mmr_size: u64,
 		segment_height: u8,
@@ -904,7 +904,7 @@ impl Desegmenter {
 		}
 
 		Self::next_required_segment_index(
-			"output",
+			SegmentType::Output,
 			local_output_mmr_size,
 			self.archive_header.output_mmr_size,
 			self.default_output_segment_height,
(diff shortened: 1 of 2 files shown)


chain/src/txhashset/desegmenter.rs
```diff
@@ -846,6 +846,55 @@ impl Desegmenter {
...
+	fn local_segment_count(local_mmr_size: u64, segment_height: u8) -> usize {
+		if local_mmr_size == 1 {
+			0
+		} else {
+			SegmentIdentifier::count_segments_required(local_mmr_size, segment_height)
+		}
+	}
+
+	fn next_required_segment_index(
+		segment_type: &str,
```
ardocrat: maybe its better to use `SegmentType` enum here

ardocrat (approved):

ardocrat (approved):

PR #3935, opened 2026-09-15
Make tx-related constants public

ardocrat:
- Make kernel and fee constants public to use them at grin-wallet
- Fix some typos
- Fix code reuse


core/src/core/committed.rs
```diff
@@ -170,7 +170,8 @@ pub fn sum_kernel_offsets(
 	}
 }
 
-fn to_secrets(bf: Vec<BlindingFactor>, secp: &secp::Secp256k1) -> Vec<SecretKey> {
+/// Convert blinding factors to private keys.
+pub fn to_secrets(bf: Vec<BlindingFactor>, secp: &secp::Secp256k1) -> Vec<SecretKey> {
```
wiesche89: The constants need to be public for the wallet, but this helper is only used inside grin_core. Could we make it pub(crate) instead?

core/src/core/transaction.rs
```diff
@@ -1656,11 +1662,7 @@ pub fn deaggregate(mk_tx: Transaction, txs: &[Transaction]) -> Result<Transactio
 			.filter(|x| *x != BlindingFactor::zero())
 			.filter_map(|x| x.secret_key(&secp).ok())
 			.collect::<Vec<_>>();
-		let negative_keys = kernel_offsets
-			.into_iter()
-			.filter(|x| *x != BlindingFactor::zero())
-			.filter_map(|x| x.secret_key(&secp).ok())
-			.collect::<Vec<_>>();
+		let negative_keys = to_secrets(kernel_offsets, &secp);
```
wiesche89, line 1655: Could we use to_secrets for positive_key as well? That would remove the remaining duplicate filter chain.

core/src/core/transaction.rs
```diff
@@ -1021,29 +1029,27 @@ impl TransactionBody {
 		self
 	}
 
+	/// Map kernels to features fees.
+	fn map_kernels_fees(
+		&self,
+	) -> FilterMap<Iter<'_, TxKernel>, fn(&'_ TxKernel) -> Option<FeeFields>> {
```
wiesche89: Could we return impl Iterator<Item = FeeFields> + '_ here? It avoids the clippy type_complexity warning and the extra imports.

core/src/core/transaction.rs
```diff
@@ -2298,6 +2300,19 @@ mod test {
 		assert_eq!(kernel2.excess_sig, sig.clone());
 	}
 
+	fn test_protocol_version(kernel: &TxKernel, commit: Commitment, sig: secp::Signature) {
```
wiesche89: Maybe rename this to assert_kernel_protocol_versions since it’s a helper, not a test. We could then reuse it in test_plain_kernel_ser_deser too.

core/src/core/transaction.rs
```diff
@@ -83,21 +86,21 @@ impl<'de> Deserialize<'de> for FeeFields {
...
 
-			fn visit_u64<E>(self, value: u64) -> Result<Self::Value, E>
+			fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
 			where
 				E: de::Error,
 			{
-				Ok(FeeFields(value))
+				let value = value
+					.parse()
+					.map_err(|_| E::custom("invalid fee field".to_string()))?;
```
wiesche89: E::custom can take the string literal directly, so .to_string() isn’t needed.

